PR #1195, opened 2014-09-15
Fix Idf import log message

kbenne:
Log was reporting translator's errors and warning log channel, now it
simply reports the untranslated objects as you would expect.

[#78851214]


DavidGoldwasser:
@kbenne I can build this branch. What is the best way for me to confirm the desired behavior? Is there a test setup that I can look at or ruby command to show this log?

PR #1321, opened 2014-12-31
Deserialize rubycontinuousvariables

elainethale:
This fixes the problem I e-mailed you about @macumber.

### DIFF
--- a/openstudiocore/src/project/ProblemRecord.cpp
+++ b/openstudiocore/src/project/ProblemRecord.cpp
@@ -27,11 +27,17 @@
 #include "MeasureGroupRecord_Impl.hpp"
 #include "OptimizationProblemRecord.hpp"
 #include "ProjectDatabase.hpp"
+#include "RubyContinuousVariableRecord.hpp"
+#include "RubyContinuousVariableRecord_Impl.hpp"
+#include "RubyMeasureRecord.hpp"
 #include "WorkflowRecord.hpp"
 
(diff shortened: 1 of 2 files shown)


openstudiocore/src/project/ProblemRecord.cpp
```diff
@@ -245,12 +251,35 @@ namespace detail {
...
     }
+    OptionalRubyMeasureRecord lastRubyMeasureRecord;
+    OptionalRubyMeasure lastRubyMeasure;
     for (int i = 0, n = ivrN + wrN; i < n; ++i) {
       if (!wrWIndex || (ivrWIndex && (*ivrWIndex < *wrWIndex))) {
         // InputVariable is next
         OS_ASSERT(*ivrWIndex == wIndex); // saved and expected index into workflow match
         OS_ASSERT(ivrIndex < ivrN);      // there is an input variable record to deserialize
-        workflow.push_back(WorkflowStep(inputVariableRecords[ivrIndex].inputVariable()));
+        InputVariableRecord ivarR = inputVariableRecords[ivrIndex];
```
macumber: @elainethale when is this code triggered?  I'll want to test it out in PAT

elainethale: This code is triggered when you open an existing project. (When a problem is deserialized from an osp.) Of course, the bulk of the code will not be triggered, because PAT doesn't support RubyContinuousVariables.

PR #1330, opened 2015-01-12
Adjust PIU fan schedule

kbenne:
When setting AirLoopHVAC::availabilitySchedule or when adding a new piu
terminal to an existing AirLoopHVAC, make sure the piu fan matches the
system availability schedule.  This is especially useful when night
cycle option is used on the AirLoopHVAC system.

A user can still set the fan availabilty schedule directly to ovrride
this default behavior.

[#84632652]

### DIFF
--- a/openstudiocore/src/model/AirLoopHVAC.cpp
+++ b/openstudiocore/src/model/AirLoopHVAC.cpp
@@ -23,6 +23,16 @@
 #include "AirLoopHVACSupplyPlenum_Impl.hpp"
 #include "AirLoopHVACReturnPlenum.hpp"
 #include "AirLoopHVACReturnPlenum_Impl.hpp"
+#include "AirTerminalSingleDuctSeriesPIUReheat.hpp"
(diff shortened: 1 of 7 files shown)


kbenne:
@asparke2 this is how it should have worked all along.  I think we both expected this.  This post from unmet hours is relevant.  https://unmethours.com/question/1997/problem-with-nightcycle-availability-manager-openstudio-1400-and-e-v-810009/

kbenne:
I offer this as proof.  Node 32 is a zone inlet node.  Node 12 the main air handler supply outlet node.

No night cycle

screen shot 2015-01-13 at 6 23 40 pm

cycle zone fans

screen shot 2015-01-13 at 6 29 34 pm

cycle full system

screen shot 2015-01-13 at 6 33 58 pm

Not really sure why this particular model needs to night cycle in chicago in the summer, but I don't think that is relevant to this PR.

asparke2:
Looks good, not sure how E+ does it, but it works!

PR #1335, opened 2015-01-12
Improve IG validation

### LINKED ISSUE #1286: Hard-sized entry of 0 reverts to Autosized
This may have already been reported but I couldn't find the issue I was looking for related to strange IP-SI conversion/rounding behavior. This happened in the Controller:OutdoorAir object while I was trying to hard size the minimum OA to 0 for an unoccupied zone with its own air loop. When returning to the field after the sim, my hard sized entry of 0 was replaced with Autosized. I hard-entered a value of 0 into the IDF and ran it through E+ instead with no errors.

kbenne:
Some autosizable fields were reported to not accept 0 as a value.
Change from autosize to hardsize with 0, and the IG would appear to
accept that.  But leave IG and come back and you will find that the
field is still set to autosize.

The root of this problem seems to be that some idd fields use exclusive
limits.  ie > not >=.  However the IG input field validators don't
discern the difference and therefore just default to inclusive limits.
This lets the IG interface lead you to believe you set a value that
really didn't validate and get written to the model.

This is a change to to add / subtract epsilon if the limit is exclusive,
so that you will know right away if a value is not accepted.

[fix #85606174]
fix #1286

### DIFF
--- a/openstudiocore/src/model_editor/InspectorGadget.cpp
+++ b/openstudiocore/src/model_editor/InspectorGadget.cpp
@@ -648,32 +648,42 @@ void InspectorGadget::layoutText( QVBoxLayout* layout,
       if(prop.minBoundType != IddFieldProperties::Unbounded)
       {
         double d = *(prop.minBoundValue);
-        if(prop.minBoundType == IddFieldProperties::ExclusiveBound) {
-          d = d + std::numeric_limits<double>::epsilon();
-        }
         if (m_unitSystem == IP) {
           OptionalQuantity minQ = convert(Quantity(d,u_si),u);
+          if(prop.minBoundType == IddFieldProperties::ExclusiveBound) {
+            if( minQ ) {
+              minQ->setValue(minQ->value() + std::numeric_limits<double>::epsilon());
+            }
+          }
           if (minQ) {
             text->setMin(minQ->value());
           }
         }
         else {
-          text->setMin( d );
+          if(prop.minBoundType == IddFieldProperties::ExclusiveBound) {
+            d = d + std::numeric_limits<double>::epsilon();
+            text->setMin( d );
+          }
         }
       }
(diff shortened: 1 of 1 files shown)


openstudiocore/src/model_editor/InspectorGadget.cpp
```diff
@@ -648,6 +648,9 @@ void InspectorGadget::layoutText( QVBoxLayout* layout,
       if(prop.minBoundType != IddFieldProperties::Unbounded)
       {
         double d = *(prop.minBoundValue);
+        if(prop.minBoundType == IddFieldProperties::ExclusiveBound) {
```
macumber: @kbenne it is possible that unit conversion could round this epsilon out, it would be better to add minInclusive and minExclusive methods on the widget.  This is probably ok for most cases though

kbenne:
I haven't looked, but I thought this used the stock validation facilities provided by Qt.  http://qt-project.org/doc/qt-4.8/qdoublevalidator.html or something like that.  That particular validator doesn't appear to have what you describe.

macumber:
I would be in here I think:

https://github.com/NREL/OpenStudio/blob/3aaa0984dbc185867261bde6305b141deb118d5e/openstudiocore/src/model_editor/IGLineEdit.cpp#L105

kbenne:
Interesting.  I think you're right, but I'm not that ambitious.

macumber:
@kbenne How about putting the +/- epsilon after unit conversion just to make sure?

kbenne:
I think that is a good compromise.

macumber:
This seems to work for OS:Coil:Cooling:DX:SingleSpeed, Rated Total Cooling Capacity.  We can merge once you have all your commits in

kbenne:
Thanks for catching that.  I think you should be able to merge now.

PR #1337, opened 2015-01-13
Fix orphan humidistat

### LINKED ISSUE #1290: deleting Humidifying Setpoint Schedule from Thermal Zone causes orphaned ZoneControlHumidistat
I tested the the humidity setpoint feature in 1.5.0 by adding a ZoneControlHumidistat schedule to a zone and then running the model. Removing the schedule from the zone and running the model again orphaned the ZoneControlHumidistat causing a fatal error.

** Severe  ** IP: IDF line~2409 Error detected in Object=ZONECONTROL:HUMIDISTAT, name=ZONE CONTROL HUMIDISTAT 1
**   ~~~   ** Field [Humidifying Relative Humidity Setpoint Schedule Name] is required but was blank

kbenne:
[fix #85579738]
fix #1290

### DIFF
--- a/openstudiocore/src/openstudio_lib/ThermalZonesGridView.cpp
+++ b/openstudiocore/src/openstudio_lib/ThermalZonesGridView.cpp
@@ -532,6 +532,9 @@ void ThermalZonesGridController::addColumns(std::vector<QString> & fields)
         [](model::ThermalZone * z) {
           if (boost::optional<model::ZoneControlHumidistat> thermostat = z->zoneControlHumidistat()) {
             thermostat->resetHumidifyingRelativeHumiditySetpointSchedule();
+            if( ! thermostat->dehumidifyingRelativeHumiditySetpointSchedule() ) {
+              thermostat->remove();
+            }
           }
         }
       );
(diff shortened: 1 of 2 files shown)


openstudiocore/src/openstudio_lib/ThermalZonesGridView.cpp
```diff
@@ -532,6 +532,9 @@ void ThermalZonesGridController::addColumns(std::vector<QString> & fields)
         [](model::ThermalZone * z) {
           if (boost::optional<model::ZoneControlHumidistat> thermostat = z->zoneControlHumidistat()) {
             thermostat->resetHumidifyingRelativeHumiditySetpointSchedule();
+            if( ! thermostat->dehumidifyingRelativeHumiditySetpointSchedule() ) {
```
macumber: @kbenne @evanweaver correctly implemented this code https://github.com/NREL/OpenStudio/commit/656a7b5f547928dcd4591274e19f59125a8cbdfe to remove objects after a bad drop.  However, I think that in this case an empty humidistat can be created if you drop a bad schedule on the object.  That is if the set fails, we may need to check for empty humidity schedules and remove the humidistat.  Checking for this in the translator as a back up is probably a good idea too

kbenne: How about just call the reset function if the drop fails?

macumber: Yeah that's kind of what I was thinking

kbenne: It doesn't look to me that Evan's commit has been merged.

macumber: You are right it is still on the multi edit branch

PR #1352, opened 2015-01-21
I had hoped to fix all the size_t to int warnings in the project but tha...

macumber:
...t is just not feasible.  Going forward I am not hot about converting all the APIs to size_t or fixing all of these errors.  Therefore, I just selectively disabled this warning so we can turn /WX back on for MSVC 64 bit builds

### DIFF
--- a/openstudiocore/src/utilities/core/Finder.hpp
+++ b/openstudiocore/src/utilities/core/Finder.hpp
@@ -103,7 +103,7 @@ namespace openstudio{
     typename std::vector<T>::const_iterator it;
     it = find_if(vec.begin(),vec.end(),finder);
     if (it != vec.end()) {
-      index = it - vec.begin();
+      index = (int)(it - vec.begin());
     }
     return index;
   }
--- a/openstudiocore/CMakeLists.txt
+++ b/openstudiocore/CMakeLists.txt
@@ -307,13 +307,18 @@ if(MSVC)
   # be linked against the same runtime
   add_definitions(/wd4251)
   
-  if(NOT CMAKE_CL_64)
-    # treat warnings as errors
-    add_definitions(/WX)
-  else()
-    add_definitions(/bigobj)
-  endif()
+  if(CMAKE_CL_64)
 
+    add_definitions(/bigobj)
+    
+    # ignore conversion from size_t to int for now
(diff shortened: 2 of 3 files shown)


macumber:
@lefticus @axelstudios @kbenne I'm curious about your thoughts on this?  Is there a similar warning on Mac?

openstudiocore/src/utilities/core/Finder.hpp
```diff
@@ -103,7 +103,7 @@ namespace openstudio{
     typename std::vector<T>::const_iterator it;
     it = find_if(vec.begin(),vec.end(),finder);
     if (it != vec.end()) {
-      index = it - vec.begin();
+      index = (int)(it - vec.begin());
```
lefticus: Isn't "never use c-style casts" in the coding standards?

macumber: Thanks @lefticus 

I didn't want to change the signature of the function, would a static cast be a better solution?  Or incrementing a counter?  I am not trying to say that this is a good function...

lefticus: Yes, `static_cast<int>(it - vec.begin())` is the "best" way to do it. You still are forcing one data type to be something different, but this is easy to search through the code for and find all the cases you are doing it.

axelstudios: FYI, I don't think there are similar warnings on mac.  The warnings that appear are generally small in number

lefticus: Yes, to answer the general question, 64bit MSVC is the only platform where `sizeof(int) != sizeof(size_t)`

The proper thing to do is to actually correct the errors, because there's probably a lurking overflow somewhere. This can generally be done with using `auto` more (which this file could benefit from immensely it appears to be at least 25% longer then it needs to be).

Practically, the overflow issues will never be a problem for you. Technically, you are arbitrarily limiting the API to working with 2 billion ish objects instead of 2^64 objects.

macumber:
Thanks everyone, sounds like we are all ok ignoring 64 bit overflow issues

We should prefer static_cast and auto to the c style casts going forward, but I'm not going to do a massive clean up now

Issue #978, opened 2014-04-13
cloning building doesn't bring over thermostatSetpointDualSetpoint


DavidGoldwasser:
Cloning building from one model to another doesn't bring over thermostats. Directly cloning thermal zones has the same results. @kbenne is this the desired behavior? It makes it tricky for me to bring in ref buildings because I need to bring in and re-wire zones to thermostats manually, (FYI  my workflow is to translate ref IDF to model2, and then clone that building into the original model.

Water use equipment doesn't come in either, but that is how I expect that worked, but still isn't ideal.

cc: @asparke2

macumber:
@kbenne is this fixed?  Can you close it?

macumber:
@kbenne nevermind, I accepted the pivotal ticket I will close

PR #1374, opened 2015-02-02
86020008 thermal zone clone

kbenne:
https://www.pivotaltracker.com/story/show/86020008


kbenne:
This should also address #978.

PR #1418, opened 2015-02-05
86127234 fan coil library

kbenne:
New fan coils and cycling PTAC for the library.  A couple of seemingly odd diffs are in here.  They are just tidying up the line endings as I saw errors during this work.  I also added some blank fields for the exponent curve so that IG doesn't show blue yuck.


kbenne:
https://www.pivotaltracker.com/story/show/86127234

kbenne:
@asparke2 done.

PR #1456, opened 2015-02-06
87764568 crash on remove water coil

kbenne:


kbenne:
https://www.pivotaltracker.com/story/show/87764568

PR #1470, opened 2015-02-17
74857180 rpg shade schedules

macumber:


macumber:
Well done @rpg777 !  Thanks @jasondegraw for the help too!

Issue #1372, opened 2015-01-31
PAT Crash when any measure fails because of a ruby error


asparke2:
1. Create a PAT project with a working seed model
2. Create a measure that fails due to a ruby error like a typo in a method name
3. Create a design alternative that includes the failing measure
4. Run the design alternatives
5. Crash

[This PAT project](https://github.com/NREL/OpenStudio-files/blob/master/Git%20Issue%20Files/Issue%201372.zip) reproduces it.
Tested with 1.6.0 on Windows.

PR #1475, opened 2015-02-19
Pat errors

macumber:


macumber:
Merging to develop will leave ticket to @asparke2 to accept and close #1372

PR #1489, opened 2015-02-26
88737258 epw ruby access

macumber:


openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -132,67 +132,119 @@ EpwDataPoint::EpwDataPoint(int year,int month,int day,int hour,int minute,
...
-  } else if(list.size() > 35) {
-    LOG_FREE(Error,"openstudio.EpwFile","Expected 35 fields in EPW data, got " << list.size() << ", additional data will be ignored");
-    return boost::optional<EpwDataPoint>();
+  // Expect 35 items in the list
+  if (list.size() < 35) {
+    if (pedantic) {
+      LOG_FREE(Error, "openstudio.EpwFile", "Expected 35 fields in EPW data, got " << list.size());
+      return boost::none;
+    } else {
+      LOG_FREE(Warn, "openstudio.EpwFile", "Expected 35 fields in EPW data, got " << list.size() << ", remaining fields will not be available");
```
macumber: @jasondegraw won't this cause an access violation down below when you try to access the fields if list is < 35?

openstudiocore/src/utilities/filetypes/EpwFile.hpp
```diff
@@ -90,167 +90,171 @@ class UTILITIES_API EpwDataPoint
...
-  static boost::optional<EpwDataPoint> fromEpwString(std::string line);
-  boost::optional<std::string> toWthString();
-  // One billion getters and setters
+  static boost::optional<EpwDataPoint> fromEpwString(const std::string &line);
+  static boost::optional<EpwDataPoint> fromEpwStrings(const std::vector<std::string> &list, bool pedantic=true);
+  std::vector<std::string> toEpwStrings() const;
+  boost::optional<std::string> toWthString() const;
+  // One billion getters
   Date date() const;
-  void setDate(Date date);
```
macumber: Hurray for me :-)

openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -480,72 +532,98 @@ static double psat(double T)
...
+  }
+  return value;
+}
+
+static double stringToDouble(const std::string &string, bool *ok)
+{
+  double value = 0;
+  *ok = true;
+  try {
+    value = std::stod(string);
```
macumber: Interesting, I didn't know about this new function.  I wonder how it performs relative to boost::lexical_cast and QString toDouble.  If this is faster we should update some of the parsing code in IdfObject, etc to use this.

openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -480,72 +532,98 @@ static double psat(double T)
...
+  catch (const std::invalid_argument) {
+    *ok = false;
+  }
+  catch (const std::out_of_range) {
+    *ok = false;
+  }
+  return value;
+}
+
+static double stringToDouble(const std::string &string, bool *ok)
```
macumber: Why is this a static function?

openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -132,67 +132,119 @@ EpwDataPoint::EpwDataPoint(int year,int month,int day,int hour,int minute,
...
+  pt.setAlbedo(list[EpwDataField::Albedo]);
+  pt.setLiquidPrecipitationDepth(list[EpwDataField::LiquidPrecipitationDepth]);
+  pt.setLiquidPrecipitationQuantity(list[EpwDataField::LiquidPrecipitationQuantity]);
   return boost::optional<EpwDataPoint>(pt);
 }
 
-boost::optional<std::string> EpwDataPoint::unitsByName(std::string name)
+std::vector<std::string> EpwDataPoint::toEpwStrings() const
+{
+  std::vector<std::string> list;
```
macumber: Could reserve 35 for speed

openstudiocore/src/utilities/filetypes/EpwFile.cpp
kbenne: That is interesting.  std = :)

openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -132,67 +132,119 @@ EpwDataPoint::EpwDataPoint(int year,int month,int day,int hour,int minute,
   setLiquidPrecipitationQuantity(liquidPrecipitationQuantity);
 }
 
-boost::optional<EpwDataPoint> EpwDataPoint::fromEpwString(std::string line)
+boost::optional<EpwDataPoint> EpwDataPoint::fromEpwString(const std::string &line)
+{
+  std::vector<std::string> list;
+  boost::split(list, line, boost::is_any_of(","));
```
macumber: Interesting, new method I didn't know about.  We do want to be careful not to introduce any new boost dependencies though.  QString might be preferred to this, unless C++11 has some spiffy method?

openstudiocore/src/utilities/filetypes/EpwFile.cpp
mbadams5: I had already done some changes from boost::lexical to a faster parsing, however this was pre-c++11 when this new method was introduced. But I agree with you about updating more of the parsing to use this probably.

openstudiocore/src/utilities/filetypes/EpwFile.cpp
kbenne: That is interesting.  boost = :(

kbenne: I think you should use std::regex.  It will take a couple of lines.  not as tidy but more portable.

macumber: Actually, @kbenne and @axelstudios we had at one point said we prefer QString to boost for string stuff.  However, with the desire to make a Qt-less utility and model target, I wonder if we should go back to boost?  Boost is at least statically linked (if not header only) and likely easier to emscripten....

mbadams5: There might be a way to do it without std::regex but i'll need to look a little more.

kbenne: let me google that for you. http://stackoverflow.com/questions/9435385/split-a-string-using-c11

kbenne: @macumber well with the adoption of c++11 with the exception of filesystem can't we get by with neither?  Well boost::optional but that is a minor header include.

mbadams5: @kbenne that is one of what I googled... and sent to you.

Also I agree with @kbenne about not needing either except in a few instances.

openstudiocore/src/utilities/filetypes/EpwFile.cpp
jasondegraw: @macumber Yes, I think it will. I think I forgot to finish this... will do that

openstudiocore/src/utilities/filetypes/EpwFile.cpp
jasondegraw: @macumber Mainly because it is just a drop in replacement for toDouble. The "right" way to do it would probably be to return an optional.

openstudiocore/src/utilities/filetypes/EpwFile.cpp
jasondegraw: I'll vote for std over boost regardless of speed pretty much every time, unless the speed difference is really significant.

openstudiocore/src/utilities/filetypes/EpwFile.cpp
jasondegraw: @macumber The vector will only ever be 35 elements, so it might be appropriate to use array instead. How about that?

openstudiocore/src/utilities/filetypes/EpwFile.cpp
jasondegraw: I'm not opposed to replacing this with that regex monstrosity, but only if all three of you think that's way to go. My understanding of anything regex is pretty limited. So I guess it is two for, with @macumber remaining to vote.

mbadams5: @jasondegraw my vote wasn't for regex, especially since I don't think we can use it... https://stackoverflow.com/questions/8561850/compile-stdregex-iterator-with-gcc/8913707#8913707

I'd rather use ...
http://stackoverflow.com/a/9436872/2358662
or 
http://stackoverflow.com/a/18564118/2358662

macumber: I don't really care my biggest concern is to not add any new boost dependencies.  If boost::split is a method we are using elsewhere then I guess it is ok to use here, if not we should use something that only requires C++11 (I would prefer to use C++11 over boost in any case).  

All the talk about boost vs Qt was a tangent that doesn't really apply here, C++11 should win in almost every case.

openstudiocore/src/utilities/filetypes/EpwFile.cpp
macumber: Vector is fine, just reserve 35 elements when you create it so it doesn't have to do a resize later.  Or even better, use the size constructor and then the enums to populate it:

    std::vector<std::string> list(35,"");
    list[EpwDataField::Year] = std::to_string(m_year);

openstudiocore/src/utilities/filetypes/EpwFile.cpp
macumber: My question was why make it static?  I ask out of curiousity, not because I know better.  Does this prevent it from conflicting with other functions of the same name or something?

http://stackoverflow.com/questions/7610668/are-there-any-advantages-to-making-a-free-static-function

openstudiocore/src/utilities/filetypes/EpwFile.cpp
macumber: Yup, kill as much boost and Qt as possible.  Dependencies are terrible, unless you want to talk about adding some Java ;-)

openstudiocore/src/utilities/filetypes/EpwFile.cpp
mbadams5: or just call...
list.reserve( 35 );

http://www.cplusplus.com/reference/vector/vector/reserve/

openstudiocore/src/utilities/filetypes/EpwFile.cpp
jasondegraw: @macumber I guess to emphasize that it shouldn't spread beyond this file?

macumber: That's cool, I didn't know about that use of static :-)  I should use that myself more often

mbadams5: @jasondegraw Does it really? Huh.

Anyways side note, why use 'bool *ok' instead of 'bool &ok' in the function parameter? Then you don't have to deal with ...
stringToDouble( "1234", &ok )

Currently looks very C-ish :smile:

jasondegraw: @mbadams5 I tend not to like that use of references even if it is safer. The right way to do it is probably to use boost::optional and get rid of the bool argument.

openstudiocore/src/utilities/core/StringHelpers.cpp
```diff
@@ -273,4 +274,17 @@ std::string formatUnderscore(const std::string& str) {
   return wStr;
 }
 
+std::vector <std::string> splitString(const std::string &string, char delimiter)
```
jasondegraw: @macumber @mbadams5 @kbenne So, how about this? (I haven't actually tested it too much yet, more looking to see if it is liked/disliked)

mbadams5: How often is this function called?

kbenne: :+1:

macumber: Looks good to me, my only genius idea is we might want to add arguments to keep empty parts or not, maybe an overload that splits on a string too:

http://qt-project.org/doc/qt-4.8/qstring.html#split

macumber:
@jasondegraw I meant for this commit to get added to this pull request.  Now that you are at NREL it's probably easiest if you start developing on NREL branches, or teach me how to use git better :-)

macumber:
@jasondegraw are you working on the performance issues?  We need to get those before merging.  Or are you just trying to increase the suspense since this will be such an awesome new feature :-)

@asparke2 FYI

jasondegraw:
@macumber I doubt that I have time now to do it right (profiling, etc). The performance isn't any worse than it was (@mbadams5 yes, yes, I know), so... 

If we really need it, then I think the best thing to do would be to remove the regexs and see what that does.

mbadams5:
@jasondegraw @macumber So I quickly profiled that test and the overwhelming majority of time is spent in EpwFile::parse(). And within that, almost all the time is due to the regexes in that function. 

Overall the 3 tests took 2549 ms to run and of that 2498 ms was spent in EpwFile::parse() with 2105 ms due to the regexes alone.

jasondegraw:
@macumber @mbadams5 Removing the regexs in the parsing of the data points reduces the Ruby test run time from about 70 seconds down to about 13. There's still probably some time to be saved in the string split function, but I think this is probably good to go for now.

macumber:
@jasondegraw did you run your big test script on all the weather files?

jasondegraw:
@macumber The number of changes per day on this branch has dropped a bit now, so it is ready for review. The TimeSeries changes are only done in one constructor right now. If you are OK with that, then I can replicate it in the other constructors and then start removing some of the superfluous data.

macumber:
Cool thanks Jason, I'll put this on my list to look at, won't be before Friday though

openstudiocore/ProjectMacros.cmake
```diff
@@ -275,10 +275,10 @@ macro(MAKE_SWIG_TARGET NAME SIMPLENAME KEY_I_FILE I_FILES PARENT_TARGET PARENT_S
 
   #message(STATUS "${${NAME}_SWIG_Depends}")
   
-  set(RUBY_AUTODOC "")
-  if(BUILD_DOCUMENTATION)
+  #set(RUBY_AUTODOC "")
```
macumber: @jasondegraw I can't commit to your branch, can you undo this change?

openstudiocore/ruby/openstudio/utilities/test/EpwFile_Test.rb
```diff
@@ -0,0 +1,102 @@
...
+  def test_data
+    path = @epwDir / OpenStudio::Path.new("USA_CO_Golden-NREL.724666_TMY3.epw")
+    epw = OpenStudio::EpwFile::load(path.to_s).get
+    data = epw.data
+    assert_equal(8760, data.size)
+    dataPoint = data[8759];
+    assert(dataPoint.is_a? OpenStudio::EpwDataPoint)
+    assert_equal(4, dataPoint.dryBulbTemperature.get)
+    assert_equal(81100, dataPoint.atmosphericStationPressure.get)
+    assert_equal(-1.0,dataPoint.fieldByName("Dew Point Temperature").get)
```
macumber: Getters that take an argument should be prefixed with 'get' 

https://github.com/NREL/OpenStudio/wiki/Cpp-Coding-Standards#other-naming-conventions

openstudiocore/src/utilities/data/TimeSeries.hpp
```diff
@@ -50,7 +50,7 @@ namespace openstudio{
 
         /// constructor from start date and time, interval length, values, and units
         /// throws if values is empty
-        TimeSeries_Impl(const DateTime& startDateTime, const Time& intervalLength, const Vector& values, const std::string& units);
```
macumber: @jasondegraw did you ever make that image that showed how to interpret time for all the different constructors?

openstudiocore/ruby/openstudio/utilities/test/EpwFile_Test.rb
jasondegraw: Not sure how that got missed, there's a couple methods that are similar that I have fixed in the next commit.

jasondegraw:
@macumber While we might want to wait and see if the SQLite fixes fix the test failure that I am seeing, this is probably as far as I can go with this right now (except for documentation and fixes, of course). The TimeSeries constructor behavior is now more uniform, and for the most part the previous behavior is preserved.

jasondegraw:
@macumber Merging in develop fixed the test failure, so except for some documentation we should be good to go. I'll add some documentation tomorrow, but don't feel like you need to wait for that...

openstudiocore/src/utilities/filetypes/EpwFile.cpp
macumber: @jasondegraw shouldn't this use the new splitString method in StringHelpers.hpp?

jasondegraw: @macumber Yes, it should. Fixed.

openstudiocore/src/utilities/data/TimeSeries.cpp
```diff
@@ -626,11 +765,40 @@ namespace openstudio{
 
     std::shared_ptr<TimeSeries_Impl> TimeSeries_Impl::operator*(double d) const {
 
-      return std::shared_ptr<TimeSeries_Impl>(new TimeSeries_Impl(m_firstReportDateTime, 
-        m_secondsFromFirstReport,
-        m_values*d, 
+      return std::shared_ptr<TimeSeries_Impl>(new TimeSeries_Impl(m_startDateTime,
+        m_secondsFromStart,
```
macumber: @jasondegraw I think this should still be the first way, constructed with m_firstReportDateTime, and m_secondsFromFirstReport, right?

jasondegraw: @macumber You are correct. This is left over from when I changed the interface. If we do get rid of some of the internals (m_secondsFromFirstReport, for example) then we will need to think about changing the impl interface some. We can do that, right?

macumber: Sure, we can change impl interfaces, is that what you are asking?  I'm guessing we need a combination of more tests (e.g. make sure multiplying by constant doesnt change dates) and simplified impl state (so there is less chance for confusion).  Think we can do that next week?

jasondegraw: Yes, that's what I meant, but lets leave that alone until we work on
cleaning out TimeSeries. I've added a test of multiplication and changed
how operator* is working (again). Now a multiplied interval series stays an
interval series.

PR #1490, opened 2015-02-27
1296 schedule day crash

kbenne:
ScheduleDay::addValue will no longer accept times less than or equal to
30 seconds.


openstudiocore/src/model/ScheduleDay.cpp
```diff
@@ -202,7 +202,7 @@ namespace detail {
 
   double ScheduleDay_Impl::getValue(const openstudio::Time& time) const
   {
-    if (time.totalDays() < 0.0 || time.totalDays() > 1.0){
+    if (time.totalMinutes() < 0.0 || time.totalDays() > 1.0){
```
macumber: @kbenne I don't think this is right here or in ::times(), we only need this in addValue I think

macumber: Oh wait I thought this was < 0.5, I think < 0.0 is correct here

Issue #1346, opened 2015-01-17
Results lost on model re-open - "Error opening measure and run data"


asparke2:
I have a model that I'm running and when I reopen it, a dialog shows up stating that the project.osp and run.db could not be opened (and therefore I have to re-run).  This is repeatable with a particular model I have.  Here is the dialog:

![1 6 0_reopen_bug](https://cloud.githubusercontent.com/assets/2009290/5789667/5c7bc3b8-9e31-11e4-960f-41f58e47b3c8.PNG)

And here is a link to a zip of the model:
https://www.dropbox.com/s/qe9r4es61em8c1i/1.6.0%20Testing.zip?dl=0

I think this model was originally created with 1.5.2, then opened in 1.5.5 (same error) and 1.6.0 (same error again).

DavidGoldwasser:
I believe that dialog was added to offer some recovery path for corrupted projects in PAT. So the underlying issue is probably the database getting corrupted. This offered a way to at least recover the measures in the workflow even if arguments can't be recovered. Not sure if this info is helpful, but thought I'd share the history of that dialog. Prior to this it wouldn't even open a PAT project. It may pre-date OS app measures? Can't remember

MatthewSteen:
I'm seeing this every time I open a model in 1.6.0 with a model that was created in a previous version, even after updating. Doesn't seem to happen with models created in 1.6.0.

macumber:
@asparke2 This allows a model to open if it has a bad-project.osp file in it:

https://github.com/NREL/OpenStudio/commit/747741a7c3e5c56012320ce4e4124e87bb75ac60

However, I don't know how or why you got the bad-project.osp in the first place.  Do either you or @MatthewSteen know what happened to cause that file to appear in the first place?

MatthewSteen:
@macumber for me it happens when I open a model created in an earlier version than 1.6.0.

macumber:
@MatthewSteen I created a new model in 1.4.0, added measures to it, then opened in 1.6.2.  I did not seen any errors and the bad-project.osp was not created.

macumber:
Unless we can figure out how the bad-project.osp 's are being created I think we will have to close this with just the one commit.

asparke2:
Confirmed that with 1.6.3, models with a bad-project.osp file will still open correctly.  Cannot figure out how to reproduce making the bad-project.osp itself.

PR #1492, opened 2015-03-02
This allows the model to open if there is a bad-project.osp file in the ...

macumber:
...companion directory

[#86721308]


macumber:
Fixes #1346

Issue #1350, opened 2015-01-21
Repeatable way to have walls fly away when adding fenestration in Plugin


DavidGoldwasser:
Followed steps described  by user here, and twice in a row was able to get similar behavior. I have never tried switching between spaces without existing space/group first, didn't even know I could.
https://unmethours.com/question/1625/wall-flying-every-where/?answer=1849#post-id-1849

Here are two before shots as I was building model. I double clicked on next space and added window in a series of spaces. I then pushed out wall as well just to see if that mattered. There were no warnings of issues while working on the model.

![clean_box_a](https://cloud.githubusercontent.com/assets/4857055/5832177/f0969b94-a0fd-11e4-9520-3881746d6e20.png)

![clean_box_b](https://cloud.githubusercontent.com/assets/4857055/5832180/f7a4ed3c-a0fd-11e4-8a24-7d210aad6709.png)

Photo re-opening the model from above.

![clean_box_c_reopen](https://cloud.githubusercontent.com/assets/4857055/5832182/01de0360-a0fe-11e4-8675-eee9cbd1fa0b.png)

It would seem that when we switch spaces this way something isn't gong correctly with the observers creating the open studio objects.FYI - tested in su2osm experimental workflow and it doesn't exhibit this behavior.

macumber:
There is a fundamental issue with SketchUp that makes it difficult to get the correct transformations when editing groups:

http://sketchucation.com/forums/viewtopic.php?f=180&t=61105

PR #1493, opened 2015-03-02
Cannot prevent the issue that causes flying walls, however this commit w...

macumber:
...ill throws an exception and alert the user if this issue occurs

#1350

[#86721074]


macumber:
@DavidGoldwasser this does not require a rebuild to test

macumber:
@DavidGoldwasser if you do accept this pull request, let's make sure to label the issue as not fixed.  We have to wait on SketchUp to add new API methods to do a real fix.

macumber:
https://github.com/NREL/OpenStudio/issues/1350

PR #1512, opened 2015-03-10
89112168 infinite measure clone

macumber:


macumber:
@DavidGoldwasser @asparke2 you should both test this out, it only copies files in the tests and resources directories when cloning a measure

PR #1517, opened 2015-03-12
Translate model to workspace on MeasureManager::updateMeasures

### LINKED ISSUE #1348: Arguments for E+ measures gets empty workspace in app
Works in PAT, reported by @mleachNREL

kbenne:
Fix #1348

### DIFF
--- a/openstudiocore/src/openstudio_lib/ScriptsTabView.cpp
+++ b/openstudiocore/src/openstudio_lib/ScriptsTabView.cpp
@@ -27,8 +27,10 @@
 #include "../shared_gui_components/MeasureManager.hpp"
 #include "../shared_gui_components/OSListView.hpp"
 #include "../shared_gui_components/SyncMeasuresDialog.hpp"
+#include "../utilities/plot/ProgressBar.hpp"
 
 #include "../analysisdriver/SimpleProject.hpp"
+#include "../energyplus/ForwardTranslator.hpp"
 
 #include <QLabel>
 #include <QVBoxLayout>
@@ -85,11 +87,13 @@ void ScriptsTabView::showEvent(QShowEvent *e)
 {
   MainTabView::showEvent(e);
 
-  boost::optional<openstudio::analysisdriver::SimpleProject> project = OSAppBase::instance()->project();
-  if (project)
-  {
-    // DLM: why is this necessary?
-    OSAppBase::instance()->measureManager().updateMeasures(*project, project->measures(), false);
+  if( auto project = OSAppBase::instance()->project() ) {
+    if( auto model = OSAppBase::instance()->currentModel() ) {
+      ProgressBar progress(this); 
+      energyplus::ForwardTranslator translator;
+      auto workspace = translator.translateModel(model.get(),&progress);
+      OSAppBase::instance()->measureManager().updateMeasures(*project, project->measures(), false, model, workspace);
+    }
   }
   variableGroupListView->refreshAllViews();
 }
--- a/openstudiocore/src/openstudio_lib/ScriptsTabView.cpp
+++ b/openstudiocore/src/openstudio_lib/ScriptsTabView.cpp
@@ -86,10 +86,11 @@ ScriptsTabView::ScriptsTabView(QWidget * parent)
 void ScriptsTabView::showEvent(QShowEvent *e)
 {
   MainTabView::showEvent(e);
+  auto app = OSAppBase::instance();
 
-  if( auto project = OSAppBase::instance()->project() ) {
-    if( auto model = OSAppBase::instance()->currentModel() ) {
-      ProgressBar progress(this); 
+  if( auto project = app->project() ) {
+    if( auto model = app->currentModel() ) {
+      ProgressBar progress(app->mainWidget()); 
       energyplus::ForwardTranslator translator;
       auto workspace = translator.translateModel(model.get(),&progress);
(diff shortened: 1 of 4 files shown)


kbenne:
@macumber I don't think this is ready yet, but I think correctly identifies the issue.  At least this brings the workspace into the context of the arguments section of the Measure.  I don't think it will address any potential issues with an empty workspace while *running* the Measure.

openstudiocore/src/shared_gui_components/MeasureManager.cpp
```diff
@@ -199,7 +201,8 @@ std::vector<ruleset::OSArgument> MeasureManager::getArguments(analysisdriver::Si
     if (appmodel)
     {
       model = appmodel;
-      idf = boost::none;
```
kbenne: These lines of code are in the context of MeasureManager::getArguments which is called by MeasureManager::updateMeasure(s).  The update is kicked off when the scripts tab is shown in OpenStudio App.  https://github.com/NREL/OpenStudio/blob/1348-IDFMeasureWorkspace/openstudiocore/src/openstudio_lib/ScriptsTabView.cpp#L92

kbenne: You really need to see the two lines of code below 202 to see the action.  The github diff has been clipped.

kbenne:
@macumber The work of this method https://github.com/NREL/OpenStudio/blob/16ee4fbf5d5520fa95f6c939199babb3755abffc/openstudiocore/src/openstudio_lib/ScriptsTabView.cpp#L84 could be moved into the widget constructor given our current architecture of one tab in memory at a time.  I'd probably error on the side of leaving things the way it is, but we probably need to think carefully about the blocking translator call in any case.  thoughts?

macumber:
@kbenne is this ready to test yet?

kbenne:
It is now.

openstudiocore/src/shared_gui_components/MeasureManager.cpp
```diff
@@ -180,7 +182,10 @@ BCLMeasure MeasureManager::insertReplaceMeasure(analysisdriver::SimpleProject &t
...
 }
 
-std::vector<ruleset::OSArgument> MeasureManager::getArguments(analysisdriver::SimpleProject &t_project, const BCLMeasure &t_measure)
+std::vector<ruleset::OSArgument> MeasureManager::getArguments(analysisdriver::SimpleProject &t_project, 
+  const BCLMeasure &t_measure,
+  const boost::optional<model::Model> &t_model,
+  const boost::optional<Workspace> &t_idf)
 {
   boost::optional<BCLMeasure> projectMeasure = t_project.getMeasureByUUID(t_measure.uuid());
   boost::optional<openstudio::model::Model> appmodel = m_app->currentModel();
```
macumber: @kbenne can you add some comments in here so we don't get confused again, like "appmodel will only be set in the OpenStudio Application.  PAT relies on this not being set and using the project instead."

openstudiocore/src/shared_gui_components/VariableList.cpp
```diff
@@ -380,7 +383,14 @@ void VariableListController::addItemForDroppedMeasureImpl(QDropEvent * event, bo
       // measure
       analysis::RubyMeasure measure(projectMeasure);
       try{
-        measure.setArguments(m_app->measureManager().getArguments(*project, projectMeasure));
+        auto model = m_app->currentModel();
+        boost::optional<Workspace> workspace;
+        if( model ) {
+          ProgressBar progress(m_app->mainWidget()); 
+          energyplus::ForwardTranslator translator;
+          workspace = translator.translateModel(model.get(),&progress);
```
macumber: @kbenne maybe these are too picky but since you are already on the scripts tab and the model can't change on the scripts tab, if you cached the workspace in the app you would not have to re translate it here

Also, you shouldn't really have to do this for a model measure but I guess it doesn't hurt

openstudiocore/src/openstudio_lib/ScriptsTabView.cpp
```diff
@@ -84,12 +86,15 @@ ScriptsTabView::ScriptsTabView(QWidget * parent)
...
-  {
-    // DLM: why is this necessary?
-    OSAppBase::instance()->measureManager().updateMeasures(*project, project->measures(), false);
+  auto app = OSAppBase::instance();
+
+  if( auto project = app->project() ) {
+    if( auto model = app->currentModel() ) {
+      ProgressBar progress(app->mainWidget()); 
+      energyplus::ForwardTranslator translator;
+      auto workspace = translator.translateModel(model.get(),&progress);
```
macumber: @kbenne why not store this in the app?

kbenne: ok.  some light caching sounds fine to me.  I'll work it up.

openstudiocore/src/shared_gui_components/VariableList.cpp
kbenne: @macumber One thing that surprised me when I was looking through this is that Model Measure arguments are populated with correct model when viewing from App.  It had to be pulling them from the t_project, but I am surprised the project keeps a current model.

kbenne:
@macumber ok I believe you merged this already, but here it is implementing the cache.

kbenne:
@macumber ok actually it is right here... https://github.com/NREL/OpenStudio/pull/1528

PR #1524, opened 2015-03-16
Drag n drop bclhvac

kbenne:


kbenne:
@axelstudios, Edwin is actually reviewing, but he doesn't have write privilege so I used your name.

Myoldmopar:
@kbenne I built this branch and tested it.  Worked great.  I was able to add a VRF system and terminal unit right from BCL and the simulation completed successfully.  

On a side note, the console output didn't update line-by-line, but rather got flushed at the end of the simulation, so yeah.....

Anyway, I don't have write access, so I can't hit the merge button, but I can say I reviewed and tested and am happy with the changes.

PR #1528, opened 2015-03-17
1348 idf measure workspace

kbenne:


macumber:
@kbenne whoops, i didn't mean to merge the last one yet, ill check this out

kbenne:
@macumber well i kinda changed direction.  Instead of passing the model and idf as an argument down through the layers, I just cache it in the app and access it globally there.  Since the cache was living at that level anyway it seemed silly and potentially confusing to also pass it down the functions.

macumber:
I saw that but figured you had a good reason, Im testing it out now, seems like it should be good

Issue #1525, opened 2015-03-16
Illuminance maps have confusing time scale


macumber:
Illuminance maps in ResultsViewer have a time scale that first goes through the design days and then the annual simulation.  This is different than the other results which have a different environment period for the sizing days.  Might want to address this if we change the SQL schema for https://github.com/NREL/OpenStudio/issues/1471

macumber:
This mirrors the way data is stored in E+ results

macumber:
This probably makes comparison of illuminance maps incorrect when comparing radiance outputs to E+

Issue #1518, opened 2015-03-13
ResultsViewer bugs from QWT update


macumber:
Found a few issues after the QWT update, I know illuminance maps have issues, not sure what else might be in there

macumber:
@DavidGoldwasser @asparke2 @axelstudios  can you guys do some testing with the 1.6.3 package?

macumber:
@axelstudios could you check out Linux and @kbenne could you check out Mac?

rpg777:
The two issues I know about with this are:
(1) Nighttime illuminance map data displays all white, instead of blue (which is the color that is supposed to represent zero values): ![Nighttime Data in RV 1.6.3](https://cloud.githubusercontent.com/assets/3161332/6644322/7a22cda4-c978-11e4-9094-edcbeff491d2.png)

(2) Stepping through timestep data causes RV to crash. I can pretty reliably reproduce this by going forward a few hours, then back a couple (using the up/down arrow toggle). Typical core dump looks like this: 
```
Process:         ResultsViewer [1628]
Path:            /Users/USER/*/ResultsViewer.app/Contents/MacOS/ResultsViewer
Identifier:      gov.nrel.openstudio.58cd57aac2.resultsviewer
Version:         1.6.3 (58cd57aac2)
Code Type:       X86-64 (Native)
Parent Process:  launchd [430]
Responsible:     ResultsViewer [1628]
User ID:         1852480870

Date/Time:       2015-03-13 12:02:58.242 -0600
OS Version:      Mac OS X 10.9.5 (13F34)
Report Version:  11
Anonymous UUID:  E032A787-21E9-9B01-43E3-3D03CE69C463

Sleep/Wake UUID: D36B07F8-03B9-41AB-8F01-6CEDA781789E

Crashed Thread:  0  Dispatch queue: com.apple.main-thread

Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000

Application Specific Information:
abort() called
*** error for object 0x7fd703f02750: pointer being freed was not allocated
 

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff97752866 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff9323935c pthread_kill + 92
2   libsystem_c.dylib             	0x00007fff8ef24b1a abort + 125
[...]

PR #1529, opened 2015-03-17
90298376 results viewer update

macumber:


macumber:
@axelstudios this should make 1.7.0

macumber:
Not done yet

macumber:
https://github.com/NREL/OpenStudio/issues/1518

macumber:
@axelstudios this is ready for testing now, can you try it out on Linux?  @kbenne can you try it on Mac?  @rpg777 can you try it out with radiance results?

macumber:
It is still slower but I will wait on that until I am done with https://github.com/NREL/OpenStudio/pull/1530

kbenne:
@macumber ok.

rpg777:
OK, using Radiance data: 
* I can step though the data with no crashes
* The floodplots look correct (and awesome):
Got Window Group Illuminance? =)

BUT:
* timesteps with all zero values have white flood plots instead of blue ones, still:
zero-value floodplot

macumber:
Thanks @DavidGoldwasser and @rpg777 my last commits change the appearance quite a bit so you should check it out, basically I am only rendering at the resolution of the data which improves performance and is arguably more "real"

macumber:
@DavidGoldwasser the slider weirdness for illuminance maps is due to https://github.com/NREL/OpenStudio/issues/1525 and is present in 1.6.0

There is slider weirdness for floodplots which is new though, the span slider starts out at 0 now when it used to start out fully maximized.  I might file a minor bug and let that go for now though, what do you think?

macumber:
I changed to rendering at twice resolution of illuminance maps density, we can discuss the right thing to do later

rpg777:
Looks good to me, Dan. I totally agree with the resolution rendering, it allows the user to appreciate the "quality" of their spatial data. The default smoothing obfuscates this unless you're really looking for it. 
flood plot, 1/2 grid spacing resolution

kbenne:
@macumber I'm not exactly sure what I need to be testing here, but I did compile it and fire up RV.  Worked fine on detailed time series line plots and flood plots.  This is the kind of stuff I normally use it for.  I didn't do any illuminance maps, but I figure Rob probably has that covered.

DavidGoldwasser:
@macumber I'm still seeing a crash this morning after overnight build, but I'm using a previously run simulation, wondering if that matters?

kbenne:
What are you testing @DavidGoldwasser?  I just pointed it at some old regression test results from over in OpenStudio-resources.

macumber:
Thanks @kbenne I just wanted some testing on Mac to make sure the update to the new Qwt version was working on Mac, if you can do line plots and flood plots on Mac and things seem fine that's all we need to check

rpg777:
@macumber @kbenne, I loaded an eplusout.sql on my Mac, looked at a few line plots and some more flood plots, and that all looks fine too. @kbenne you probably don't need to do anything.

DavidGoldwasser:
@kbenne I'm going to test an E+ run with illuminance maps in ResultsViewer. I mistakenly build develop last night, I'm building the correct branch this morning.

DavidGoldwasser:
@macumber The crash is gone when I use the branch now. I still do see a white screen with "NO DATA" vs. a blue background if i go to a time where there is no data.@rpg777 seems to see blue in this situation. Maybe this is not a mac vs. PC but becuase my map came from E+ run vs. Radiance?

macumber:
@DavidGoldwasser yes that is radiance vs E+ runs, that is working as designed

macumber:
Thanks all, sounds like this is good to go, I'm going to merge to develop.  @axelstudios can you do some testing with ResultsViewer on Linux once you make the packages?

PR #1540, opened 2015-03-19
@rpg777 can you check this out for all the various configurations (no re...

macumber:
...veal, no sill, etc, etc) I did not look at the output of this so I am committing "BLIND"-ly, get it! GUH-HYUK-HYUK!


macumber:
@rpg777 I don't know how to set the name for these new polygons either, do you know how to do that when we are writing this out stream of consciously here https://github.com/NREL/OpenStudio/blob/3647f03e55cfe2e811d8d4c314ce326052408564/openstudiocore/src/radiance/ForwardTranslator.cpp#L1451?

macumber:
@axelstudios let's try to get this in the package, may need a round or two of testing between @rpg777 and I

Issue #1180, opened 2014-08-29
Zone Infiltration Effective Leakage Area bug


macumber:
Generally I thought in SI units, we match what IDF editor shows, which does seem to be cm^2. But what we have isn’t even m^2, but m(m^2) which looks like m^3.

I attached some screenshots. If one of you has an answer related to this can you post it to the forum.
https://openstudio.nrel.gov/forums/openstudio-application/bug-reports-and-feature-requests/unit-effective-air-leakage-area-param#comment-3871

Thanks,
David

![8-14-2014 8-37-44 am](https://cloud.githubusercontent.com/assets/1327850/4093119/20dd16f6-2f99-11e4-9a90-c4b96d3e6e51.png)
![8-14-2014 8-34-29 am](https://cloud.githubusercontent.com/assets/1327850/4093121/20fcbf9c-2f99-11e4-941d-ad55f3fd6fe2.png)
![8-14-2014 8-34-43 am](https://cloud.githubusercontent.com/assets/1327850/4093120/20ee2d24-2f99-11e4-85d4-f5c1a01b3b4c.png)

PR #1575, opened 2015-04-05
[#91646056] Change effective air leakage area from m^2 to cm^2.

evanweaver:


kbenne:
@evanweaver so this shows up in the inspector gadget?

evanweaver:
@kbenne yes, via SpaceTypesGridView

DavidGoldwasser:
@kbenne talking with Dan and Evan, may want to look at if we handle this at all in version translation, when someone opens a 1.7.0 model that already has one of these objects. It isn't a heavily used object so probably not many models like this, but there are some people using it. Just got post on unmet hours this morning.
https://unmethours.com/question/4772/wrong-unit-in-the-openstudio-space-infiltration-effective-leakage-area-dialog/?answer=4776#post-id-4776

FYI - it is very hard to make this object now. If there isn't already a ticket, I'll add one to have add an object to our library so users can drag that in.

DavidGoldwasser:
FYI - this seems related to #1180. When I closed that I must not have tested forward translation to confirm unit translation. Seems like E+ may not have changed. We changed from m(m^2) to M^2

evanweaver:
@kbenne any thoughts here?

kbenne:
Well I can help with the version translation if needed.  I'm wrestling that beast right now.  leave it up to @DavidGoldwasser to tell us if it is necessary or not.  Based on the fact that we've had a few bug reports I'd say yes.

dstuder:
@evanweaver @kbenne @DavidGoldwasser 

This hasn't been fixed in 1.9.0...

Forward translating (from 1.8.4 to 1.9.0) causes the "Effective Air Leakage Area" parameter to be multiplied by a factor of 1,000. However, the conversion from m2 to cm2 should be 10,000 (100 x 100). Also, the units shown for this parameter in 1.9.0 are "m(m2)" but should be "cm2".

evanweaver:
Below is a model built in OS v1.8.5:

image

image

Below is the model above, now displayed in OS v1.9.0

image

image

As I have stated in the past, cm2 is now used in the OS idd, and corresponds with the E+ idd. As I have also said in the past, the SI units displayed were m2, and are now m(m2). Rather than display cm2, our units class displays m(m2), meaning milli * meters squared. Given milli is defined by 10^-3, 0.1 m2 = 100 m(m2).

With this, I am closing this ticket. If someone wants cm2 to be displayed, open a new ticket as a feature request, rather than a bug.

PR #1593, opened 2015-04-10
Refactoring modules available in Embedded Ruby into a single location, m...

macumber:
...aking that list the same as in openstudio.rb


macumber:
@DavidGoldwasser out of curiousity, did the measure you had trouble with have a "require 'openstudio'" at the top?  I am wondering if we should just make the embedded ruby load all of openstudio, then there is less to keep in synch.  Does anyone remember why we didn't do that to begin with?  Speed?

@kbenne @lefticus @axelstudios

PR #1631, opened 2015-04-21
Energy plus dir

macumber:
In order to test this we will need to have installers so we can run code using installed versions.  Since this is an addition, can we add the code, merge to develop, test installers, then proceed?  Just need sign off that these methods are ok


macumber:
This seems to be the install command for EnergyPlus:

https://github.com/NREL/OpenStudio/blob/16c6dd1b52eb78d15e9efb91d947dd4a496af21e/openstudiocore/CMakeLists.txt#L1047

macumber:
I'm going to just merge this so Andrew can start testing next iteration

PR #1652, opened 2015-05-13
Design days grid view

evanweaver:


asparke2:
@evanweaver I tested out this pull request and here are the things I found.  My guess is that the first 3 crashes are caused by the same thing.

1. Crash when Set Weather File
  1. Create a new model, save it
  2. Click Set Weather File
  3. Select Chicago-Midway
  4. Crash

2. Crash when opening a model with the weather file path set but file itself not at that location
  1. Create a model with 1.7.0 or other, set the weather file and design days
  2. Delete the folder with the same name as the osm.  This will delete the weather file, but the path will still be set in the .osm.
  3. Open the file with the design day grid view build
  4. Crash

3. Crash when opening a model with the weather file set and file is at that location
  1. Create a model with 1.7.0 or other, set the weather file and design days
  2. Open the file with the design day grid view build
  3. Crash

4. Design day grid doesn't respect IP/SI unit preferences (see Temperature sub-tab)

5. Why is the import button duplicated next to the Name: field?

6. Stretch: if a DDY file happens to contain 0 design days, we should show a dialog that warns the user

evanweaver:
Thanks @asparke2, I'll clean it up and get back to you.

evanweaver:
@asparke2  all of the above noted issues have now been resolved and checked into this feature branch.  Develop has been merged in, so this pull request should be ready for your eval.  Thanks.

asparke2:
@evanweaver I checked it out.  The crash bugs are gone now, here is what I see:

1. The behavior to check the weather file path is missing (in design doc)
2. When a weather file is imported and you populate the site object, pull the name from the City field in the weather file object.
3. The warning dialog doesn't work if you already have design days in the model
    1.  Import a valid DDY file with design days inside

evanweaver:
@asparke2 this should be ready to merge to develop.  Thanks!

Issue #1655, opened 2015-05-15
SQLFile queries much slower for E+ 8.3 results


macumber:
Timeseries queries are up to 1000x slower for SQLFiles produced by E+ 8.3 than for E+ 8.2.  This seems to be the root cause of this issue:

https://github.com/macumber/openstudio-vA3C/issues/14

@mbadams5 are you able to take a look at this?  There is an example model attached to the ticket in the vA3C repo.

macumber:
@kbenne @DavidGoldwasser FYI

macumber:
This is the query that is much slower

openstudio::OptionalTimeSeries SqlFile_Impl::timeSeries(const std::string& envPeriod, const std::string& reportingFrequency, const std::string& timeSeriesName, const std::string& keyValue)

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/sql/SqlFile_Impl.cpp#L2434

macumber:
Ultimately ends up at this sql query

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/sql/SqlFile_Impl.cpp#L2297

macumber:
@mbadams5 can you look into this test failure too?

SqlFileFixture.CreateSqlFile

unknown file: error: C++ exception with description &quot;Error executing SQL statement: insert into reportdatadictionary (ReportDataDictionaryIndex, IsMeter, VariableType, IndexGroup, TimestepType, KeyValue, VariableName, ReportingFrequency, ScheduleName, VariableUnits) values (1, &apos;0&apos;,&apos;Sum&apos;, &apos;Zone&apos;, &apos;Zone&apos;, &apos;DAYLIGHTING WINDOW&apos;, &apos;Daylight Luminance&apos;, &apos;Hourly&apos;, null, &apos;lux&apos;); no such table: reportdatadictionary&quot; thrown in the test body.

PR #1659, opened 2015-05-18
Fix sqlite queries

mbadams5:
This addresses #1655 

The cause of the slow query was due to a removed index from E+ sqlite generation. This index is now automatically added each SqlFile instance if it is not already in the database.

This also fixes a couple remaining query issues and failing sql unit tests.

I also updated the SQLite library included with OpenStudio to 3.10.1. This improves performance immensely across the board for queries and index creation.

@macumber


macumber:
Blazing fast, back to 9.4 seconds :-)  I'll review and merge tomorrow

openstudiocore/src/utilities/sql/SqlFile_Impl.cpp
```diff
@@ -82,46 +83,50 @@ namespace openstudio{
...
-            "CREATE TABLE Strings (StringIndex INTEGER PRIMARY KEY, StringTypeIndex  INTEGER, Value TEXT);"
-            "CREATE TABLE Surfaces (SurfaceIndex INTEGER PRIMARY KEY, SurfaceName, ConstructionIndex INTEGER, ClassName TEXT, Area REAL, GrossArea REAL, Perimeter REAL, Azimuth REAL, Height REAL, Reveal REAL, Shape INTEGER, Sides INTEGER, Tilt REAL, Width REAL, HeatTransferSurf INTEGER, BaseSurfaceIndex INTEGER, ZoneIndex INTEGER, ExtBoundCond INTEGER,  ExtSolar INTEGER, ExtWind INTEGER);"
-            "CREATE TABLE SystemSizes (SystemName TEXT, Description TEXT, Value REAL, Units TEXT);"
-            "CREATE TABLE TabularData (ReportNameIndex INTEGER, ReportForStringIndex INTEGER, TableNameIndex INTEGER, SimulationIndex INTEGER, RowNameIndex INTEGER, ColumnNameIndex INTEGER, RowId INTEGER, ColumnId INTEGER, Value TEXT, UnitsIndex INTEGER);"
+            "CREATE TABLE EnvironmentPeriods ( EnvironmentPeriodIndex INTEGER PRIMARY KEY, SimulationIndex INTEGER, EnvironmentName TEXT, EnvironmentType INTEGER, FOREIGN KEY(SimulationIndex) REFERENCES Simulations(SimulationIndex) ON DELETE CASCADE ON UPDATE CASCADE );"
```
macumber: @mbadams5 this does not require any change but maybe we should put the new index keys at the end of the column list?  this might affect people using index based queries?

macumber: Hmm looks like this was just wrong in that OpenStudio did not match the EnergyPlus schema.

openstudiocore/src/utilities/sql/SqlFile_Impl.cpp
```diff
@@ -138,31 +143,37 @@ namespace openstudio{
       if (m_connectionOpen)
       {
         try {
-          execAndThrowOnError("DROP INDEX rmdTI;");
```
macumber: @mbadams5 why would you ever call removeIndexes?  I don't think we need to remove this method just curious

openstudiocore/src/utilities/sql/SqlFile_Impl.cpp
```diff
@@ -174,31 +185,37 @@ namespace openstudio{
       if (m_connectionOpen)
       {
         try {
-          execAndThrowOnError("CREATE INDEX rmdTI ON ReportMeterData (TimeIndex ASC);");
```
macumber: @mbadams5 seems like we could switch on E+ version here, these queries will all fail on e+ 8.2 or earlier

macumber:
We need a regression type test to make sure we can read sql files from all versions of E+ 7.0 and up.  I'll take a crack at this on this branch.

@kbenne FYI

openstudiocore/src/utilities/sql/SqlFile_Impl.cpp
mbadams5: I don't understand your first comment, but the second is true but for E+ 8.3 and higher.

openstudiocore/src/utilities/sql/SqlFile_Impl.cpp
mbadams5: I don't know why you'd want to remove indexes.

openstudiocore/src/utilities/sql/SqlFile_Impl.cpp
mbadams5: I don't know if they will fail but they definitely won't all work on E+ 8.2 and earlier

mbadams5:
Yeah there might be changes here that could fail on E+ 8.2 and earlier but i'm not entirely sure. Regression tests and/or unit tests would probably be good.

PR #1677, opened 2015-06-08
Fix AirToAirComponent being counted twice on OA EquipmentList

### LINKED ISSUE #1676: Two HeatExchangerAirToAirSensibleAndLatent are output to EquipmentList
I came across this working on the prototype buildings but it is even seen in the heatexchanger_airtoair_sensibleandlatent.rb simulation test.

One HeatExchangerAirToAirSensibleAndLatent is created on the Outdoor Air Loop but when forwarded translated there are two instances output to AirLoopHVAC:OutdoorAirSystem:EquipmentList (but both are the same single HeatExchangerAirToAirSensibleAndLatent).

I'm surprised EnergyPlus allows this but it simulates, however, it causes the energy use to approximately double for Heat Recovery.

@kbenne

mbadams5:
Fixes #1676 

Since AirToAirComponents are on both OA and RA loops in the OA system, it was being output twice to the OA EquipmentList.

### DIFF
--- a/openstudiocore/src/model/AirLoopHVACOutdoorAirSystem.cpp
+++ b/openstudiocore/src/model/AirLoopHVACOutdoorAirSystem.cpp
@@ -344,7 +344,6 @@ namespace detail {
       }
       else if( boost::optional<AirToAirComponent> comp = modelObject->optionalCast<AirToAirComponent>() )
       {
-        modelObjects.push_back(*comp);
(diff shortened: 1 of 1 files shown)


kbenne:
This changes the API behavior in an undesirable way.  I'd like to do this at translator level.  I'll push up an alternative.

PR #1678, opened 2015-06-08
Fix a single air to air showing up on equipment list

### LINKED ISSUE #1676: Two HeatExchangerAirToAirSensibleAndLatent are output to EquipmentList

kbenne:
close #1676

In my tests this didn't change the energy numbers at all, but it is still is a good change to avoid improper idf.


kbenne:
@mbadams5 back to you for review.  I think this is a better way but let me know if you have concerns.

PR #1681, opened 2015-06-10
Fix simulation failure for CAV/VAV with no supply fan schedule in hvac library

### LINKED ISSUE #1605: Error in AirLoopHVACUnitarySystem Library
** Severe  ** AirLoopHVAC:UnitarySystem = UNITARY - CW COOLING - HW HEATING - CAV - HW REHEAT
   **   ~~~   ** Supply Fan Object Type = FAN:CONSTANTVOLUME
   **   ~~~   ** Fan type must be Fan:OnOff when Supply Air Fan Operating Mode Schedule Name = Blank.

mbadams5:
Fixes #1605

This solves the issue with CAV and VAV fans with a blank Supply Air Fan Operating Model Schedule. These hvac_library objects will simulate now.

It is currently defaulted to an Always On Discrete schedule, but the user can change a schedule in the GUI where they can not change a Fan Type.

### DIFF
--- a/openstudiocore/src/openstudio_app/Resources/hvaclibrary/hvac_library.osm
+++ b/openstudiocore/src/openstudio_app/Resources/hvaclibrary/hvac_library.osm
@@ -5707,7 +5707,7 @@ OS:AirLoopHVAC:UnitarySystem,
   ,                                       !- Air Outlet Node Name
   {057b7168-bdc2-418a-9961-3bfe683a200e}, !- Supply Fan Name
   BlowThrough,                            !- Fan Placement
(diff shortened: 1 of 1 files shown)


kbenne:
@mbadams5 did you use the existing always on schedule or make a new one? Using the existing would be good or we are going to accumulate a lot of always on in the hvac library.

mbadams5:
It should be the existing always on schedule. I think it is the one from the top of the library. I didn't add any new schedule if that is what you are wondering. This is only a two line fix.

kbenne:
Good point.  If you had added a new schedule it would show up in the diff wouldn't it.

Issue #1667, opened 2015-05-27
SketchUp Inspector does not properly create ThermalZone


kbenne:
https://unmethours.com/question/8069/openstudio-application-crash-when-clicking-on-facility-tab/

ThermalZone instances created in SektchUp do not have the required ZoneHVACEquipmentList or PortList that are normally created by the ThermalZone constructor.

macumber:
Changed to blocker since this causes a crash when clicking on facility tab

macumber:
Adding user file to files repo

macumber:
@kbenne here is the magic that fixes up zones added using the plugins inspector:

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/ruby/openstudio/sketchup_plugin/lib/interfaces/ThermalZone.rb#L62

kbenne:
@macumber  Who calls ```check_model_object``` and when?  It seems like one explanation might be that events to trigger that method are not happening deterministically.

macumber:
It's quite a call stack:

Kicked off here:
https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/ruby/openstudio/sketchup_plugin/lib/watchers/PluginModelWatcher.rb#L55

Goes here:
https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/ruby/openstudio/sketchup_plugin/lib/interfaces/ModelInterface.rb#L1026

Then here:
https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/ruby/openstudio/sketchup_plugin/lib/interfaces/ThermalZone.rb#L39

Then I think this gets called when the object is drawn:
https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/ruby/openstudio/sketchup_plugin/lib/interfaces/ThermalZone.rb#L62

So yes it is possible something goes wonky

DavidGoldwasser:
On a fresh model check_model_object does appear to get called. (I checked making new zone via the inspector, a user script, and the space attributes dialog). Here is the resulting model object. 

OS:ThermalZone,
  {bebcb225-dd00-41b1-afd9-2b9bc502c121}, !- Handle
  Thermal Zone 1,                         !- Name
  ,                                       !- Multiplier
  ,                                       !- Ceiling Height {m}
  ,                                       !- Volume {m3}
  ,                                       !- Floor Area {m2}
  ,                                       !- Zone Inside Convection Algorithm
[...]

PR #1685, opened 2015-06-15
Disabled ability to add or remove zones with inspector

DavidGoldwasser:
This should only affect the SketchUp Plugin. This is should resolve issue #1667 unless we find other causes of mal-formed thermal zones.

### DIFF
--- a/openstudiocore/ruby/openstudio/sketchup_plugin/lib/dialogs/PluginInspectorDialog.rb
+++ b/openstudiocore/ruby/openstudio/sketchup_plugin/lib/dialogs/PluginInspectorDialog.rb
@@ -68,6 +68,48 @@ module OpenStudio
     #  super
     #end
 
+    def show_error
+      msg  = "An error occurred in the OpenStudio SketchUp plug-in.\n\n"
+      msg += "It advised that you save a backup of your current OpenStudio model and restart SketchUp."
+      UI.messagebox(msg)
+    end
+    
+    def onPushButtonNew(checked)
+      model_interface =  Plugin.model_manager.model_interface
+      if model_interface.model_watcher.enabled
+        super(checked)
+      else
+        show_error
+      end
+    end
+    
+    def onPushButtonCopy(checked)
+      model_interface =  Plugin.model_manager.model_interface
+      if model_interface.model_watcher.enabled
+        super(checked)
+      else
+        show_error
+      end
+    end
+    
+    def onPushButtonDelete(checked)
+      model_interface =  Plugin.model_manager.model_interface
+      if model_interface.model_watcher.enabled
+        super(checked)
+      else
+        show_error
+      end
+    end
+    
+    def onPushButtonPurge(checked)
+      model_interface =  Plugin.model_manager.model_interface
+      if model_interface.model_watcher.enabled
+        super(checked)
+      else
+        show_error
+      end
+    end
+    
     def show
       Plugin.log(OpenStudio::Trace, "#{current_method_name}, enabled? = #{enabled?}")
       update
--- a/openstudiocore/ruby/openstudio/sketchup_plugin/lib/interfaces/ConstructionBase.rb
+++ b/openstudiocore/ruby/openstudio/sketchup_plugin/lib/interfaces/ConstructionBase.rb
@@ -31,7 +31,7 @@ module OpenStudio
       if not model_object.empty? and (handle.to_s == model_object.get.handle.to_s)
         model_object = model_object.get
       else
-        puts "ConstructionBase: model_object is empty for #{handle.class}, #{handle.to_s}, #{Plugin.model_manager.model_interface.openstudio_model}"                    
+        puts "ConstructionBase: model_object is empty for #{handle.class}, #{handle.to_s}, #{Plugin.model_manager.model_interface.openstudio_model}"
         model_object = nil
       end
       return model_object
--- a/openstudiocore/ruby/openstudio/sketchup_plugin/lib/interfaces/DaylightingControl.rb
+++ b/openstudiocore/ruby/openstudio/sketchup_plugin/lib/interfaces/DaylightingControl.rb
@@ -41,7 +41,7 @@ module OpenStudio
       if not model_object.empty?
         model_object = model_object.get
       else
-        puts "DaylightingControl: model_object is empty for #{handle.class}, #{handle.to_s}, #{Plugin.model_manager.model_interface.openstudio_model}"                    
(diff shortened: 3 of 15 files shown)


macumber:
@DavidGoldwasser I did not want to do this in a knee jerk reaction, I think you were right that this happened because the model watchers stopped working somehow, I don't think this big hammer solution is needed

@kbenne

kbenne:
@macumber are you suggesting this is an isolated incident and we take no action at this time?  Or can we move that initialization code to somewhere that does not depend on model watcher?

macumber:
I don't think we should disable adding and removing zones in the inspector yet.  I do think we should figure out and fix the orphaned connection when deleting a zone in the inspector. 

What I would really like is in-app analytics that get phoned home so we know which buttons people are pushing in our apps... Then we'd know if we can kill off a particular button and we could get crash reports with call stacks

macumber:
There are several operations where we disable the model watcher and then re-enable it later.  If someone pushes the new/remove/copy buttons in the inspector we should be able to check if the model watcher is enabled.  If it is not we should pop the plugin crashed warning.

kbenne:
ick.  Removing zones... I don't think that is something practical without model support.  https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/ThermalZone.cpp#L1395

macumber:
@kbenne Looks like the + button calls addObject, https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model_editor/InspectorDialog.cpp#L355, which does not call stuff in the public ctor, https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/ThermalZone.cpp#L2169.  I wonder if any of this could be moved into the impl ctor?  If not, I can continue to just mimic it in the watcher handler.

However, the remove and copy buttons should be calling clone, https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/ThermalZone.cpp#L2169, and remove, https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model_editor/InspectorDialog.cpp#L384, so I wonder what the real problem is?

kbenne:
@macumber do we know the issue #1667 was on remove?  Could it have been incomplete add or was that ruled out?  It seems like maybe it was, I can't remember.  

The most likely thing is of course some bad equipment interactions.  Equipment was removed badly by the api or something.  (basically a bug in my code)  But I was thinking we ended up here in plugin land because the model hadn't been exposed to OS App yet.

macumber:
@DavidGoldwasser noted an orphan connection when removing thermal zone in the inspector.  However, I get the same result doing this with the API:

irb(main):002:0> m = OpenStudio::Model::Model.new
irb(main):003:0> t = OpenStudio::Model::ThermalZone.new(m)
irb(main):004:0> t.remove
irb(main):005:0> puts m

OS:Version,
  {a8028979-32db-440a-99fb-d13cbe58ff6f}, !- Handle
  1.8.0;                                  !- Version Identifier

OS:Connection,
  {fe741d7d-6aaa-47d9-874f-4aa63603ef5c}, !- Handle
  {6cf1bac5-5a0a-464f-b4db-fdb602e61b89}, !- Name
  ,                                       !- Source Object
  11,                                     !- Outlet Port
  ,                                       !- Target Object
  2;                                      !- Inlet Port

macumber:
@kbenne given all that, I'm inclined to add my new error trapping code but keep the new, remove, and clone buttons working

kbenne:
I agree.

macumber:
@DavidGoldwasser I'm assigning this pull back to you, can you merge tomorrow morning after some quick testing?

@kbenne, up to you if you want to look into the orphan connection issue

macumber:
@kbenne I was able to cause a crash byt cloning several zones, then deleting them:

 	openstudio_model.dll!boost::assertion_failed(const char * expr, const char * function, const char * file, long line) Line 51	C++
 	openstudio_model.dll!openstudio::model::detail::ZoneHVACEquipmentList_Impl::thermalZone() Line 252	C++
 	openstudio_model.dll!openstudio::model::ZoneHVACEquipmentList::thermalZone() Line 399	C++
 	openstudio_model.dll!openstudio::model::detail::ThermalZone_Impl::zoneHVACEquipmentList() Line 1834	C++
 	openstudio_model.dll!openstudio::model::detail::ThermalZone_Impl::equipment() Line 1867	C++
>	openstudio_model.dll!openstudio::model::detail::ThermalZone_Impl::remove() Line 1410	C++
 	openstudio_utilities.dll!openstudio::WorkspaceObject::remove() Line 1269	C++
 	openstudio_modeleditor.dll!InspectorDialog::onPushButtonDelete(bool __formal) Line 384	C++
 	openstudiomodeleditor.so!_wrap_InspectorDialog_onPushButtonDelete(int argc, unsigned __int64 * argv, unsigned __int64 self) Line 9550	C++

I verified that the clone and remove model methods were being called.  I was able to reproduce in the API:

irb(main):006:0> m = OpenStudio::Model::Model.new
irb(main):007:0> t = OpenStudio::Model::ThermalZone.new(m)
irb(main):008:0> t2 = t.clone
irb(main):009:0> t2.remove
irb(main):010:0> t.remove # will crash

kbenne:
I'm going to open a new issue for ThermalZone::clone

macumber:
Is the orphaned connection on normal zone remove not an issue?  I honestly don't know 

irb(main):002:0> m = OpenStudio::Model::Model.new
irb(main):003:0> t = OpenStudio::Model::ThermalZone.new(m)
irb(main):004:0> t.remove
irb(main):005:0> puts m

OS:Version,
{a8028979-32db-440a-99fb-d13cbe58ff6f}, !- Handle
1.8.0; !- Version Identifier

OS:Connection,
{fe741d7d-6aaa-47d9-874f-4aa63603ef5c}, !- Handle
{6cf1bac5-5a0a-464f-b4db-fdb602e61b89}, !- Name
, !- Source Object
11, !- Outlet Port
, !- Target Object
2; !- Inlet Port

kbenne:
Not a functional issue.  It just looks bad.  I'll file a separate issue for that.

kbenne:
@macumber After reviewing, I think the zone::clone is the original corruption of the user model.  It is stealing the zone air node, which I believe was the original symptom.  I'll have a fix shortly.  I'll leave it up to @axelstudios the change should get into 1.8.

axelstudios:
I'm fine with this making it in.

DavidGoldwasser:
@macumber and @axelstudios I'm kicking off build now. I'll do basic testing and then merge to develop.

Issue #1388, opened 2015-02-04
Allow users to specify costs and ECMs on external models imported to PAT


macumber:
Do not need to support IDF import at this time.
Do need to support cloud run
Do not need to export bare OSM from OSM with measures in app
Do not allow adding external model with measures

Implementation details:
Create a new measure called “AddCostsToExternalModel” which takes arguments “Measure Type”, “Taxonomy”, and “Cost”, this measure will be shipped with OpenStudio.  When user creates alternative from external model, allow user interface to add multiple copies of this new measure.

macumber:
Checking if an alternative model has measures is too complex so I will not prevent a user from loading an alternative with measures (all alternatives have the built in measures)

PR #1694, opened 2015-06-23
90664130 alternative model pat

macumber:
@asparke2 I still want to fix a few bugs in PAT but this new feature is ready for initial testing


macumber:
@asparke2 are you able to build and test?

asparke2:
@macumber Yes, I will test

asparke2:
@macumber After merging develop into this branch, it works all the way through with the Xcel Measures.  I also tested importing an existing file with external measures, and that also works. There is one crash (not convinced it is related to your changes), and one change request.

Crash:
If you open a PAT project from a previous version, then try to save it after the upgrades, PAT crashes.  However, after the crash, if you re-open PAT and re-open the same project, it works fine (no crash on save).  I have a PAT file you can use to reproduce if you want.

crash message

Change request:
If a user creates a design alternative from external file and has 0 user-defined measures, the XML will leave out the /measures element.  The XML is still valid, but EDAPT will choke on it.  If there are 0 measures, can we add that to the warning dialog message and not export that design alternative to the XML?  Otherwise I will get questions about why isn't EDAPT working with this file.  You can imagine how I already know that this particular thing breaks EDAPT :)

macumber:
Thanks @asparke2 I'll look into the crash and can add that warning

macumber:
@asparke2 this is ready for you to review again.  Can you also look at the two issues that are still unchecked on the pivotal ticket?  I believe those can be closed in github.  Finally, are you able to write some documentation for this new feature to include in the PAT reference guide?

macumber:
Closes  #1388

asparke2:
@macumber This looks good, the change request to not export the extra thing and the file opening crash appear to be fixed.  I was able to reproduce the other two bugs you mentioned on the pivotal ticket, but I don't think those should hold up this work unless you think they are easy fixes.

Issue #1699, opened 2015-06-24
Meters are missing time indices in sql file


macumber:
Testing by @asparke2 seems to indicate that meters are not sufficient to create entries in the time table, only variables seem to do that.  Therefore, if your model has only meters at timestep frequency you will not have entries in the time table for time intervals that the meter is reported at.  Adding a variable at the same timestep seems to create the entries in the time table and these are associated with the meter.  As a workaround we will make sure that there is an output variable at the corresponding timestep for each model.

@mbadams5 FYI

macumber:
@MatthewSteen FYI

macumber:
Confirmed that this fix works for me

PR #1700, opened 2015-06-24
Preliminary fix for #1699, needs testing

macumber:


macumber:
@axelstudios , @asparke2 is going to test then we will want to merge this into release

macumber:
@nllong you may want to apply a similar fix to the server, I'm not sure if there is an equivalent place to do this?

asparke2:
@macumber This now passes my test for EDAPT export.

asparke2:
@macumber @axelstudios I merged this last bugfix, so I think it's really ready for 1.8.0

PR #1716, opened 2015-07-09
95574472 ip measure

DavidGoldwasser:
This should be ready to put into 1.8.1 installer. The measure itself will be posted to BCL for use with OpenStudio 1.7.5 or later. There may be some code that can be cleaned up related to calls to local .css and .js files.


openstudiocore/resources/analysis/pat_export/scripts/StandardReports/measure.rb
```diff
@@ -1,139 +1,165 @@
 require 'erb'
+require 'json'
+
+require "#{File.dirname(__FILE__)}/resources/os_lib_reporting"
+require "#{File.dirname(__FILE__)}/resources/os_lib_schedules"
+require "#{File.dirname(__FILE__)}/resources/os_lib_helper_methods"
 
 #start the measure
-class StandardReports < OpenStudio::Ruleset::ReportingUserScript
+class OpenStudioResults < OpenStudio::Ruleset::ReportingUserScript
```
macumber: Like the name :-)

DavidGoldwasser: Thanks, I didn't want to step on name of old one, so I just followed the name for E+ results. I also wanted the class name to match the display name, which wasn't true for the old measure; although on BCL it did snake_case the name when I uploaded it.

macumber: Hmm, now that I look at this you seem to have overwritten the StandardReports measure with this new one.  I thought you were adding a new measure and not overwriting this one?   Now the class name does not match the file name.  You did seem to give this a new UUID.

macumber: Here are the cases to consider:

1) existing OSM/OSP, this will reference the old standard report measure by name and.or uuid.  I thought we would want to keep the old one around for this reason.  When an old OSP is upgraded the code in SimpleProject would look for the new reporting measure and add it as well.
2) new OSM/OSP.  The new measure would be added only, the old measure would be around for people to drag into their workflows if they wanted to

openstudiocore/resources/analysis/pat_export/scripts/StandardReports/measure.xml
```diff
@@ -1,53 +1,476 @@
 <measure>
-  <name>Standard Reports</name>
-  <uid>fc337100-8634-404e-8966-01243d292a79</uid>
-  <version_id>1da31602-8ce8-4d8d-8cf7-6dc4206c433b</version_id>
-  <description>Change me</description>
-  <modeler_description>Change me</modeler_description>
+  <schema_version>3.0</schema_version>
+  <name>open_studio_results</name>
+  <uid>7910afa8-3a94-49ea-afa7-f09b185ac679</uid>
```
macumber: Changed the name and uuid but not the file name?

openstudiocore/resources/analysis/pat_export/scripts/StandardReports/measure.rb
```diff
@@ -1,139 +1,165 @@
 require 'erb'
+require 'json'
+
```
macumber: The files in openstudiocore/resources/analysis/pat_export/scripts seem to only be used in this unit test, https://github.com/NREL/OpenStudio/blob/bf5f68c76b79a69deaff25aae05cbb8a798b31b1/openstudiocore/src/analysis/test/Analysis_GTest.cpp.  So it is fine to update them but not critical.  We should make sure the unit tests still pass.

openstudiocore/resources/analysis/pat_export/scripts/StandardReports/measure.rb
macumber: I was expecting to see a new measure (OpenStudio Results) listed here:

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/bcl/BCLMeasure.hpp#L101

Then, code to update existing PAT projects here:

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/analysisdriver/SimpleProject.cpp#L442

This can get pretty tricky, can we just have the measure on BCL for the moment and not include in the package for 1.8.1?  The files in the measure should be renamed to match the class name in the measures repo (if that is not already the case).

macumber: @DavidGoldwasser if you want to go the easy way then I suggest just updating the existing standard reports to the new measure, preserving the old name and UUID.  If you do this on the BCL too, is the new measure backwards compatible (minus off line operation) to 1.7.0?

DavidGoldwasser: @macumber it needs 1.7.5 or later I would have had to add some code to support pre 8.3 E+ due to SQL query changes.

mbadams5: @DavidGoldwasser are you sure? I tried hard to keep reading from the 8.3+ E+ SQLite database the same as pre-8.3. It is inserting new data that is slightly different.

macumber: @DavidGoldwasser which query does not work in 8.2?

PR #1719, opened 2015-07-13
Evans hvac objects

evanweaver:
@kbenne


openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFluidCoolerSingleSpeed.cpp
```diff
@@ -0,0 +1,147 @@
...
+  ///////////////////////////////////////////////////////////////////////////
+  s = modelObject.name();
+  if (s)
+  {
+    idfObject.setName(*s);
+  }
+  ///////////////////////////////////////////////////////////////////////////
+
+  ///////////////////////////////////////////////////////////////////////////
+  mo = modelObject.waterInletNode();
```
kbenne: This method should not exist.  It should be from the base StraightComponent "inletModelObject".  It was implemented in FluidCooler when instead StraightComponent::inletModelObject should have been reimplemented.  Using the keyword override would help ensure.  I'll comment when I get to the class.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFluidCoolerSingleSpeed.cpp
```diff
@@ -0,0 +1,147 @@
...
+    s = mo->name();
+    if (s)
+    {
+      idfObject.setString(openstudio::FluidCooler_SingleSpeedFields::WaterInletNodeName, *s);
+    }
+  }
+  ///////////////////////////////////////////////////////////////////////////
+
+  ///////////////////////////////////////////////////////////////////////////
+  mo = modelObject.waterOutletNode();
```
kbenne: Same as above.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFluidCoolerSingleSpeed.cpp
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+  ///////////////////////////////////////////////////////////////////////////
+
+  ///////////////////////////////////////////////////////////////////////////
+  idfObject.setString(openstudio::FluidCooler_SingleSpeedFields::PerformanceInputMethod, modelObject.performanceInputMethod());
+  ///////////////////////////////////////////////////////////////////////////
+
+  ///////////////////////////////////////////////////////////////////////////
+  d = modelObject.designAirFlowRateUfactorTimesAreaValue();
```
kbenne: This field is autosizable and that needs to be handled.  I believe there are other fields in this object that are autosizable too.  I'll leave it for you to figure out which ones.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFluidCoolerTwoSpeed.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <utilities/idd/IddEnums.hxx>
+#include <utilities/idd/FluidCooler_TwoSpeed_FieldEnums.hxx>
+
+using namespace openstudio::model;
+
+namespace openstudio {
+
+namespace energyplus {
+
+  boost::optional<IdfObject> ForwardTranslator::translateFluidCoolerTwoSpeed(FluidCoolerTwoSpeed & modelObject)
```
kbenne: All of my comments on single speed fluid cooler apply here.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslatePipeIndoor.cpp
```diff
@@ -0,0 +1,148 @@
...
+    s = mo->name();
+    if (s)
+    {
+      idfObject.setString(openstudio::Pipe_IndoorFields::ConstructionName, *s);
+    }
+  }
+  ///////////////////////////////////////////////////////////////////////////
+
+  ///////////////////////////////////////////////////////////////////////////
+  mo = modelObject.fluidInletNode();
```
kbenne: Same comment I made about StraightComponent::inletModelObject / outletModelObject apply here.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslatePipeOutdoor.cpp
```diff
@@ -0,0 +1,118 @@
...
+    s = mo->name();
+    if (s)
+    {
+      idfObject.setString(openstudio::Pipe_OutdoorFields::ConstructionName, *s);
+    }
+  }
+  ///////////////////////////////////////////////////////////////////////////
+
+  ///////////////////////////////////////////////////////////////////////////
+  mo = modelObject.fluidInletNode();
```
kbenne: Same comment as previously.

openstudiocore/src/model/Duct.cpp
```diff
@@ -0,0 +1,174 @@
...
+    if (result.empty()){
+    }
+    return result;
+  }
+
+  IddObjectType Duct_Impl::iddObjectType() const {
+    return Duct::iddObjectType();
+  }
+
+  boost::optional<Connection> Duct_Impl::inletNode() const {
```
kbenne: This should not be here.  You only need to define the [ports](https://github.com/NREL/OpenStudio/blob/df20ac3652c3c3bf1d1172fee3d65434f0780124/openstudiocore/src/model/FanConstantVolume.cpp#L111) and let the base class do the rest of this inlet outlet stuff.

openstudiocore/src/model/Duct.cpp
```diff
@@ -0,0 +1,174 @@
...
+    return getObject<ModelObject>().getModelObjectTarget<Connection>(OS_DuctFields::InletNode);
+  }
+
+  boost::optional<Connection> Duct_Impl::outletNode() const {
```
kbenne: Remove

openstudiocore/src/model/Duct.cpp
```diff
@@ -0,0 +1,174 @@
...
+  {
+    return OS_DuctFields::InletNode;
+  }
+
+  unsigned Duct_Impl::outletPort()
+  {
+    return OS_DuctFields::OutletNode;
+  }
+
+  bool Duct_Impl::setInletNode(const boost::optional<Connection>& connection) {
```
kbenne: Remove

openstudiocore/src/model/Duct.cpp
```diff
@@ -0,0 +1,174 @@
...
+    }
+    return result;
+  }
+
+  void Duct_Impl::resetInletNode() {
+    bool result = setString(OS_DuctFields::InletNode, "");
+    OS_ASSERT(result);
+  }
+
+  bool Duct_Impl::setOutletNode(const boost::optional<Connection>& connection) {
```
kbenne: Remove

openstudiocore/src/model/Duct.cpp
```diff
@@ -0,0 +1,174 @@
...
+      result = setPointer(OS_DuctFields::OutletNode, connection.get().handle());
+    }
+    else {
+      resetOutletNode();
+      result = true;
+    }
+    return result;
+  }
+
+  void Duct_Impl::resetOutletNode() {
```
kbenne: Remove... and similar.

openstudiocore/src/model/Duct.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+  //@}
+
+  static IddObjectType iddObjectType();
+
+  /** @name Getters */
+  //@{
+
+  // TODO: Check return type. From object lists, some candidates are: Connection.
+  boost::optional<Connection> inletNode() const;
```
kbenne: Remove

openstudiocore/src/model/Duct.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+  // TODO: Check return type. From object lists, some candidates are: Connection.
+  boost::optional<Connection> outletNode() const;
+
+  //@}
+  /** @name Setters */
+  //@{
+
+  // TODO: Check argument type. From object lists, some candidates are: Connection.
+  bool setInletNode(const Connection& connection);
```
kbenne: Remove.

openstudiocore/src/model/Duct_Impl.hpp
```diff
@@ -0,0 +1,107 @@
...
+namespace model {
+
+// TODO: Check the following class names against object getters and setters.
+class Connection;
+class Connection;
+
+namespace detail {
+
+  /** Duct_Impl is a StraightComponent_Impl that is the implementation class for Duct.*/
+  class MODEL_API Duct_Impl : public StraightComponent_Impl {
```
kbenne: Remove all get / set node stuff.

openstudiocore/src/model/Duct_Impl.hpp
```diff
@@ -0,0 +1,107 @@
...
+#define MODEL_DUCT_IMPL_HPP
+
+#include <model/ModelAPI.hpp>
+#include "StraightComponent_Impl.hpp"
+
+namespace openstudio {
+namespace model {
+
+// TODO: Check the following class names against object getters and setters.
+class Connection;
```
kbenne: Remove any and all Connection declarations or includes from the headers.  Actually you shouldn't need any connection includes they will come from the base class.

openstudiocore/src/model/FluidCoolerSingleSpeed.cpp
```diff
@@ -0,0 +1,498 @@
...
+{
+  OS_ASSERT(getImpl<detail::FluidCoolerSingleSpeed_Impl>());
+
+  // TODO: Appropriately handle the following required object-list fields.
+  bool ok = true;
+   //ok = setHandle();
+  OS_ASSERT(ok);
+   ok = setPerformanceInputMethod("NominalCapacity"); // UFactorTimesAreaAndDesignWaterFlowRate, NominalCapacity
+  OS_ASSERT(ok);
+  ok = setDesignAirFlowRateUfactorTimesAreaValue(0.1);
```
kbenne: These don't look like very good defaults to me.  Use E+ idd default or better yet us material from an example file.

openstudiocore/src/model/FluidCoolerSingleSpeed.cpp
```diff
@@ -0,0 +1,498 @@
...
+  OS_ASSERT(ok);
+   ok = setNominalCapacity(0.1);
```
kbenne: If a field is autosizeable then initialize it to autosize.

openstudiocore/src/model/FluidCoolerSingleSpeed.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+bool FluidCoolerSingleSpeed::setDesignAirFlowRateFanPower(double designAirFlowRateFanPower) {
+  return getImpl<detail::FluidCoolerSingleSpeed_Impl>()->setDesignAirFlowRateFanPower(designAirFlowRateFanPower);
+}
+
+void FluidCoolerSingleSpeed::autosizeDesignAirFlowRateFanPower() {
+  getImpl<detail::FluidCoolerSingleSpeed_Impl>()->autosizeDesignAirFlowRateFanPower();
+}
+
+bool FluidCoolerSingleSpeed::setOutdoorAirInletNode(const Node& node) {
```
kbenne: This getter/setter is one that can stay.  It is just a reference as opposed to being connected into the hvac topology through Connection.

openstudiocore/src/model/FluidCoolerSingleSpeed_Impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+
+    bool setDesignAirFlowRate(boost::optional<double> designAirFlowRate);
+
+    void autosizeDesignAirFlowRate();
+
+    bool setDesignAirFlowRateFanPower(boost::optional<double> designAirFlowRateFanPower);
+
+    void autosizeDesignAirFlowRateFanPower();
+
+    // TODO: Check argument type. From object lists, some candidates are: Node.
```
kbenne: Remove all of the TODO comments.

openstudiocore/src/model/FluidCoolerTwoSpeed.hpp
```diff
@@ -0,0 +1,218 @@
...
+
+  boost::optional<double> lowFanSpeedAirFlowRate() const;
+
+  bool isLowFanSpeedAirFlowRateAutocalculated() const;
+
+  double lowFanSpeedAirFlowRateSizingFactor() const;
+
+  boost::optional<double> lowFanSpeedFanPower() const;
+
+  bool isLowFanSpeedFanPowerAutocalculated() const;
```
kbenne: I should have said something on the IDD review but I try to just use the keyword autosize so that the interfaces are consistent.  I find that introducing a distinction between autosize and autocalculateable pedantic and it makes the api more complicated.

kbenne:
Many of my comments are attached to the first line where I saw the occurrence, but there are typically several other places where the same comment applies.  Here is a summary of what I found.

* These all derive from StraightComponent.  There should be no implementing inlet/outletModelObject/Node methods.  Just reimplement StraightComponent::inletPort and ::outletPort and be done.  There are a couple of exceptions where there is a simple reference to a node for control purposes (get set outdoor air node for instance).  I tried to comment where I saw those exceptions.  If you see a get /set plant/fluid/airInletNode or ...outletNode it needs to be removed.

* Many objects have autosizeable fields that are not accounted for in the translator.

* Many objects have autocalculate fields which I would prefer to see just as autosize for simplicity in the API.

* The default initialization needs improvement.  Look for some example files for source material.  Autosize all autosizeable / autocalculatable fields.

* Add entires for all objects that have schedules in the schedule type registry.

* Remove all include and forward declarations of Connection.  That is an implementation thing that the base class takes care of.

* Ideally these would be added to the UI.  At minimum we need icons so you don't see a blank spot in the hvac layout UI if someone uses this stuff through the api.

* Need simulation tests like [this](https://github.com/NREL/OpenStudio-resources/blob/develop/model/simulationtests/vrf.rb)  To me this is probably the most important thing because I normally catch so many of my own errors here.  You really don't know if it works until you send these new ModelObjects through E+.  I can help on this one.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFluidCoolerSingleSpeed.cpp
```diff
@@ -0,0 +1,147 @@
...
+  if (d)
+  {
+    idfObject.setString(openstudio::FluidCooler_SingleSpeedFields::DesignAirFlowRateUfactorTimesAreaValue, toString(*d));
```
mbadams5: @evanweaver You should use setDouble here and in all other places instead of using setString and calling toString(). I don't even know where it is making that call to? Maybe it is converting it to a QString? Either way, not the correct approach.

evanweaver: @mbadams5 thanks, careless oversight on my part.

evanweaver:
@kbenne  do you want to merge this into this iteration?

PR #1727, opened 2015-07-16
1712 availability managers

kbenne:
@asparke2 I didn't leave you with a very good description about what is going on here. I added 

AvailabilityManager_HybridVentilation ()
AvailabilityManager_OptimumStart ()
AvailabilityManager_DifferentialThermostat ()
AvailabilityManager_NightVentilation ()

You can see these demonstrated [here](https://github.com/NREL/OpenStudio-resources/blob/develop/model/simulationtests/availability_managers.rb#L29).

The api allows one AVM at a time, but requires an availability schedule so if there is no AVM explicitly provided the osm->idf translator will provide a scheduled AVM based on the availability manager attached to the system.

Stacking multiple AVMs is not supported.  Lets look to EMS for those types of complicated control strategies.

My design document is [here](https://docs.google.com/document/d/1lCbmzkL_tLjAd99B88n5nxP5vE5IeoasJecJjF7KjPI/edit#heading=h.ve83x9n67ogo) if you want a little more detail.

The new API is [here](https://github.com/NREL/OpenStudio/blob/8ec7bef3a79dbd3b43075b5b7dbeff00d04976b6/openstudiocore/src/model/AirLoopHVAC.hpp#L258).

Old APIs to conveniently set night cycle is still there and functioning the same.  Although now you could explicitly set and access the night cycle manager object to control more detailed settings like the thermostat offset.


asparke2:
@kbenne Everything looks logical to me.  One question: where is the code to translate the NightCycle SPM?  It looks like it was removed from ForwardTranslateAirLoopHVAC, but I don't see where it landed.  Just trying to make sure it is still somewhere.

kbenne:
@asparke2 

uhh.... I was going to point you here 

https://github.com/NREL/OpenStudio/blob/1712-AvailabilityManagers/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateAirLoopHVAC.cpp#L335

and then point you to the night cycle translator. All of the managers now have explicit translate functions, like https://github.com/NREL/OpenStudio/blob/1712-AvailabilityManagers/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateAvailabilityManagerNightVentilation.cpp

Except I see I didn't actually commit the new night cycle translator.  One moment I think I have that.

Although I now realize that the simulation test doesn't actually test the most important manager.

https://github.com/NREL/OpenStudio-resources/blob/develop/model/simulationtests/availability_managers.rb#L30

I will fix that too.

PR #1733, opened 2015-07-21
TimeSeries Fix

jasondegraw:
@axelstudios @macumber This should hopefully fix at least part of the time series issue that came up today. Part of the problem is in SqlFile's timeSeries method. I have changed it to use the new TimeSeries creation behavior, but it still creates "RunPeriod" time series using the detailed ctor rather than the interval ctor. That is not ideal, but is acceptable for now as far as I can tell.


axelstudios:
This appears to have fixed the problem.  @jasondegraw Do you expect any more commits, since you said this might only fix part of the issue?

jasondegraw:
@axelstudios Not if it fixes the problem. I can't run what you were running, so I wasn't sure this would do it. I have some concerns about SqlFile::timeSeries, but I am not inclined to mess with it until I fully understand it.

axelstudios:
Nevermind, I think it's still failing.  Let me know if you want to step through the debugger with me.

PR #1745, opened 2015-07-29
Fix component setpoint scheme

### LINKED ISSUE #1709: Bug in plant operation schemes object for component setpoint
Translator for plant operation schemes set the demand calculation node for plant equipment as the component outlet node, in the 'PlantEquipmentOperation:ComponentSetpoint' object.
This instead should be the component inlet node.

Specifying the demand calculation node as the component outlet node results in high loop temperatures and high number of unmet load hours.

kbenne:
When generating component setpoint operation schemes, OpenStudio
incorrectly used the component outlet node as the demand calculation
node, when it should be the component inlet node.

fix #1709


kbenne:
Fix verified by @supriyagoel

PR #1760, opened 2015-08-09
100880588 Fix EpwFile regex issue by removing regex

### LINKED ISSUE #1758: OpenStudio::EpwFile.new() crashes on Linux
To repeat (bug on linux only, works fine on windows and mac):

> irb
> require 'openstudio'
> OpenStudio::EpwFile.new('USA_TX_Houston-Bush.Intercontinental.AP.722430_TMY3.epw')

result:

>regex_error

This is a blocker because the Create Prototype Building Measure (and in turn the PTool work) both need this method to work in order to dynamically set the weather file on the server.

@macumber @jasondegraw FYI

jasondegraw:
Fix the EpwFile Linux regex issue by removing the regex usage. It is not clear exactly what the issue was, but using regex here was a bit of overkill.

This will close #1758

This isn't quite ready to go yet. Need to build it on Linux tomorrow and make sure the unit tests pass. But this bug is a blocker, so want to get this in as soon as possible.

@macumber Can you review this?

### DIFF
--- a/openstudiocore/src/utilities/filetypes/EpwFile.cpp
+++ b/openstudiocore/src/utilities/filetypes/EpwFile.cpp
@@ -2511,55 +2511,60 @@ bool EpwFile::parse(bool storeData)
 
 bool EpwFile::parseLocation(const std::string& line)
 {
-  bool result = true;
+  bool result = false;
 
   // LOCATION,Chicago Ohare Intl Ap,IL,USA,TMY3,725300,41.98,-87.92,-6.0,201.0
-  std::regex locationRegex("^LOCATION,(.*?),(.*?),(.*?),(.*?),(.*?),(.*?),(.*?),(.*?),([^,]*).*?$");
-  std::smatch matches;
-  if (std::regex_search(line, matches, locationRegex)){
-    std::string city = std::string(matches[1].first, matches[1].second); boost::trim(city);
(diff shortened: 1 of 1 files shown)


mbadams5:
@jasondegraw I can almost guarantee [this is the issue](http://stackoverflow.com/a/12665408/2358662). Basically gcc didn't implement \<regex\> until 4.9.0 even though it is in the library and you can call it in < 4.9.0...

Myoldmopar:
@jasondegraw @macumber @brianlball 

Built this branch on Ubuntu 14.04 and it works:

```
irb(main):002:0> OpenStudio::EpwFile.new('/tmp/USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw')
[openstudio.EpwFile] <0> Successive data points (1986-Jan-31 to 1977-Feb-01, ending on line 753) are greater than 1 day apart in EPW file '/tmp/USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw'. Data will be treated as typical (TMY)
=> #<OpenStudio::EpwFile:0x00000002677620 @__swigtype__="_p_openstudio__EpwFile">
```

jasondegraw:
@axelstudios @macumber I've tested this on Windows and Linux (gcc and clang) and it works on all three where previously both Linux versions failed. It is ready to merge once it is reviewed.

@evanweaver @asparke2 @brianlball FYI

axelstudios:
Looks good to me.

Issue #1755, opened 2015-08-06
OpenStudio App crash on user file


macumber:
Hello,

This file is in reference to the following post:

https://unmethours.com/question/9983/any-experience-with-the-following-openstudio-crash-an-unhandled-exception-has-occurred-access-violation-no-rtti-data/

I was assigning spaces to thermal zones and received the error when I went into OS and attempted to access the Facilities tab.  There are several simple images attached that I hope will describe the space types and zoning.  Note that the thermal zoning is incomplete at this point.

The v3-12 (3) *.osm file does not trigger the error message is OS.  The v3-12 (4) *.osm file does trigger the error in OS.

Please let me know what additional information is required and I'll happily provide it!

Many Thanks,
Nick

macumber:
The call stack for this crash in OS 1.8.0 is the same as in #1763

kbenne:
I repaired Nick's model and sent it back. The underlying issue is similar to https://github.com/NREL/OpenStudio/issues/1667 but almost the opposite. There we were seeing zones that did not get initialized. Here it is as if the initializers are over zealous and being called an extra time on a phantom zone. 

@jmarrec's [description](https://unmethours.com/question/9983/any-experience-with-the-following-openstudio-crash-an-unhandled-exception-has-occurred-access-violation-no-rtti-data/) of the repair is very accurate.

kbenne:
@macumber and @DavidGoldwasser are we good to close?

Issue #1763, opened 2015-08-12
An unhandled exception has occurred. Access violation - no RTTI data


jmarrec:
I'm using Openstudio 1.8.0 build 97b69c8a8d.

I am getting this exception repeatedly, similar to #1755

![2015-08-12 12_08_12-unhandled exception](https://cloud.githubusercontent.com/assets/5479063/9221470/dc164d44-40ea-11e5-94da-d36959ce4e8e.jpg)

I tried pretty much everything: delete the folder that goes with the osm file, trying to copy out the code to paste it in another file that had a working folder, etc.

The error is coming when clicking on any of the following tabs (the rest works fine)
- Facility
- Thermal Zone
- Measures

At some point, I clicked on Measure and got the following error message (having pointers as to what's causing the problem sure is helpful...)

![error](https://cloud.githubusercontent.com/assets/5479063/9219430/2cf1962a-40db-11e5-81c4-11d4d86f4312.jpg)

I did find an OS:Sizing:Zone without a zone, so I deleted it out. But the problem persists.

I tried the diagnostic script, there are no problem left, but yet I keep getting the RTTI error.

is there a way to display more information when getting the RTTI data error?

How can I fix this?

Additional info: I emailed the file to openstudio.at.nrel.gov.
I don't have any measures or scripts in there.

jmarrec:
Similar to #1755

macumber:
Added user file from email

jmarrec:
Any idea if someone will be able to take a look at it soon and if it's usually a fixable problem? (As with most problems, this one happens while I'm on an impossible deadline, so I just want to know if I should pretty much restart from scratch or not.)

macumber:
This will be on our fix list for 1.9.0.  let me spend a few minutes and see if I can find a workaround for you now

jmarrec:
Thanks!

macumber:
The call stack for this crash in OS 1.8.0 is:

    msvcr120d.dll!_wassert(const wchar_t * expr, const wchar_t * filename, unsigned int lineno) Line 369	C
    openstudio_model.dll!boost::assertion_failed(const char * expr, const char * function, const char * file, long line) Line 51	C++
    openstudio_model.dll!openstudio::model::detail::ZoneHVACEquipmentList_Impl::thermalZone() Line 252	C++
    openstudio_model.dll!openstudio::model::ZoneHVACEquipmentList::thermalZone() Line 399	C++
[...]

PR #1768, opened 2015-08-14
Update DiagnosticScript to remove cruft from badly initialized zones

macumber:
Add catch if object is passed to check_model_object but not yet available in workspace (I think this was the underlying issue but have not been able to reproduce), checking if error occurs during ThermalZone initialization

Update thermal zone methods for performance (also works around this crash)


macumber:
@kbenne can you give this second version a review?

macumber:
Fixes #1755 and #1763

macumber:
I just noticed that the ThermalZone diff includes all that white space once again!

macumber:
@kbenne don't merge this yet, I will remove the + button in inspector

PR #1769, opened 2015-08-14
Plant operation schemes

kbenne:
https://github.com/NREL/OpenStudio-resources/blob/NewTestsForEvan/model/simulationtests/plant_op_schemes.rb#L29

https://github.com/NREL/OpenStudio/blob/PlantOperationSchemes/openstudiocore/src/model/PlantLoop.hpp#L116

https://github.com/NREL/OpenStudio/blob/PlantOperationSchemes/openstudiocore/src/model/PlantEquipmentOperationRangeBasedScheme.hpp#L35


kbenne:
@asparke2 I'm going to merge this into develop. I still want and need your review, but this has a version translation so I don't want the branch to get stale or it will cause me more work. Plus I have a mountain of  things to review so I want to start integrating it all together.

asparke2:
sounds good.

PR #1782, opened 2015-08-19
Zone ventilation design flow rate

kbenne:


DavidGoldwasser:
@kbenne looks like forward translation isn't making ZoneVentilationDesignFlowRate objects in IDF.
8-20-2015 10-20-30 am

In the OSM the zone ventilation is being added to a ZoneHVAC Equipment List object.

PR #1785, opened 2015-08-20
Zone mixing3

macumber:
@kbenne previously delta temperature and temperature units were evaluating to equal.  This fixes that, will result in more strict checking.  Not sure if we want that or not


macumber:
@kbenne looks like there are some new unit test failures on this one, I'll look into those

macumber:
@kbenne I'd advise we not pull this in to the 1.8.4 release build

PR #1789, opened 2015-08-20
98318570 remove radiance checkbox

macumber:


macumber:
@kbenne can you check this branch to make sure it does not include any problematic commits on it?  I resolved several conflicts including some EOL issues. My git environment is set up to check out windows style and check in unix style so I am not sure how this is even happening.

Issue #1730, opened 2015-07-20
ModelObjects not translated to E+


macumber:
Several model objects are not being translated to E+, https://github.com/NREL/OpenStudio/commit/2ccd1315eeb05d0a9eddc0277842fc12b18b6ff0.  I have not done a full audit, suspect there may be others.

DavidGoldwasser:
We can constrain this to just simulation settings for now, but we need to do a full audit that every model object is being translated.

macumber:
The following objects do not have translate functions in the ForwardTranslator, emphasis on ones that are a problem (@kbenne @mbadams5 can you edit to include HVAC ones that are an issue?):

AirTerminalDualDuctVAV
AirTerminalSingleDuctConstantVolumeCooledBeam
BuildingStory - ok, no E+ counterpart
ClimateZones - ok, no E+ counterpart
CoilCoolingCooledBeam
CoilCoolingDXMultiSpeedStageData
CoilCoolingLowTempRadiantConstFlow
CoilCoolingLowTempRadiantVarFlow
CoilCoolingDXVariableSpeedSpeedData
CoilCoolingWaterToAirHeatPumpVariableSpeedEquationFitSpeedData
CoilHeatingDXMultiSpeedStageData
CoilHeatingGasMultiStageStageData
CoilHeatingLowTempRadiantConstFlow
CoilHeatingLowTempRadiantVarFlow
CoilHeatingDXVariableSpeedSpeedData
CoilHeatingWaterToAirHeatPumpVariableSpeedEquationFitSpeedData
CoilHeatingWaterBaseboard
*ComponentCostAdjustments*
ComponentData - ok, no E+ counterpart
Connection
ConnectorMixer
ConnectorSplitter
DefaultConstructionSet - ok, included in other translation
DefaultScheduleSet - ok, included in other translation
DefaultSubSurfaceConstructions - ok, included in other translation
DefaultSurfaceConstructions - ok, included in other translation
*DesignSpecificationZoneAirDistribution*
ElectricEquipmentDefinition - ok, included in other translation
ExteriorLightsDefinition - ok, included in other translation
Facility - ok, no E+ counterpart
GasEquipmentDefinition - ok, included in other translation
GenericModelObject - ok, no E+ counterpart
*GlareSensor*
HotWaterEquipmentDefinition - ok, included in other translation
InternalMassDefinition - ok, included in other translation
*LifeCycleCostUsePriceEscalation*
[...]

PR #1796, opened 2015-08-21
Added case statements for missing IDD enums, added warnings for objec…

macumber:
…ts that are not being translated but should be

Addresses #1730


macumber:
@DavidGoldwasser I think this is good enough for now, we will want to implement all these translations in the future

@kbenne @evanweaver @mbadams5 FYI

kbenne:
please don't merge this yet. I'd like to review and discuss

macumber:
@kbenne ok to merge this?

kbenne:
yes

Issue #1601, opened 2015-04-14
PAT ruby script path not set


macumber:
User reported this issue: https://unmethours.com/question/4404/primary-ruby-script-path-not-set-for-passed-in-measure/

macumber:
Link to error message in code https://github.com/NREL/OpenStudio/blob/3aaa0984dbc185867261bde6305b141deb118d5e/openstudiocore/src/runmanager/lib/RubyJobUtils.cpp#L1212

DavidGoldwasser:
Closed with 1601.

Issue #1617, opened 2015-04-16
Repeatable crash when loading specific seed model


DavidGoldwasser:
I'm using develop build. The seed model doesn't have any measures in it, and neither does the PAT project. It is is a freshly made PAT project.

1. Launch PAT and make new project
2. choose "0416_lab_a.osm" as seed

Should crash with error about "Passed in measure does not have primary RubyScript Path set, which is required for RubyJobBuilder. It is an odd message since no measures are involved yet.

![4-16-2015 8-35-29 am](https://cloud.githubusercontent.com/assets/4857055/7183405/8052f170-e414-11e4-8434-21734e56775d.png)

I tried scanning for tools, that didn't fix this. Next I'll try throwing away the database with the osm file. The osm file does open up fine in OpenStudio app.

Here is link to seed file I'm trying to load
https://github.com/NREL/OpenStudio-files/tree/master/Git%20Issue%20Files/Issue1617

DavidGoldwasser:
Throwing away the database didn't make a difference. Neither did doing a simulation run in the OSM file (which ran successfully) and then saving the OSM.

I did load another OSM as a seed model to confirm that this doesn't always happen.

macumber:
Great, seems related to https://github.com/NREL/OpenStudio/issues/1601

DavidGoldwasser:
Also tried changing "My Measures" directory and throwing away my "BCL" directory. Neither of those did anything.

macumber:
I think this is a duplicate of https://github.com/NREL/OpenStudio/issues/1617

DavidGoldwasser:
I'll confirm if this still happens and assign to @macumber if it does

DavidGoldwasser:
@macumber so I confirmed this happens both in 1.8.1 installed version of OpenStudio and in my current development build (from earlier today). Throwing away everything but the OSM doesn't change what happens.

macumber:
Last update on this thread had a similar issue, failure when no user defined measures in the workflow:

https://unmethours.com/question/4404/primary-ruby-script-path-not-set-for-passed-in-measure/?comment=10483#comment-10483

User's osm had utility bills in it so it is trying to use calibration measure, maybe your test model had this too @DavidGoldwasser ?

PR #1799, opened 2015-08-21
101698856 issue1601

macumber:


macumber:
Fixes #1601 and #1617

PR #1800, opened 2015-08-22
Fix python binding issue

### LINKED ISSUE #1786: Issue with Python bindings for v1.8.0
I get an import error:

```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
---> 12 import openstudio

/home/ehale/Projects/openstudio/openstudiocore/python/openstudio.py in <module>()
     15 from openstudioutilities import *
     16 
---> 17 import openstudioenergyplus as energyplus
     18 
     19 import openstudiomodel as model

/home/ehale/Projects/openstudio/build-release/OSCore-prefix/src/OSCore-build/Products/python/openstudioenergyplus.py in <module>()
    179 import openstudioutilitiesidf
    180 import openstudioutilitiesfiletypes
--> 181 import openstudiomodel
    182 class ErrorLevel(_object):
    183     __swig_setmethods__ = {}

/home/ehale/Projects/openstudio/build-release/OSCore-prefix/src/OSCore-build/Products/python/openstudiomodel.py in <module>()
     98 
     99 
--> 100 from openstudiomodelcore import *
    101 from openstudiomodelsimulation import *
    102 from openstudiomodelresources import *

/home/ehale/Projects/openstudio/build-release/OSCore-prefix/src/OSCore-build/Products/python/openstudiomodelcore.py in <module>()
   1526 
   1527 
-> 1528 def compareInputAndOutput(object, attributeName, inputResult, outputResult, tol, logLevel=Debug):
   1529     return _openstudiomodelcore.compareInputAndOutput(object, attributeName, inputResult, outputResult, tol, logLevel)
   1530 compareInputAndOutput = _openstudiomodelcore.compareInputAndOutput

NameError: name 'Debug' is not defined
```

Looking around at .i files, etc., I don't see any recent (last 6 month) changes that should have caused this, nonetheless, there it is. There is a workaround that I will use for now: adding 

```
#if defined SWIGPYTHON
  %ignore openstudio::model::compareInputAndOutput;
#endif
```

to `src/model/ModelCore.i`.

mbadams5:
Fixes #1786 

@elainethale This fix works for me when I tested it locally. Can you test this and see if it fixes the issue for you as well? I think this is a better solution for that issue instead of suppressing the function, but either approach is fine.


elainethale:
I agree that that looks better. My C++ is so rusty, I had some vague recollection that something like that should be possible but the syntax didn't jump immediately to mind. I'll try it out.

elainethale:
I tried it, and it works. Let's go with this one and not mine. Thanks, @mbadams5! @macumber, should I go ahead and merge?

elainethale:
@macumber, the build failure for i386 on Windows seems to be abnormal ... I tried hitting the 'delete results and rerun' button on the details page, but that didn't work for me. Suggestions?

macumber:
Hey @elainethale that line is a little suspect to me:

https://github.com/NREL/OpenStudio/blob/7647ba8149a0f39244a3492a950b80c371f826ae/openstudiocore/src/model/AirLoopHVAC.cpp#L480

I'm not sure auto is correct in that case, @lefticus or someone else might know.  It looks like it uses unsigned in 1.8.4:

https://github.com/NREL/OpenStudio/blob/v1.8.4/openstudiocore/src/model/AirLoopHVAC.cpp#L480

I'd merge develop into this branch and see if that fixes this.

lefticus:
Correct - auto doesn't work well in that kind of a loop. `size_t` is the most correct (and still typeable) answer. The "real" answer is `std::vector<DemandInet>::size_type` or maybe `decltype(demandInletNodes)::size_type`.

Using `unsigned` just leads to more pain with the plethora of 64bit warnings that MSVC gives (and it is absolutely correct to warn about it).

kbenne:
@lefticus I find your use of the word "typeable" confusing. Was that a double entendre?

lefticus:
@kbenne How's "size_t is the only not un-succinct option"?

kbenne:
k

lefticus:
It's not guaranteed that `std::vector<...>::size_type` be `size_t` but it is with every compiler that I know of.

kbenne:
You know i never actually looked at the line in question until just now. 

auto was bad cause it was no doubt auto defining type based on the literal 0 used to initialize i. 

I wrote that then I believed Evan changed it to unsigned instead of auto. I was thinking when I saw Evan's change that it really should be size_t but I didn't comment and here is why.

size_t is great at fixing up the loop, but then that index is used as an argument into the demandInletNodes(unsigned ) api. So using size_t might be inaccurate too because the compiler has to go from size_t to unsigned on the api call. In any case I didn't say anything because I was busy and was trying to avoid unraveling this whole conversation cause we really only expect size() to be 2 at most so it probably doesn't matter a whole lot.

lefticus:
Do what @kbenne said then. But auto is the wrong answer - yes. That will warn on every platform.

kbenne:
I don't know maybe the api is just ill conceived. But we have it so...

I think the most correct answer is some variation of

```
    auto demandInletNodes = t_airLoopHVAC->demandInletNodes();
    auto plenumIndex = 0;
    for( size_t i = 0; i < demandInletNodes.size(); ++i ) {
      thermalZone.removeSupplyPlenum(plenumIndex);
      ++plenumIndex
    }
```

See why i didn't say anything :)

lefticus:
```
auto plenumIndex = 0;
```

Still initializes to a signed integer 0.

You could do:

```
auto plenumIndex = 0u; // I think
```

But since you really only just want a count of items to remove:

```
auto plenumIndex = 0u;
for( const auto & : t_airLoopHVAC->demandInletNodes()) {
  thermalZone.removeSupplyPlenum(plenumIndex);
  ++plenumIndex
}
```

We could do this all day.

kbenne:
we have a winner :trophy:

elainethale:
Ok, people. I have no idea how this turned into an esoteric discussion of a line in AirLoopHVAC.cpp, and my c++ is too rusty to know about 'auto'. Oh well, it is nice to see a @lefticus and @kbenne back and forth in any case ... So I take it from @macumber's comment that mostly I should merge develop in and hope this little issue goes away. On it.

brianlball:
unsubscribe

elainethale:
Alright, that seemed to clear things up :) Merging ...

Issue #1522, opened 2015-03-16
Issues with Workspace::swap


macumber:

macumber:
Several people have reported issues with Workspace::swap, including apparent differences when using measures on the server and with the app. 

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/idf/Workspace.cpp#L145

We should consider deprecating this function if it does not work correctly, a work around is:

    # remove existing objects
    handles = OpenStudio::UUIDVector.new
    workspace.objects.each do |obj|
      handles << obj.handle
    end
    workspace.removeObjects(handles)
    
    # add new file
    workspace.addObjects(new_workspace.toIdfFile.objects)

macumber:
There is also a Workspace::swap method that takes a single object.  However, this doesn't seem like it uses the clone and remove methods so I think we might want to remove this from the public interface as well

https://github.com/NREL/OpenStudio/blob/3aaa0984dbc185867261bde6305b141deb118d5e/openstudiocore/src/utilities/idf/Workspace.cpp#L2885

macumber:
@kbenne @asparke2 @nllong @phylroy FYI

PR #1802, opened 2015-08-25
Fixes #937.  I've fixed this issue for COFEE and Phylroy, I thought t…

### LINKED ISSUE #937: Create From External File + facility.exteriorLights crashes ruby
If you have a model and run the "AddExteriorLights" measure, the measure works fine.  However, if you create a PAT project, load this same model as a design alternative using the "Create From External File" feature, and have "AddExteriorLights" as an always-run measure, ruby crashes on the last line of this code inside the "AddExteriorLights" measure:
```ruby
    facility = model.getFacility
    starting_exterior_lights_power = 0
    starting_exterior_lights = facility.exteriorLights
```
This was reproduced with 1.2.4 and 1.2.5
@elainethale Perhaps something to do with how "ReplaceModel" is copying the model?

macumber:
…he fix had made its way back to OpenStudio.  Basically, Workspace::swap does not work (#1522), using remove objects and then add objects is the best way I know of.

### DIFF
--- a/openstudiocore/src/pat_app/Measures/ReplaceModel/measure.rb
+++ b/openstudiocore/src/pat_app/Measures/ReplaceModel/measure.rb
@@ -107,7 +107,8 @@ class ReplaceModel < OpenStudio::Ruleset::ModelUserScript
     end
     model.removeObjects(handlesToRemove)
     
-    model.addObjects(newModel.objects)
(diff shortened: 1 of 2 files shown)


macumber:
@asparke2 @DavidGoldwasser this change should get applied to the measures repo once it is verified

PR #1811, opened 2015-08-27
1810 osversionfail

kbenne:


openstudiocore/src/osversion/VersionTranslator.cpp
```diff
@@ -2638,18 +2639,15 @@ std::string VersionTranslator::update_1_8_3_to_1_8_4(const IdfFile& idf_1_8_3, c
       m_refactored.push_back( std::pair<IdfObject,IdfObject>(object,newObject) );
       ss << newObject;
     } else if(iddname == "OS:AvailabilityManager:Scheduled") {
-      m_untranslated.push_back(object);
-      continue;
+      m_deprecated.push_back(object);
```
macumber: @kbenne did the test pass after these changes?  I agree that the tests were rubbish but they were passing :-/

kbenne: No I probably should have made two issues. The above fixes 1955 - OSVersionFixture.VersionTranslator_ExampleModel_1_1 (Failed) a legitimate thing.

The rubbish default tests are failing because of [this assertion](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/osversion/VersionTranslator.cpp#L2604) which depends on a required relationship to another object. Naturally these required relationships are not established in the default test methodology so the assertion fails. What is surprising is that we went this far without uncovering a problem with this tests. So it is.

openstudiocore/src/osversion/VersionTranslator.cpp
```diff
@@ -2550,6 +2550,7 @@ std::string VersionTranslator::update_1_7_4_to_1_7_5(const IdfFile& idf_1_7_4, c
 
 std::string VersionTranslator::update_1_8_3_to_1_8_4(const IdfFile& idf_1_8_3, const IddFileAndFactoryWrapper& idd_1_8_4)
 {
+  LOG(Error,"Here is the model" << idf_1_8_3);
```
macumber: @kbenne this seems like it is not supposed to be here?  Was this just a mistake?

openstudiocore/src/osversion/test/VersionTranslator_GTest.cpp
```diff
@@ -228,100 +228,6 @@ TEST_F(OSVersionFixture, VersionTranslator_ExampleComponent_1_5) {
...
-  osversion::VersionTranslator translator;
-
-  // iterate through osversion subfolders
-  openstudio::path resources = resourcesPath() / toPath("osversion");
-  for (boost::filesystem::directory_iterator it(resources); it != boost::filesystem::directory_iterator(); ++it) {
-    if (boost::filesystem::is_directory(it->status())) {
-
-      QString stem = toQString(it->path().stem()).replace("_", ".");
-      VersionString vs(toString(stem));
-      if (vs.major() == major && vs.minor() == minor && vs.patch() == 3){
```
macumber: @kbenne what was up with the 'vs.patch() == 3' test?

kbenne: Uhhh??? No clue, I didn't really critique the stuff I was removing because it was my estimation that the very idea of doing it was bad. But if I had to guess some of the stuff WS needed was only available in a patch. Then maybe later (or at the same time but overlooked) this function was made generic to handle other major minor?

openstudiocore/src/osversion/VersionTranslator.cpp
kbenne: yes a mistake. I was doing some testing to figure out what the heck was going on.

kbenne: FYI idf_1_8_3 is a model with every object in it, with a handle but every other field blank. At that point I thought I was losing my mind.

macumber: Can you remove that line before I merge this?  Otherwise I think we'd have some pretty excessive logging going on that could slow things down.

openstudiocore/src/osversion/test/VersionTranslator_GTest.cpp
macumber: No prob, just trying to figure out if there was more to this test than meets the eye.

openstudiocore/src/osversion/VersionTranslator.cpp
kbenne: hehe. Yes of course. And it would be excessive.

openstudiocore/src/osversion/VersionTranslator.cpp
kbenne: hehe. Yes of course. And it would be excessive.

kbenne:
@macumber back to you.

PR #1821, opened 2015-08-31
Improve packaging

### LINKED ISSUE #1496: Installer puts 64 bit E+ in 32 bit OS installer.
Probably need to reconfigure the build machine and watch out for this in the future, as there is no programatic catch for this in the cmake finder.  On that note consider adding guards against this in the finder.

https://unmethours.com/question/3110/openstudio-160-unable-to-run-energyplus-82/

### LINKED ISSUE #1746: OpenStudio 1.8.0 does not work with installed E+
Users have reported they have to delete the copy installed with OS and install E+ 8.3 on their own:

https://unmethours.com/question/9286/openstudio-180-unable-to-run-energyplus-83/

### LINKED ISSUE #1815: Install Radiance with OpenStudio
We get the message "Radiance was found but it is not the NREL version" when we install radiance with OpenStudio

axelstudios:
Resolves #1496, resolves #1746, resolves #1815

This PR removes the need for developers to install EnergyPlus or Radiance - the correct versions (for the current platform/architecture) will be downloaded automatically via CMake.

Packaging EnergyPlus and Radiance is now mandatory and effortless.


openstudiocore/CMake/FindEnergyPlus.cmake
```diff
@@ -50,9 +50,9 @@ else()
   file(GLOB ENERGYPLUS_POSSIBLE_PATHS "/usr/local/[eE]nergy[pP]lus?*")
 endif()
 
-# sort possible paths in reverse order
-list(APPEND ENERGYPLUS_POSSIBLE_PATHS $ENV{ENERGYPLUSDIR})
+# sort possible paths in reverse order, favoring the ENERGYPLUSDIR environment variable
```
macumber: Do we ever need this finder any more?  Can you use this to create an OpenStudio build with a custom E+ build?

openstudiocore/CMakeLists.txt
```diff
@@ -1045,30 +1041,22 @@ if(NOT UNIX)
...
-
-  if(APPLE)
-    install(DIRECTORY "${RADIANCE_LOCATION}" DESTINATION "sharedresources" USE_SOURCE_PERMISSIONS)
-  elseif(MSVC)
-    install(DIRECTORY "${RADIANCE_LOCATION}" DESTINATION "share/openstudio" USE_SOURCE_PERMISSIONS)
-  else()
-    install(DIRECTORY "${RADIANCE_LOCATION}" DESTINATION "share/openstudio-${OPENSTUDIO_VERSION}" USE_SOURCE_PERMISSIONS)
-  endif()
+# Package Radiance
+set(RADIANCE_LOCATION "" CACHE PATH "Location of Radiance to Package")
```
macumber: @axelstudios can't RADIANCE_LOCATION be set here since we are downloading this file?

openstudiocore/CMake/FindEnergyPlus.cmake
macumber: The ability to use a custom E+ build might be needed on alternative platforms right?

axelstudios: Is that something people do?  OpenStudio is currently locked to the EnergyPlus SHA, so someone would have to change that too.

openstudiocore/CMakeLists.txt
axelstudios: Nah, it's the same as `ENERGYPLUS_DIR`.  All of the logic is handled at the top level, so that variable is already set in the CMake cache by the time it gets to this line.

macumber:
@axelstudios I am going to merge this branch in to https://github.com/NREL/OpenStudio/compare/85723528_Issue1815 and include in https://github.com/NREL/OpenStudio/pull/1822

openstudiocore/CMake/FindEnergyPlus.cmake
macumber: @axelstudios the second use case is that you want to build on Peregrine and there is no package available for you to point to so you have to build E+ (and now Radiance) yourself.  So the SHA would be the same, just different build.

PR #1822, opened 2015-09-01
Adding methods to find radiance install [#85723528]

macumber:


macumber:
After talking to @axelstudios and @evanweaver I am going to merge this so we can get it in the 1.8.5 packages

PR #1837, opened 2015-09-03
Fix Failing ScheduleVariableInterval Unit Tests

jasondegraw:
There were actually two related problems causing the failures.

* ScheduleVariableInterval isn't storing enough information to set the start time of a TimeSeries object. The fix for that is to grotesquely assume that the schedule starts at the start of the day. 
* The schedules in the unit tests were not being made correctly and did not start at the start of a day. Correcting the construction errors (the first report times were at the start of a day) fixes the failures. 

Since the original time series in the tests are perfectly legitimate, I have renamed the original tests and kept them. I'll submit an issue to address the poor handling of time series that don't start at the start of a day. 

@macumber I also fixed some formatting errors that I ran across

### DIFF
--- a/openstudiocore/src/energyplus/Test/ScheduleInterval_GTest.cpp
+++ b/openstudiocore/src/energyplus/Test/ScheduleInterval_GTest.cpp
@@ -48,18 +48,21 @@ using namespace openstudio;
 
 TEST_F(EnergyPlusFixture,ForwardTranslator_ScheduleFixedInterval_Hourly)
 {
+  // Create the values vector
   Vector values = linspace(1, 8760, 8760);
 
+  // Create an hourly interval time series that begins on 1/1 00:00
   TimeSeries timeseries(DateTime(Date(MonthOfYear::Jan, 1), Time(0,1,0)), Time(0,1,0), values, "");
 
   Model model;
 
+  // Create a schedule and make sure that it worked
   boost::optional<ScheduleInterval> scheduleInterval = ScheduleInterval::fromTimeSeries(timeseries, model);
   ASSERT_TRUE(scheduleInterval);
   EXPECT_TRUE(scheduleInterval->optionalCast<ScheduleFixedInterval>());
 
+  // Forward translate the schedule
   ForwardTranslator ft;
-
   Workspace workspace = ft.translateModel(model);
 
   std::vector<WorkspaceObject> objects = workspace.getObjectsByType(IddObjectType::Schedule_Compact);
@@ -68,8 +71,10 @@ TEST_F(EnergyPlusFixture,ForwardTranslator_ScheduleFixedInterval_Hourly)
   boost::regex throughRegex("^Through:\\s*(.*)/\\s*(.*)\\s*");
   boost::regex untilRegex("^Until:\\s*(.*):(.*)\\s*");
 
-  workspace.save(toPath("./ForwardTranslator_ScheduleFixedInterval_Hourly.idf"), true);
+  // Write out the schedule - keep this around for now
+  //workspace.save(toPath("./ForwardTranslator_ScheduleFixedInterval_Hourly.idf"), true);
 
+  // Check the contents of the output
   unsigned N = objects[0].numFields();
   boost::optional<Date> lastDateThrough;
   bool until24Found = false;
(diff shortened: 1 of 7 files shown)


openstudiocore/src/energyplus/Test/ScheduleInterval_GTest.cpp
```diff
@@ -144,9 +147,112 @@ TEST_F(EnergyPlusFixture,ForwardTranslator_ScheduleFixedInterval_Hourly)
...
+  TimeSeries timeseries(DateTime(Date(MonthOfYear::Jan, 1)), Time(0, 1, 0), values, "");
+
+  Model model;
+
+  boost::optional<ScheduleInterval> scheduleInterval = ScheduleInterval::fromTimeSeries(timeseries, model);
+  ASSERT_TRUE(scheduleInterval);
+  EXPECT_TRUE(scheduleInterval->optionalCast<ScheduleFixedInterval>());
+
+  TimeSeries ts = scheduleInterval->timeSeries();
+  EXPECT_EQ(DateTime(Date(MonthOfYear::Jan, 1)), ts.firstReportDateTime());
```
macumber: Shouldn't this be:

    EXPECT_EQ(DateTime(Date(MonthOfYear::Jan, 1), Time(0,1,0)), ts.firstReportDateTime());

The first interval starts at time 0 but ends at interval length, right?
https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/data/TimeSeries.hpp#L157

jasondegraw: @macumber No, the time series was created with a first report date time of 1/1 00:00 and a 1 hour interval, so the time series starts at 12/31 23:00. The constructor that is getting used is this one:

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/data/TimeSeries.hpp#L161

Since the time series isn't translated properly into the schedule object, it can't return the right time series. I'm not sure what the right time series is, but I don't think it is this one.

macumber: Oh cool, I see.  I wonder if we should remove that constructor with the DateTime since it isnt handled well :-/

macumber: Maybe make the test more explicit though:

    EXPECT_EQ(DateTime(Date(MonthOfYear::Jan, 1), Time(0,0,0)), ts.firstReportDateTime());

macumber: And we probably need to comment these tests better, especially since it will be us getting confused the next time these break

jasondegraw: @macumber I don't think we should remove that ctor, since the problem is really the schedule. Your other suggestions are good, though. I'll add in some comments and make the test more explicit.

macumber:
I'm getting test failures in:

    EnergyPlusFixture, ForwardTranslator_ScheduleFixedInterval_Hourly_Shifted
    EnergyPlusFixture, ForwardTranslator_ScheduleVariableInterval_500_Shifted

are these expected?  Do we know what the right answers are or should we punt on these tests until we tackle the IDF schedule for timeseries that don't start at the beginning of the day problem?

jasondegraw:
@macumber Yes, those are expected, but there should be three:

    EnergyPlusFixture.ForwardTranslator_ScheduleFixedInterval_Hourly_Shifted
    EnergyPlusFixture.ForwardTranslator_ScheduleVariableInterval_Hourly_Shifted
    EnergyPlusFixture.ForwardTranslator_ScheduleVariableInterval_500_Shifted

I don't think we know what the right answers are yet, so this should be addressed as part of the TimeSeries revamp. In my opinion.

openstudiocore/src/model/ScheduleVariableInterval.cpp
```diff
@@ -176,7 +176,18 @@ namespace detail {
...
+    // Grotesquely assume that the schedule starts at the start of the first day
+    OptionalInt startHour = 0;
+    OptionalInt startMinute = 0;
+    OS_ASSERT(startMonth);
+    OS_ASSERT(startDay);
+    OS_ASSERT(startHour);
+    OS_ASSERT(startMinute);
+
     DateTimeVector dateTimes;
+    dateTimes.push_back(DateTime(Date(MonthOfYear(*startMonth), *startDay), Time(0, *startHour, *startMinute)));
```
macumber: @jasondegraw if the timeseries out has the same convention of reporting at end of intervals and not reporting the start time then is this the right thing to do?  What would happen if you kept converting back and forth from timeseries to schedule?  I'll add a test for that.

macumber: @jasondegraw I added some unit tests around this, but I'm still not sure how this is the right thing to do.  Can you explain it to me tomorrow?

PR #1851, opened 2015-09-10
102941444 correct bsd fs

macumber:
@rpg777 all of the content I would want to merge for 1.9.0 is in here now, however it needs a lot of testing before merging.  We should run that unit test with each type of shading controls (or no controls) and make sure it is doing the right thing


openstudiocore/src/pat_app/Measures/RadianceMeasure/measure.rb
```diff
@@ -382,7 +382,7 @@ def calculateDaylightCoeffecients(radPath, sim_cores, t_catCommand, options_treg
 				
 				rad_command = ""
 				
-				if wg[2].to_i == 0	# uncontrolled windows
+				if wg[2] == 'n/a' || wg[2] == 'AlwaysOff'	# uncontrolled windows
```
macumber: @rpg777 is this one of the lines causing trouble?

openstudiocore/src/pat_app/Measures/RadianceMeasure/measure.rb
```diff
@@ -492,7 +492,7 @@ def runSimulation(t_space_names_to_calculate, t_sqlFile, t_simCores, t_options_s
 				wg = row.split(",")[0]
 		
 				# do uncontrolled windows (WG0)
-				if wg[2].to_i == 0
+				if wg[2] == "n/a" || wg[2] == "AlwaysOff"
```
macumber: @rpg777 or this one?

openstudiocore/src/pat_app/Measures/RadianceMeasure/measure.rb
rpg777: It's both. Problem resolved in an upcoming commit.

PR #1872, opened 2015-09-17
Photovoltaics

macumber:
Creating pull request to start getting review comments


openstudiocore/src/model/PlanarSurface.hpp
```diff
@@ -218,6 +219,8 @@ class MODEL_API PlanarSurface : public ParentObject {
   /// Returns any solar hot water collectors associated with this surface.
   std::vector<ModelObject> solarCollectors() const;
 
+  /// Returns any generator photovoltaics associated with this surface.
+  std::vector<GeneratorPhotovoltaic> generatorPhotovoltaics() const;
```
macumber: Currently I am allowing multiple photovoltaic generators to reference a surface, I am not sure if this is supported by EnergyPlus or if it is a good idea.

openstudiocore/src/model/ElectricLoadCenterInverterLookUpTable.hpp
```diff
@@ -0,0 +1,184 @@
...
+  //@}
+
+  static IddObjectType iddObjectType();
+
+  /** @name Getters */
+  //@{
+
+  boost::optional<Schedule> availabilitySchedule() const;
+
+  // DLM: TODO the thermalZone methods should be moved to the Inverter base class
```
macumber: @evanweaver did you see the todo's in these headers?

openstudiocore/src/model/PlanarSurface.hpp
macumber: This is what is suggested by the data model and doing anything else would take a lot of model logic to enforce and introduce potential problems, thoughts?

openstudiocore/src/model/SolarCollectorIntegralCollectorStorage.cpp
```diff
@@ -189,17 +189,10 @@ namespace detail {
...
 
-    if (surface.isAirWall()){
-      return false;
-    }
-
     // DLM: check for existing solar collectors or photovoltaic generators?
 
     if (surface.optionalCast<Surface>()){
-      Surface s = surface.cast<Surface>();
-      if (istringEqual("SunExposed", s.sunExposure())){
```
macumber: I had been enforcing that solar collectors could only go on sun exposed surfaces.  However, it seemed like that might be too strict plus would require a lot of observer stuff to happen to ensure that the model never became invalid.  In the end I went for looser rules, we can always throw more warnings on translation.

openstudiocore/src/model/GeneratorPhotovoltaic.hpp
```diff
@@ -0,0 +1,143 @@
...
+} // detail
+
+/** GeneratorPhotovoltaic is a Generator that wraps the OpenStudio IDD object 'OS:Generator:Photovoltaic'. */
+class MODEL_API GeneratorPhotovoltaic : public Generator {
+ public:
+  /** @name Constructors and Destructors */
+  //@{
+
+  /// Creates a GeneratorPhotovoltaic object with PhotovoltaicPerformanceSimple performance
+  static GeneratorPhotovoltaic simple(const Model& model);
```
macumber: These factory methods are a little different pattern than other objects, they construct a required performance object of the right type.  Curious what @kbenne and others think of this?

mbadams5: @macumber Why not just create the required object in the GeneratorPhotovoltaic constructor? That is usually what we've/i've done with required objects. This is also similar to Curves for Coils. You can default construct the required object or pass it in through an overload.

macumber: The difference is that there are several types of performance objects so you'd have to indicate to the constructor which one to make.

macumber: And I didn't want the constructors for the pv performance objects to be public, I think that is the different from coil ctors that can take different curves.

macumber:
@DavidGoldwasser I think we'll need you to update the StandardReport measure to display energy produced by PV

developer/ruby/test/ElectricLoadCenterInverterLookUpTable_GTest.cpp
```diff
@@ -0,0 +1,29 @@
+/**********************************************************************
```
macumber: @evanweaver this is a cruft file too

openstudiocore/src/model/ElectricLoadCenterInverterLookUpTable.cpp
```diff
@@ -0,0 +1,494 @@
...
+    }
+    OS_ASSERT(result);
+  }
+
+  void ElectricLoadCenterInverterLookUpTable_Impl::resetRatedMaximumContinuousOutputPower() {
+    bool result = setString(OS_ElectricLoadCenter_Inverter_LookUpTableFields::RatedMaximumContinuousOutputPower, "");
+    OS_ASSERT(result);
+  }
+
+  void ElectricLoadCenterInverterLookUpTable_Impl::setNightTareLossPower(boost::optional<double> nightTareLossPower) {
```
macumber: @evanweaver all of these impl functions that take optionals instead of real types are just leftovers from the old qattributes, I typically convert these to not take optionals, the optional signature is redundant with the reset method.  All that said, what really matters is the public interface so I won't be too picky about impls.

openstudiocore/src/model/Inverter_Impl.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+    boost::optional<ElectricLoadCenterDistribution> electricLoadCenterDistribution() const;
+
+    virtual boost::optional<ThermalZone> thermalZone() const = 0;
+
+    //@}
+    /** @name Setters */
+    //@{
+
+    virtual bool setThermalZone(const boost::optional<ThermalZone>& thermalZone) = 0;
```
macumber: @evanweaver I'd make this take a ThermalZone like the public interface but its ok either way

openstudiocore/src/model/test/PhotovoltaicPerformanceSimple_GTest.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "../PhotovoltaicPerformanceSimple.hpp"
+#include "../PhotovoltaicPerformanceSimple_Impl.hpp"
+
+using namespace openstudio;
+using namespace openstudio::model;
+
+TEST_F(ModelFixture, PhotovoltaicPerformanceSimple_Instantiate) {
+  Model model;
+  // TODO: Check constructor.
+  PhotovoltaicPerformanceSimple photovoltaicPerformanceSimple(model);
```
macumber: @evanweaver if you have remaining time some additional tests would be good

openstudiocore/src/model/PhotovoltaicPerformanceSimple.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+} // detail
+
+/** PhotovoltaicPerformanceSimple is a PhotovoltaicPerformance that wraps the OpenStudio IDD object 'OS:PhotovoltaicPerformance:Simple'. */
+class MODEL_API PhotovoltaicPerformanceSimple : public PhotovoltaicPerformance {
+ public:
+  /** @name Constructors and Destructors */
+  //@{
+
+  explicit PhotovoltaicPerformanceSimple(const Model& model);
```
macumber: @evanweaver this object doesn't make a lot of sense on its own, it is designed to be owned by a generatorphotovoltaic object.  i had originally made this a protected ctor to avoid weirdness about ownership when assigning to a generatorphotovoltaic.  This is one place we should chat with @kbenne and @mbadams5

openstudiocore/src/model/PhotovoltaicPerformanceEquivalentOneDiode.hpp
```diff
@@ -0,0 +1,245 @@
...
+
+} // detail
+
+/** PhotovoltaicPerformanceEquivalentOneDiode is a PhotovoltaicPerformance that wraps the OpenStudio IDD object 'OS:PhotovoltaicPerformance:EquivalentOne-Diode'. */
+class MODEL_API PhotovoltaicPerformanceEquivalentOneDiode : public PhotovoltaicPerformance {
+ public:
+  /** @name Constructors and Destructors */
+  //@{
+
+  explicit PhotovoltaicPerformanceEquivalentOneDiode(const Model& model);
```
macumber: @evanweaver this was also protected

DavidGoldwasser:
@macumber I used the plugin user script for PV on small model for two shading surfaces. I get these error in E+ run

   ** Severe  ** IP: IDF line~3457 Object=ELECTRICLOADCENTER:GENERATORS, name=ELECTRIC LOAD CENTER DISTRIBUTION 1 GENERATORS, Required Field=[Generator 1 Name] was blank.
   ** Severe  ** IP: IDF line~3457 Object=ELECTRICLOADCENTER:GENERATORS, name=ELECTRIC LOAD CENTER DISTRIBUTION 1 GENERATORS, Required Field=[Generator 1 Object Type] was blank.

In my IDF the ElectricLoadCenter:Generators object is empty except for name

ElectricLoadCenter:Generators,
  Electric Load Center Distribution 1 Generators; !- Name

Here is a list to documentation on the object
http://bigladdersoftware.com/epx/docs/8-2/input-output-reference/group-electric-load-center.html#electricloadcentergenerators

ElectricLoadCenter:Generators isn't in the OSM.

DavidGoldwasser:
@macumber hit a different simulation error when using the test in OpenStudio resources
https://github.com/NREL/OpenStudio-resources/blob/Photovoltaics/model/simulationtests/photovoltaics.rb

   ** Severe  ** Invalid Conversion Efficiency Input Mode = 
   **   ~~~   ** Entered in PhotovoltaicPerformance:Simple = PHOTOVOLTAIC PERFORMANCE SIMPLE 1
   **   ~~~   ** Field cannot be blank

DavidGoldwasser:
I'm hitting same issue in my measure as in the simulation test in resources, but there doesn't seem to be a method for performance.setConverstionEfficiencyInputMode. I assume then that it should set itself based on which value type was lasted entered, but that isn't working.
https://github.com/NREL/OpenStudio-measures/blob/595b7c3f99c8d9d2cc9f400abaf62111e80df96e/NREL%20working%20measures/add_rooftop_pv/measure.rb#L125

macumber:
@DavidGoldwasser you are correct, we should not be able to set conversionEfficiencyInputMode directly, it should be set when you call setFixedEfficiency or setEfficiencySchedule (also setting fixed efficiency should reset efficiency schedule and vice versa).  Finally the object should be initialized to fixed efficiency.  @evanweaver can you make these changes, David can answer questions about what the behavior should be, you can use this object as a reference:

https://github.com/NREL/OpenStudio/blob/9aca14214889f0f10d0e047c10cb08a9ebe14940/openstudiocore/src/model/SolarCollectorPerformancePhotovoltaicThermalSimple.cpp#L98

evanweaver:
@macumber yep, no problem.

DavidGoldwasser:
Thanks @evanweaver. Just to confirm I did hand edit the IDF file and the simulation ran and generated electricity with the PV after this.

@macumber I think the error I saw in user script in Plugin in "ELECTRICLOADCENTER:GENERATORS" is probably something I can fix in the script itself, and shouldn't require any additional C++ work. I did run simulation test on solar collector, and that ran fine.

DavidGoldwasser:
@evanweaver and @macumber  the plugin was failing for a different reason. 

It just creates and inverter but doesn't explicitly set the efficiency.

```ruby 
inverter = OpenStudio::Model::ElectricLoadCenterInverterSimple.new(model)
```

I can fix it by adding this line, but would it be better to always initialize this field with default value from IDD when a new one is made?

```ruby
inverter.setInverterEfficiency(0.98)
```

evanweaver:
Completed, @DavidGoldwasser , but @macumber there was no mention of the desired behavior when resetEfficiencySchedule() is called.  For the moment, I have added a call in reset to set the efficiencyInputMode to schedule, as well.

macumber:
I'd reset efficiency mode to fixed and set the efficiency to what the object is constructed with, same thing when you reset the fixed efficiency

jmarrec:
Shoot, I didn't this thread before I opened https://github.com/NREL/OpenStudio/issues/1923.

PR #1910, opened 2015-10-19
Add support for design oa spec on terminals

### LINKED ISSUE #1907: Enable Design OA spec on terminal
We have the design oa spec, and we support all of the E+ terminals now, but there is one configuration where the design oa spec can be attached directly to the terminal as opposed to the zone. Please support this case in OS.

kbenne:
AirTerminalSingleDuctVAVReheat and AirTerminalSingleDuctVAVNoReheat are
the two types of terminals supported by OS that E+ allows to assign a
design oa spec to directly. When doing this it means the terminal will
actively attempt to request flow to meet oa requirements for the
specific zone. This change brings OS up to speed with the E+
capability.

close #1907


asparke2:
Used the GUI to toggle OA control on/off.  Ran simulations and looks like it's working.

asparke2:
@kbenne I tested this and it looks good, you can go ahead and resolve whatever conflicts and merge this pull request.

kbenne:
@asparke2 I had to do a transition. Now that the iteration has rolled over, that transition needs a little tuneup. I was expecting this....

asparke2:
Ah ok, I assume this transition doesn't affect this actual feature?

PR #1913, opened 2015-10-19
Fixing potentially serious bug in measure xml updating, I cannot find…

macumber:
… where this was introduced, seems to have been here since 1.5.0 https://github.com/NREL/OpenStudio/commit/bbdce4fe1802461456f3a4e38cb7543693bf245a

This was causing the value of my choice arguments to be replaced by blank values in the measure.xml when the measure was changed, has anyone else seen this?


macumber:
@asparke2 @DavidGoldwasser have you seen this happen?  It might be that we are reading it from disk wrong (what happens in this bug) but then immediately running the embedded interpreter and pushing the new values in.  I saw the corruption in a special case where i had set my user measures dir to be the same directory as my application measures dir.  It might be worth filing a bug to prevent users from doing that.

DavidGoldwasser:
@macumber what are choice values even used for in measure.XML. I don't trust how the auto-generated measure.json choice value. I have had to hand alter those.

PR #1914, opened 2015-10-20
Fixes #1900 by removing limitation of > 1 user-defined measure on alt…

### LINKED ISSUE #1900: User has to specify at least one measure on alternate model in PAT
Otherwise EDAPT XML won't export, should not have to do this if alternate model is the baseline.  Reported by Elizabeth Gillmor

asparke2:
Fixes #1900 by removing limitation of > 1 user-defined measure on alternatives created from external file.


asparke2:
@macumber Tested on files from Yijia and Elizabeth

macumber:
Looks good to me

Issue #1923, opened 2015-10-29
ElectricalLoadCenter:Generators is missing


jmarrec:
I'm using 1.9.2 and have been playing with Photovoltaic.

I saw there's a sketchup plugin measure "Add Photovoltaics", I believe from @macumber . It renders and everything, but it's missing a critical object: `ElectricalLoadCenter:Generators` which is especially where you define the wattage of the array.

Right now when you add a generator to an `Electricloadcenter:Distribution`, it uses an `OS:ModelObjectList`, for example:

OS:ModelObjectList,
  {ea26b7f8-273b-4345-9316-37bd80eaff7d}, !- Handle
  Model Object List 4,                    !- Name
  {b46284f3-0457-4e01-84e6-ab3a9359f99e}; !- Model Object 1

When can we expect this missing object to be added?

jmarrec:
Hum, the OpenStudio::Model::GeneratorPhotovoltaic has a [ratedElectricPowerOutput](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-1.9.2-doc/model/html/classopenstudio_1_1model_1_1_generator_photovoltaic.html#aa527ec48b8bfc879fbb4f8f9dddbea1e) (and the OpenStudio::Model::Generator also has a ratedThermalToElectricalPowerRatio)

Did you just choose not to create an `OS:ElectricalLoadCenter:Generators` object and to have the wattage be defined at the GeneratorPhotovoltaic level and just use an OS:ModelObjectList to string them together?

I find this slightly confusing as I'm use to see a 1-1 correspondence between 'IDF' objects and their 'OS:IDF' equivalent

macumber:
That's right @jmarrec I think I just need to improve the documentation on these objects

jmarrec:
@macumber, what I was surprised with in your sketchup plugin measure is that nowhere there's rated power of array defined.

It's actually not a required field for EnergyPlus, and it makes sense.

Surface Net Area * Area Covered with Cells * Efficiency of simple generator * Solar irradiance

PR #1927, opened 2015-10-30
Added checks to delete potentially empty ElectricLoadCenterDistributi…

jmarrec:
…on (and orphan inverter) to avoid getting an E+ fatal error.


jmarrec:
@macumber I hope I did this pull request right... This is somewhat related to https://github.com/NREL/OpenStudio/issues/1923

macumber:
Thanks for the pull request @jmarrec!  I may move these checks to the EnergyPlus translator instead but the idea is the same.

PR #1934, opened 2015-11-04
Cbecc reintegrate

kbenne:
@macumber take a look at this. It is the current cbecc branch with current develop merge in. I'm thinking it would be nice to get this back into develop. Mostly confined to the sdd directory, but there are a few odds and ends outside of there. Let me know what you think.


macumber:
@kbenne looks good to me, I'd say hit merge

PR #1938, opened 2015-11-06
Date time fixes

macumber:


macumber:
@evanweaver this one is worth building a new package for

@brianlball @nllong @henryHorsey this should fix the issue without needing to call GC.start, I would still call GC after each measure in the workflow gem

PR #1940, opened 2015-11-08
Dereference open studio

macumber:
Removing some return by references

### DIFF
--- a/openstudiocore/resources/energyplus/ProposedEnergy+.idd
+++ b/openstudiocore/resources/energyplus/ProposedEnergy+.idd
@@ -10203,8 +10203,8 @@ RoomAir:Node:AirflowNetwork,
        \object-list AllHeatTranSurfNames
 
 RoomAir:Node:AirflowNetwork:InternalGains,
-      \extensible:4
-      \min-fields 5
+      \extensible:3
+      \min-fields 4
       \memo define the internal gains that are associated with one particular RoomAir:Node
   A1, \field Name
       \type Alpha
--- a/openstudiocore/src/model/Gas.cpp
+++ b/openstudiocore/src/model/Gas.cpp
@@ -961,7 +961,7 @@ Gas::Gas(const Model& model,
   OS_ASSERT(ok);
 }
 
-const std::vector<std::string>& Gas::validGasTypes() {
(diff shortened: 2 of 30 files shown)


macumber:
@mbadams5 I gave this pull request to you, it is just removing other return by references before they cause us problems

mbadams5:
@macumber I wonder if there are any performance implications in these changes? If solely within C++ these are fine and avoid a copy, but the issue is with SWIG. Do all these extra copies decrease performance noticeably and is avoiding SWIG issues more important? I don't know about the performance, but I think avoiding SWIG issues might be more important. We can deal with the performance later if it is a problem.

macumber:
Those were my thoughts.  The changes in Model are almost certainly worth it as they are called infrequently.  The IddObject ones are not often called in Ruby and they are called often in C++ so they might not be worth it.  I didn't compare performance

macumber:
PS: I think that all of these things are currently in the SWIG API but they probably don't have to be.  Removing them from SWIG could allow us to make C++ only performance improvements.  I'm sure @kbenne can add this to his list of reasons to slim down the public API

ljbrackney:
FYI - Kyle will be following up with @lefticus when he's back from Africa to set up a SOW.  Should he be putting some SWIG work into that subcontract?

kbenne:
@macumber you know me too well. Elimination is exactly where my head was going.

kbenne:
@ljbrackney I also agree that a small allocation for Jason toward weird swig issues is prudent. For when the next oddball thing pops up.

macumber:
@mbadams5 after thinking about it, I think the thing to do here is revert everything in IdfObject and IddObject to avoid performance issues.  If I do that, you ok if I merge the branch?

macumber:
@mbadams5 regarding your comment about performance for the radoutFile.illuminanceMap method, I think this is the main place this is used:

https://github.com/NREL/OpenStudio/blob/3b0d232b5f99a2a331804a8f9762f32a99b1a56b/openstudiocore/ruby/openstudio/radiance/DaylightMetrics.rb#L196

Because this is primarily being used in Ruby I was thinking it would be better to be SWIG safe than performant, just my two cents

PR #1941, opened 2015-11-09
Single subtab in memory

evanweaver:
@kbenne thanks for looking this over.


openstudiocore/src/openstudio_lib/ConstructionsTabController.cpp
```diff
@@ -32,29 +32,80 @@
 namespace openstudio {
 
 ConstructionsTabController::ConstructionsTabController(bool isIP, const model::Model& model)
-  : MainTabController(new ConstructionsTabView(model, "Constructions"))
+  : MainTabController(new ConstructionsTabView(model, "Constructions")),
+  m_model(model),
+  m_isIP(isIP)
 {
```
kbenne: Are these controllers actually removed? that would be awesome!

openstudiocore/src/openstudio_lib/ConstructionsTabController.cpp
```diff
@@ -32,29 +32,80 @@
...
+{
+  if (m_currentIndex == index) {
+    return;
+  }
+  else {
+    m_currentIndex = index;
+  }
 
-  connect(this, &ConstructionsTabController::toggleUnitsClicked, static_cast<ModelSubTabView*>(m_materialsController->subTabView()), &ModelSubTabView::toggleUnitsClicked);
+  if (qobject_cast<DefaultConstructionSetsController *>(m_currentController)) {
```
kbenne: yeah ok this is all weird, but i can't argue with improvement. I would still like to understand better.

openstudiocore/src/openstudio_lib/ConstructionsTabController.cpp
```diff
@@ -32,29 +32,80 @@
...
+  else if (m_currentController) {
+    // Oops! Should never get here
+    OS_ASSERT(false);
+  }
 
-  connect(m_defaultConstructionSetsController.get(), &DefaultConstructionSetsController::downloadComponentsClicked, this, &ConstructionsTabController::downloadComponentsClicked);
+  switch (index){
+  case 0:
+  {
+    auto defaultConstructionSetsController = new DefaultConstructionSetsController(m_model);
```
kbenne: I see they aren't gone. just created on demand. aka single tab in memory. doh!

evanweaver: None the less, it was exciting for a moment.

PR #1950, opened 2015-11-13
Sketchup2016

macumber:
This changes the SketchUp installers to have security certificates that will work with SU 2016.  It would be nice to include this in 1.10.0 but this will require heavy testing of the new installers on both Windows and Mac

@axelstudios @evanweaver FYI, I'm not sure who should evaluate the pull request


axelstudios:
The changes look good to me.  Merging prior to testing.

PR #1960, opened 2015-11-19
Electric load center improvements

macumber:


macumber:
@jmarrec here is your pull request with a few more additions, thanks for the contribution!

macumber:
This will be available in the 1.9.5 build

PR #1961, opened 2015-11-19
Idd clean up

macumber:


macumber:
The messages about "has 'autosize' as its value even though it is autocalculable." only occur when parsing an Idf object (IdfObject_Impl::fieldDataIsCorrectType) so it isn't possible to get these in the Idd cleanup.

PR #1970, opened 2015-11-25
This change appears to fix the OpenSSL load issue for SketchUp howeve…

macumber:
…r I have not been able to test completely
ly


macumber:
@axelstudios I think this fixes the issue without the need to downgrade OpenSSL versions

@evanweaver can you include this in the build tomorrow?

axelstudios:
is the front/back/ENV['PATH'] manipulation even worth keeping then?  Otherwise, this looks good.

macumber:
I'm not sure if the PATH manipulation is doing anything or not but it does seem to be the right thing to do.

PR #1975, opened 2015-12-01
Misc standard report measure issues

DavidGoldwasser:
@asparke2 all the changes are in the measure. No C++ changes  you should be able to test by replacing the measure in your installed OpenStudio.

The only conflict I'm aware of in this branch is the measure.xml file. We want the one from this branch, not from develop.


DavidGoldwasser:
@macumber didn't know if you were interested in looking at this as well.

macumber:
My main question would be if you tested this out with the spreadsheet?  Do the new attribute names work?

DavidGoldwasser:
@macumber I can test it on the server piggy backed on top of the default measure. I'll have to copy it to measures repo under the other uuid so they don't step on each other. I can see the attributes when I run the test locally. They are all snake case. I did find a change in SQL data that I think came in with EnergyPlus 8.4. Could have added logic to handle old and new SQL but that gets messy.

DavidGoldwasser:
@macumber and @henryHorsey to follow up on question about server. In the past when using spreadsheet I had to manually add time series data that the measure used as a row on spreadsheet. The part of reporting measure that stuck in output requests didn't work.

Looking at the server running the develop version of the standard report. It does now add in all of the time series data without me adding output measures manually except for hourly zone humidity. See how that isn't populated on the zone conditions section below.
http://bball-130553.nrel.gov:8080/data_points/e473182a-3f98-432c-a49f-6bd8e16d4053/view_report?file=open_studio_results_report.html#Zone_Conditions

Should look more like attached image from local run.

screen shot 2015-12-02 at 11 51 06 am

DavidGoldwasser:
Ahh, @henryHorsey looks like it is requesting humidity, but is doing it daily while it should be hourly. Only other time series data other than hourly temp and humidity, are monthly outdoor air temp and heating and cooling loads (they are all working, except for humidity).

screen shot 2015-12-02 at 11 53 29 am

macumber:
@henryHorsey does this mean that the workflow gem now implements the output requests method for reporting measures?

DavidGoldwasser:
@macumber I expect he hard coded certain ones to be there. Otherwise humidity would be hourly.

PR #1982, opened 2015-12-02
Updated templates to 1.9.5

DavidGoldwasser:
Develop wasn’t yet at 1.10.

### DIFF
--- a/openstudiocore/ruby/openstudio/sketchup_plugin/resources/templates/MinimalTemplate.osm
+++ b/openstudiocore/ruby/openstudio/sketchup_plugin/resources/templates/MinimalTemplate.osm
@@ -1,7 +1,7 @@
 
 OS:Version,
-  {5a05d0ff-62e7-4d58-8760-b977a20f97ff}, !- Handle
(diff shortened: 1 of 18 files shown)


DavidGoldwasser:
@macumber and @kbenne how about hvac_library.osm. I have been afraid to touch that in case odd things happen in forward translation. Should I open that in GUI to update to 1.10 or leave alone?

kbenne:
I appreciate your sense of caution. It can be temperamental. It appears to be at 1.9.3. I generally only update when necessary, which is when I add something new. I can see the argument for updating it for each major release though. 

I'd say we should make this a policy going forward but leave it as is for right now on 1.10.

DavidGoldwasser:
@kbenne sounds good. @macumber I made commit for other files, so if they look good you can merge.

kbenne:
I know this isn't a perfect fit, but maybe this warrants an item in [here](https://github.com/NREL/OpenStudio/blob/develop/doc/Building%20and%20Testing%20OpenStudio.md). We need to document these policies somewhere.

Issue #1979, opened 2015-12-01
OpenStudio Crashes When Clicking Measures Tab If Material Conductivity Is Not Set


pflaumingo:
When creating a new material in the OpenStudio plug-in for SketchUp it does not set a default material conductivity. If the model is then opened in the GUI and the measures tab is selected then an unhandled exception is thrown per the attached image. OpenStudio 1.9.3 was used.
![matconductivityexception](https://cloud.githubusercontent.com/assets/15233892/11516210/c8f38e2c-9836-11e5-9ad9-54fea3b4584e.PNG)

pflaumingo:
Tested in v1.9.5 and the same issue occurs.

DavidGoldwasser:
@macumber looks like other objects like load definition also appear to have blank required values, however having an empty light definition doesn't crash like an empty material does. For the crash to happen as described by @pflaumingo you don't have to even have any measures in your workflow, just going to the tab will cause the crash. 

Not sure what you think easiest fix, I thought API request for new material starts with default values for required fields, but maybe not? or maybe the plugin by-passes this.

macumber:
Thanks guys.  There are two possible fixes: 

1. Disable the add button for these objects in SketchUp
2. Try to provide sensible default values in SketchUp and keep these synchronized with defaults in the OS App (wondering where the default values in the app come from too @evanweaver ?)
3. Add default values to the IDD for everything (again where to get good default values?)

@kbenne and I have been down route 2 before and it is a constant source of bugs. I'd vote to add IDD defaults where they make sense but then go with option 1. Removing the + button for all objects with required fields that don't have defaults is going to severely reduce the number of objects you can make in SketchUp but that is fine with me.

macumber:
Changed to blocker because of crash

macumber:
This also crashes on the materials tab (which I would expect).  The crash on the measures tab is due to translating the current model to idf for computing arguments of EnergyPlus measures

macumber:
Here is a script that prints all the objects which have this issue if created in the plugin:

    require 'openstudio'

    file = nil
    File.open('skp_iddobjects.txt', 'r') do |f|
      file = f.read.split("\n")
    end

    factory = OpenStudio::IddFileAndFactoryWrapper.new()
    file.each do |object_name|
      object_name = object_name.strip
      next if object_name.empty?
      object = factory.getObject(object_name)
      next if object.empty?
      object = object.get
      
      object.nonextensibleFields.each do |field|
[...]

PR #1992, opened 2015-12-07
Removing ability to create and edit simulation control objects, const…

macumber:
…ructions, materials, schedules, loads, and other objects from SketchUp InspectorDialog #1979


macumber:
@DavidGoldwasser can you review the objects removed from the inspector here?

@asparke2 @pflaumingo FYI

PR #2009, opened 2015-12-21
Cbecc reintegration

kbenne:
This is the CBECC branch merged on top of develop. This is a staging area to get CBECC in shape to reintegrate to develop.


kbenne:
There are two significant features commingled in this PR outside of CBECC translator.

1) tertiary plant feature.
2) new SizingZone fields.

Typically I try to make PRs more single topic, but this is the culmination of some intersecting code and the CBECC branch which is large and long lived.

PR #2018, opened 2016-01-04
2012 glare sensor fix

rpg777:


rpg777:
@macumber if you have a sec to put eyes on this, great. I've attached a couple of models here.
[f_1_3-axis-blinds.osm.txt](https://github.com/NREL/OpenStudio/files/77721/f_1_3-axis-blinds.osm.txt)
[glare_sensor_2space.osm.txt](https://github.com/NREL/OpenStudio/files/77722/glare_sensor_2space.osm.txt)

openstudiocore/src/pat_app/Measures/RadianceMeasure/measure.rb
```diff
@@ -835,21 +846,25 @@ def parseResults(t_cmds, t_space_names_to_calculate, t_spaceWidths, t_spaceHeigh
...
         index = 0;
-        splitvalues = Hash.new
+        splitvalues = {}
 
         t_space_names_to_calculate.each do |space_name|
           space_size = t_spaceWidths[space_name] * t_spaceHeights[space_name]
           space = []
           illum = []
-          glaresensors = nil
+          glaresensors ||= {} # TODO: you can probably remove this
```
macumber: @nllong what is this operator?

openstudiocore/src/pat_app/Measures/RadianceMeasure/measure.rb
```diff
@@ -858,23 +873,34 @@ def parseResults(t_cmds, t_space_names_to_calculate, t_spaceWidths, t_spaceHeigh
...
+                  print_statement("### DEBUG: index is #{index}; view_index is #{view_index}", runner) if hour == 0 && debug_mode         
+                  t_radGlareSensorViews[space_name][sensor][hour] ||= {}
+                  t_radGlareSensorViews[space_name][sensor][hour]["#{sensor_index}_#{view_index}"] ||= {}
+                  view_values = values.slice(index, 1).first                     
+                  
+                  adjustedval = 0.00
+                  if view_values[hour].to_f != 0.00
+                    adjustedval = [(0.0000622*view_values[hour].to_f)+0.184, 0].max.round(2)
+                  end
+                  t_radGlareSensorViews[space_name][sensor][hour]["#{sensor_index}_#{view_index}"]['dgp'] = adjustedval.round(2)
```
macumber: @nllong @rpg777 probably not the most efficient indexing

macumber:
Both files do run for me and there seem to be the correct number of reports in the .glr files, I can't really say if these results are right or not.  

The glare sensor results are not in the radout.sql file, I assume that is intentional.  

The daylight sensor illuminance values are all 0 in the glare_sensor_2space radout.sql file (using testing settings), there are non-zero values in the f_1_3-axis-blinds radout.sql file.  I don't know if this is correct, due to testing settings, or a bug because there are two spaces.  It seems like the daylight sensor illuminance results should be in a file like the .glr file but I don't think there is a file for those results, is that right?

The global, direct, and diffuse illuminance timeseries are repeated for each space in the radout.sql file, I think these are all the same for the entire building so we could save some space in the file by not repeating these.

macumber:
Using model settings I see that there are results for Daylighting Control 1 in the radout.sql file, but Daylighting Control 2 is all 0.  It is possible that this is because the second sensor is not referenced by the thermal zone?

macumber:
Setting secondary daylight sensor did not produce results.  I was able to get glare data out of E+ but wasn't able to compare it because E+ reports glare index and I don't think that is output in our radiance report

rpg777:
Thanks, @macumber. 
* Yes, the lighting control support is still (very) limited, to one primary sensor per thermal zone. While the Radiance measure *does* calculate the illuminance maps for all spaces -- and by extension the daylight metrics -- the lighting schedules getting pushed back to the OSM are based on one primary daylighting control. At this point, this is a major problem and probably should be addressed...
* Yes the exterior daylight data is repeated in each dataset. Makes it easier to do comparative visualizations in Results Viewer. When we are doing visualizations some other way, we will definitely refactor the data structures a bit more. 
* Comparisons of "glare" data between E+ and Radiance as produced by the OS Radiance measure are not possible because 1) as you noted the metrics are different, 2) the sensors are different, and 3) glare data from E+ is garbage.

PR #2023, opened 2016-01-09
Geography

macumber:
This pull request adds a new dependency on [GeographicLib](http://geographiclib.sourceforge.net/), the library is included as an external project at the top level - @kbenne @mbadams5 @jasondegraw any comments on this dependency?

@axelstudios we would need to host the zip file on amazon and get a hash for it.


jasondegraw:
@macumber Looks OK to me, the code doesn't look as complex as some of the alternatives.

openstudiocore/src/utilities/geometry/Geometry.hpp
```diff
@@ -82,14 +83,14 @@ namespace openstudio{
 
   /// compute triangulation of vertices, holes are removed in the triangulation
   /// requires that vertices and holes are in clockwise order on the z = 0 plane (i.e. in face coordinates but reversed) 
-  UTILITIES_API std::vector<std::vector<Point3d> > computeTriangulation(const Point3dVector& vertices, const std::vector<std::vector<Point3d> >& holes, double tol = 0.001);
+  UTILITIES_API std::vector<std::vector<Point3d> > computeTriangulation(const std::vector<Point3d>& vertices, const std::vector<std::vector<Point3d> >& holes, double tol = 0.001);
 
   /// move all vertices towards point by distance, pass negative distance to move away from point
   /// no guarantee that resulting polygon will be valid
```
kbenne: This is a weird method. I'm not sure what your use case is, but it seems like you could make some strange shapes with this.

macumber: This is for doing offsets and things like that, but yeah pretty weird

PR #2031, opened 2016-01-18
Fix issue with curve min/max

### LINKED ISSUE #1897: Biquadratic curve min/max value edits not retained
@kbenne because I believe you work on HVAC.

In OS 1.9 (Windows x64): edits to the min/max values of x/y in the Curve:Biquadratic object attached to OS:Boiler:HotWater (also OS:CoolingTower:VariableSpeed, but those are the ones I checked) appear editable, but are reset to 0/1 when leaving the edit pane.

kbenne:
This takes care of an issue with curve limits and switching to IP units.
Prior to this fix, the min/max would be limited to 0/1 when in IP units.

close #1897

### DIFF
--- a/openstudiocore/resources/model/OpenStudio.idd
+++ b/openstudiocore/resources/model/OpenStudio.idd
@@ -16979,42 +16979,30 @@ OS:Table:MultiVariableLookup,
        \note This field is not allowed to be set equal to 0.
   N3 , \field Minimum Value of X1
        \type real
-       \unitsBasedOnField A8
   N4 , \field Maximum Value of X1
        \type real
-       \unitsBasedOnField A8
   N5 , \field Minimum Value of X2
        \type real
-       \unitsBasedOnField A9
   N6 , \field Maximum Value of X2
        \type real
-       \unitsBasedOnField A9
   N7 , \field Minimum Value of X3
        \type real
-       \unitsBasedOnField A10
   N8 , \field Maximum Value of X3
        \type real
-       \unitsBasedOnField A10
   N9 , \field Minimum Value of X4
        \type real
-       \unitsBasedOnField A11
   N10, \field Maximum Value of X4
        \type real
-       \unitsBasedOnField A11
   N11, \field Minimum Value of X5
        \type real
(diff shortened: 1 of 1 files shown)


kbenne:
@macumber

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -1015,7 +1015,6 @@ OS:SizingPeriod:DesignDay,
        \note Humidity Indicating Type
        \note required-field if Humidity Indicating Day Schedule is not used
        \type real
-       \units BasedOnField A3
```
macumber: @kbenne what was happening here?  OS thought this thing had units in SI and was confused it was missing an IP units?

kbenne: See #1897. Turns out that was only a bug when in IP units, which brought me to these changes.

macumber:
Weird, I knew there was code that handled those "UnitsBasedOn" things but I guess there were side effects.  Did you intend to also clean up OS:Table:MultiVariableLookup, looks like that object also has some of the UnitsBasedOn markup (although it's missing a space between units so your find replace might have missed it).

macumber:
I guess we don't need to know why there were side effects if we don't allow this anymore.  [This](https://github.com/NREL/OpenStudio/blob/a477c2f00aa05e37813d0aedab3ddbaf8d2881be/openstudiocore/src/utilities/idd/IddField.cpp#L147) is the workaround code that was supporting this.

[Search results for unitsBasedOnOtherField](https://github.com/NREL/OpenStudio/search?utf8=%E2%9C%93&q=unitsBasedOnOtherField), I guess one of these was causing the issue.

macumber:
@kbenne this is ready to merge, I just assume that you also want to clean up 'OS:Table:MultiVariableLookup' first?

kbenne:
I didn't know about the code to handle "units based on" fields. I guess there is an argument to get that working.

macumber:
I don't think it is worth getting it to work, "UnitsBasedOnOtherField" is just about as useful as no units from the code's perspective

PR #2032, opened 2016-01-19
111639676 unique thermostats

kbenne:


kbenne:
wtf. how are there conflicts in this? I'll follow up. Are you done @DavidGoldwasser?

kbenne:
heh. I'm the source of conflicts. :( 

hitting a different version translation for Sizing:Zone.

DavidGoldwasser:
@kbenne I'm done with commits.

openstudiocore/ruby/openstudio/sketchup_plugin/lib/dialogs/SpaceAttributesInterface.rb
```diff
@@ -174,16 +177,27 @@ module OpenStudio
...
-
-          # set space's parent zone values
-          # (I could probalby make array of unique values and loop outside of main loop so I don't reset the string over and over for spaces that share zones)
+          
+          # get thermostat object from input name
+          # make thermostat list
+          thermostat_objects = model.getThermostats
+          thermostat_objects.each do |object|
+            if object.name.to_s == thermostat
+              thermostat = object
```
macumber: @DavidGoldwasser this is a style comment, here you are reusing the thermostat variable which used to be a string to now be the actual model object.  I would normally use a second variable initialized to nil to hold the model object, that way i could check if i found the object or not.  I guess here you know you will find the object because the list of names is populated right before this runs, but in general you can't always know that you are going to find the object you are looking for.

openstudiocore/src/osversion/VersionTranslator.cpp
```diff
@@ -3004,9 +3004,36 @@ std::string VersionTranslator::update_1_10_0_to_1_10_1(const IdfFile& idf_1_10_0
...
+              referencingZones.push_back(zone);
+            }
+          }
+        }
+      }
+
+      // Clone the thermostat for every zone after the first
+      if( referencingZones.size() > 1u ) {
+        for( auto it = (referencingZones.begin() + 1); it != referencingZones.end(); ++it ) {
+          auto newThermostat = object.clone();
```
macumber: @kbenne no name munging?

openstudiocore/src/osversion/VersionTranslator.cpp
```diff
@@ -3004,9 +3004,36 @@ std::string VersionTranslator::update_1_10_0_to_1_10_1(const IdfFile& idf_1_10_0
...
+
   for (const IdfObject& object : idf_1_10_0.objects()) {
     auto iddname = object.iddObject().name();
-    if (iddname == "OS:Sizing:Zone") {
+
+    if (iddname == "OS:ThermostatSetpoint:DualSetpoint") {
+      // Get all of the zones that point to this thermostat
+      std::vector<IdfObject> referencingZones;
+      for( const auto & zone : zones ) {
+        if( auto thermostateHandle = zone.getString(19) ) {
```
macumber: @kbenne couldnt you just check if string 19 equals current object's handle?

openstudiocore/src/osversion/VersionTranslator.cpp
```diff
@@ -3004,9 +3004,36 @@ std::string VersionTranslator::update_1_10_0_to_1_10_1(const IdfFile& idf_1_10_0
   IdfFile targetIdf(idd_1_10_1.iddFile());
   ss << targetIdf.versionObject().get();
 
+  auto zones = idf_1_10_0.getObjectsByType(idf_1_10_0.iddFile().getObject("OS:ThermalZone").get());
+
   for (const IdfObject& object : idf_1_10_0.objects()) {
     auto iddname = object.iddObject().name();
-    if (iddname == "OS:Sizing:Zone") {
+
+    if (iddname == "OS:ThermostatSetpoint:DualSetpoint") {
```
macumber: @kbenne what about other thermostat types?  not worried about them because they aren't common?

openstudiocore/ruby/openstudio/sketchup_plugin/lib/dialogs/SpaceAttributesInterface.rb
```diff
@@ -174,16 +177,27 @@ module OpenStudio
...
+              break
+            end
+          end
 
           parent_zone = drawing_interface.model_object.thermalZone
           if not parent_zone.empty?
 
+            # clone and rename thermostat
+            new_thermostat = thermostat.clone(model_interface.openstudio_model).to_Thermostat.get
```
macumber: @DavidGoldwasser reuse model from above

openstudiocore/src/osversion/VersionTranslator.cpp
kbenne: Well magically the IdfFile / Workspace gears are appending a integer to guarantee uniqueness. (I just confirmed) But perhaps you would like to see the thermostat name derive from the zone name? I can be on board with that. Honestly I hadn't considered that since the name isn't visible in that app.

DavidGoldwasser: I'm fine either way with how the version translator handle this. Ideally this won't be triggered much down the road as users start from models made with 1.11 or later. @kbenne the template OSM files we currently use have thermostats that are not associated with any thermal zones. I assume those would be un-touched by version translation.

openstudiocore/ruby/openstudio/sketchup_plugin/lib/dialogs/SpaceAttributesInterface.rb
DavidGoldwasser: @macumber thanks, I added the model but forgot to use it in the loop.

openstudiocore/src/osversion/VersionTranslator.cpp
macumber: Not a big deal at all, i didnt know if the workspace stuff would get triggered in the translator

openstudiocore/ruby/openstudio/sketchup_plugin/lib/dialogs/SpaceAttributesInterface.rb
DavidGoldwasser: @macumber I'll fix that, I also realized I can move this code above the selection loop so it just runs once.

openstudiocore/src/osversion/VersionTranslator.cpp
kbenne: Yeah you're right. I was probably thinking I was going to need to store the thermostat or something, but that is not true it is the zone I'm storing. I'll make this change.

macumber:
Also noticed that the list of thermostats you can select from in the plugin is now going to be a lot more cluttered with each thermal zone's thermostat in the list, just a consequence of not having a concise list of resources.

macumber:
@kbenne final comment on this, when you create a new thermal zone in the app and set a cooling schedule, the thermostat will be named 'Thermostat Setpoint Dual Setpoint 1' or such.  This makes it hard to know what thermostat is what when you look at the list by name in the plugin.  Not a big deal just wanted to mention it, you can fix it if you want

macumber:
@kbenne @DavidGoldwasser I'm ready to merge when you guys give the all clear

kbenne:
Hang on. Let me do something about this naming situation.

kbenne:
I'm done.

PR #2035, opened 2016-01-21
Add new options for ControllerMechanicalVentilation

### LINKED ISSUE #1968: Add new oa method to ControllerMechanicalVentilation
Add new options 

ProportionalControlBasedonOccupancySchedule
ProportionalControlBasedOnDesignOccupancy

kbenne:
System Outdoor Air Method:
  ProportionalControlBasedonOccupancySchedule
  ProportionalControlBasedOnDesignOccupancy

This brings with it a new object, ZoneControlContaminantController, and
also enables translation support for ZoneAirContaminantBalance, which is
also required. ZoneControlContaminantController defines the contaminant
levels in the zone, ZoneAirContaminantBalance defines the contaminant
(CO2) outside.

close #1968


openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -15708,6 +15710,47 @@ OS:Controller:MechanicalVentilation,
...
        \object-list DesignSpecificationZoneAirDistributionNames
 
+OS:ZoneControl:ContaminantController,
+       \min-fields 8
+  A1, \field Handle
+       \type handle
+       \required-field
+  A2 , \field Name
+       \required-field
+  A3 , \field Controlled Zone Name
```
macumber: This is one of those data model mismatch things, it's possible for multiple ZoneControl:ContaminantControllers to reference the same zone, this just has to get enforced in the API I guess

kbenne: Well reversing the relationship may be worse. In that case you could have multiple zones referencing the same controller.

macumber: Yup, I really think this is a good use case for a nested data structure, make it an actual child and put a restriction of 0-1 on it

openstudiocore/src/model/ThermalZone.cpp
```diff
@@ -2269,6 +2276,34 @@ namespace detail {
...
+        }
+      }
+    }
+
+    return boost::none;
+  }
+
+  bool ThermalZone_Impl::setZoneControlContaminantController(const ZoneControlContaminantController & contaminantController)
+  {
+    auto tz = getObject<ThermalZone>();
```
macumber: Remove existing controllers here?

openstudiocore/src/model/ThermalZone.cpp
```diff
@@ -2269,6 +2276,34 @@ namespace detail {
...
+    return contaminantController.getImpl<detail::ZoneControlContaminantController_Impl>()->setControlledZone(tz);
+  }
+
+  void ThermalZone_Impl::resetZoneControlContaminantController()
+  {
+    if( auto controller = zoneControlContaminantController() ) {
+      controller->getImpl<detail::ZoneControlContaminantController_Impl>()->resetControlledZone();
```
macumber: Should you delete it?

kbenne: I debated that. I'm really not sure what the policy is.

macumber: Otherwise you get orphans that you will never see right?  These things aren't shown in the model list right, just the library list?  I'd go with what you did for thermostats

openstudiocore/src/model/test/ZoneControlContaminantController_GTest.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <gtest/gtest.h>
+#include "ModelFixture.hpp"
+#include "../ZoneControlContaminantController.hpp"
+#include "../ZoneControlContaminantController_Impl.hpp"
+
+using namespace openstudio;
+
+TEST_F(ModelFixture,ZoneControlContaminantController)
+{
+  ::testing::FLAGS_gtest_death_test_style = "threadsafe";
```
macumber: What's up with all the fancy stuff here?  Why not just:

    model::Model m;
    model::ZoneControlContaminantController controller(m);

I'm not asking a leading question, I really don't know what you are trying to test here?

kbenne:
@macumber. is this mergable?

macumber:
Last thing was protecting against inadvertent deletes when calling set with same input twice, were you planning on fixing that?

kbenne:
I wasn't but since you're pushing I'll go ahead. I think this is a bit of a rabbit hole. What about passing a thing that is already owned by another thing? Or in other cases besides this one, what if you pass in a type that the set ends up failing on? Like if you pass an incompatible derived type? Anyway I'll go ahead and do this one.

macumber:
Well my argument was really that these things should be consistent.  I was looking at setThermostat's implementation as a template, maybe this could be factored into some sort of a helper method in ThermalZone?

macumber:
I was wrong, I guess you aren't protecting against that in the setThermostat either.  These 'rabbit holes' are an example of what I keep seeing when the API and data model don't line up.  The point about sharing is valid too, I thought that was fixed on the unique thermostat branch but I guess not, https://github.com/NREL/OpenStudio/blob/111639676-UniqueThermostats/openstudiocore/src/model/ThermalZone.cpp#L997

If the point of that branch was to ensure unique thermostats when using the API but you can do this:

    Thermostat t(model);
    ThermalZone z1(model);
    ThermalZone z2(model);
    z1.setThermostat(t);
    z2.setThermostat(t);
    EXPECT_EQ(z1.thermostat()->handle(), z2.thermostat()->handle());

that doesn't seem like it is enforcing unique thermostats right?

macumber:
Your call on the merge @kbenne I don't want to hold up progress just to be a jerky code reviewer

kbenne:
"API but you can do this..." 

I think that pseudo code is fiction. Besides that fact that Thermostat is a base and can't be instantiated. ThermalZone::setThermostat first resets, and if you look at reset it removes. Not to say there isn't trouble lurking. if you pass the same thermostat that is already established, it will be removed and then attempted to be set. I'll write some tests, but i think your example will crash, not pass the EXPECT_EQ.

macumber:
Well only one way to find out :-)  The code is passing the same thermostats to two different zones so I dont think it will get reset.  If you passed the thermostat to the same zone twice it would get deleted and cause a crash.

kbenne:
Indeed. You are correct. I should know better. Although I think what I said is also true. setting the same thermostat will crash.

So what is the desired behavior if a thermostat is already in use? steel it away or fail the setter?

macumber:
If the idea is that thermostats are not shared then i guess you would steal it, probably a clone is better than actually stealing it

kbenne:
i like clone.

kbenne:
i think this is just the tip of the iceberg. I have to think to drum up examples. Here is an easy one. https://github.com/NREL/OpenStudio/blob/5264d43d9085b989df35073360c3b51170decc05/openstudiocore/src/model/CoilCoolingDXSingleSpeed.cpp#L231 This is actually close to correct. At least now that curve is a resource. Prior to that, not so much.

kbenne:
Here this one. ptac should "own" the fan. I think that is pretty obvious. But....

https://github.com/NREL/OpenStudio/blob/5264d43d9085b989df35073360c3b51170decc05/openstudiocore/src/model/ZoneHVACPackagedTerminalAirConditioner.cpp#L431

kbenne:
i guess this is one reason to make everything a ResourceObject!

PR #2050, opened 2016-01-26
Fix zone controller ownership

### LINKED ISSUE #2049: fix ownership issues with zone controllers
This includes thermostats, humidistats, and new ZoneControlContaminantController. Make the API more robust and enforce our will.

kbenne:
close #2049 

Hey @macumber. This should make you happy. I bet we can still pick something out, but far more robust than where we were.

### DIFF
--- a/openstudiocore/src/model/ThermalZone.cpp
+++ b/openstudiocore/src/model/ThermalZone.cpp
@@ -977,18 +977,13 @@ namespace detail {
 
   bool ThermalZone_Impl::setThermostatSetpointDualSetpoint(const ThermostatSetpointDualSetpoint & thermostat)
   {
-    resetThermostat();
-    return setPointer(OS_ThermalZoneFields::ThermostatName, thermostat.handle());
+    return setThermostat(thermostat);
   }
 
   void ThermalZone_Impl::resetThermostatSetpointDualSetpoint()
   {
-    if( boost::optional<ThermostatSetpointDualSetpoint> thermostat = this->thermostatSetpointDualSetpoint() )
-    {
-      thermostat->remove();
-    }
-
-    setString(OS_ThermalZoneFields::ThermostatName, "");
+    // This will reset other thermostat types, but I think that is ok
+    resetThermostat();
   }
 
   boost::optional<Thermostat> ThermalZone_Impl::thermostat() const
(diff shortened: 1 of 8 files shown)


openstudiocore/src/model/ThermalZone.cpp
```diff
@@ -998,14 +993,32 @@ namespace detail {
 
   bool ThermalZone_Impl::setThermostat(const Thermostat & thermostat)
   {
-    resetThermostat();
-    return setPointer(OS_ThermalZoneFields::ThermostatName, thermostat.handle());
+    auto t_model = model();
+
+    if( t_model != thermostat.model() ) {
```
macumber: This check seems ridiculous but honestly I've considered it, it is legit error checking.  I guess normally this is included if you check the return value of set methods but this could save you headache to do it first.

openstudiocore/src/model/ThermalZone.cpp
```diff
@@ -998,14 +993,32 @@ namespace detail {
...
+      return false;
+    }
+
+    if( auto currentZone = thermostat.thermalZone() ) {
+      if( currentZone->handle() == handle() ) {
+        // or should it be false?
+        // I think this is similar to what you would see in 
+        // Lights::setSpace() under similar conditions
```
macumber: True this can happen in the setSpace but there setting the same thing twice is benign, except I guess you get unnecessary change signals emitted

macumber: I'd say should return true because the condition that the function was supposed to do is true at the end of the function

openstudiocore/src/model/ZoneControlHumidistat.cpp
```diff
@@ -119,6 +123,20 @@ namespace detail {
     OS_ASSERT(result);
   }
 
+  boost::optional<ThermalZone> ZoneControlHumidistat_Impl::controlledZone() const
+  {
+    Handle h = handle();
+    auto zones = model().getModelObjects<model::ThermalZone>();
```
macumber: Probably a faster look up using get sources

openstudiocore/src/model/test/ThermalZone_GTest.cpp
```diff
@@ -636,33 +623,138 @@ TEST_F(ModelFixture, ThermalZone_Ports)
...
+
+  {
+    Model m;
+    ThermalZone zone(m);
+    ThermostatSetpointDualSetpoint thermostat(m); 
+    zone.setThermostat(thermostat);
+    auto zone2 = zone.clone().cast<model::ThermalZone>();
+
+    EXPECT_TRUE(zone2.thermostat());
+    EXPECT_EQ(2u,m.getModelObjects<model::Thermostat>().size());
```
macumber: Is the second thermostat hooked up to the cloned zone?  It looked like the thermostat field got blanked in the ThermalZone::clone:

      tz.setString(OS_ThermalZoneFields::ThermostatName,"");

macumber: Huh I guess it is since you are checking 

    EXPECT_TRUE(zone2.thermostat());

maybe just pop in a 

    ASSERT_TRUE(zone1.thermostat());
    ASSERT_TRUE(zone2.thermostat());
    EXPECT_NE(zone.thermostat()->handle(), zone2.thermostat()->handle());

macumber: PS: I'll do this, I owe you that much :-)

openstudiocore/src/model/ThermalZone.cpp
kbenne: Yeah and the other thing is that the set method return value can come too late. Like say for instance if you do a reset first.

Issue #2051, opened 2016-01-27
Mac packages fail at runtime due to geo lib


kbenne:
Prefer to static link if possible.

kbenne:
https://github.com/NREL/OpenStudio/pull/2053

macumber:
@kbenne is this issue closed?

kbenne:
yep.

PR #2053, opened 2016-01-27
Static link geo lib and use project config file - again

kbenne:
Fix a problem in the first attempt at static link


kbenne:
close #2051

PR #2066, opened 2016-02-05
2064 fix version translator

kbenne:


openstudiocore/src/osversion/VersionTranslator.cpp
```diff
@@ -445,6 +446,42 @@ void VersionTranslator::initializeMap(std::istream& is) {
...
+    if( currentVersion == VersionString(1,9,0) ) {
+      if( oIdfFile ) {
+        auto sizingObjects = oIdfFile->getObjectsByType(iddFile.getObject("OS:Sizing:Zone").get());
+
+        // Figure out if the OS:Sizing:Zone object looks like it is from CBECC
+        // it will have extra fields that would were not in the "real" 1.9.0 IDD
+        bool fromCBECC = false;
+        for( auto const & object : sizingObjects ) {
+          if( auto value = object.getString(2) ) {
+            if( istringEqual("SupplyAirTemperature", value.get()) ||
```
macumber: @kbenne is this test robust enough?  Could that field be blank?

openstudiocore/src/osversion/VersionTranslator.cpp
```diff
@@ -445,6 +446,42 @@ void VersionTranslator::initializeMap(std::istream& is) {
...
+        if( fromCBECC ) {
+          // Remove the sizing objects so that the translation will proceed smoothly
+          oIdfFile->removeObjects(sizingObjects);
+
+          // Get a special CBECC idd file,
+          // load the idf file again against that idd,
+          // get the sizing objects and save them for later,
+          // we will reintrodce the sizing objects in the version 1.10.2 phase of the translation
+          // when they were officially part of OS
+          auto cbeccIddFile = IddFile::load( getSharedResourcesPath() / "osversion/1_9_0_CBECC/OpenStudio.idd");
```
macumber: Is it actually required to load this again?  Could you just keep the IdfObjects you removed?  You could stash them in a string or something.

openstudiocore/src/osversion/VersionTranslator.hpp
```diff
@@ -162,6 +162,7 @@ class OSVERSION_API VersionTranslator {
   int m_nObjectsFinalIdf;
   int m_nObjectsFinalModel;
   bool m_isComponent;
+  std::vector<IdfObject> m_cbeccSizingObjects;
```
macumber: maybe tag with a version number in case this happens again?  m_cbeccSizingObjects_190?

macumber:
Overall not too bad, just made a few comments.  I do think @evanweaver already made the iteration branch, did you want to try to get this on there?

openstudiocore/src/osversion/VersionTranslator.cpp
kbenne: I wanted to do that. The trouble is that they are truncated at the max fields of the idd.

openstudiocore/src/osversion/VersionTranslator.cpp
kbenne: it is certainly a heuristic. I'll check but I think it is a required field and therefore not blank.

openstudiocore/src/osversion/VersionTranslator.hpp
kbenne: well the very idea of what I'm doing here is pretty horrific in my mind, but I think this is decent considering the options. Users should be none the wiser and it should just work for them.

kbenne:
well the very idea of what I'm doing here is pretty horrific in my mind, but I think this is decent considering the options. Users should be none the wiser and it should just work for them.

openstudiocore/src/osversion/VersionTranslator.cpp
macumber: Oh gross

macumber:
@kbenne I merged this, @evanweaver is this too late to make the 1.10.3 packages?

evanweaver:
@macumber @kbenne  1.10.3 was built, tested, and published Friday afternoon.

PR #2068, opened 2016-02-08
Update OpenStudio::DateTime to include a utc offset member and increa…

macumber:
…se support of ISO8601 formatted strings


macumber:
@henryHorsey @nllong @jasondegraw  FYI

This pull request should not be merged lightly, I tried addressing places where time strings were written to databases and used for comparison but may not have gotten all of them

brianlball:
is the calibration report part of the testing framework for this?  That uses the DateTime functionality

macumber:
I don't think there are enough automated tests of the calibration report, but you're right that is a good test case

macumber:
@kbenne can you look this over and merge for 1.11?

openstudiocore/src/project/ProjectDatabase.cpp
```diff
@@ -186,6 +186,7 @@ namespace detail {
     // make sure we are the last one using database connection
     OS_ASSERT(m_qSqlDatabase.use_count() == 1);
 
+    QString connectionName = m_qSqlDatabase->connectionName();
```
macumber: @kbenne I was attempting to get rid of the error messages about closing QSqlDatabase connections still in use.  I did not solve this issue.

openstudiocore/src/project/Test/ProjectFixture.cpp
```diff
@@ -88,7 +88,11 @@ openstudio::project::ProjectDatabase ProjectFixture::getDatabase(
 void ProjectFixture::removeAllDatabases() {
   for (const openstudio::path& dbDir : databaseDirs) {
     if (boost::filesystem::exists(dbDir)) {
-      boost::filesystem::remove_all(dbDir);
+      try{
```
macumber: This was failing and causing several tests to fail.  I did not understand why, but it was causing unit test failures

PR #2102, opened 2016-02-26
Combine EPW and DDY folder locations to same. Continue to use absolut…

evanweaver:
…e path as originally implemented, but add relative path for OpenStudio model.

I debated the best path forward: broad-sweeping, holistic change, or finite, lower risk tweak. Because 1) we had numerous problems with weather files in the past, and they appeared to be almost completely solved with a November commit, and 2) we are completely refactoring this code with OS2, I opted for what I considered to be a less invasive (albeit less emotionally satisfying) solution.

That said, @macumber  and @DavidGoldwasser this should be well-tested, and in a manner likely consistent with Elizabeth G's workflow. Please let me know your thoughts.

### DIFF
--- a/openstudiocore/src/model/StraightComponent_Impl.hpp
+++ b/openstudiocore/src/model/StraightComponent_Impl.hpp
@@ -56,7 +56,7 @@ namespace detail {
 
     virtual std::vector<ModelObject> children() const override;
 
-    virtual std::vector<HVACComponent> edges(const boost::optional<HVACComponent> & prev);
+    virtual std::vector<HVACComponent> edges(const boost::optional<HVACComponent> & prev) override;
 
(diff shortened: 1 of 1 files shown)


macumber:
Just a note, this issue also exists in PAT:

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/analysisdriver/SimpleProject.cpp#L1921

However, PAT has all kinds of complicated relative/absolute path checking going on.  Will have to see if this is an issue for Elizabeth

macumber:
Here are the places makeUrlAbsolute is called, thankfully not too many

https://github.com/NREL/OpenStudio/search?utf8=%E2%9C%93&q=makeUrlAbsolute

Issue #2055, opened 2016-02-01
PAT datapoints with comma's in name fail to run


macumber:
Reported by @egillmor

macumber:
![untitled](https://cloud.githubusercontent.com/assets/1327850/12722892/420983bc-c8c4-11e5-93a3-d556f689af5e.png)

eringold:
Related to https://github.com/NREL/OpenStudio/issues/1597?

macumber:
Maybe, there are [several comma related bugs](https://github.com/NREL/OpenStudio/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+comma)

macumber:
I am not able to reproduce this either, @egillmor can you reproduce this bug with a simple project?

egillmor:
Yes.  Create a run from an external file.  Add a user-defined measure name with commas in it.  Fails every time.

![image](https://cloud.githubusercontent.com/assets/16869106/13410400/cd7f1e8e-def4-11e5-9a9d-be3fdfe04d14.png)

eringold:
I've found this happens because the name of the User Defined Measure is set as the name a Life Cycle Cost idf objects created in the run. Since the name has a comma, it messes up the idf syntax.

macumber:
That is exactly what is happen @eringold , thanks!

eringold:
@macumber we just ran into another problem, possibly related to this issue. A string with a newline character was copied into the User Defined Measure name field, which then becomes the name of a LifeCycleCost object. Something within EnergyPlus then replaces the newline with a comma, which caused the object to fail because where the first numeric field should be now has a alphanumeric string. Is it possible to add `\n` to your encoding check? 

PS i'm not sure how that character ended up in the string; possibly it came from excel.

kbenne:
I have a corresponding pivotal issue on this and I see the branch has been merged. Can we close?

macumber:
I think so, although it would be good for @asparke2, @DavidGoldwasser, @eringold or anyone else to test in 1.10.5 to confirm

eringold:
Just tested with 1.10.5: Pat alternatives with user-defined measure names that include commas and newline characters run without issue. Seems to me that this can be closed.

Issue #1597, opened 2015-04-13
API allows you to enter invalid object names


DavidGoldwasser:
I ran acros this testing fix for issue #1591. It seems like the fix is more likely related to changes made between 1.7.1 and 1.7.2 than fixes made for BCL attributes. Below are screenshots of the issue. It doesn't seem to matter what weather file I chose. I hit the same thing. If you look at screenshots from #1591 you can see that it didn't hit the issue then?

Screenshots in apply measures now
![4-13-2015 9-45-12 am](https://cloud.githubusercontent.com/assets/4857055/7119868/73661cc6-e1c6-11e4-89c2-3264f6e31e76.png)

![4-13-2015 9-45-35 am](https://cloud.githubusercontent.com/assets/4857055/7119877/829382b0-e1c6-11e4-9e57-7b6efc567509.png)

![4-13-2015 10-25-56 am](https://cloud.githubusercontent.com/assets/4857055/7120048/819b416c-e1c7-11e4-9216-7f998e5c15fa.png)

Similar issue in PAT. It happens in model to IDF, but I assume I would hit it sooner if we had more OSM measures.
![4-13-2015 10-19-37 am](https://cloud.githubusercontent.com/assets/4857055/7120055/86abb114-e1c7-11e4-86e7-c3c551f9bd10.png)

Here is the text I see when I try to open the OSM made by the measures ruby measure test.
``
Error: Model with Version 1.7.2 IDD is not valid to draft strictness level.

Error: The collection is INVALID at strictness level 'Draft', because of the errors:
Field      level data error of type DataType          .
Error is in an object of type 'OS:Site', named 'Denver Centennial  Golden   Nr', in field 2.
Additional information about the error type: field-level data is of an incorrect type.
Field      level data error of type NumericBound      .
Error is in an object of type 'OS:Site', named 'Denver Centennial  Golden   Nr', in field 4.
Additional information about the error type: numeric data violates a min or max bound.
``

DavidGoldwasser:
Here is the site object from the ruby test.
OS:Site,
  {34fceb3c-6e0d-46e1-8ddf-19844eea695a}, !- Handle
  Denver Centennial  Golden   Nr, CO, USA, !- Name
  39.74,                                  !- Latitude {deg}
  -105.18,                                !- Longitude {deg}
  -7,                                     !- Time Zone {hr}
  1829,                                   !- Elevation {m}
  ;                                       !- Terrain

[...]

PR #2107, opened 2016-03-01
Ecoding commas, semicolons, and exclamation marks in IDF format.  If …

macumber:
…this is too much to do at Idf level we could implement it at ModelObject level instead.

### DIFF
--- a/openstudiocore/src/model/test/LifeCycleCostParameters_GTest.cpp
+++ b/openstudiocore/src/model/test/LifeCycleCostParameters_GTest.cpp
@@ -23,6 +23,12 @@
 
 #include "../LifeCycleCostParameters.hpp"
 #include "../LifeCycleCostParameters_Impl.hpp"
+#include "../LifeCycleCost.hpp"
+#include "../LifeCycleCost_Impl.hpp"
+#include "../Building.hpp"
+#include "../Building_Impl.hpp"
+
+#include <utilities/idd/OS_LifeCycleCost_FieldEnums.hxx>
 
 #include "../../utilities/time/Date.hpp"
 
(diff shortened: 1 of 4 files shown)


macumber:
@kbenne @mbadams5 would appreciate some eyes on this

macumber:
Should this have a version translation to go with it?  I'm not sure that is even possible?

mbadams5:
Shouldn't `[!,;]` be rejected as invalid inputs for a field? They are invalid in EnergyPlus. I know you are getting around that by encoding and then decoding those characters.

Anyways, I am curious how this will potentially affect performance. `setString()` and `name()` are both used very often internally and I would say 99.99% of the time this encoding/decoding is not needed. Also I am leery of those 3 `boost::replace_all` calls but I don't want to speculate too much without profiling.

macumber:
I guess checking for them would be faster than replacing them but I'm not sure if that is the desired behavior.  I could check for the presence of the &# combo before doing any replacements

mbadams5:
That would still involve the same amount of looping as just replacing them. Don't the strings already get validated or verified somewhere else? Is there some regex that they go through?

macumber:
Fixes #2055

macumber:
Fixes #1597

macumber:
I took a look at this and decided I don't want to do any performance optimizations without some profiling, otherwise I'm shooting in the dark.  I would like to get this in the 1.11 release, so 1) do you think performance testing is required to merge and 2) what is the best test case for profiling?

macumber:
@mbadams5 I think most of the validity checking happens here https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/idf/WorkspaceObject.cpp#L387

macumber:
I could see adding additional tests to make sure that objects can point to each other by names with special characters, find objects by name works with special characters, etc.  Any specific other test cases I should add?

PR #2108, opened 2016-03-01
Adding baseline pdrop adjustment to air loop hvac

jmarrec:
I'd welcome (more than) a few pointers as this is the first time I try to do anything in C++

### DIFF
--- a/openstudiocore/resources/osversion/1_10_4/OpenStudio.idd
+++ b/openstudiocore/resources/osversion/1_10_4/OpenStudio.idd
@@ -8276,9 +8276,14 @@ OS:AirLoopHVAC,
        \type object-list
        \required-field
        \object-list ConnectionNames
-  A14; \field Return Air Bypass Flow Temperature Setpoint Schedule Name
+  A14, \field Return Air Bypass Flow Temperature Setpoint Schedule Name
        \type object-list
        \object-list ScheduleNames
+  N2; \field Baseline Pressure Drop Adjustment 
+       \note Calculate it as a function of the total Supply CFM for the loop.
+       \type real
+       \units Pa.s/m3
+       \default 0
 
 OS:AirLoopHVAC:ControllerList,
        \memo List controllers in order of control sequence
--- a/openstudiocore/src/model/AirLoopHVAC.cpp
+++ b/openstudiocore/src/model/AirLoopHVAC.cpp
@@ -1770,7 +1770,7 @@ double AirLoopHVAC::baselinePressureDropAdjustment() const {
 }
 
 bool AirLoopHVAC::setBaselinePressureDropAdjustment(double baselinePressureDropAdjustment) {
-  getImpl<detail::AirLoopHVAC_Impl>()->setBaselinePressureDropAdjustment(baselinePressureDropAdjustment);
+  return getImpl<detail::AirLoopHVAC_Impl>()->setBaselinePressureDropAdjustment(baselinePressureDropAdjustment);
 }
 
 void AirLoopHVAC::resetBaselinePressureDropAdjustment() {
--- a/openstudiocore/src/model/AirLoopHVAC.cpp
+++ b/openstudiocore/src/model/AirLoopHVAC.cpp
@@ -923,8 +923,8 @@ namespace detail {
   
   // Set the Baseline Pdrop
   bool AirLoopHVAC_Impl::setBaselinePressureDropAdjustment(double baselinePressureDropAdjustment) {
-    result = setDouble(OS_AirLoopHVACFields::BaselinePressureDropAdjustment, baselinePressureDropAdjustment);
-    OS_ASSERT(result);
+    bool result = setDouble(OS_AirLoopHVACFields::BaselinePressureDropAdjustment, baselinePressureDropAdjustment);
+    return result;
   }
   
   // Reset Baseline Pdrop
--- a/openstudiocore/src/model/AirLoopHVAC.cpp
+++ b/openstudiocore/src/model/AirLoopHVAC.cpp
@@ -916,30 +916,22 @@ namespace detail {
   }
   
   /** Baseline Pressure Drop Adjustment **/
-  // Return the basleine Pdrop
-  boost::optional<double> AirLoopHVAC_Impl::baselinePressureDropAdjustment() const {
+  // Return the Baseline Pdrop
+  double AirLoopHVAC_Impl::baselinePressureDropAdjustment() const {
     return getDouble(OS_AirLoopHVACFields::BaselinePressureDropAdjustment,true);
   }
   
-  
   // Set the Baseline Pdrop
-  void AirLoopHVAC_Impl::setBaselinePressureDropAdjustment(boost::optional<double> baselinePressureDropAdjustment) {
-    bool result(false);
-    if (baselinePressureDropAdjustment) {
-      result = setDouble(OS_AirLoopHVACFields::BaselinePressureDropAdjustment, baselinePressureDropAdjustment.get());
-    }
-    else {
-      resetBaselinePressureDropAdjustment();
-      result = true;
-    }
+  bool AirLoopHVAC_Impl::setBaselinePressureDropAdjustment(double baselinePressureDropAdjustment) {
+    result = setDouble(OS_AirLoopHVACFields::BaselinePressureDropAdjustment, baselinePressureDropAdjustment);
     OS_ASSERT(result);
   }
   
   // Reset Baseline Pdrop
   void AirLoopHVAC_Impl::resetBaselinePressureDropAdjustment() {
-    bool result = setString(OS_AirLoopHVACFields::BaselinePressureDropAdjustment, "");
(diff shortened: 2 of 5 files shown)


jmarrec:
Related to https://github.com/NREL/openstudio-standards/blob/baseline_automation/openstudio-standards/lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb#L214

openstudiocore/resources/osversion/1_10_4/OpenStudio.idd
```diff
@@ -8276,9 +8276,14 @@ OS:AirLoopHVAC,
        \type object-list
        \required-field
        \object-list ConnectionNames
-  A14; \field Return Air Bypass Flow Temperature Setpoint Schedule Name
+  A14, \field Return Air Bypass Flow Temperature Setpoint Schedule Name
```
macumber: Adding fields that are not in the E+ Idd to OSM requires some discussion, @kbenne should definitely weigh in.  How will this get translated to E+?

openstudiocore/src/model/AirLoopHVAC.cpp
```diff
@@ -914,6 +914,25 @@ namespace detail {
     bool result = setString(OS_AirLoopHVACFields::DesignSupplyAirFlowRate, "autosize");
     OS_ASSERT(result);
   }
+  
+  /** Baseline Pressure Drop Adjustment **/
+  // Return the Baseline Pdrop
+  double AirLoopHVAC_Impl::baselinePressureDropAdjustment() const {
+    return getDouble(OS_AirLoopHVACFields::BaselinePressureDropAdjustment,true);
```
macumber: Does this compile?  I would expect you to have to do this:

    boost::optional<double> result = getDouble(OS_AirLoopHVACFields::BaselinePressureDropAdjustment,true);
    OS_ASSERT(result); // ok to assert since there is a default in the IDD, want to assert because .get below will cause a crash if result is empty
    return result.get();

openstudiocore/src/model/AirLoopHVAC.cpp
```diff
@@ -914,6 +914,25 @@ namespace detail {
...
+  }
+  
+  // Set the Baseline Pdrop
+  bool AirLoopHVAC_Impl::setBaselinePressureDropAdjustment(double baselinePressureDropAdjustment) {
+    result = setDouble(OS_AirLoopHVACFields::BaselinePressureDropAdjustment, baselinePressureDropAdjustment);
+    OS_ASSERT(result);
```
macumber: Here you don't want the assert, result could be true or false (e.g. if user passes in -9999), we return the value to the user and don't really care which way it goes.  You do need to return a value or it won't compile, also you have to declare the result variable:

    bool result = setDouble(OS_AirLoopHVACFields::BaselinePressureDropAdjustment, baselinePressureDropAdjustment);
    return result;

openstudiocore/src/model/AirLoopHVAC.cpp
```diff
@@ -914,6 +914,25 @@ namespace detail {
...
+  }
+  
+  // Reset Baseline Pdrop
+  void AirLoopHVAC_Impl::resetBaselinePressureDropAdjustment() {
+    bool result = setDouble(OS_AirLoopHVACFields::BaselinePressureDropAdjustment, 0);
```
macumber: Looks good, often "reset" means resetting the field to an empty string, that would be equivalent in this case since there is a default of 0.  Personally I would reset to an empty string since that is the way the object would be default constructed but I think @kbenne or @mbadams5 might have different views.

openstudiocore/src/model/AirLoopHVAC.cpp
```diff
@@ -1741,6 +1760,24 @@ void AirLoopHVAC::autosizeDesignSupplyAirFlowRate() {
...
 
+
+
+/** Presure Drop Adjustments **/  
+double AirLoopHVAC::baselinePressureDropAdjustment() const {
+  return getImpl<detail::AirLoopHVAC_Impl>()->baselinePressureDropAdjustment();
+}
+
+bool AirLoopHVAC::setBaselinePressureDropAdjustment(double baselinePressureDropAdjustment) {
+  getImpl<detail::AirLoopHVAC_Impl>()->setBaselinePressureDropAdjustment(baselinePressureDropAdjustment);
```
macumber: You have to return a value unless the function is void, C++ does not do the ruby "return the last evaluated thing" thing:

    return getImpl<detail::AirLoopHVAC_Impl>()->setBaselinePressureDropAdjustment(baselinePressureDropAdjustment);

openstudiocore/src/model/AirLoopHVAC.hpp
```diff
@@ -89,6 +89,14 @@ class MODEL_API AirLoopHVAC : public Loop
   void resetDesignSupplyAirFlowRate();
 
   void autosizeDesignSupplyAirFlowRate();
+  
+  
+  /** Baseline Pressure Drop Adjustment **/
+  double baselinePressureDropAdjustment() const;
```
macumber: This API looks correct

openstudiocore/resources/osversion/1_10_4/OpenStudio.idd
jmarrec: It does not get translated to E+ at all as pointed out in the first commit message. This is just needed for the baseline automation work

openstudiocore/src/model/AirLoopHVAC.cpp
jmarrec: @mbadams5 is the one that told me to ditch the boost::optional thing?!

mbadams5: I did tell you to ditch it for the return type, but that doesn't mean that `getDouble` does not return a boost::optional.

macumber: He was correct, the boost optional should not be in the public interface.  But inside the function (which no one else can see) you do have to use boost optional, you just call .get on it before you return it to the user.  The user never sees the optional.

jmarrec: Ah I see. Well I actually started by copying word for word this:

```
  boost::optional<double> AirLoopHVAC_Impl::designSupplyAirFlowRate() const {
    return getDouble(OS_AirLoopHVACFields::DesignSupplyAirFlowRate,true);
  }
```

openstudiocore/resources/osversion/1_10_4/OpenStudio.idd
macumber: Hmm, I'm not sure @kbenne is gonna go for that.  I'll get out of the way and let you guys discuss.  @asparke2 should also chime in.

asparke2:
My 2 cents: this field is necessary for baseline automation, but I think we need to follow the pattern of adding the field to a standards object attached to the airloop instead of directly to the OS object.  The main worry is that people will see this field and think that it impacts the E+ simulation (which it does not), and will not be able to find documentation for it in the E+ I/O ref.  One day, we need to implement something in the GUI that shows which fields (like this) are not used by E+ directly in the simulation.

kbenne:
I agree with @asparke2. I think he can point you to a place in the standards ruby code to put this. It would look pretty much like what you have here but implemented in Ruby.

openstudiocore/src/model/AirLoopHVAC.cpp
kbenne: No I agree with you here @macumber. Reset to empty field so you are back to the state you constructed at.

jmarrec:
Ok, you guys can take care of it then, I don't know how to do that yet and don't have time to learn right now.

I had also thought about using a separate object (OS:BaselinePresureDrop or whatever) where you could define in more detail the type of filter and stuff like that, and have it reference the airloop.


As far as how to display this information that doesn't affect the simulation, I think kyle can help setup some colored field like the dreaded blue ones :)

jmarrec:
Wait, are you talking about just doing this in pure ruby in the openstudio-standards? just adding an attribute in Standards.AirLoopHVAC and have a getter and a setter?

kbenne:
@jmarrec well yeah i was saying that, but it doesn't work because you need to save this information.

jmarrec:
Right, what I was thinking too. Also, won't be possible to display that in the GUI, which I hope one day would be feasible...

kbenne:
Just so the thread is aware of the back conversation. We are going to abandon this PR because to do this we want to add an entirely new standards object to support AirLoopHVAC. That is a lot more work then anyone has time or appetite for right now. 

Not to be dismayed, I have swindled @jmarrec into adding new E+ 8.5 fields to the pump object!

jmarrec:
I've never signed anything, verbal contracts don't exist my side of the ocean.

But yes, I'll try to implement the design watts per flow thing, since I request it already on [uservoice](https://energyplus.uservoice.com/forums/258860-energyplus/suggestions/9592662-add-the-possibility-to-define-design-watts-per-flo)

PR #2110, opened 2016-03-01
2088 xml export measure names

### LINKED ISSUE #1349: Measure Groups Rename on Reorder
Tested with 1.6.0:

1. Create a PAT project
2. Create 2 measure groups (model measures)
3. Rename the measure groups
4. Reorder the measure groups.
Result:
The names you entered got lost

### LINKED ISSUE #2088: PAT XML export is not exporting user-defined names
Reported by Elizabeth.  Her PAT project has nice names in the GUI, but on export, everything in the XML has the default names like "Measure Whatever Alternative 1."

Suspect that this was introduced during the cleanup of name vs. displayName that @macumber did when adding user-defined Measures.  I should have caught this during testing.

asparke2:
Fixes #2088, fixes #1349.  Needs to be in 1.10.5 for Elizabeth.

### DIFF
--- a/openstudiocore/src/pat_app/ExportXML.cpp
+++ b/openstudiocore/src/pat_app/ExportXML.cpp
@@ -84,7 +84,7 @@ bool ExportXML::exportXML(const analysisdriver::SimpleProject project, QString x
   analysis::DataPointVector dataPoints = analysis.dataPoints();
   std::vector<std::string> alternativeNames;
   for (const analysis::DataPoint& datapoint : dataPoints) {
-    alternativeNames.push_back(datapoint.name());
+    alternativeNames.push_back(datapoint.displayName());
   } 
 
   // open modal dialog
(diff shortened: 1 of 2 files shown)


openstudiocore/src/pat_app/ExportXML.cpp
```diff
@@ -137,10 +137,10 @@ bool ExportXML::exportXML(const analysisdriver::SimpleProject project, QString x
       if ( boost::optional<QDomElement> alternativeElem = exportAlternative(doc, *reportAttr, datapoint, jobs, edaBaselineName, proposedBaselineName, certificationBaselineName) ) {
         alternativesElem.appendChild(*alternativeElem);
       } else{
-        errors << toQString(datapoint.name());
+        errors << toQString(datapoint.displayName());
       }
     }else{
-      errors << toQString(datapoint.name());
+		errors << toQString(datapoint.displayName());
```
macumber: @asparke2 looks like you are committing tabs instead of two spaces

Issue #2005, opened 2015-12-16
Run Simulation Tab Says "Hello World" When Using Apply Measure Now


pflaumingo:
When in the 'Run Simulation' tab if you use 'Apply Measure Now' the screen goes blank and displays "Hello World" - I've tried this with a few measures. Clicking on any other tab seems to fix the issue so it's a very minor bug, but thought I would note it.
![addreportvariablehelloworld](https://cloud.githubusercontent.com/assets/15233892/11851758/31e8ec16-a3e9-11e5-9fb4-afcc39539468.PNG)

macumber:
Weird, possibly related to this code:

https://github.com/NREL/OpenStudio/blob/2d29315446347a3f46ea04996f5ab902ef18b0c9/openstudiocore/src/openstudio_lib/MainTabView.cpp#L55

https://github.com/NREL/OpenStudio/blob/3b0d232b5f99a2a331804a8f9762f32a99b1a56b/openstudiocore/src/shared_gui_components/EditView.cpp#L423

MatthewSteen:
[easter egg](https://en.wikipedia.org/wiki/Easter_egg_(media)) from the developers? what happens if you enter :arrow_up: :arrow_up: :arrow_down: :arrow_down: :arrow_left: :arrow_right: :arrow_left: :arrow_right: B A?

evanweaver:
@MatthewSteen :+1:

MatthewSteen:
In 1.10.4 this also occurs in the Spaces tab if the user accepts the changes when using Apply Measure Now.

evanweaver:
Thanks, I'm investigating that general issue now. It is an architectural issue related to using Apply Measures Now.

macumber:
I see that the hello world message is gone.  However, when I start the app it doesn't look right, when i click on the one button that does show the rest of the app suddenly loads

![3-3-2016 10-07-09 am](https://cloud.githubusercontent.com/assets/1327850/13502687/32403700-e129-11e5-822f-fffc5bd2844f.jpg)

PR #2111, opened 2016-03-01
[#114600781] Run Simulation Tab Says "Hello World" When Using Apply M…

evanweaver:
…easure Now #2005


openstudiocore/src/openstudio_lib/OSDocument.cpp
```diff
@@ -604,6 +598,11 @@ namespace openstudio {
 
   void OSDocument::createTab(int verticalId)
   {
+    if (m_runningApplyMeasureNow) {
+      m_runningApplyMeasureNow = false;
```
macumber: This seems to get reset when accepting changes during apply measure but not if you hit cancel on the dialog

PR #2114, opened 2016-03-02
Standard report area check tol

DavidGoldwasser:
@asparke2 now only warns if off by 1 m^2 or more.


DavidGoldwasser:
@evanweaver any thoughts on failed test, just .rb and .xml changed.

evanweaver:
@DavidGoldwasser this is really the bailiwick of @lefticus and perhaps @myoldmopar. For your purposes, I wouldn't worry about zero tests being run, but thanks for asking.

PR #2121, opened 2016-03-03
Help cmake find patch in its new location

lefticus:


lefticus:
re @Myoldmopar this patch is needed because the CI build machines use a newer version of git for windows.

PR #2130, opened 2016-03-08
Escaping newline and carriage return as suggested by @eringold

macumber:


macumber:
We need to test running files with these characters in the IDF, I am not sure if they will break reporting measures

@DavidGoldwasser can you test running an OSM with commas, semicolons, and exclamation points in the name in 1.10.5 when it comes out?

@eringold FYI

DavidGoldwasser:
@macumber I'll do that.I know apostrophes in object names breaks the SQL query, there is an issue filed for that.

This seems related to issue #1597

eringold:
Thanks, Dan!

Issue #2081, opened 2016-02-18
Dropping Schedule


Machbo:
I am losing a schedule name on a Service Hot Water Fixture when running a simulation in OS.  When I enter it into the idf, it shows up in OS, but I lose it again every time I run the simulation.

jmarrec:
Check your schedule's TypeLimits

Machbo:
I'm just using one of the occupancy schedules to test.  It is fractional and between 0 and 1 at all times.

Machbo:
I can't see any reason for this happening, I believe it to be a bug in OS,

Machbo:
I have just run the EP idf editor and entered a name for the schedule and then run the simulation in EP.  It ran through with no problem.

asparke2:
@Machbo can you upload your .osm file so we can reproduce this?

Machbo:
This issue was resolved at Unmethours by @kbenne.
https://unmethours.com/question/15131/i-cant-get-a-service-hot-water-plant-loop-to-run/?answer=15143#post-id-15143

kbenne:
Resolved yes, but I think I'm going to make some changes to make things more robust. I'm going leave this issue open.

Machbo:
Thanks Kyle, I have it up and running and don't need anymore help. I understand what you did and actually figured out how I made it happen. In the facility tab, I somehow managed to get the Always on Discrete schedule in an exterior equipment load. In that area the fraction is editable I believe. 

Thanks again, 
Mark 

----- Original Message -----

macumber:
@kbenne you want to tag this with 1.11?

PR #2135, opened 2016-03-09
2081 op scheme schedule

kbenne:
This branch is derived from the 8.5 update branch so hold the merge until after 8.5 is in develop. 

The commit log stinks. This change just makes sure the always on schedule is translated into the idf file before the plant op scheme list tries to use it.

See #2081 for details of the bug.


macumber:
@kbenne is this the fix that has a version translator issue in it?  we want to wait until 1.10.6 or 1.11.0?

PR #2143, opened 2016-03-10
Fix bug translating heat pump to idf format

### LINKED ISSUE #2043: CoilHeatingWaterToAirHeatPump:EquationFit forward translator sets hard sized rated water flow rate as air flow rate
Found this in v1.10.0, Windows x64. Looks like this is happening [here](https://github.com/NREL/OpenStudio/blob/a477c2f00aa05e37813d0aedab3ddbaf8d2881be/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilHeatingWaterToAirHeatPumpEquationFit.cpp#L112). This hurts the simulation. 

@kbenne @macumber

kbenne:
close #2043

@eringold take a look and see if you think this takes care of it.


eringold:
Thanks, that's how I thought it would be fixed.

PR #2148, opened 2016-03-15
Translate ZoneAirContaminantBalance

### LINKED ISSUE #2109: OS:ZoneAirContaminantBalance not translated
It appears that this issue is back based on this [unmethours post](https://unmethours.com/question/15137/cannot-enable-co2-concentration-calculation-in-openstudio/) from Tiejun Wu. I confirmed the issue in 1.10.0. The OS:ZoneAirContaminantBalance object shows up in the .osm, but not in the exported IDF. Also, for this particular object, the string input field is no go; should really be a dropdown.

kbenne:
close #2109

### DIFF
--- a/openstudiocore/src/energyplus/ForwardTranslator.cpp
+++ b/openstudiocore/src/energyplus/ForwardTranslator.cpp
@@ -355,6 +355,14 @@ Workspace ForwardTranslator::translateModelPrivate( model::Model & model, bool f
     globalGeometryRules.setString(openstudio::GlobalGeometryRulesFields::RectangularSurfaceCoordinateSystem, "Relative");
     m_idfObjects.push_back(globalGeometryRules);
 
+    {
+      auto zoneAirContaminantBalances = model.getConcreteModelObjects<ZoneAirContaminantBalance>();
+      for( auto & balance : zoneAirContaminantBalances ) {
+        // We should only have one of these, but the ModelObject is not unique, so there could be multiple
+        translateAndMapModelObject(balance);
+      }
+    }
+
     // create meters for utility bill objects
     std::vector<UtilityBill> utilityBills = model.getConcreteModelObjects<UtilityBill>();
     for (UtilityBill utilityBill : utilityBills){


openstudiocore/src/energyplus/ForwardTranslator.cpp
```diff
@@ -355,6 +355,14 @@ Workspace ForwardTranslator::translateModelPrivate( model::Model & model, bool f
     globalGeometryRules.setString(openstudio::GlobalGeometryRulesFields::RectangularSurfaceCoordinateSystem, "Relative");
     m_idfObjects.push_back(globalGeometryRules);
 
+    {
+      auto zoneAirContaminantBalances = model.getConcreteModelObjects<ZoneAirContaminantBalance>();
```
macumber: @kbenne I didn't understand why the following lines were not working:

https://github.com/NREL/OpenStudio/blob/2109-trans-contaminant-balance/openstudiocore/src/energyplus/ForwardTranslator.cpp#L2727

https://github.com/NREL/OpenStudio/blob/2109-trans-contaminant-balance/openstudiocore/src/energyplus/ForwardTranslator.cpp#L400

Did you test that this change works?  It seems like it would be the same to me.

kbenne: Thats a good question. Yes I tested that it now works and you only get one instance in the idf file. The question is was it already working before i did this and if it wasn't then why not.

I generally don't like that catch all, but this seems like a good application of it.

macumber: This wasn't the catch all, OS_ZoneAirContaminantBalance was explicitly added to the list of objects to translate:

    result.push_back(IddObjectType::OS_ZoneAirContaminantBalance);

I wonder if the issue was the app, if you enter a schedule name in that text box, it isn't saved to the OSM (since it is not a real schedule) but it looks like it is there until you change tabs

macumber: I made an OSM in 1.10.5 and exported to IDF, the ZoneAirContaminantBalance object came through.

kbenne: well maybe we reject this PR and take no action?

macumber: That's what I am thinking, the real problem is the UI

PR #2151, opened 2016-03-15
Fix translation of fluid to fluid HX sepoint nade

### LINKED ISSUE #2037: HeatExchanger:FluidToFluid problems
Can you please either:

1. Expose the 'Heat Exchanger Setpoint Node Name' field (allow for greater flexibility I guess...)
2. Do a smarter forward translation: if I didn't supply a setpointManager on the outlet of the HX, then you should assume it's the Loop Supply Side Outlet Node and not the HX Supply Side Outlet because that'll crash E+, like you do perfectly for components such as chillers.

Thanks!

@kbenne

kbenne:
If there is no spm on the hx outlet node, then set the
HeatExchangerSetpointNode to the plant outlet node

close #2037 

@jmarrec is this the right change


jmarrec:
I guess that would do it for most situations while being in accordance with the rest of the app (the field isn't truly exposed anywhere), so I'm going to say yes!

Envoyé de mon iPhone

PR #2154, opened 2016-03-16
Fix failing IDF import

### LINKED ISSUE #2153: Fails to import IDF when CondenserLoop is part of IDF
The proposedEnergy+.idd does not have the updated key choices for CondenserLoop, Load Distribution Scheme. The other `Load Distribution Scheme` fields have been updated, except for this one (since it is not an OpenStudio object).

This causes the Import IDF functionality to fail due to invalid draft strictness, even though HVAC objects are not imported into OSM.

mbadams5:
Fixes #2153


kbenne:
thanks @mbadams5

PR #2156, opened 2016-03-16
Fix crash after apply measure now

### LINKED ISSUE #2155: 1.10.5 closes after apply measure now
Steps to reproduce:

1. Navigate to Thermal Zone tab
2. Apply Measure Now
3. Accept Changes
4. Click a string field, e.g. thermal zone name or zone equipment

Model will unexpectedly close. May be related to the measure I ran which renames zone equipment based on zone name, which I can provide if needed.

kbenne:
close #2155


macumber:
Passes my limited test, I say we merge it for testing in 1.10.6

PR #2157, opened 2016-03-17
Replace QUUID with boost::uuids::uuid

lefticus:
@kbenne note: aparently boost::uuid has two problems

  1. Abhorent speed if you recreate the generator on each call
  2. default constructor makes no guarantee on value, nil generator
     needs to be used instead

We fix those issues by making our own UUID type that inherits
from boost::uuids::uuid and sets the default construction.

We also make global thread safe generators that can be called as
necessary. That code can be moved to C++11's thread_local at some point
in the future if you move to MSVC2015 as a requirement.

Important notes:

  * The built in QUUID::toString was leading to much duplicated and
    inefficient string/qstring/uuid conversions. This patch reduces that
  * We rename methods as necessary to get it looking as much like QUUID as
    possible.

### DIFF
--- a/openstudiocore/src/utilities/core/UUID.i
+++ b/openstudiocore/src/utilities/core/UUID.i
@@ -12,16 +12,13 @@
 namespace openstudio{
 
   #ifdef SWIGRUBY
-    %rename("nil?") isNull();
+    %rename("nil?") is_nil();
   #endif
 
   class UUID {
   public:
-    bool operator!= ( const UUID & other ) const;
-    bool operator< ( const UUID & other ) const;
-    bool operator== ( const UUID & other ) const;
-    bool operator> ( const UUID & other ) const;
-    bool isNull () const;
+    bool is_nil() const;
+    bool isNull() const;
     ~UUID();
 
   protected:
@@ -41,7 +38,27 @@ namespace openstudio{
     }
     
     std::string __str__() const{
-      return openstudio::toString(self->toString());
+      return openstudio::toString(*self);
+    }
+
+    bool isNull() const {
+      return self->is_nil();
+    }
+
+    bool operator!= ( const UUID & other ) const {
+      return *self != other;
+    }
+
+    bool operator< ( const UUID & other ) const {
+      return *self < other;
+    }
+
+    bool operator== ( const UUID & other ) const {
+      return *self == other;
+    }
+
+    bool operator> ( const UUID & other ) const {
+      return *self > other;
     }
   };
 
--- a/openstudiocore/src/utilities/core/UUID.i
+++ b/openstudiocore/src/utilities/core/UUID.i
@@ -12,37 +12,32 @@
 namespace openstudio{
 
   #ifdef SWIGRUBY
-    %rename("nil?") is_nil();
+    %rename("nil?") isNull();
   #endif
 
   class UUID {
   public:
-    bool is_nil() const;
     bool isNull() const;
     ~UUID();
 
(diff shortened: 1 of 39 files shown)


kbenne:
Thank you @lefticus. We are wrapping up the 1.11.0 release right now, but as soon as that is out I want to merge this before it gets stale.

openstudiocore/src/utilities/core/UUID.i
```diff
@@ -12,16 +12,13 @@
 namespace openstudio{
 
   #ifdef SWIGRUBY
-    %rename("nil?") isNull();
+    %rename("nil?") is_nil();
```
macumber: Does this work?

openstudiocore/src/utilities/core/UUID.i
```diff
@@ -12,16 +12,13 @@
...
   #endif
 
   class UUID {
   public:
-    bool operator!= ( const UUID & other ) const;
-    bool operator< ( const UUID & other ) const;
-    bool operator== ( const UUID & other ) const;
-    bool operator> ( const UUID & other ) const;
-    bool isNull () const;
+    bool is_nil() const;
```
macumber: I'd prefer we only export one of these, let's just stick with isNull

openstudiocore/src/utilities/core/UUID.i
lefticus: I cannot imagine why it wouldn't work? It's been that way for eons AFAICT. All I did was change the name to match the new uuid class.

lefticus: Plus, I of course verified that all existing ruby uuid related unit tests still pass.

macumber: I was originally confused thinking:

    if !model_object.handle

would work, but I forgot that Ruby doesn't implicitly call `.nil?` in that case.  As for the name, I've never called `.nil?`, I always use `isNull`, so we should keep all of those.  What is the benefit of also adding `is_nil`?  Is there any reason to add a third method for that?

lefticus: `is_nil` is the native name for the `boost::uuids::uuid` "check if null" method. So I added it just for completeness. We can probably safely remove it.

The rest of the API remains unchanged from the ruby perspective.

nrel-bot-2:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-2:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

PR #2166, opened 2016-03-22
Correctly display construction inheritance in the gridview, and apply…

evanweaver:
… new policies to resource objects to allow viewing (but not editing) in Inspector Gadget.


macumber:
@evanweaver this looks good, it would be good to be able to disable the "Add/Remove Extensible Groups" using IG policy but I don't think that is possible, right @kbenne ?

PR #2170, opened 2016-03-24
Red hat build

macumber:


dependencies/Boost/gcc_5_no_cxx11.patch
```diff
@@ -0,0 +1,37 @@
+From eec808554936ae068b23df07ab54d4dc6302a695 Mon Sep 17 00:00:00 2001
```
macumber: @lefticus I applied this patch to get Boost 1.55 to build with gcc 5.2 on RedHat 6.6.  Do you see any issues applying this patch on Ubutnu too?

@axelstudios @kbenne FYI

openstudiocore/src/utilities/geometry/PointLatLon.cpp
```diff
@@ -20,9 +20,9 @@
 #include "PointLatLon.hpp"
 #include "Point3d.hpp"
 
-#include "../../GeographicLib/Geodesic.hpp"
-#include "../../GeographicLib/LocalCartesian.hpp"
-#include "../../GeographicLib/UTMUPS.hpp"
+#include <GeographicLib/Geodesic.hpp>
```
macumber: @axelstudios GeographicLib is built and installed in the build directory, I'm not sure how this was working anywhere

@kbenne FYI

dependencies/Boost/gcc_5_no_cxx11.patch
lefticus: One question: why was this necessary? gcc 5.2 supports C++11/14/17

And to answer your question, I'd strongly recommend against applying where not needed because it could have a significant impact on compile times (variadic templates are much faster than the alternative).

macumber: [This is the error ](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=197594) I was seeing

When I tried to apply the patch in that first report it failed for some reason that I can't remember.  I found the other patch [here](https://github.com/bitcoin/bitcoin/pull/6280/files) and it fixed my compile errors.  I think we need to update boost anyway pretty soon.  If compile time is the only downside then I'd be inclined to merge this and then delete it when we upgrade boost.

openstudiocore/src/utilities/geometry/PointLatLon.cpp
macumber: @evanweaver

axelstudios: ¯\_(ツ)_/¯

nrel-bot-3:

PR #2173, opened 2016-03-28
112181489 remove q file from open studio libraries readyaimfire 3

lefticus:
With these commits all filesystem access is organized in the `openstudio::filesystem` namespace. Unit tests pass (as well as they did before the changes).

This code is primed for moving to `std::filesystem` when possible. It currently relies on `boost::filesystem` but all code was written while referencing the documentation for `std::filesystem` and only very minor issues were found. These issues have been accounted for for future use.

Notes
 - Some QFile removals had to be reversed because they were references
   to embedded internal files
 - Had to remove usage of file created timestamp. However, I don't think
   this is actually used anywhere and the entire FileReference class can
   probably be removed
 - All required FilesystemHelpers implemented


openstudiocore/src/utilities/core/FileReference.cpp
```diff
@@ -150,7 +150,7 @@ void FileReference::setPath(const openstudio::path& newPath) {
 bool FileReference::makePathAbsolute(const openstudio::path& searchDirectory) {
```
kbenne: I'm curious about this FileReference class in general. I wonder who is using it. Can its function be rolled into the new openstudio::filesystem?

openstudiocore/src/utilities/core/Path.hpp
```diff
@@ -22,11 +22,12 @@
```
kbenne: OK i think the same comment applies. Can we do some consolidation here? Path, PathHelpers, Tempdir... I would like one place (namespace, source code directory) where all file related things are at, with no duplicity or confusion. I think I'm going to have to review this body of code outside of the PR. It is hard to see the big picture from the PR diff. BUT if there is consolidation and cleanup that needs to happen I do not think it needs to be part of this PR or @lefticus task.

openstudiocore/src/utilities/data/Test/CalibrationResult_GTest.cpp
```diff
@@ -30,8 +30,8 @@ TEST_F(DataFixture, CalibrationResult)
 {
 
   openstudio::path xmlPath = toPath("./calibration_report.xml");
-  if(boost::filesystem::exists(xmlPath)){
```
kbenne: This is like OpenStudio archaeology. I bet we can trim some of this stuff out of utilities for OS 2.0.

openstudiocore/src/utilities/core/FileReference.cpp
macumber: FileReference should be deprecated, I think it was mostly used in the project database

openstudiocore/src/utilities/core/Path.hpp
macumber: All these things are mostly free function helpers that could be added into an OpenStudio Path class or a single header with free helper methods if we don't wrap filesystem

openstudiocore/src/utilities/core/FileReference.cpp
kbenne: I hope by deprecated you mean we can eliminate it. I think we'll just have to see how tangled it really is.

openstudiocore/src/utilities/data/Test/CalibrationResult_GTest.cpp
macumber: What's wrong with this?

openstudiocore/src/utilities/core/FileReference.cpp
macumber: Yeah kill it

openstudiocore/src/airflow/contam/ForwardTranslator.cpp
```diff
@@ -434,10 +434,18 @@ boost::optional<contam::IndexModel> ForwardTranslator::translateModel(model::Mod
   m_logSink.setThreadId(QThread::currentThread());
   m_logSink.resetStringStream();
 
-  m_prjModel.read(std::string(":/templates/template.prj"));
-  if (!m_prjModel.valid()) {
-    return boost::none;
+  {
+    QFile f(":/templates/template.prj");
```
kbenne: @lefticus What is going on here? one would assume you are on the pursuit of getting rid of QFile, but here obviously you can't since a q resource is involved. Are you trying to stage for the future in some way i don't entirely understand.

openstudiocore/src/airflow/contam/ForwardTranslator.cpp
```diff
@@ -434,10 +434,18 @@ boost::optional<contam::IndexModel> ForwardTranslator::translateModel(model::Mod
...
+    f.open(QFile::ReadOnly);
+    QString s = f.readAll();
+    Reader r(&s);
```
kbenne: New adaptor class?

openstudiocore/src/airflow/contam/ForwardTranslator.cpp
macumber: It would be nice to have something to replace embedded resources, although if we can't have that we could use resources on disk.  That makes openstudio.exe less self contained though

openstudiocore/src/utilities/core/Path.hpp
lefticus: One note of caution regarding rolling some of these things together. Some of the path.hpp stuff was carefully crafted so that it could be included in GenerateIddFactory and other top level things and not accidentally cause rebuilds of the entire system & have few dependencies. This is one reason I only very lightly touched those headers.

openstudiocore/src/airflow/contam/ForwardTranslator.cpp
lefticus: I had to explicitly change this to create a QFile because it (`read` function) was previously creating a QFile internally and I removed that logic from PrjModel. So this limits the use of QFile to exactly the one function that cares about it.

openstudiocore/src/airflow/contam/ForwardTranslator.cpp
lefticus: Nope, Reader was there all along, but Reader no longer deals in terms of QFiles.

lefticus: Reader is used inside of the `.read` method of m_prjModel

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-2:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @lefticus @lgentile it has been 7 days since this pull request was last updated.

PR #2202, opened 2016-04-21
Adds symlinks to openstudio-standards gem for Mac

asparke2:
@evanweaver I think the only way to test this change is to create and run an installer on Mac.  Can you create one and @DavidGoldwasser will test?


macumber:
@kbenne you have any thoughts on this?

asparke2:
@macumber @kbenne is the one who suggested this.  I don't know bout no symlinks.

macumber:
Hehe ok by me :-)

kbenne:
Well this was basically my suggestion. One thing @mbadams5 and I were talking about is how this plays out on el capitan with tighter security. I think the general idea is ok as a bandaid, but maybe we should be linking into some site_ruby directory instead. We've already been linking openstudio.rb for years so...

evanweaver:
Let's give it a shot.

PR #2214, opened 2016-05-03
Adding object Site:GroundTemperature:Deep and others

nmerket:
@macumber @shorowit

This should address [this issue](https://github.com/NREL/OpenStudio-Beopt/issues/31#issuecomment-212463242) for BEopt, and [this pivotal ticket](https://www.pivotaltracker.com/story/show/118043955). I'm adding classes for `Site:GroundTemperature:FCfactorMethod`, `Site:GroundTemperature:Shallow`, and `Site:GroundTemperature:Deep`.

### `Site:GroundTemperature:Deep`
- [x] add new classes
- [x] unit tests
- [x] forward translator
- [x] add to `Site::children`

### Update `Site:GroundTemperature:BuildingSurface`
- [x] update classes
- [x] unit tests

### `Site:GroundTemperature:Shallow`
- [x] add new classes
- [x] unit tests
- [x] forward translator
- [x] add to `Site::children`

### `Site:GroundTemperature:FCFactorMethod`
- [x] add new classes
- [x] unit tests
- [x] forward translator
- [x] add to `Site::children`


openstudiocore/src/model/SiteGroundTemperatureDeep.hpp
```diff
@@ -0,0 +1,183 @@
...
+namespace model {
+
+namespace detail {
+
+  class SiteGroundTemperatureDeep_Impl;
+
+} // detail
+
+/** SiteGroundTemperatureDeep is a ModelObject that wraps the OpenStudio IDD object 'OS_Site_GroundTemperature_Deep'. */
+class MODEL_API SiteGroundTemperatureDeep : public ModelObject {
```
nmerket: @macumber this is mostly a copy of the `SiteGroundTemperatureBuildingSurface` files. I plan to add two more. I'm thinking it would be beneficial to have a base class that implements all these methods that all four of the models can derive from instead of duplicating them all. I'm a little fuzzy on how to implement all that, though. Is this being done somewhere else in OpenStudio that I could copy the pattern?

openstudiocore/src/model/Model.cpp
```diff
@@ -1786,6 +1788,21 @@ void addExampleModelObjects(Model& model)
   groundTemp.setNovemberGroundTemperature(19.802);
   groundTemp.setDecemberGroundTemperature(19.633);
 
+  // add SiteGroundTemperatureDeep
```
macumber: @nmerket this is good but not required

openstudiocore/src/model/SiteGroundTemperatureDeep_Impl.hpp
```diff
@@ -0,0 +1,198 @@
...
+#include "ModelObject_Impl.hpp"
+
+namespace openstudio {
+namespace model {
+
+namespace detail {
+
+  /** SiteGroundTemperatureDeep_Impl is a ModelObject_Impl that is the implementation class for SiteGroundTemperatureDeep.*/
+  class MODEL_API SiteGroundTemperatureDeep_Impl : public ModelObject_Impl {
+    Q_OBJECT;
```
macumber: We do not want to use Q_OBJECT and Q_PROPERTY for new objects as we are reducing our dependence on Qt

openstudiocore/src/model/CMakeLists.txt
```diff
@@ -1400,6 +1403,7 @@ set(${target_name}_moc
   Site_Impl.hpp
   SiteGroundReflectance_Impl.hpp
   SiteGroundTemperatureBuildingSurface_Impl.hpp
+  SiteGroundTemperatureDeep_Impl.hpp
```
macumber: Should remove this line after removing the Q_OBJECT and Q_PROPERTY macros

openstudiocore/src/model/SiteGroundTemperatureDeep.hpp
macumber: GeneratorPhotovoltaic inherits from Generator and is a recent class that implements inheritence

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/GeneratorPhotovoltaic.hpp

macumber: Would the base class just have methods to get monthly temperatures?

macumber:
@nmerket nice job, this is a good looking pull request.  One thing that is missing is a unit test, can you copy and modify one of the GTests in the tests folder?  You will have to add the test file to CMakeLists.txt

mbadams5:
@nmerket You also need to do the forward translator so this gets translated into EnergyPlus. 

Also I know @kbenne has been advocating for fewer defaults and instead setting those default values in the constructor instead, but i'll let him chime in.

openstudiocore/src/model/SiteGroundTemperatureDeep.hpp
nmerket: I would think it would have getters and setters and maybe some other convenience methods such as being able to get or set the temperatures as a vector or have a getter/setter for each month by a number.

nmerket:
@mbadams5 would the lack of a forward translator explain why I'm getting build failures?

mbadams5:
@nmerket No, your build errors are for a different reason. You are accessing the wrong field in your getters/setters.

The fields have this text:
`N1 , \field January Deep Ground Temperature`
Which will get translated to `JanuaryDeepGroundTemperature` for the field accessors.

You have `JanuaryGroundTemperature` currently.

kbenne:
Re: "Also I know @kbenne has been advocating for fewer defaults and instead setting those default values in the constructor instead, but i'll let him chime in."


Yes I see a future where OpenStudio may not be backed for by IDD. In OpenStudio I like to see E+ fields that have defaults turned into \required-field and then initialize the field to a default value in the C++ constructor. I think this gives us more flexibility in the future. Also using required field eliminates all of that boost optional nonsense making the API more user friendly.

openstudiocore/src/model/Model.cpp
```diff
@@ -1786,6 +1788,21 @@ void addExampleModelObjects(Model& model)
...
+  SiteGroundTemperatureDeep groundTempDeep = model.getUniqueModelObject<SiteGroundTemperatureDeep>();
```
kbenne: I think this is good material for the default constructor.

nmerket:
It sounds like the `SiteGroundTemperatureBuildingSurface` class I am basing this on is using some pretty out-of-date paradigms in OpenStudio land. Would the `GeneratorPhotovoltaic` be a good example to see a more modern example, or should I look at somethings else?

kbenne:
Well we haven't established hard rules about \defuault. I just try to avoid it if I can. [Absorption Chiller](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/ChillerAbsorption.hpp) is pretty modern in my opinion. Yes there are optionals returned for the autosized fields, but the most of the fields are simply required fields initialized in the ctor.

openstudiocore/src/model/Model.cpp
macumber: Adding these numbers to the example model is fine (although as I said before) it isn't necessary to add every object to the example model.

Adding defaults to the constructor is a little more dangerous.  Unless you are sure that these numbers are better than E+ defaults for almost all models I would stick with using E+ defaults just for traceability.

macumber:
Putting \default in the IDD or making the fields required and initializing them in the constructor are two options each with pros/cons.  If you get rid of the IDD default then you have to put the number in the constructor AND in the reset methods and there is no way to know if the field is defaulted.  The OS GUIs have some code to show defaulted fields in green.

In either case your accessors in the C++ public class are correct, we are just quibbling about how to define the IDD and where to put the defaults.  I think Kyle and I have different opinions on this topic.

macumber:
In this case I'd suggest sticking to patterns in the existing similar object, that ok to you @kbenne ?

nmerket:
I think this is good for `SiteGroundTemperatureDeep`. I wanted to get all your eyes on it before I update `SiteGroundTemperatureBuildingSurface` to be the same and add `SiteGroundTemperatureShallow` and `SiteGroundTemperatureFCFactorMethod`. Right now I'm leaning towards not creating a base class for these four.

openstudiocore/src/model/SiteGroundTemperatureDeep.cpp
```diff
@@ -0,0 +1,653 @@
...
+      case 3 : retval = marchDeepGroundTemperature(); break;
+      case 4 : retval = aprilDeepGroundTemperature(); break;
+      case 5 : retval = mayDeepGroundTemperature(); break;
+      case 6 : retval = juneDeepGroundTemperature(); break;
+      case 7 : retval = julyDeepGroundTemperature(); break;
+      case 8 : retval = augustDeepGroundTemperature(); break;
+      case 9 : retval = septemberDeepGroundTemperature(); break;
+      case 10 : retval = octoberDeepGroundTemperature(); break;
+      case 11 : retval = novemberDeepGroundTemperature(); break;
+      case 12 : retval = decemberDeepGroundTemperature(); break;
```
nmerket: I'd like to throw an exception in a default clause here in case someone enters an invalid month (i.e. 13). Would I do that with `OS_ASSERT` or something else?

macumber: OS_ASSERT shouldn't be used for user input validation, you could use the [LOG_AND_THROW](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/core/Logger.hpp#L48) macro to log the error and throw an exception. 

Another option would be to return an optional double but I agree that sucks.

Final note is that you should provide an override that takes a [MonthOfYear](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/time/Date.hpp#L75) object.  These enum values line up with your month case statement values.  Advantage of these enums is that it is harder to pass in a bad month.  However, I do think people like being about to use a simple type for this stuff too.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateSiteGroundTemperatureDeep.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+namespace energyplus {
+
+boost::optional<IdfObject> ForwardTranslator::translateSiteGroundTemperatureDeep( SiteGroundTemperatureDeep& modelObject )
+{
+  IdfObject idfObject( openstudio::IddObjectType::Site_GroundTemperature_Deep );
+
+  m_idfObjects.push_back(idfObject);
+
+  if (!modelObject.isJanuaryDeepGroundTemperatureDefaulted()){
```
macumber: @nmerket we've started standardizing on translating the openstudio value to e+ even when it is defaulted.  That way the OS defaults are always used instead of the E+ ones

macumber: So just set all the fields no matter what :-)

openstudiocore/src/model/SiteGroundTemperatureDeep.cpp
```diff
@@ -0,0 +1,653 @@
...
+  }
+
+  void SiteGroundTemperatureDeep_Impl::resetAllMonths() {
+    for (int i = 1; i <= 12; ++i) {
+      resetTemperatureByMonth(i);
+    }
+  }
+
+  void SiteGroundTemperatureDeep_Impl::setAllMonthlyTemperatures(const std::vector<double> &monthly_temperatures) {
+    OS_ASSERT(monthly_temperatures.size() == 12);
```
macumber: I wouldn't use an ASSERT here.  Suggest returning a bool and return false if the wrong number of arguments are passed

openstudiocore/src/model/test/SiteGroundTemperatureDeep_GTest.cpp
```diff
@@ -0,0 +1,191 @@
...
+  );
+
+  // create a model to use
+  Model m;
+
+  // create a SiteGroundTemperatureDeep object
+  SiteGroundTemperatureDeep ground_temp(m);
+
+  // Check to make sure all the fields are defaulted as expected
+  ASSERT_TRUE(ground_temp.isJanuaryDeepGroundTemperatureDefaulted());
```
macumber: Minor note, you can use EXPECT_TRUE here instead, ASSERT_TRUE stops the test on failure, EXPECT_TRUE continues on

openstudiocore/src/model/test/SiteGroundTemperatureDeep_GTest.cpp
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// Test setting and getting fields
+TEST_F(ModelFixture, SiteGroundTemperatureDeep_SetGetFields)
+{
+  // Create model
+  Model m;
+
+  // Create the object
+  SiteGroundTemperatureDeep ground_temp(m);
```
macumber: In C++ we use lowerCamelCase for variable names instead of snake_case

openstudiocore/src/model/SiteGroundTemperatureDeep.cpp
```diff
@@ -0,0 +1,653 @@
...
+
+  SiteGroundTemperatureDeep_Impl::SiteGroundTemperatureDeep_Impl(const SiteGroundTemperatureDeep_Impl& other,
+                                                                                       Model_Impl* model,
+                                                                                       bool keepHandle)
+    : ModelObject_Impl(other,model,keepHandle)
+  {}
+
+  boost::optional<ParentObject> SiteGroundTemperatureDeep_Impl::parent() const
+  {
+    boost::optional<Site> result = this->model().getOptionalUniqueModelObject<Site>();
```
macumber: If Site is the parent then you should add this to Site's children:

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/Site.cpp#L97

macumber:
Looks pretty good @nmerket !  Made a few minor comments

openstudiocore/src/model/SiteGroundTemperatureBuildingSurface_Impl.hpp
```diff
@@ -30,31 +30,7 @@ namespace detail {
 
   /** SiteGroundTemperatureBuildingSurface_Impl is a ModelObject_Impl that is the implementation class for SiteGroundTemperatureBuildingSurface.*/
   class MODEL_API SiteGroundTemperatureBuildingSurface_Impl : public ModelObject_Impl {
-    Q_OBJECT;
```
mbadams5: @macumber Does the GUI still rely on these Q_PROPERTY? With the change to grid view, I don't know off hand.

macumber: No not anymore, the new GUI objects use function pointers instead of Q_PROPERTY

openstudiocore/src/model/SiteGroundTemperatureDeep.cpp
nmerket: So, would I just create a [function overload](http://www.learncpp.com/cpp-tutorial/76-function-overloading/) for that which accepts that enum then?

openstudiocore/src/model/SiteGroundTemperatureDeep.cpp
nmerket: Okay, so `false` = something went wrong and `true` = all is well? For some reason I seem to think that the usual pattern is the opposite, but as long as I'm consistent with other places in OS I can do this.

openstudiocore/src/model/SiteGroundTemperatureDeep.cpp
macumber: Yeah, the overload can just call the version that takes an int.  You can get the int value using [enum.value()](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/core/EnumBase.hpp#L90)

openstudiocore/src/model/SiteGroundTemperatureDeep.cpp
macumber: Yeah false means something went wrong and true is things worked out.  A lot of other APIs return 0 for success and an error code otherwise, we have been using exceptions instead of error codes

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateSiteGroundTemperatureDeep.cpp
nmerket: So, this is copied from `Site:GroundTemperature:BuildingSurface` should I change those to use the OS default too?

macumber: Yeah, we used to not set defaulted fields but have since learned that we should

nmerket:
This should be ready to merge. @macumber

macumber:
Thanks @nmerket I'll check it out and try to merge by tomorrow's iteration.  Show up to the iteration meeting so you can get your props :-)

openstudiocore/src/model/test/SiteGroundTemperatureBuildingSurface_GTest.cpp
```diff
@@ -0,0 +1,239 @@
...
+  groundTemp.setTemperatureByMonth(openstudio::MonthOfYear::Jan, 15.0);
+  ASSERT_EQ(groundTemp.getTemperatureByMonth(openstudio::MonthOfYear::Jan), 15.0);
+  groundTemp.resetTemperatureByMonth(openstudio::MonthOfYear::Jan);
+  ASSERT_TRUE(groundTemp.isMonthDefaulted(openstudio::MonthOfYear::Jan));
+
+  // Test exceptions on an invalid month
+  try {
+    groundTemp.getTemperatureByMonth(13);
+  } catch (openstudio::Exception const & err) {
+    ASSERT_TRUE(boost::algorithm::ends_with(err.message(), "Invalid Month 13"));
```
mbadams5: @nmerket Are you wanting to just make sure the function throws an exception? Or do you also want to make sure the error message ends with that string? If it is only the former then you can just call
`EXPECT_THROWS_ANY(groundTemp.getTemperatureByMonth(13));`

If it is the latter as well then I would encourage you to find a non-boost way to check the error message. We are trying to reduce the usages of Qt and Boost as much as possible.

nmerket: I do actually want to check that the error message ends with that string. It makes sure that it's throwing the error in the place I expect. 

Why is boost included as a dependency if we're not supposed to use it? It's all over the place in OpenStudio, including plenty of `boost::optional` in the very classes this test is testing. It seems totally arbitrary to decide it's not okay to use here.

macumber: I had not seen this before, http://en.cppreference.com/w/cpp/regex, we should probably start moving our boost regex over to std regex.

@nmerket we do distinguish between header only boost libraries (like optional) and compiled ones (like regex).  In the future we'd like to reduce our use of boost as much as possible, if we can get to only using headers we can just plop boost in our source.  Much of boost is being added to C++ std lib, if something goes from boost to std namespace we'd rather use the std one.

Maybe you could use [find_last_of](http://www.cplusplus.com/reference/string/string/find_last_of/) here?

macumber: Also to give more context, there is a whole effort on the os 2.0 branch to remove as many dependencies as possible (boost and Qt especially).  That is why @mbadams5 is sensitive to this

mbadams5: @nmerket You could try something like this...

```c++
std::string expectedErrorMsg("Invalid Month 13");
ASSERT_TRUE(std::equal(expectedErrorMsg.rbegin(), expectedErrorMsg.rend(), err.message().rbegin()));
```
I think this is functionally the same, but definitely test it.

mbadams5: @macumber
Since we are still targeting GCC 4.8.4, we can not use STL regex yet. GCC does not have a fully functioning regex implementation until 4.9.0, even though ALL function stubs are there. So many functions compile but do nothing.

nmerket:
@mbadams5 @macumber boost has been banished from my tests.

PR #2224, opened 2016-05-10
Change translation logic for ControllerMechanicalVentilation avail

### LINKED ISSUE #2223: Controller Mechanical Ventilation Availability Schedule
User reports...

> I was trying to set an availability schedule for the ControllerMechanicalVentilation object in my translator program to disable the mechanical ventilation controller, but found in the generated IDF file that the availability schedule for the ControllerMechanicalVentilation object always uses the “Minimum Outdoor Air Schedule” of its associated “Controller:OutdoorAir” object. Looks like that is done by the forward translator of the ControllerMechanicalVentilation (https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateControllerMechanicalVentilation.cpp ,  lines 66-70 ).
>  
> Is it possible to change the forward translator code so that the forward translator does not automatically override the availability schedule I set for the controllerMechanicalVentilation?

Yes I think it is possible. Tricky though. I don't like the magic happening on lines 66-70, but it is there as a result of [this E+ issue](https://github.com/NREL/EnergyPlus/issues/4633). 

I propose that we make a change to OS and go ahead and translate the ControllerMechanicalVentilation::AvailabilitySchedule as given *provided* there is not a ControllerOutdoorAir::MinimOutdoorAirSchedule given *AND* the ControllerMechanicalVentilation availability is not still set to the default always on schedule.

Basically if a user provides any ControllerMechanicalVentilation::AvailabilitySchedule other than the default always on schedule, use what is given no matter what. This probably adds slightly more trickiness to already opaque logic, but I think it captures what people want to do without them having to know an obnoxious amount of detail.

kbenne:
close #2223 

@TiejunWu what do you think?

### DIFF
--- a/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateControllerMechanicalVentilation.cpp
+++ b/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateControllerMechanicalVentilation.cpp
@@ -69,6 +69,7 @@ boost::optional<IdfObject> ForwardTranslator::translateControllerMechanicalVenti
   // Note that this scheme will not support fractions (schedule values above 0) because anything greater than 0 will
   // make the mechanical ventilation controller avaiable and thus taking precedence.
   bool useAvailabiltySchedule = true;
+  auto availabilitySchedule = modelObject.availabilitySchedule();
 
   // Find the associated oa controller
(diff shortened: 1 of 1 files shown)


nrel-bot-2:
@kbenne @lgentile it has been 7 days since this pull request was last updated.

PR #2246, opened 2016-05-26
Fix UnitVentilator bad clone

### LINKED ISSUE #2184: unit ventilator object is incomplete in 1.11.0
The ZoneHVACUnitVentilator objects do not expose the fan component in the inspector.

https://unmethours.com/question/13699/supply-fan-name-type-in-for-an-unit-ventilator/

A FanConstantVolume object is added to a model for each unit ventilator, but according to the IO Ref this object should have all three fan type options.

http://bigladdersoftware.com/epx/docs/8-4/input-output-reference/group-zone-forced-air-units.html#field-supply-air-fan-object-type-1

mbadams5:
Fixes #2184

This fixes the bad clone implementation in UnitVentilator.

This also adds more options for ZoneHVAC:UnitVentilator in the HVAC Library.


mbadams5:
@kbenne ping

PR #2248, opened 2016-05-27
Add Generator:MicroTurbine to OpenStudio

jmarrec:
@macumber: nothing but a work in progress so far...


jmarrec:
Generator:MicroTurbine must have a base class "Generator".

But if Heat Recovery is used, it also needs to be a StraightComponent.

Need to implement Generator:MicroTurbine that inherits from Generator and Generator:MicroTurbine:HeatRecovery that is a StraightComponent and has the heat recovery-related fields: the plant loop nodes and other values:

```
Generator:MicroTurbine:HeatRecovery
A7, \field Heat Recovery Water Inlet Node Name
  A8, \field Heat Recovery Water Outlet Node Name
 N12, \field Reference Thermal Efficiency Using Lower Heat Value
 N13, \field Reference Inlet Water Temperature
  A9, \field Heat Recovery Water Flow Operating Mode
 N14, \field Reference Heat Recovery Water Flow Rate
 A10, \field Heat Recovery Water Flow Rate Function of Temperature and Power Curve Name
 A11, \field Thermal Efficiency Function of Temperature and Elevation Curve Name
 A12, \field Heat Recovery Rate Function of Part Load Ratio Curve Name
 A13, \field Heat Recovery Rate Function of Inlet Water Temperature Curve Name
 A14, \field Heat Recovery Rate Function of Water Flow Rate Curve Name
 N15, \field Minimum Heat Recovery Water Flow Rate
 N16, \field Maximum Heat Recovery Water Flow Rate
 N17, \field Maximum Heat Recovery Water Temperature


```

macumber:
First of all @jmarrec you rock!  Secondly, I'll take a look at this and comment when you ask for it

jmarrec:
Ok, so I have added both new objects and wrote the forwardtranslator mostly (didn't test it)

if you have some time to look at it and see if I'm at least going in the right direction I'll take any comment you can give...

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -24728,6 +24728,282 @@ OS:SolarCollector:FlatPlate:PhotovoltaicThermal,
...
+      \min-fields 12
+  A1, \field Handle
+       \type handle
+       \required-field
+  A2, \field Name
+      \required-field
+      \type alpha
+      \reference GeneratorNames
+      \reference MicroTurbineGeneratorNames
+      \reference ConnectionObject
```
macumber: Is this a ConnectionObject?  Seems like this might be for the heat recovery part?

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -24728,6 +24728,282 @@ OS:SolarCollector:FlatPlate:PhotovoltaicThermal,
...
+  N1, \field Reference Electrical Power Output
+      \required-field
```
macumber: Anything required without a default needs to get somewhere, where will you set this?

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -24728,6 +24728,282 @@ OS:SolarCollector:FlatPlate:PhotovoltaicThermal,
...
+      \memo the HeatRecovery (StraightComponent) from the Generator:MicroTurbine(Generator)
+      \min-fields 12
+  A1, \field Handle
+      \type handle
+      \required-field
+  A2, \field Name
+      \required-field
+      \type alpha
+      \reference MicroTurbineHeatRecoveryNames
+      \reference ConnectionObject
```
macumber: @kbenne is this the correct reference group for a straight component?

openstudiocore/src/model/GeneratorMicroTurbine.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+  /** @name Getters */
+  //@{
+
+  double referenceElectricalPowerOutput() const;
+
+  double minimumFullLoadElectricalPowerOutput() const;
+  bool isMinimumFullLoadElectricalPowerOutputDefaulted() const;
+
+  boost::optional<double> maximumFullLoadElectricalPowerOutput() const;
```
macumber: I would make this return a double, if it is not set just return referenceElectricalPowerOutput like E+ does.  You would also add an isMaximumFullLoadElectricalPowerOutputDefaulted

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -24728,6 +24728,282 @@ OS:SolarCollector:FlatPlate:PhotovoltaicThermal,
...
+      \note current operating conditions
+  A6, \field Fuel Type
+      \type choice
+      \key NaturalGas
+      \key PropaneGas
+      \default NaturalGas
+  N8, \field Fuel Higher Heating Value
+      \type real
+      \units kJ/kg
+      \default 50000
```
macumber: Weird this is not a function of fuel type....

openstudiocore/src/model/GeneratorMicroTurbine.hpp
```diff
@@ -0,0 +1,249 @@
...
+  bool isStandbyPowerDefaulted() const;
+
+  double ancillaryPower() const;
+  bool isAncillaryPowerDefaulted() const;
+
+  // TODO: Check return type. From object lists, some candidates are: QuadraticCurves, UniVariateTables.
+  boost::optional<Curve> ancillaryPowerFunctionofFuelInputCurve() const;
+
+  // Optional Generator:MicroTurbine:HeatRecovery
+  boost::optional<StraightComponent> generatorMicroTurbineHeatRecovery() const;
```
macumber: I'd return an optional<GeneratorMicroTurbineHeatRecovery>

openstudiocore/src/model/GeneratorMicroTurbine.cpp
```diff
@@ -0,0 +1,1004 @@
...
+GeneratorMicroTurbine::GeneratorMicroTurbine(const Model& model)
+  : Generator(GeneratorMicroTurbine::iddObjectType(),model)
+{
+  OS_ASSERT(getImpl<detail::GeneratorMicroTurbine_Impl>());
+
+  // TODO: Appropriately handle the following required object-list fields.
+  //     OS_Generator_MicroTurbineFields::ElectricalPowerFunctionofTemperatureandElevationCurveName
+  //     OS_Generator_MicroTurbineFields::ElectricalEfficiencyFunctionofTemperatureCurveName
+  //     OS_Generator_MicroTurbineFields::ElectricalEfficiencyFunctionofPartLoadRatioCurveName
+  bool ok = true;
```
macumber: Where will you set the fields that are required but have no default?  @kbenne @mbadams5 can advise

referenceElectricalPowerOutput
referenceElectricalEfficiencyUsingLowerHeatingValue
electricalPowerFunctionofTemperatureandElevationCurve
electricalEfficiencyFunctionofTemperatureCurve
electricalEfficiencyFunctionofPartLoadRatioCurve

openstudiocore/src/model/GeneratorMicroTurbineHeatRecovery.hpp
```diff
@@ -0,0 +1,186 @@
...
+  // TODO: Check return type. From object lists, some candidates are: Connection.
+  // boost::optional<Connection> heatRecoveryWaterInletNode() const;
+
+  // TODO: Check return type. From object lists, some candidates are: Connection.
+  // boost::optional<Connection> heatRecoveryWaterOutletNode() const;
+
+  double referenceThermalEfficiencyUsingLowerHeatValue() const;
+  bool isReferenceThermalEfficiencyUsingLowerHeatValueDefaulted() const;
+
+  boost::optional<double> referenceInletWaterTemperature() const;
```
macumber: IDD says this is required with no default, should return double in that case

openstudiocore/src/model/GeneratorMicroTurbineHeatRecovery.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+  std::string heatRecoveryWaterFlowOperatingMode() const;
+
+  bool isHeatRecoveryWaterFlowOperatingModeDefaulted() const;
+
+  boost::optional<double> referenceHeatRecoveryWaterFlowRate() const;
```
macumber: IDD says this is required with no default, should return double in that case

openstudiocore/src/model/GeneratorMicroTurbineHeatRecovery.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+  // TODO: Check argument type. From object lists, some candidates are: Connection.
+  //bool setHeatRecoveryWaterOutletNode(const Connection& connection);
+  //void resetHeatRecoveryWaterOutletNode();
+
+  bool setReferenceThermalEfficiencyUsingLowerHeatValue(double referenceThermalEfficiencyUsingLowerHeatValue);
+  void resetReferenceThermalEfficiencyUsingLowerHeatValue();
+
+  void setReferenceInletWaterTemperature(double referenceInletWaterTemperature);
+  void resetReferenceInletWaterTemperature();
```
macumber: No reset if this is required

macumber:
Currently our only generators were PV so we didn't have to check bus type for Electric Load Center.  I assume microturbine is AC bus so we have to be more careful now.  Basically we should not allow DC and AC generators on the same bus:

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/ElectricLoadCenterDistribution.cpp#L183

macumber:
Also, some features of ELCD are not implemented yet, do you need any of these?

boost::optional<double> demandLimitSchemePurchasedElectricDemandLimit() const;
boost::optional<Schedule> trackScheduleSchemeSchedule() const;
boost::optional<std::string> trackMeterSchemeMeterName() const;
boost::optional<ElectricalStorage> electricalStorage() const;
boost::optional<Transformer> transformer() const;

macumber:
Biggest question I have is around required fields with no defaults, @kbenne and @mbadams5 can advise

openstudiocore/resources/model/OpenStudio.idd
jmarrec: Yeah I just forgot to remove it when I broke it between generator and straight component.

openstudiocore/src/model/GeneratorMicroTurbineHeatRecovery.hpp
jmarrec: Hum yeah, Technically it's not a required field in E+, it will default to 0, which is dumb with this separation of MicroTurbine and its heat recovery module... Why would you create one to have it at 0 water flow rate...

Goes back to your required withh no default question, how and when do I set them?

openstudiocore/src/model/GeneratorMicroTurbineHeatRecovery.hpp
jmarrec: Mmkay, done. Still needs to figure out where to set it

jmarrec:
I do need a bunch of these ELCD objects ultimately, most of all electrical storage that needs to get created as well... depending on how this ends up going I might tackle that next.

I've tried to make some of the changes you suggested, still a bunch more I guess especially that question that keeps poping about what (and where) to do with required fields that have no default

jmarrec:
I can run the unit tests now. I think it fails on the first one (construction of a generator), I have no idea why. I'm just giving up for today.

If someone has any time to help, it'll be extremely appreciated

jmarrec:
@macumber Generator class has ratedThermalToElectricalPowerRatio but doesn't have `setRatedThermalToElectricalPowerRatio`

jmarrec:
I'm having ruby issues since I added GeneratorMTHeatRecovery to ModelGenerators.i (I couldn't see the constructor for this in ruby bindings, after investigation it was missing from this file).

I also wrote a reverse translator, I need to go fetch some info from the ElectricLoadCenter:GeneratorLists in order for it to be 100% complete (availability schedule, and rated Thermal To Electrical Power Ratio).
I'm also ensuring that the curve objects are set to the correct type on reverse translation, not sure if that's really necessary. (If the IDF you import refers to a curve that is Cubic when the IDD says Biquadratic, it won't set it and fill throw a warning message)

jmarrec:
[ ] Todo: implement the  `mchp.setRatedThermaltoElectricalPowerRatio`

I wrote a ruby test, I can confirm that it works properly. 

Here is a [pastebin](http://pastebin.com/FgTN2gsT)

openstudiocore/src/model/GeneratorMicroTurbine_Impl.hpp
```diff
@@ -0,0 +1,266 @@
...
+    
+    virtual std::vector<ScheduleTypeKey> getScheduleTypeKeys(const Schedule& schedule) const override;
+
+    virtual std::string generatorObjectType() const;
+
+    virtual boost::optional<double> ratedElectricPowerOutput() const;
+
+    virtual boost::optional<Schedule> availabilitySchedule() const override;
+
+    virtual boost::optional<double> ratedThermaltoElectricalPowerRatio() const override;
```
macumber: @jmarrec looks correct

openstudiocore/src/model/GeneratorPhotovoltaic.cpp
```diff
@@ -136,7 +136,7 @@ namespace detail {
     return "Generator:Photovoltaic";
   }
 
-  boost::optional<double> GeneratorPhotovoltaic_Impl::ratedThermalToElectricalPowerRatio() const
+  boost::optional<double> GeneratorPhotovoltaic_Impl::ratedThermaltoElectricalPowerRatio() const
```
macumber: @jmarrec meh, I'm not crazy about either way here, I guess I prefer the original casing

jmarrec:
@kbenne  I added it to the OS App as well, seems to be working. Can you review maybe?

recovery on plant loop

jmarrec:
Problems:

[ ] 1. Can't add it to the demand side
[ ] 2. generator:Microturbine (parent) object isn't dragged with it
[ ] 3. I'd also want/need to have a method to return the generator associated with the heat recovery module (I might have to have the MT:HeatRecovery reference its generator rather than the other way around in order to fix 2.)

jmarrec:
2 and 3 might be linked: I might have to have the MT:HeatRecovery reference its generator rather than the other way around in order to fix 2.

If I don't need to do that, I guess I could write a method that would like kinda like this:

    boost::optional<GeneratorMicroTurbine> GeneratorMicroTurbine_Impl::generatorMicroTurbine() const {
      Handle handle = this->handle();
      boost::optional<GeneratorMicroTurbine> linkedmchp;
      for (const GeneratorMicroTurbine& mchp : this->model().getConcreteModelObjects<GeneratorMicroTurbine>()){
        boost::optional<GeneratorMicroTurbineHeatRecovery> mchpHR = mchp.generatorMicroTurbineHeatRecovery();
        if (mchpHR){
          if (mchpHR->handle() == handle){
            linkedmchp = mchp;
          }
        }
      }
      return linkedmchp;
    }

Thoughts @kbenne?

jmarrec:
I fixed the last problems I think. Here's an updated test suites that uses MiniTest (testing OS model object, forward translate and reverse translate).

[GeneratorMicroTurbine_test.txt](https://github.com/NREL/OpenStudio/files/313662/GeneratorMicroTurbine_test.txt)

For reverse translator test:

[Capstone-C65.txt](https://github.com/NREL/OpenStudio/files/313663/Capstone-C65.txt)


The last thing on my list is to make sure the `Generator:MicroTurbine` is carried over when dragging a `Generator:MicroTurbine:HeatRecovery` from the HVAC library. Do I need to modify the MTHR clone method to also clone the parent MT?

(I also need to redo the hvac library modification but I'll do that right now)

jmarrec:
New unit test to include clone testing:

[GeneratorMicroTurbine_test.txt](https://github.com/NREL/OpenStudio/files/314428/GeneratorMicroTurbine_test.txt)

nrel-bot-2:

nrel-bot-2:

PR #2250, opened 2016-05-28
Don't export "Baseline" checks unless the alternative is actually the…

### LINKED ISSUE #2249: Handle "Baseline" checks on PAT XML export
In the PAT export, we should only export a checks with the category "Baseline" if the design alternative is indeed the baseline.

asparke2:
… baseline.  Fixes #2249


asparke2:
@macumber I tested this with the new Energy Trust of Oregon reporting and QAQC Measure and it works as expected.

openstudiocore/src/pat_app/ExportXML.cpp
```diff
@@ -781,17 +784,38 @@ boost::optional<QDomElement> ExportXML::exportMonthly(QDomDocument& doc,
...
+      if (isBaselineCheck && !isBaseline) {
+        continue; // Skip to the next check
+      }
+
+      // Export the check
       QDomElement checkElem = doc.createElement("check");
       checksElem.appendChild(checkElem);
       for ( const Attribute & checkValAttr : checkAttr.valueAsAttributeVector()) {
-        if ( checkValAttr.name() == "name" ) {
+		    if ( checkValAttr.name() == "name" ) {
```
macumber: @asparke2 you should check your tab settings, should be 2 spaces

PR #2258, opened 2016-06-06
Fixes #2240

### LINKED ISSUE #2240: Unit conversion error - in^2 to cm^2
OpenStudio's unit conversion doesn't handle "cm^2" correctly.  Users will see this as incorrect unit conversion in IG, particularly in the SpaceInfiltration:EffectiveLeakageArea object.  Because I can reproduce this easily using ruby, I don't think it is an issue with the OS App.

In reality, 1 in^2 = 6.45 cm^2

`puts OpenStudio.convert(1.0, 'in^2', 'cm^2').get => 0.645`

OpenStudio is off by a factor of 10.

**Workaround**: manually convert your area to cm2, switch the OS App to SI, and enter the value.  This will make sure that the IDF gets the right numbers.

Here is 1.0 in2 in the inspector:
leakage_ip

Switch to SI units:
leakage_si

The IDF contains the value you entered as cm2 in SI mode:
leakage_idf

macumber:
Basically, there is some real weirdness when multiplying scales.  If the resulting scale cannot be found, the closest one is returned with an adjustment factor, code was ignoring the adjustment factor because there is no way to use it.  I added more scales, this issue can still occur with other units.

### DIFF
--- a/openstudiocore/src/utilities/units/Scale.cpp
+++ b/openstudiocore/src/utilities/units/Scale.cpp
@@ -114,6 +114,20 @@ namespace openstudio{
     static Scale M = {"M", "mega",6,1.0E6};
     return M;
   }
+  
+  /** Static constant defining SI prefix for 10^5. */
+  UTILITIES_API const Scale& hectokilo()
+  {
+    static Scale hk = {"_hk", "hectokilo",5,1.0E5};
+    return hk;
+  }
+
+  /** Static constant defining SI prefix for 10^4. */
+  const Scale& myria()
+  {
+    static Scale k = {"_ma", "myria",4,1.0E4};
+    return k;
+  }
 
   /** Static constant defining SI prefix for 10^3. */
   const Scale& kilo()
@@ -122,11 +136,32 @@ namespace openstudio{
     return k;
   }
 
+  /** Static constant defining SI prefix for 10^2. */
+  UTILITIES_API const Scale& hecto()
+  {
+    static Scale h = {"_h", "hecto",2,1.0E2};
+    return h;
+  }
+
+  /** Static constant defining SI prefix for 10^1. */
+  UTILITIES_API const Scale& deka()
+  {
+    static Scale da = {"_da", "deka",1,1.0E1};
+    return da;
+  }
+
   /** Static constant defining no scale (1.0). */
   const Scale& one()
   {
     static Scale o = {"","",0,1.0};
     return o;
+  }
+
+  /** Static constant defining SI prefix for 10^{-1} */
+  const Scale& deci()
+  {
+    static Scale d = {"_d", "deci",-1,1.0E-1};
+    return d;
   }
 
   /** Static constant defining SI prefix for 10^{-2} */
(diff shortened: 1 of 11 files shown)


macumber:
@asparke2 @DavidGoldwasser there may be really weird unintended consequences here.  we need to test unit conversions in scripts and the App carefully

openstudiocore/src/utilities/units/Scale.cpp
```diff
@@ -114,6 +114,20 @@ namespace openstudio{
...
   }
+  
+  /** Static constant defining SI prefix for 10^5. */
+  UTILITIES_API const Scale& hectokilo()
+  {
+    static Scale hk = {"_hk", "hectokilo",5,1.0E5};
+    return hk;
+  }
+
+  /** Static constant defining SI prefix for 10^4. */
```
asparke2: @macumber I think you should dump myria.  First, because it apparently uses "k", which is confusing when we have "kilo" and second because it is [obsolete](https://en.wikipedia.org/wiki/Myria-)

macumber: Well the main thing is we need a named scale for 1E4, I don't really care what it is named.  The actual short name used is "_ma" which should be enough to prevent it from being found in any unit strings, k is just the local variable.

macumber: Really we should be able to have unnamed scales but but the ScaleFactory doesn't support that

asparke2: Ah ok I missed that k was just a local variable.  I don't think we need to put in something for 1E4 if the [official NIST guide to SI](http://www.nist.gov/pml/pubs/sp811/sec04.cfm) doesn't even have one.

macumber: That's the thing though, not having a named scale for 1E-4 was the bug.  When you do cm^2 you get centi = 1E-2 raised to the 2 power = 1E-4.  When the ScaleFactory looked up this scale it didn't find anything so it returned the next closest scale (milli) plus a fudge factor.  The fudge factor got dropped because there was no place for it to get added and you ended up with the wrong conversion.

I added a couple of other scales to prevent this from happening in other situations.

macumber: Note that none of this was my code and it is all crazy complicated

openstudiocore/src/utilities/units/test/IPUnit_GTest.cpp
```diff
@@ -110,6 +110,64 @@ TEST_F(UnitsFixture, IPUnit_convert)
   EXPECT_NEAR(10.7639, value.get(), 0.0001);
 }
 
+TEST_F(UnitsFixture, IPUnit_convert2)
+{
+  boost::optional<double> value;
+
+  value = convert(1.0, "in", "cm");
```
macumber: @asparke2 check out the unit tests, these might be more interesting to you

openstudiocore/src/utilities/units/test/SIUnit_GTest.cpp
```diff
@@ -120,7 +125,7 @@ TEST_F(UnitsFixture,SIUnit_ArithmeticOperators)
   testStreamOutput("cm^6",u7);
   EXPECT_EQ("p",u7.scale().abbr);
   u7.pow(1,3); // (cm)^2
-  testStreamOutput("m(m^2)",u7); // no 10^-4 scale--moves to 10^-3
```
macumber: @asparke2 this one too

openstudiocore/src/utilities/units/Scale.cpp
asparke2: That makes sense.  We can have it internally but people won't use it externally because it isn't a thing.

macumber: Yes, that is why I prefixed all these new units with '_' so they won't get used accidentally

PR #2261, opened 2016-06-07
120391243 reporting measure template

DavidGoldwasser:


openstudiocore/src/utilities/bcl/templates/ReportingMeasure/measure.rb
```diff
@@ -110,7 +110,7 @@ def run(runner, user_arguments)
     if ann_env_pd
 
       # get desired variable
-      key_value =  "" # when used should be in all caps. In this case I'm using a meter vs. an output variable, and it doesn't have a key
+      key_value =  "Environment" # when used should be in all caps. In this case I'm using a meter vs. an output variable, and it doesn't have a key
```
macumber: @DavidGoldwasser the comments in the code are out of sync here, you are using an output value in this case, can you change the comments to match?

PR #2262, opened 2016-06-09
I think this loads qwindows plugin statically

macumber:


openstudiocore/src/qtwinmigrate/qwinwidget.cpp
```diff
@@ -56,6 +56,10 @@
 
 #include <qevent.h>
 
+// DLM: not sure how to do this right, think I only need this for static builds
+#include <QtPlugin>
+Q_IMPORT_PLUGIN(qwindows)
```
kbenne: my understanding was that the only difference between static plugin vs dynamic in qt 5.6 was that with static you needed to link the .lib plugin like you are doing here for qwindows.lib. I don't think there are more or less requirements for what you do in the code.

qsqlite seems to work without special attention.

macumber:
@kbenne this seems like it is working

thanks for the help @axelstudios, FYI Alex has some useful examples for linking static Qt [here](https://github.com/axelstudios/MouseSim/blob/master/CMakeLists.txt)

openstudiocore/src/qtwinmigrate/qwinwidget.cpp
```diff
@@ -56,6 +56,10 @@
 
 #include <qevent.h>
 
+// DLM: not sure how to do this right, think I only need this for static builds
```
macumber: @kbenne this probably needs to be inside an #if defined STATIC_BUILD or something in the case we are linking with dynamic Qt, not sure what we are going with

nrel-bot-3:

PR #2263, opened 2016-06-09
Add Meter:Custom and Meter:CustomDecrement to the SDK

jmarrec:


jmarrec:
@macumber: not sure I'm following.

----


Here's a unit test for the ruby bindings, testing both the model implementation and the reverse translating. Everything works as expected.

[MeterCustom_test.txt](https://github.com/NREL/OpenStudio/files/307162/MeterCustom_test.txt)

openstudiocore/src/model/MeterCustom.cpp
```diff
@@ -0,0 +1,298 @@
...
+    : ModelObject_Impl(other,model,keepHandle)
+  {}
+
+  // This logically doesn't produce any output variables
+  const std::vector<std::string>& MeterCustom_Impl::outputVariableNames() const
+  {
+    static std::vector<std::string> result;
+    if (result.empty()){
+    }
+    LOG(Info, "Meter:Custom does not produce any output variables.");
```
macumber: @jmarrec I'd remove this, will just create a bunch of noise

openstudiocore/src/model/MeterCustom.cpp
```diff
@@ -0,0 +1,298 @@
...
+  IddObjectType MeterCustom_Impl::iddObjectType() const {
+    return MeterCustom::iddObjectType();
+  }
+
+  // fuel Type
+  boost::optional<std::string> MeterCustom_Impl::fuelType() const {
+    return getString(OS_Meter_CustomFields::FuelType,true);
+  }
+
+  bool MeterCustom_Impl::setFuelType(boost::optional<std::string> fuelType) {
```
macumber: @jmarrec this should be `bool MeterCustom_Impl::setFuelType(const std::string& fuelType) {` the optional input is an error from the class generator.  This method takes a string and the reset method resets it

openstudiocore/src/model/MeterCustom.cpp
```diff
@@ -0,0 +1,298 @@
...
+    return ok;
+  }
+
+  // Remove all the (Key, Var) groups
+  void MeterCustom_Impl::removeAllKeyVarGroups() {
+    getObject<ModelObject>().clearExtensibleGroups();
+  }
+
+  // Return a vector of (Key, Var) pairs
+  std::vector< std::pair<std::string, std::string> > MeterCustom_Impl::keyVarGroups() {
```
macumber: @jmarrec I would prefer to define a simple class:

`struct MeterCustomKeyVarGroup{
  std::string key;
  std::string variable;
}`

than use `std::pair<string, string>` it will be much easier to work with in SWIG

openstudiocore/src/model/MeterCustom.cpp
```diff
@@ -0,0 +1,298 @@
...
+  getImpl<detail::MeterCustom_Impl>()->removeAllKeyVarGroups();
+}
+
+// // Return a vector of (Key, Var) pairs
+std::vector< std::pair<std::string, std::string> > MeterCustom::keyVarGroups() {
+  return getImpl<detail::MeterCustom_Impl>()->keyVarGroups();
+}
+
+// Return the number of (KeyName, OutputVariableorMeterName) groups
+unsigned MeterCustom::numKeyVarGroups() const {
```
macumber: @jmarrec sometimes calling it `variable` other times `var`, can we make this consistent?

openstudiocore/src/model/MeterCustom.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+  //@}
+  /** @name Type Casting */
+  //@{
+
+
+  //@}
+ protected:
+  /// @cond
+  typedef detail::MeterCustom_Impl ImplType;
```
macumber: @jmarrec here is that typedef that allows the casting methods to work

macumber:
@jmarrec nice job on this!

macumber:
Please feel free to add MeterCustomDecrement on this same branch

jmarrec:
Added MeterCustomDecrement. I did that very quickly so I expect a couple of kinks here and there...

jmarrec:
Updated Ruby unit test for MeterCustom (use a string rather than a file to test reverse translating)

[MeterCustom_test.txt](https://github.com/NREL/OpenStudio/files/308889/MeterCustom_test.txt)


Created Ruby Unit test for MeterCustomDecremement:

[MeterCustomDecrement_test.txt](https://github.com/NREL/OpenStudio/files/308891/MeterCustomDecrement_test.txt)

Change extension back to .rb

nrel-bot-3:

openstudiocore/src/model/MeterCustom.cpp
macumber: @jmarrec another option would be to change the IDD to have a single field in each extensible group. this field would point to either a variable or meter in the model.  what do you think of that?  in idf do you have to create an output:variable or output:meter in order to use the custom meter or does one get magically created for you if you reference it in the custom meter?

jmarrec: Is this what your are proposing?

    # Current
    meter_custom.addKeyVarGroup(key, var)

    # Yours
    group = OpenStudio::Model::MeterCustomKeyVarGroup.new(key, var)
    meter_custom.addKeyVarGroup(group)
    
If so, I don't see how that's easier to work with in SWIG

As far as your second option, it does "magically" adds them*. You can put anything from the mtd or rdd without having to also add it as an Output:Meter or Output:Variable.
For example, when I aim to meter my apartements versus common area, I add a Meter:Custom that lists the apartments electric conso by picking it from the mtd directly, and I will create a Meter:CustomDecrement by referencing "Electricili:Facility" and subtracting the apartment meter.

* As far as I know, all meters and variables you see in the mtd and rdd are necessary calculated during the simulation. Adding an Output:Meter or Output:Variable just means "please don't trash the results, just save them at this frequency"

macumber: Option 1:
```
meter_custom.addKeyVarGroup(key, var)
meter_custom.keyVarGroups.each do |key_var|
    # key_var is a std::pair<std::string, std::string>  
    key = key_var.first
    var = key_var.second
end
```

Option 2:
```
group = OpenStudio::Model::MeterCustomKeyVarGroup.new(key, var)
meter_custom.addKeyVarGroup(group)
meter_custom.keyVarGroups.each do |group|
    # group is a MeterCustomKeyVarGroup  
    key = group.key
    var = group.var
end
```

Option 3:
```
meter = OpenStudio::Model::Meter(model)
meter_custom.addVariable(meter)
variable = OpenStudio::Model::OutputVariable(var, model)
meter_custom.addVariable(variable)
meter_custom.meterOrVariables.each do |object|
    # object is a ModelObject
    if object.to_Meter.is_initialized?
        meter = object.to_Meter.get
    else
        variable = object.to_Variable.get
    end
end
```

I think Option 2 is a bit more clear than Option 1.  Option 3 does seem unnecessarily complicated, especially since it would require a lot of extra output.  It does seem like we should rename Meter to OutputMeter to avoid confusion.  We can create a SWIG alias for Meter -> OutputMeter.  

Finally, when you make a MeterCustom do you have to make a corresponding OutputMeter to write it to the file?  Our current Meter object (really OutputMeter) doesn't support arbitrary meter names so that might not work?

nrel-bot-3:

PR #2276, opened 2016-06-15
Add ElectricLoadCenter:Storage:Simple and expose rest of ELC:Distribution fields

jmarrec:
As part of this works, I created a base class called "ElectricalStorage" from which this object and the to-do one ElectricLoadCenter:Storage:Battery will inherit.

I still have some work to do, including modifying @macumber 's ElectricLoadCenter:Distribution object to implement all the missing methods (to set the storage, pick the track mode, etc).

So far I wrote I the base class, the storage:simple class and its forward and reverse translator as well as the Gtest (untested test right now).


jmarrec:
I have a problem: for some reason my class doesn't have "thermalZone", "setThermalZone" and "resetThermalZone" methods available, when I clearly defined them...

jmarrec:
@macumber: this works. But I need to expose the rest of the ElectricLoadCenter:Distribution fields.

I'm going to push the start of that, but the build fails so far.

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -25053,10 +25113,141 @@ OS:ElectricLoadCenter:Distribution,
...
+       \units W
+  A16; \field Storage Control Utility Demand Target Fraction Schedule Name
+       \note Modifies the target utility service demand power over time.
+       \note Schedule values should be fractions from -1.0 to 1.0, inclusive.
+       \note if omitted a schedule value of 1.0 is used. Negative values indicate export to grid
+       \note Schedule is used if Storage Operation Scheme is set to FacilityDemandLeveling.
+       \type object-list
+       \object-list ScheduleNames
+
+OS:ElectricLoadCenter:Storage:Converter,
```
macumber: @jmarrec does this object get translated?

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateElectricLoadCenterDistribution.cpp
```diff
@@ -88,38 +80,222 @@ boost::optional<IdfObject> ForwardTranslator::translateElectricLoadCenterDistrib
...
+  // NEW
+
+  // Logic based on Electrical Buss Type to translate or not translate inverters, storage
+  std::string bussType = modelObject.electricalBussType();
+ 
+  // Translate the buss Type already
+  idfObject.setString(ElectricLoadCenter_DistributionFields::ElectricalBussType, bussType);
+  
+  /// Inverter and Buss Type
+  boost::optional<Inverter> inverter = modelObject.inverter();
```
macumber: @jmarrec sorry i wasn't around when you were asking questions.  my preference is to try to keep object state correct using model logic but it does get complicated.  there is also the fact that your model objects may have to go through incorrect states in the process of being worked on

openstudiocore/resources/model/OpenStudio.idd
jmarrec: I' just starting out on this one. I implemented storage, expanded the elcd, that good.
But realized I also need the converter so I just added it to the idd but I've got everything else to do on this one

Envoyé de mon iPhone

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateElectricLoadCenterDistribution.cpp
macumber: My thought would be that setting an inverter should trump what the bus type is.  I also had the idea that we should detect AC or DC by the generators on the list.  The model should prevent you from mixing AC and DC generators.

nrel-bot-2:

PR #2285, opened 2016-06-27
Improvements to the class generator

macumber:


macumber:
This removes the setters taking optional in the impl classes, passes strings by const ref, better identifies schedule fields

@jmarrec can you remember any other issues you had with the class generator?  anything that was particularly annoying or confusing to fix?

jmarrec:
Not off the top of head no, but I need to add `ElectricLoadCenter:Storage:Converter`, so I can merge this into my branch and try it. I'll keep an eye out for problems.

jmarrec:
Should be able to fix these:

// TODO: Check the following class names against object getters and setters.
class Schedule;
class UniVariateCurves; //replace with Curve;
class Zone; // replace with ThermalZone

jmarrec:
The .cpp class also ends up with the setXXX methods having a type "void" while they clearly return a bool

jmarrec:
Need to delete the Q_OBJECT reference

```
  class MODEL_API XXXXXX_Impl : public ParentObject_Impl {
    Q_OBJECT;
   public:
```

jmarrec:
@macumber maybe reopen this one?

PR #2286, opened 2016-06-28
Updates openstudio-standards to 0.1.4 to enable Hospital

asparke2:
This should probably wait until 1.12.1


evanweaver:
@asparke2 Yep, we can incorporate this next week.

PR #2288, opened 2016-06-30
Adding OSW to the model dir was easier than I thought it would be

macumber:


macumber:
@kbenne this was easier than i thought it would be, does this work for you?

PR #2290, opened 2016-07-01
OS 2.0 dependency reduction

mbadams5:
This removes Qt signals/slots from Model and Utilities. We are using nano-signal-slot library instead which is a header-only, c++11 library. This library is faster than other signal/slot libraries and hopefully faster than Qt (https://github.com/NoAvailableAlias/signal-slot-benchmarks/tree/master/#performance).

This also removes all Q_PROPERTY and Q_OBJECT macros from Model. This eliminates the need for Qt entirely in Model and also the need to MOC files in Model. This should hopefully help to slowly drive down compile times for model. 

Removing Qt entirely from Model and most of Utilities has far reaching effects for compile times, performance, and most importantly, the CLI bundling.

This branch also includes performance improvements to the GUI, specifically grid view. With the change to Nano Signal/Slot, we lost the queued signals and QTimer so the signals were firing more often, causing underlying (slow) functions to be called many times. It was straightforward to improve these slow functions and improve overall grid view performance. For my test file, I see ~24% overall improvement (52 sec to 40 sec), with most of the remaining time being driven by version updating (25 sec).

##### 1.12.0 Release OpenStudio
Running Time    Self (ms)        Symbol Name
12809.0ms   24.2%    0.0         openstudio::SpacesTabController::setSubTab(int)

##### Release build of this branch
Running Time    Self (ms)        Symbol Name
3465.0ms    8.5%    0.0            openstudio::SpacesTabController::setSubTab(int)

There are two remaining bugs in the GUI: 
- In Radiance Parameters, when switching between Fine and Coarse settings, the non-combo-box fields don’t update like they should
- in Zone Air Contaminant Balance, the Outdoor Carbon Dioxide Schedule Name isn’t modifiable because I couldn’t find the corresponding field to link it to.


kbenne:
Monster!

macumber:
Wow!  I'll give this a go on Windows

kbenne:
I'm not even sure where to begin on this review, but I'll just say.... awesome!

kbenne:
Here is what I've found so far. 

1. For Mac I had to add shared qt to the build system. Not really an issue with this branch but with the 2.0 develop branch in general.
1. Qt Plugins are not configured correctly. I had to add QT_PLUGIN_PATH and set a hard path to get the app to launch. This is now on my todo list to investigate.
1. The app launches a lot faster. Not sure why because I thought most of the time was spent loading the library. It is nice though.
1. There are a few small visual changes/bugs in layout with qt 5.7 change.

It would be nice to stress test to see what kind of performance improvements have been made. If anyone has a good test model that pushes the limits (of grid especially) I'm very interested.

kbenne:
@mbadams5 and @macumber I think this is now good enough to merge into os 2.0 develop. Does anyone object? I have tested and tweaked builds on Mac and Windows.

macumber:
Go for it, I'll try a clean build once you merge

kbenne:
I'm thinking this iteration build of 2.0 should include CLI *and* OS App. Therefore marching toward something we can release. I suspect I'm going to have to clean up the install commands after all of the changes, and I will add an installer option for the App so if you only want the cli you can just get that.

macumber:
This is awesome!  Looking at what Qt is left in utilities it seems like QUrl and QUUID might be the next targets?  Here was the list I came up with, @mbadams5, what were you thinking about doing next or are you done for FY16?

Try to Remove:
  E:\openstudio-2-0d\openstudiocore\src\utilities\core\Json.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\core\Url.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\data\Attribute.hpp (2 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\data\Attribute_Impl.hpp (2 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\cloud\AWSProvider_Impl.hpp (2 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\cloud\CloudProvider_Impl.hpp (2 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\cloud\OSServer_Impl.hpp (3 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\cloud\VagrantProvider_Impl.hpp (2 hits)
  
Move to Utilities_Qt:
  E:\openstudio-2-0d\openstudiocore\src\utilities\bcl\BCL.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\bcl\RemoteBCL.hpp (3 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\core\Application.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\core\PathWatcher.hpp (4 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\core\UpdateManager.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\plot\ProgressBar.hpp (4 hits)
  
Surgery:
  E:\openstudio-2-0d\openstudiocore\src\utilities\core\FilesystemHelpers.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\core\Optional.hpp (3 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\core\RubyException.hpp (2 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\core\RubyInterpreter.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\core\String.hpp (3 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\core\UUID.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\units\Quantity.hpp (3 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\idd\IddEnums.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\idf\IdfObject.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\idf\IdfObject_Impl.hpp (2 hits)
  E:\openstudio-2-0d\openstudiocore\src\utilities\idf\URLSearchPath.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\idf\Workspace.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\idf\Workspace_Impl.hpp (1 hit)
  E:\openstudio-2-0d\openstudiocore\src\utilities\units\OSOptionalQuantity.hpp (1 hit)

PR #2292, opened 2016-07-06
Optimize the isPlenum method

macumber:


macumber:
@kbenne, @asparke2 noted that this method was a performance bottleneck

openstudiocore/src/model/ThermalZone.cpp
```diff
@@ -1998,18 +1998,13 @@ namespace detail {
   {
     boost::optional<AirLoopHVACSupplyPlenum> result;
 
-    std::vector<AirLoopHVACSupplyPlenum> plenums = model().getModelObjects<AirLoopHVACSupplyPlenum>();
+    std::vector<AirLoopHVACSupplyPlenum> plenums = getObject<ModelObject>().getModelObjectSources<AirLoopHVACSupplyPlenum>();
```
mbadams5: @macumber should use `model().getConcreteModelObjects<AirLoopHVACSupplyPlenum>()` instead, it'll be higher performance and is probably driving most of slowness here.

openstudiocore/src/model/ThermalZone.cpp
```diff
@@ -1998,18 +1998,13 @@ namespace detail {
...
-    {
-      if( boost::optional<ThermalZone> tz = plenum.thermalZone() )
-      {
-        if( tz->handle() == handle() )
-        {
-          result = plenum;
-          break;
-        }
-      }
+    if (plenums.size() == 1){
```
mbadams5: @macumber If you switch to using `getConcreteModelObjects` you'll need revert to using the old approach.

openstudiocore/src/model/ThermalZone.cpp
macumber: Either one should be faster, we could experiment or just pick one, I don't have a big preference

kbenne: Never been a fan of seeing the method getModelObjectSources in use. If the performance gain can be achieved using getConcreteModelObjects but leaving the method the same, I'm more in favor of that.

nrel-bot-3:
@kbenne @macumber @lgentile it has been 7 days since this pull request was last updated.

nrel-bot-3:
@kbenne @macumber @lgentile it has been 7 days since this pull request was last updated.

Issue #561, opened 2013-10-09
Adjacent stacked spaces in same zones may result in incorrect area and likely loads as well.


DavidGoldwasser:
A user pointed this out on the forum.
https://openstudio.nrel.gov/forums/openstudio-plug-sketchup/general-discussion/surface-matching-and-thermal-zone-area

I think we do things correctly when two or more  spaces are in the same zone, and the only shared surfaces are walls (vs. roof to ceiling). I also think that things would work as expected if two or more spaces are not adjacent and are in the same zone, but to be honest I'm not sure how EnergyPlus would deal with two non-adjacent volumes in the same zone. I'm also not sure how EnergyPlus would deal with my work around suggestion where two adjacent spaces with shared floor/ceiling are set to adiabatic. Might have to make a number test models to experiment with this.

We need to decide how we want this to go to EnergyPlus. Should it be a single volume where we just override the "area" floor area field, or do we need to do we need to leave the surfaces in place.

Another concern is if OpenStudio would come up with the same area as EnergyPlus does. This is important for ECM's that rely on having an accurate area.

@macumber and @asparke2 just wanted to copy you on these so you are aware of this.

I will make tests of a varity of combinations of geometry, boundary condition and construction (specifically air wall or not air wall). I should also do this when they are the same space types, different space types, and a variety of internal load calculations (per area or total power). I should also look at perfect stack vs. something with partial overlap.

Another thought is that we could issue a warning on ModelToIdf if we hit a non wall/wall matched surface pair in two spaces of within a single zone. That would probably be good to have if our solution is trying to infer what the user's intent is.

DavidGoldwasser:
Some screenshots from my testing of pull request #2311 

Looked at expected floor area by zone and building area  in OS app and E+ output. Also looked at changes to area based electric and lighting loads due to changed floor area.

![screenshot 2016-08-01 13 22 01](https://cloud.githubusercontent.com/assets/4857055/17306553/e1a40e48-57ed-11e6-9d35-b61b688f944b.png)

![screenshot 2016-08-01 13 22 17](https://cloud.githubusercontent.com/assets/4857055/17306556/e7d6d64c-57ed-11e6-8014-d6752b557625.png)

![screenshot 2016-08-01 13 27 49](https://cloud.githubusercontent.com/assets/4857055/17306559/ec6f9d42-57ed-11e6-8e91-bd97aa7dad77.png)

[...]

PR #2311, opened 2016-07-23
Do not include area of air wall surfaces in floor area

macumber:
@DavidGoldwasser @asparke2 what do you think about this?

### DIFF
--- a/openstudiocore/src/model/ThermalZone.cpp
+++ b/openstudiocore/src/model/ThermalZone.cpp
@@ -1179,7 +1179,6 @@ namespace detail {
           // air wall floors do not count in floor area
           if (surface.isAirWall()){
             needToSetFloorArea = true;
-            spaceType.reset();
             break;
           } 
 
@@ -1193,7 +1192,6 @@ namespace detail {
                 {
                   // this surface is completely inside the zone, need to set floor area since this surface will be removed 
                   needToSetFloorArea = true;
-                  spaceType.reset();
                   break;
                 }
               }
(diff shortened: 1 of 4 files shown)


macumber:
There are probably more things I'd need to do here (like make sure to remove loads on E+ translation) but what do you think about the general idea of not counting air walls in floor area?  I think David may have a ticket or two about this

DavidGoldwasser:
@macumber I like this idea. Issue #561 about stacked spaces relate to this. It is important for per area loads to be accounted for as well.

macumber:
@asparke2 @DavidGoldwasser this is ready for testing, should I merge this in for 1.12.3 for testing?  Should I ask @evanweaver to roll a separate build?

openstudiocore/src/model/ThermalZone.cpp
```diff
@@ -1246,11 +1274,16 @@ namespace detail {
...
     if (needToSetFloorArea){
+      
+      // do not allow per area loads in the space type since we are overriding floor area
+      spaceType.reset();
+
+      // don't override if user provided zone floor area
       if (isEmpty(OS_ThermalZoneFields::FloorArea)){
-        // DLM: do we want to do this?
-        //this->setDouble(OS_ThermalZoneFields::FloorArea, totalFloorArea);
+        this->setDouble(OS_ThermalZoneFields::FloorArea, totalFloorArea);
```
macumber: This will write to the zone's floor area field in E+

evanweaver:
Test builds under way.

PR #2315, opened 2016-07-28
Implement Storage Converter

jmarrec:
And fix a couple of issues with the GeneratorMicroTurbine (HVAC related, will need @kbenne's help on this)


openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslatePlantEquipmentOperationSchemes.cpp
```diff
@@ -511,6 +519,14 @@ boost::optional<IdfObject> ForwardTranslator::translatePlantEquipmentOperationSc
 
       for( auto setpointComponent : t_setpointComponents )
       {
+        // TODO: Find the right way to deal with this
+        // For now, "dirty" (?) fix for Generator:MicroTurbine
+        // @kbenne, FYI
+        if (boost::optional<GeneratorMicroTurbineHeatRecovery> mchp = setpointComponent.optionalCast<GeneratorMicroTurbineHeatRecovery>())
+        {
+          setpointComponent = mchp->generatorMicroTurbine();
```
kbenne: I think this is exactly what you're going to have to do.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslatePlantEquipmentOperationSchemes.cpp
```diff
@@ -385,6 +385,14 @@ ComponentType componentType(const HVACComponent & component)
     {
       return ComponentType::COOLING;
     }
+    case openstudio::IddObjectType::OS_Generator_MicroTurbine_HeatRecovery :
+    {
+      // Maybe that should be both, in the case of an absorption chiller?
```
kbenne: Well we probably aren't going to have a one size fits all solution. Use your best judgment about what you think the most common control scheme is and go with that. 

For custom stuff that is what the operation scheme API is for.

openstudiocore/src/model/ElectricLoadCenterDistribution.cpp
```diff
@@ -33,8 +33,8 @@
 #include "ElectricalStorage_Impl.hpp"
```
kbenne: I'm assuming the changes in this file have nothing to do with your op scheme troubles.

jmarrec:
Here is the ruby unit test for the ElectricLoadCenter:Storage:Converter object:

[ElectricLoadCenterStorageConverter_test.rb](https://github.com/NREL/OpenStudio/files/391126/ElectricLoadCenterStorageConverter_test.txt)

(save as .rb back).

@macumber I'm done with this PR, ready for review

openstudiocore/src/energyplus/ForwardTranslator.cpp
```diff
@@ -290,6 +290,22 @@ Workspace ForwardTranslator::translateModelPrivate( model::Model & model, bool f
...
+  // Remove orphan Storage
+  for (auto& storage : model.getModelObjects<ElectricalStorage>()) {
+    if (!storage.electricLoadCenterDistribution()){
+      LOG(Warn, "Electrical Storage " << storage.name().get() << " is not referenced by any ElectricLoadCenterDistribution, it will not be translated.");
+      storage.remove();
+    }
+  }
+  
+  // Remove orphan Converters
+  for (auto& converter : model.getModelObjects<ElectricLoadCenterStorageConverter>()){
```
macumber: @jmarrec if this is a concrete type (opposed to a base class) you can use model.getConcreteModelObjects which will be faster

openstudiocore/src/model/ElectricLoadCenterStorageConverter.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+  void ElectricLoadCenterStorageConverter_Impl::resetAvailabilitySchedule() {
+    bool result = setString(OS_ElectricLoadCenter_Storage_ConverterFields::AvailabilityScheduleName, "");
+    OS_ASSERT(result);
+  }
+
+  // This method will automatically switch the PowerConversationEfficiencyMethod to "SimpleFixed" and reset the DesignMaximumContinuousInputPower and EfficiencyFunctionofPowerCurveName fields
+  bool ElectricLoadCenterStorageConverter_Impl::setSimpleFixedEfficiency(double simpleFixedEfficiency) {
+    // Switch the Method
+    bool result = setString(OS_ElectricLoadCenter_Storage_ConverterFields::PowerConversionEfficiencyMethod, "SimpleFixed");
```
macumber: nice

openstudiocore/src/model/ElectricLoadCenterStorageConverter.cpp
```diff
@@ -0,0 +1,407 @@
...
+    OS_ASSERT(result);
+    // Reset the Simple Fixed Efficiency
+    result = setString(OS_ElectricLoadCenter_Storage_ConverterFields::SimpleFixedEfficiency, "");
+    OS_ASSERT(result);
+    return result;
+  }
+
+  bool ElectricLoadCenterStorageConverter_Impl::setAncillaryPowerConsumedInStandby(double ancillaryPowerConsumedInStandby) {
+    bool result = setDouble(OS_ElectricLoadCenter_Storage_ConverterFields::AncillaryPowerConsumedInStandby, ancillaryPowerConsumedInStandby);
+    OS_ASSERT(result);
```
macumber: @jmarrec don't really need an assert here, if it fails you can just return that.  asserts are fairly big deals, they are for things that would be worse than throwing an exception

macumber:
nice @jmarrec you are getting good at this!

openstudiocore/src/model/ElectricLoadCenterDistribution.cpp
```diff
@@ -479,8 +481,14 @@ namespace detail {
 
 
   // Storage Converter Object Name
-  //bool ElectricLoadCenterDistribution_Impl::setStorageConverter(const ElectricLoadCenterStorageConverter& converter);
-  //void ElectricLoadCenterDistribution_Impl::resetStorageConverter();
+  bool ElectricLoadCenterDistribution_Impl::setStorageConverter(const ElectricLoadCenterStorageConverter& converter) {
```
macumber: Could add some magic here to clone the converter if it is used by another ElectricLoadCenterDistribution

macumber: Really E+ should have made the storage converter point to the electric load center

jmarrec: Why do you say this?

I think it kinda makes sense in this way, you can easily see the ELCD with his inverter, transformer an storage devices. It'd be more abstract and hard to comprehend the other way around. Also this allows you to see the bussType of the elcd and quickly see if that makes sense considering your objects.

Speaking about this, we've scratch the subject before, but maybe we should switch the bussType automatically based on whether you set/reset inverters and converters. I don't really need that (my validityCheck method is enough for me), and I have little time to do this now, but this is food for thoughts.



Envoyé de mon iPhone

macumber: Well the real right solution is to embed the storage converter (or any other true child object) into the ELCD.  This isn't really possible in IDF but could be done in JSON.

Switching the bussType automatically is great but it is complicated.  You have to decide which methods can change it and in which circumstances, then you'd have to document it all for users.  Totally fine to skip this for now.

jmarrec: Like I said I already coded a validityCheck method for the elcd, it's just a matter of doing the opposite of that, so neither complicated nor hard to do since I already coded the logic. Documenting would need a bit of work though

Envoyé de mon iPhone

nrel-bot-2:
@macumber @jmarrec @lgentile it has been 7 days since this pull request was last updated.

jmarrec:
@macumber are we good on this PR? Do you want me to rebase on the current develop before merging that in (I kinda already did it because of the WaterUseConnection/Equipment bug introduced in 1.12.2)

macumber:
I'll review and let you know!

nrel-bot-3:
@macumber @jmarrec @lgentile it has been 7 days since this pull request was last updated.

jmarrec:
The tests failing is because of the WaterUseEquipment bug introduced in 1.12.2

macumber:
Merged!  Thanks for another great set of model objects!

jmarrec:
Glad I could help and contribute!

Envoyé de mon iPhone

PR #2323, opened 2016-08-01
Should fix round tripping of air walls

macumber:


macumber:
@DavidGoldwasser this should fix the issue you found, will be in 1.12.3

PR #2325, opened 2016-08-06
Os 2 0 kill super build gems

macumber:


openstudiocore/CMakeLists.txt
```diff
@@ -482,7 +482,7 @@ endif()
 # openstudio gems
 
 set(OPENSTUDIO_GEMS_ZIP_FILENAME "openstudio-gems.tar.gz")
-set(OPENSTUDIO_GEMS_ZIP_EXPECTED_MD5 "bc574f74016c5ebb6fb3e970e4370d77")
+set(OPENSTUDIO_GEMS_ZIP_EXPECTED_MD5 "6ea8fdeecc7fbac93731e31da64e284e")
```
macumber: @kbenne currently we don't have any binary gem dependencies that would require us to have platform specific builds, let's try to keep it that way :-)

kbenne:
So if I ```require 'openstudio-standards'``` from cli it will work now? Fantastic! Thank you! Thank you again!

PR #2340, opened 2016-08-24
Add sketchup plugin

kbenne:
@macumber there are a few things to point out in here. 

One is that I removed the openstudio_gui ruby binding and just put it all in the openstudio binding. Not my first choice, but I hit issues with the QApplication singleton being statically linked into two different modules. Each module ended up with its own instance of the singleton. If you know of a solution to that type of problem we could change back, but for expediency I think this is fine for now.

I also had to mess with the OpenStudioLib binding. It had OSDocument and FileOperations in it. OSDocument was apparently only used in the developer menu of the plugin so I removed it from the OpenStudioLib binding because it brought in the webkit DLL dependency. I thought I should point out this change in case you object. If you do we will either have to heavily refactor OSDocument or live with a DLL depend on openstudio ruby binding. (or i guess figure out the app singleton issue and put OSDocument back in openstudio_gui binding where we accept the webkit dll).

I have monkeyed with how the plugin is installed *and* tested. For the installed plugin I have adjusted the load paths to be relative and based on the structure of the install tree, wherever it may be situated. With these adjustments I am expecting the entire plugin to stay within the OS install tree. The installers may run a post flight script to generate a small stub in the sketch up/plugins dir that requires the plugin entry point, but since I'm not really focused on installers yet, that post flight script does not exist. Anyway, this is slightly different then how it used to be where we had a handful of plugin files installed in the sketch up plugins directory plus a config openstudio file that was post flight generated. Take a look and see if this will work for you and discuss as needed. Basically I'm trying to minimize work in the installers since they are platform specific, and I'm also trying to keep the OS install tree as self contained as possible. It was pretty self contained before, now with this change it is even more so.

There is a sketchup_plugin_test.rb file generated into the products directory. You can copy or link this into the sketch up plugins directory to do development. If you don't build static, then you are going to hit the issue about missing dlls when the plugin tries to load 'openstudio', but we can fix that separately as it is a more general issue beyond just the plugin. 

Thanks for looking!


openstudiocore/ruby/CMakeLists.txt
```diff
@@ -65,33 +69,50 @@ else()
...
-set_target_properties(openstudio_gui_rb PROPERTIES PREFIX "")
-set_target_properties(openstudio_gui_rb PROPERTIES OUTPUT_NAME openstudio_gui)
-if(APPLE)
-  set_target_properties(openstudio_gui_rb PROPERTIES SUFFIX ".bundle" )
-else()
-  set_target_properties(openstudio_gui_rb PROPERTIES SUFFIX ".so" )
-endif()
+install(DIRECTORY openstudio/sketchup_plugin DESTINATION share/openstudio COMPONENT "SketchUp Plugin")
+install(FILES openstudio/sketchup_plugin/Startup.rb DESTINATION share/openstudio/Plugins/OpenStudio/ COMPONENT "SketchUp Plugin")
+install(FILES openstudio/sketchup_plugin/OpenStudio.hash DESTINATION share/openstudio/Plugins/OpenStudio/ COMPONENT "SketchUp Plugin")
```
kbenne: I see a bug in the component name here....

openstudiocore/ruby/openstudio/sketchup_plugin/OpenStudio.rb
```diff
@@ -20,20 +20,22 @@
 require 'extensions.rb'   # defines the SketchupExtension class
 
 # look for the new plugin
-new_plugin = Sketchup.find_support_file("OpenStudio-config", "Plugins/OpenStudio")
+# This doesn't work anymore. Consider using openstudio -v from the cli
+#new_plugin = Sketchup.find_support_file("OpenStudio-config", "Plugins/OpenStudio")
```
kbenne: ping @macumber

openstudiocore/ruby/openstudio/sketchup_plugin/OpenStudio.rb
```diff
@@ -20,20 +20,22 @@
...
+new_plugin = true
```
kbenne: Need to properly detect "new" plugin.

macumber: This is all installer stuff we can address it later

openstudiocore/ruby/openstudio/sketchup_plugin/OpenStudio.rb
kbenne: kind of. but my main point is that I'm shooting for the installer not to have to do anything beyond dropping the files into the install tree. The openstudio-config should be a thing of the past and instead openstudio ruby bindings are located by relative path based on an assumption of the install tree layout. 

Without this config file though, there is not a handy file to grep for the os version. We may see that version appear in the install tree directory structure when everything gets finalized, but I think the most authentic place to get version is from cli.

openstudiocore/ruby/openstudio/sketchup_plugin/OpenStudio.rb
kbenne: aside from a link from the plugins directory to the os install tree, i have tested this from install and it works.

macumber:
Awesome Kyle, I will try this out now:

"I hit issues with the QApplication singleton being statically linked into two different modules".  Can't we instantiate QApplication in openstudio.so and extern it in openstudio_gui.so?

I think [this](https://github.com/NREL/OpenStudio/blob/8296251db964cfdac272468381950d6327b42500/openstudiocore/ruby/openstudio/sketchup_plugin/lib/MenuManager.rb#L755) is the only use of OSDocument in the plugin.  That seems like it was trying to start a window with basically the OS App inside of SketchUp, we can live with out that.  FileOperations was doing some of the crazy moving files around between temp dirs.  There might be stuff in there we need but we should refactor it out of OS Lib.

Let's just punt on installers for now.

I am trying to build a release static os/static qt build right now but the link times are really bad (linker keeps crashing actually).  I think we should add dir of openstudio.so to ENV['PATH'] and whatever else is necessary to get this to work with shared libs, it should all be in the old openstudio.rb file.

kbenne:
@macumber one thing I want to do right after we get this 2.0 release out the door is look into the link time and instability. I think step one is try VS 2015 and see if the linker has magically gotten better. Rumor suggest that to be true. Step 2 might be to reach out to my new contacts at MS.

kbenne:
I'm hoping that [here](https://github.com/NREL/OpenStudio/blob/0b947275e12b782d217826f89b0fd5146d30afec/openstudiocore/ruby/openstudio_rb.cpp#L28) would be the place to adjust path so that dlls are resolved. I'm afraid that is too late though. I'm betting the system will try to resolve the libraries before the init function is even entered. Perhaps we can get the symbols to lazy load giving us a chance to adjust path. or perhaps we can catch a load exception, adjust the path in the exception handler and try again. Option 1 seems easier. I believe there is a simple method for lazy loading.

kbenne:
ah yes. https://msdn.microsoft.com/en-us/library/151kt790.aspx

PR #2342, opened 2016-08-26
@kbenne will this work for the initialization?

macumber:


kbenne:
works for me @macumber. Style sheets work too now.  Thank you!

PR #2353, opened 2016-09-13
Os 2 0 develop linux

macumber:


openstudiocore/src/cli/main.cpp
```diff
@@ -229,28 +260,177 @@ int main(int argc, char *argv[])
...
+
+    Init_pathname();
+    rb_provide("pathname");
+    rb_provide("pathname.so");
+    
+    //Init_psych();
+    //rb_provide("psych");
+    //rb_provide("psych.so");
+
+#ifndef _MSC_VER
```
macumber: @kbenne this might now work with your current ruby build

openstudiocore/CMakeLists.txt
```diff
@@ -105,7 +112,7 @@ option(BUILD_R "Build R package" OFF)
 option(BUILD_SHARED_LIBS "Build OS libs as shared libraries" ON)
 
 # Build the OpenStudio Application
-# This option will require shared libs
+# This option will require shared Qt
```
kbenne: not that shared qt is an option. unless you added an option.

openstudiocore/CMakeLists.txt
```diff
@@ -551,46 +561,32 @@ if(BUILD_WITH_OPENSSL)
...
-      INACTIVITY_TIMEOUT 120
-      SHOW_PROGRESS
-      EXPECTED_MD5 ${SWIG_ZIP_EXPECTED_MD5})
-      execute_process(COMMAND ${CMAKE_COMMAND} -E tar xfz ${SWIG_ZIP_LOCAL_PATH} WORKING_DIRECTORY "${CMAKE_BINARY_DIR}")
-  endif()
-
-  set(SWIG_INSTALL_DIR "${CMAKE_BINARY_DIR}/Swig-install/")
+if(UNIX)
+  # We patch it up with a version of pcre we provide to avoid having to have the requirement locally
+  ExternalProject_Add(SWIG
```
kbenne: I think this is a mistake to use external project. (Now one more level deep in the inception of projects) How about just a custom build rule? You are already predicting where the swig executable is on line 577. I think the reason to external project would be to have the opportunity to cmake configure to whatever the external project products where. Here you're not doing that so why add such complexity? Perhaps there is no real tangible bad side effect of this, but when i was building this the other day my configure failed where you where trying to you use "SWIG" as a target level dependency on another target. I just seems bad. Just add_custom_command with output being the swig executable and use that as a dependency on the targets that need swig.

Or better yet. just use the silly swig lib environment variable or cli argument or whatever.

I know I introduced EP. But I regret it. Still has a place. like where we are still using it to create the binary packs of dependencies we upload to s3. But it has caused pain, and now I'm afraid of it. Here I particularly don't see the value.

openstudiocore/CMakeLists.txt
```diff
@@ -1320,15 +1378,15 @@ set(TARGETS_FILE ${CMAKE_BINARY_DIR}/OpenStudioTargets.cmake)
 set(all_lib_targets
```
kbenne: Im not even sure what this does anymore. Can we remove it?

openstudiocore/CMakeLists.txt
```diff
@@ -1361,8 +1417,8 @@ set(all_exe_targets
   #ResultsViewer
 )
 
-if(NOT Qt5Static)
-  list(APPEND all_exe_targets "OpenStudio")
+if(BUILD_OS_APP)
+  list(APPEND all_exe_targets "OpenStudioApp")
```
kbenne: ditto. why do we need these all_foo_targets? I've overlooked this for a while, but it has been on my mind to review what all it might be doing.

openstudiocore/src/cli/EmbeddedHelp.hpp
```diff
@@ -7,36 +7,21 @@
 #include <windows.h>
 #endif
 
-extern "C" {
```
kbenne: Can't exactly tell what you're up to but i think you're just removing the shenanigans to delay init this json stuff, and instead you're just doing it on os load. Fine by me. I like the cleanup. This always seems yucky to me.

openstudiocore/src/cli/main.cpp
```diff
@@ -229,28 +260,177 @@ int main(int argc, char *argv[])
...
+    Init_pty();
+    rb_provide("pty");
+    rb_provide("pty.so");
+#endif
+
+    Init_rational();
+    rb_provide("rational");
+    rb_provide("rational.so");
```
kbenne: are you sure you need to list foo and foo.so separately? I guess maybe where ruby require statements use require 'foo.so' explicitly instead of require 'foo'. Not a fan, but if you are certain you need it then by all means. Smells fishy.

openstudiocore/src/cli/main.cpp
```diff
@@ -229,28 +260,177 @@ int main(int argc, char *argv[])
...
   }
   
   #ifdef QT_STATIC
   Q_IMPORT_PLUGIN(QSQLiteDriverPlugin);
-  #endif
-
-  // Need embedded_help for requiring files out of the embedded system
-  auto embedded_extensions_string = embedded_files::getFileAsString(":/embedded_help.rb");
-  rubyInterpreter.evalString(embedded_extensions_string);
-
```
kbenne: I'm sure you've worked this out, but how have you removed this and not broken everything?

openstudiocore/CMakeLists.txt
macumber: This was the quickest way to move it into core since 1) binary packages weren't portable and 2) SWIG is a small dependency not worth messing around maintaining at the super level.  

I guess the thing to do on Unix is to download the source tar at configure time, then patch and build SWIG as a custom command with output being swig.exe, then you would still need to make those other commands depend on swig.exe.  You could so something similar on Windows but download the binary instead of source.  To me external project does all these steps in an easy way, only real difference is it downloads at build time instead of configure, right?

Other thing to look into is if we even need a custom SWIG build anymore, if Jason's commits were merged into SWIG maybe we can just use the normal releases?

openstudiocore/src/cli/main.cpp
macumber: Kept hitting errors when parts of ruby would require foo.so instead of foo.  I just figured I'd do this for everything just in case to avoid fixing errors one by one as they get exercised.

macumber: require 'pathname.so' was the one that Noel and I hit

openstudiocore/CMakeLists.txt
kbenne: I'm pretty sure we can use official binary. But note they too will have swig lib compiled in, but just to a more settling but still not helpful location like /usr/local/

openstudiocore/src/cli/EmbeddedHelp.hpp
macumber: Yeah it sounds like you found some sort of rb_delayed_provide, this would be better way to do delayed loads, I didn't want to write a custom load wrapper for each extension

openstudiocore/CMakeLists.txt
kbenne: You're also right that the way you are using external project here probably isn't much different than a custom command. But you're also not really using external project in the way (i think) it is intended, where there is a downstream configure step. The point of EP being to designate two distinct cmake configure times.

All in all i think i should acquiesce on this use of EP. I have probably developed an irrational hostility to it.

PR #2355, opened 2016-09-21
Updates openstudio-standards gem to 0.1.5, which supports baseline au…

asparke2:
…tomation.


asparke2:
@evanweaver Here's the PR for updating the standards gem so we can get baseline automation working.

evanweaver:
Super. I’ll start the builds.

aaron-boranian:
Problem: Performance energy models for code compliance or rating programs (T24 CBECC-COM, 90.1, LEED), require specific model inputs for a baseline and proposed model. Other simulation software on the market is able to build these baseline models, and generate standard compliance reports. Often, these simulation softwares are selected for projects, not because they are nicer to work with or have greater modeling capability, but simply because the can generate a baseline model and report with all the embedded rules. OpenStudio/EnergyPlus would get much greater adoption if it could provide this functionality.

Solution: Add a component to the OS App to set up the code baseline and baseline HVAC systems (assigning code space type definitions for example). This could also be a stand-alone application.

Migrated from [UserVoice](http://openstudio.uservoice.com/forums/308629/suggestions/14849835)

PR #2356, opened 2016-09-23
Os 2 0 fix static build

macumber:
This cleans up the C# bindings and examples


kbenne:
it might conflict a little but, I certainly agree with these changes. Actually the plugin thing I already have locally. Im going to merge.

PR #2363, opened 2016-09-29
BuildingUnit

nmerket:
This adds a `BuildingUnit` object that is optionally referenced by a space such that it can group together spaces into a unit such as an apartment in a multifamily building. It includes an extensible "features" list that specifies a list of user definable key, datatype, value sets that can be specified on the `BuildingUnit` such as "NumberOfBedrooms", "NumberOfBathrooms", etc.

@shorowit @macumber


openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -1980,6 +1980,39 @@ OS:BuildingStory,
...
+       \key Residential
+       \key NonResidential
+       \default Residential
+  A5,  \field Building Unit Feature Name
+       \type alpha
+       \begin-extensible
+       \required-field
+  A6,  \field Building Unit Feature Data Type
+       \type choice
+       \key alpha
```
macumber: @nmerket can you align these keys with the enums [here](https://github.com/NREL/OpenStudio/blob/os_2_0_develop/openstudiocore/src/utilities/data/Variant.hpp#L54):

  ((Boolean)(Boolean)(0))
  ((Double)(Double)(1))
  ((Integer)(Integer)(2))
  ((String)(String)(3))

openstudiocore/src/model/BuildingUnit.cpp
```diff
@@ -0,0 +1,430 @@
...
+  }
+
+  std::vector<std::string> BuildingUnit_Impl::suggestedFeatures() const
+  {
+    std::set<std::string> availableFeatureNames;
+    availableFeatureNames.insert("NumberOfBedrooms");
+    availableFeatureNames.insert("NumberOfBathrooms");
+    for (const BuildingUnit& bldgUnit : this->model().getConcreteModelObjects<BuildingUnit>()) {
+      for (const std::string& featureName : bldgUnit.featureNames()) {
+        availableFeatureNames.insert(featureName);
```
macumber: nice!

openstudiocore/src/model/BuildingUnit.cpp
```diff
@@ -0,0 +1,430 @@
...
+    std::vector<std::string> retvals;
+    retvals.assign(availableFeatureNames.begin(), availableFeatureNames.end());
+    return retvals;
+  }
+
+  bool BuildingUnit_Impl::setFeatureGroupDataTypeAndValue(const std::string& name, const std::string& dataType, const std::string& value)
+  {
+    boost::optional<ModelExtensibleGroup> group(getFeatureGroupByName(name));
+    if (group) {
+      const bool dataTypeOK = group->setString(OS_BuildingUnitExtensibleFields::BuildingUnitFeatureDataType, dataType);
```
macumber: probably should blank out the dataType and value if either set fails, otherwise you can be left in a bad state

macumber: or revert to the previous value if there was one

openstudiocore/src/model/BuildingUnit.cpp
```diff
@@ -0,0 +1,430 @@
...
+    return names;
+  }
+
+  boost::optional<ModelExtensibleGroup> BuildingUnit_Impl::getFeatureGroupByName(const std::string &name) const {
+
+    for (ModelExtensibleGroup& group : castVector<ModelExtensibleGroup>(extensibleGroups())) {
+      const boost::optional<std::string> featureName(group.getString(OS_BuildingUnitExtensibleFields::BuildingUnitFeatureName));
+      OS_ASSERT(featureName);
+      if (*featureName == name) {
+        boost::optional<ModelExtensibleGroup> modelGroup(group);
```
macumber: you can just return group if you want

openstudiocore/src/model/BuildingUnit.cpp
```diff
@@ -0,0 +1,430 @@
...
+  }
+
+  boost::optional<RenderingColor> BuildingUnit_Impl::renderingColor() const
+  {
+    return getObject<ModelObject>().getModelObjectTarget<RenderingColor>(OS_BuildingUnitFields::RenderingColor);
+  }
+
+  bool BuildingUnit_Impl::setRenderingColor(const RenderingColor& renderingColor)
+  {
+    return setPointer(OS_BuildingUnitFields::RenderingColor, renderingColor.handle());
```
macumber: uh oh, guess im gonna have to add this to the rendering modes in sketchup

macumber: i guess we will also need to extend the app, we usually do that after adding the model object

openstudiocore/src/model/BuildingUnit.cpp
```diff
@@ -0,0 +1,430 @@
...
+  std::vector<Space> BuildingUnit_Impl::spaces() const
+  {
+    return getObject<ModelObject>().getModelObjectSources<Space>(Space::iddObjectType());
+  }
+
+  std::vector<std::string> BuildingUnit_Impl::featureNames() const
+  {
+    std::vector<std::string> names;
+    for (const ModelExtensibleGroup& group : castVector<ModelExtensibleGroup>(extensibleGroups())) {
+      boost::optional<std::string> name = group.getString(OS_BuildingUnitExtensibleFields::BuildingUnitFeatureName);
```
macumber: duplicate names will be returned twice, setter prevents duplicates

openstudiocore/src/model/BuildingUnit.cpp
macumber: also, should be aware that signals will fire after each setString, any slots attached to that signal may fire after the first set and get an out of state object.  There is a badly named third argument to setString called checkValidity, if that is passed in as false the setter will not emit change signals.:

https://github.com/NREL/OpenStudio/blob/os_2_0_develop/openstudiocore/src/utilities/idf/WorkspaceObject_Impl.hpp#L195

openstudiocore/src/model/BuildingUnit.hpp
```diff
@@ -0,0 +1,133 @@
...
+  static IddObjectType iddObjectType();
+
+  /** @name Getters */
+  //@{
+
+  // Gets the rendering color
+  boost::optional<RenderingColor> renderingColor() const;
+
+  // Gets the building unit type
+  boost::optional<std::string> buildingUnitType() const;
```
macumber: most often this would return the default if there is nothing set.  I'd suggest either making the return type a string and return the default or remove the idd default.

openstudiocore/src/model/BuildingUnit.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+  std::vector<Space> spaces() const;
```
macumber: do we want a floor area method?

openstudiocore/src/model/Space.cpp
```diff
@@ -2496,6 +2514,15 @@ namespace detail {
     return result;
   }
 
+  boost::optional<ModelObject> Space_Impl::buildingUnitAsModelObject() const {
```
macumber: these methods were to support Q_PROPERTY access, we don't support those anymore so there is no need to implement this method

openstudiocore/src/model/Space.cpp
```diff
@@ -2644,6 +2671,16 @@ namespace detail {
     return false;
   }
 
+  bool Space_Impl::setBuildingUnitAsModelObject(const boost::optional<ModelObject>& modelObject) {
```
macumber: ditto, not needed

openstudiocore/src/model/BuildingUnit.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+class MODEL_API BuildingUnit : public ModelObject {
+ public:
+  /** @name Constructors and Destructors */
+  //@{
+
+  explicit BuildingUnit(const Model& model);
+
+  virtual ~BuildingUnit() {}
+
```
macumber: any thoughts on re-implimenting clone or remove?

openstudiocore/src/model/Space.cpp
```diff
@@ -734,6 +736,22 @@ namespace detail {
     OS_ASSERT(result);
   }
```
macumber: Check if we need to handle unit in clone or remove, just copy what is done for building story

openstudiocore/src/model/BuildingUnit.cpp
```diff
@@ -0,0 +1,430 @@
...
+    for (ModelExtensibleGroup& group : castVector<ModelExtensibleGroup>(extensibleGroups())) {
+      const boost::optional<std::string> featureName(group.getString(OS_BuildingUnitExtensibleFields::BuildingUnitFeatureName));
+      OS_ASSERT(featureName);
+      if (*featureName == name) {
+        boost::optional<ModelExtensibleGroup> modelGroup(group);
+        return modelGroup;
+      }
+    }
+    boost::optional<ModelExtensibleGroup> nullModelGroup(boost::none);
+    return nullModelGroup;
```
macumber: can just return boost::none

macumber:
Nice job @nmerket !

openstudiocore/src/model/BuildingUnit.cpp
nmerket: So, should I do some sort of check for duplicate names here and log an error if there are duplicates?

openstudiocore/src/model/BuildingUnit.cpp
nmerket: Good to know. `Boost::optional` is kind of mysterious to me. I'll change that.

openstudiocore/src/model/BuildingUnit.cpp
nmerket: Noted.

openstudiocore/src/model/BuildingUnit.cpp
nmerket: Thanks

openstudiocore/resources/model/OpenStudio.idd
macumber: Not yet, that Variant class is only on OpenStudio 2.X branch for now.  Eventually I think your class should have a method:

    boost::optional<openstudio::Variant> getFeature(const std::string& name) const;

openstudiocore/src/model/BuildingUnit.cpp
macumber: You had stuffed everything into a set then copied the set to a vector in another method, you could do that here to ignore duplicates

openstudiocore/src/model/BuildingUnit.cpp
nmerket: That all made sense to me except that last comment about the signals and out of state objects.

nmerket: Problem: the IDFExtensibleGroup class which implements the `setString` method in this case has the [one with the check validity flag set as a private method](https://github.com/NREL/OpenStudio/blob/buildingunit/openstudiocore/src/utilities/idf/IdfExtensibleGroup.hpp#L243). Should I make it public? Also, how would I finally tell it to send the update signal?

macumber: Yeah, let's make that method public.  The method you want is [emitChangeSignals](https://github.com/NREL/OpenStudio/blob/buildingunit/openstudiocore/src/utilities/idf/IdfObject_Impl.hpp#L376).  There are a few examples of using it [here](https://github.com/NREL/OpenStudio/search?utf8=%E2%9C%93&q=emitChangeSignals&type=Code)

openstudiocore/src/model/BuildingUnit.hpp
nmerket: I don't see anything in `BuildingStory` or `BuildingStory_Impl` about cloning or removing.

openstudiocore/src/model/BuildingUnit.hpp
nmerket: As far as I can tell, there isn't anything on BuildingStory or BuildingStory_Impl about removing or cloning.

macumber: Looks like you are correct, let's just do the same for BuildingUnit then

nmerket:
@macumber Have another look. I think I addressed all your concerns.

nmerket:
@macumber Just checking in on this. Do you need anything else from me?

macumber:
Thanks for the reminder, I'll check it out now

PR #2380, opened 2016-10-16
removed CAV and VAV fan limit from air loops

### LINKED ISSUE #2377: remove limit on number of fans on air loops
Currently OS limits the number of fans that can be assigned to the following objects to (2), which is not present in EnergyPlus. One use case would be [approximating a DOAS to AHU system](https://unmethours.com/question/20545/approximating-doas-to-ahus-in-energyplus/).

* AirLoopHVAC
* AirLoopHVACOutdoorAirSystem
* AirLoopHVAC + AirLoopHVACOutdoorAirSystem 

image

image

image

MatthewSteen:
closes #2377 

image


kbenne:
Thanks!

PR #2387, opened 2016-10-29
Gbxml sizing

macumber:
Adds autosized loads to gbXML export, use case is using OS for sizing then exporting to other tools such as http://www.hvacsolution.com/

@asparke2 @kbenne I'd appreciate a review, I'm not sure how much work we need to put in to handle hard sized equipment, etc or how much of the sizing logic from the standards gem should be pulled into model.  I think the basic use case is to autosize everything, then export to gbxml, the fields I output are what Joe Simmons at HVAC Solution requested.

The easiest way to do this export may be as a measure, either a model measure that does a sizing run or a reporting measure


openstudiocore/src/gbxml/ForwardTranslator.cpp
```diff
@@ -1171,6 +1343,52 @@ namespace gbxml {
     result.appendChild(nameElement);
     nameElement.appendChild(doc.createTextNode(QString::fromStdString(name)));
 
+    // heating setpoint
+    boost::optional<double> designHeatT;
+    boost::optional<double> designCoolT;
+    boost::optional<model::Thermostat> thermostat = thermalZone.thermostat();
+    if (thermostat && thermostat->optionalCast<model::ThermostatSetpointDualSetpoint>()){
```
macumber: Would be great to have a method like Thermostat::coolingDesignSetpoint and Thermostat::heatingDesignSetpoint

asparke2:
@macumber I think this makes sense.  For now, since you are only getting the zone loads, it isn't a big deal to keep the queries here.  However, I would eventually like to move all of the code to retrieve equipment's autosized values from openstudio-standards into the C++.

Also, I think it would be nice to add a generalized method to different HVAC classes to get common things like heating and cooling capacity.  These would simply call the very E+-centric methods that each object has that maps to specific field names.

nrel-bot-2:

PR #2398, opened 2016-11-28
New method in E+ reverse translator to only translate objects in DDY files.

IainMacdonald:
The generalised idf reverse translator produces warning messages due to missing gepmetry objects that this now method does not. The new method only converts site and design day objects.


macumber:
Thanks @IainMacdonald I think I am going to do this another way, I'll link when done so you can see

IainMacdonald:
No worries – it was a bit of a hack! I wanted to identify the DDY file and then act on that…

macumber:
@IainMacdonald please see https://github.com/NREL/OpenStudio/pull/2405 for my proposed solution (the first commit is the solution, second commit is fixing an unrelated bug)

macumber:
Hmm that was messy, here is the actual commit https://github.com/NREL/OpenStudio/pull/2405/commits/cc55e60a2edf3f57d0a888bd848a153db156ac54

IainMacdonald:
Much nicer than my hack!

PR #2408, opened 2016-12-01
Fixes #1519, makes line edits work much better.  I want to be a littl…

### LINKED ISSUE #1519: People Def can have values for multiple unit types, other loads can't
I thought for all internal loads that at least the GUI (if not the API) only allowed you to enter a value for one of the unit types. This is true for other load types, but for some reason not for people. @macumber I assume this is an un-intentinal oversight vs. a design. Not clear what would happen on forward translation if you have multiple values. See screenshot below

3-13-2015 11-34-05 am

Note, that outdoor air does by design allow you to enter multiple values, and then uses a choice option to choose method handle the value. Infiltration is similar but the choice method in the end just picks one value. While not the subject of this bug it may be cleaner if it also just excepted one value so there is not confusion.

3-13-2015 11-41-48 am

macumber:
…e careful with this since line edits are everywhere, hoping @DavidGoldwasser or @asparke2 can test (in the loads definition tab and also grid views) before merging


asparke2:
@macumber @DavidGoldwasser I'm building this to test.

macumber:
@asparke2 I merged this so @evanweaver can put it in the package

PR #2411, opened 2016-12-05
Ems

brianlball:
This adds the ability to do weather file actuation.  
If either a Site or WeatherFile is used as the modelObject in the Actuator constructor, then the actuated component is set to "Environment" which is what the E+ EMS code uses to alter the weather file variables.


macumber:
Looks good to me, merging in for 1.14.0 :-)

PR #2424, opened 2017-01-04
I believe this fixes #2419

### LINKED ISSUE #2419: PAT 1.14.0 issue, results are deleted when project is opened
I confirmed on mac what UH user posted. This isn't a 1.13 to 1.14 transition issue. Happens with fresh 1.14.0 files.
https://unmethours.com/question/21807/pat-1140-deletes-old-results-on-reopening-project-file/

Steps to reproduce:

1. Create a new PAT project and set seed file.
2. Add a measure and create one or more design alternatives (this is probably optional step)
3. Run Analysis
4. Look at results page (to confirm you can see results)
4. Quit PAT (look and confirm that datapoint directories still exist)
5. Re-launch PAT and load the project
6. At this point, the datapoint directories are gone and if you go to the results page the data is gone there.

macumber:


openstudiocore/src/pat_app/Measures/StandardReports/measure.xml
```diff
@@ -389,7 +389,6 @@
       </choices>
     </argument>
   </arguments>
-  <outputs/>
```
macumber: @DavidGoldwasser I think this outputs element indicates that the xml had been written by OS 2.0

PR #2448, opened 2017-01-20
Wrapping SurfaceProperty:ConvectionCoefficients

nmerket:
@macumber @shorowit

What does this still need?


openstudiocore/src/energyplus/CMakeLists.txt
```diff
@@ -292,6 +292,7 @@ set(${target_name}_src
   ForwardTranslator/ForwardTranslateSteamEquipment.cpp
   ForwardTranslator/ForwardTranslateSubSurface.cpp
   ForwardTranslator/ForwardTranslateSurface.cpp
+  ForwardTranslator/ForwardTranslateSurfacePropertyConvectionCoefficients.cpp
```
macumber: This object would be a candidate for reverse translation, not saying you have to do it only that it would be possible

openstudiocore/src/model/SurfacePropertyConvectionCoefficients.hpp
```diff
@@ -0,0 +1,171 @@
...
+
+  boost::optional<SubSurface> surfaceAsSubSurface() const;
+
+  boost::optional<InternalMass> surfaceAsInternalMass() const;
+
+  boost::optional<std::string> convectionCoefficient1Location() const;
+
+  boost::optional<std::string> convectionCoefficient1Type() const;
+
+  boost::optional<double> convectionCoefficient1() const;
```
macumber: My inclination would have been to make a little more friendly interface like 'setInsideCoefficient' and 'setOutsideCoefficient' with more smarts (e.g. reject setting a schedule if it is the wrong type) but I really can't fault you for sticking closer to E+.  It's just easy to shoot yourself in the foot using this object, probably ok since it is a bit of an advanced feature.

openstudiocore/src/model/SurfacePropertyConvectionCoefficients.cpp
```diff
@@ -0,0 +1,399 @@
...
+    }
+    return result;
+  }
+
+  IddObjectType SurfacePropertyConvectionCoefficients_Impl::iddObjectType() const {
+    return SurfacePropertyConvectionCoefficients::iddObjectType();
+  }
+
+  bool SurfacePropertyConvectionCoefficients_Impl::setSurface(const ModelObject &surface) {
+    return setPointer(OS_SurfaceProperty_ConvectionCoefficientsFields::SurfaceName, surface.handle());
```
macumber: Maybe remove existing surface coefficients on the surface?

macumber: We should add a method boost::optional<SurfacePropertyConvectionCoefficients> PlanarSurface::surfacePropertyConvectionCoefficients() const;  Also want to make sure these are listed as a child of the planar surface.

macumber: If we did these things the only way to get two of these objects pointing at the same surface is to clone it.  We can make clone blank out the surface field.

nmerket:
Thanks @macumber, I'll go through these and make some changes.

openstudiocore/src/model/SurfacePropertyConvectionCoefficients.cpp
nmerket: How do I list this object as a child of planar surface?

nmerket: So, I would implement a custom clone that blanks out the planar surface then?

macumber: You can implement a children method in PlanarSurface similar to this one in Space, https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/Space.cpp#L175

macumber: And correct, you would implement a custom clone that blanks out the planar surface

nmerket: A couple of things:

 - I can't add that method to `PlanarSurface` because it needs to be on `Surface`, `SubSurface`, and `IntMass`. Not all of those inherit from `PlanarSurface`. I'm planning on just putting in a separate getter on each of those.
 - So, when I `setSurface` on the `SurfacePropertyConfectionCoefficients`, if I have it remove the existing surface coefficients, should I just delete them? Otherwise, they'll be left without a surface to point to, which I specifically designed this API to never allow. 
 - Same concern with blanking out the surface field on a clone, then we'll have one of these objects without a surface, which I was trying to avoid. Am I missing something?

openstudiocore/src/model/SubSurface.cpp
```diff
@@ -735,6 +737,27 @@ namespace detail {
     }
   }
 
+  boost::optional<SurfacePropertyConvectionCoefficients> SubSurface_Impl::surfacePropertyConvectionCoefficients() const {
+    std::vector<SurfacePropertyConvectionCoefficients> allspccs(model().getConcreteModelObjects<SurfacePropertyConvectionCoefficients>());
+    std::vector<SurfacePropertyConvectionCoefficients> spccs;
```
macumber: You can also use [getModelObjectSources](https://github.com/NREL/OpenStudio/blob/72e4961a2d6547dcb40be48f722345b5691e3a74/openstudiocore/src/model/ModelObject.hpp#L217)

openstudiocore/src/model/SurfacePropertyConvectionCoefficients.cpp
macumber: @nmerket you did not implement any custom clone logic right?

openstudiocore/src/model/SubSurface.cpp
nmerket: @macumber I haven't written any custom clone methods. I was planning on it, but ran out of time. Is it too late to get this in?

nmerket:
@macumber Did this make it into the iteration build? It looks like it hasn't been merged. Also, where is the clone you made?

PR #2473, opened 2017-02-14
Preserve results

macumber:


macumber:
@DavidGoldwasser @asparke2 I figured out that the issue of results being deleted on save was due to the fact that we are running in the user's model dir and not the temp dir, so when the temp dir is copied to the user dir results are lost.  Not ideal, but my solution is to go back to running in the temp dir.  I am keeping the open simulation results button to open the user's dir, I will disable this during the run so users aren't confused about why simulation results aren't updated in real time.

macumber:
@kbenne @rpg777 FYI

rpg777:
Are you telling me that the Radiance measure needs to play ball with the temp directory as well?

macumber:
I'm telling you...... don't worry it's all good :-)

You just won't be able to see simulation output (other than the status messages in the run window) until the run is complete

rpg777:
Gotcha, thank you.

PR #2482, opened 2017-02-23
Revised MAXILKSIZE

arif-hanif:
Revised MAXILKSIZE to 0x7FFFFFFF to remove LNK1320 Fatal Error


macumber:
Thanks @arif-hanif !  Looks like your editor made some extra whitespace changes but we can live with that

PR #2534, opened 2017-03-21
Fix problem opening sql in RV from OS App

### LINKED ISSUE #2525: ResultsViewer button in App starts RV but doesn't load data
RV launches, but doesn't automatically open the correct sql file.  Manually navigating to the sql file and opening it works okay.

kbenne:
The issue was that the temp directory where OS puts a copy of the sql
file is actually a symbolic link (this is how the OS sets up the tmp
location). RV has a lot of logic to keep track of open files and
aliases, and some of this logic uses path comparison operations. The
trouble starts when some paths being compared are "canonical" (links
expanded) and others are not expanded. So the solution was just expand
the paths right from the get go. This seemed to only impact files opened
from the cli because it seems the Qt file open dialog already expands
the file paths.

close #2525

### DIFF
--- a/openstudiocore/src/resultsviewer/MainWindow.cpp
+++ b/openstudiocore/src/resultsviewer/MainWindow.cpp
@@ -1395,11 +1395,12 @@ namespace resultsviewer{
     {
       for (QString file : fileList)
       {
-        file = QString::fromStdString(openstudio::filesystem::canonical(toPath(file)).string());
         if (!QFile::exists(file)) {
           QMessageBox::information(this, tr("File Open"), tr("File not found:\n" + file.toUtf8()));
           continue;
         }
+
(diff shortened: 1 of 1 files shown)


openstudiocore/src/resultsviewer/MainWindow.cpp
```diff
@@ -1394,6 +1395,7 @@ namespace resultsviewer{
     {
       for (QString file : fileList)
       {
+        file = QString::fromStdString(openstudio::filesystem::canonical(toPath(file)).string());
```
macumber: @kbenne need to be careful with canonical since i think it will throw if the file does not exist:

http://en.cppreference.com/w/cpp/experimental/fs/canonical

I ended up writing this function in WorkflowJSON:

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/filetypes/WorkflowJSON.cpp#L53

PR #2538, opened 2017-03-21
Install radiance for pat

kbenne:


rpg777:
THanks Kyle, that added Radiance to PAT packages. We are not out of the woods yet, though. OS_RAYPATH is still not getting set...

PR #2542, opened 2017-03-24
Improve simulation reporting in app

macumber:


macumber:
@kbenne @rpg777 FYI

I think this should go into 2.1.0, without it measure results are not shown in the app

rpg777:
This works in the App now, but all the messages print to the App run console after the measure has completed.

PR #2577, opened 2017-04-18
fixing Propane/PropaneGas bug when translating to EnergyPlus

nmerket:
EnergyPlus recently had an idd change in `Coil:Heating:Fuel` for the fuel type to be `Propane` instead of `PropaneGas`. OpenStudio was still translating to `PropaneGas`. Now, it accepts either in the OS model (warning about the deprecation of `PropaneGas`) and forward translates to `Propane` in EnergyPlus.

Fixes https://github.com/NREL/OpenStudio-BEopt/issues/114

@shorowit

### DIFF
--- a/openstudiocore/resources/model/OpenStudio.idd
+++ b/openstudiocore/resources/model/OpenStudio.idd
@@ -14740,6 +14740,7 @@ OS:Coil:Heating:Gas,
   A8;  \field Fuel Type
        \type choice
        \key NaturalGas
+       \key Propane
        \key PropaneGas
        \key Diesel
        \key Gasoline
--- a/openstudiocore/resources/model/OpenStudio.idd
+++ b/openstudiocore/resources/model/OpenStudio.idd
@@ -14738,6 +14738,7 @@ OS:Coil:Heating:Gas,
        \type real
        \units W
   A8;  \field Fuel Type
+       \note PropaneGas is deprecated in favor of Propane
        \type choice
        \key NaturalGas
        \key Propane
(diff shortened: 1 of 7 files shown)


openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -14740,6 +14740,7 @@ OS:Coil:Heating:Gas,
   A8;  \field Fuel Type
        \type choice
        \key NaturalGas
+       \key Propane
        \key PropaneGas
```
macumber: Can you add a comment that PropaneGas is deprecated?

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilHeatingGas.cpp
```diff
@@ -73,7 +73,12 @@ boost::optional<IdfObject> ForwardTranslator::translateCoilHeatingGas( CoilHeati
   ///////////////////////////////////////////////////////////////////////////
   // Field: Fuel Type
   ///////////////////////////////////////////////////////////////////////////
-  idfObject.setString(openstudio::Coil_Heating_FuelFields::FuelType, modelObject.fuelType());
+  if (modelObject.fuelType() == "PropaneGas") {
+    idfObject.setString(openstudio::Coil_Heating_FuelFields::FuelType, "Propane");
```
macumber: You can add a LOG(Warn, "'PropaneGas' is deprecated for Coil_Heating_FuelFields:FuelType, use 'Propane' instead")

macumber:
Few minor comments then this is good to go, thanks!

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilHeatingGas.cpp
nmerket: I already have this in the OpenStudio model. Should I have it again here in the forward translator?

macumber: Yeah, I'd do it here in case people have this in existing models

PR #2578, opened 2017-04-18
Adding autosizing capability and new E+ 8.7 fields to water to water …

### LINKED ISSUE #2575: Enable autosizing of WaterToWaterHeatPumps
`WaterToWaterHeatPumpEquationFitHeating` and `WaterToWaterHeatPumpEquationFitCooling` are [now autosizable](http://bigladdersoftware.com/epx/docs/8-7/input-output-reference/group-plant-equipment.html#field-rated-load-side-flow-rate).  Each object has 4 autosizable fields, plus an extra rated COP field and field that links the heating and cooling heat pump objects for sizing purposes.  It looks likely that we'll need this capability for district system modeling.

asparke2:
…heat pumps fixes #2575

### DIFF
--- a/openstudiocore/resources/model/OpenStudio.idd
+++ b/openstudiocore/resources/model/OpenStudio.idd
@@ -11060,12 +11060,12 @@ OS:HeatPump:WaterToWater:EquationFit:Heating,
         \note COP = Reference Heating Capacity / Reference Heating Power Consumption
         \type real
         \minimum> 0.0
-        \default 7.5
+        \required-field
    N16, \field Sizing Factor
         \note Multiplies the autosized capacity and flow rates
         \type real
         \minimum> 0.0
-        \default 1.0
+        \required-field
    A7 ; \field Companion Cooling Heat Pump Name
         \type object-list
         \object-list WWHPCoolingNames
--- a/openstudiocore/resources/model/OpenStudio.idd
+++ b/openstudiocore/resources/model/OpenStudio.idd
@@ -10980,6 +10980,7 @@ OS:HeatPump:WaterToWater:EquationFit:Heating,
        \required-field
    A2,  \field Name
         \type alpha
+        \reference WWHPHeatingNames
         \required-field
         \reference ConnectionObject
    A3,  \field Source Side Inlet Node Name
@@ -10998,26 +10999,32 @@ OS:HeatPump:WaterToWater:EquationFit:Heating,
        \type object-list
        \required-field
        \object-list ConnectionNames
-   N1,  \field Rated Load Side Flow Rate
+   N1,  \field Reference Load Side Flow Rate
         \required-field
         \type real
         \minimum> 0.0
         \units m3/s
-   N2,  \field Rated Source Side Flow Rate
+        \ip-units gal/min
+        \autosizable
+   N2,  \field Reference Source Side Flow Rate
         \required-field
         \type real
         \minimum> 0.0
         \units m3/s
-   N3,  \field Rated Heating Capacity
+        \ip-units gal/min
+        \autosizable
+   N3,  \field Reference Heating Capacity
         \required-field
         \type real
         \minimum> 0.0
         \units W
-   N4,  \field Rated Heating Power Consumption
+        \autosizable
+   N4,  \field Reference Heating Power Consumption
         \required-field
         \type real
         \minimum> 0.0
         \units W
+        \autosizable
    N5,  \field Heating Capacity Coefficient 1
         \required-field
         \type real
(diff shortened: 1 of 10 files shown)


nrel-bot-3:
@kbenne @asparke2 @lgentile it has been 7 days since this pull request was last updated.

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -10998,26 +10999,32 @@ OS:HeatPump:WaterToWater:EquationFit:Heating,
        \type object-list
        \required-field
        \object-list ConnectionNames
-   N1,  \field Rated Load Side Flow Rate
+   N1,  \field Reference Load Side Flow Rate
```
kbenne: At first glance I was going to protest that this was going to be an API change. When I looked at the related Class implementation I found that the old name is still used in the API and the implementation maps to the new IDD name in the ModelObject. So I was almost on board. There is still a breaking API change that I will comment on in the header.

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -11045,16 +11052,31 @@ OS:HeatPump:WaterToWater:EquationFit:Heating,
...
+        \minimum> 0.0
+        \default 7.5
+   N16, \field Sizing Factor
+        \note Multiplies the autosized capacity and flow rates
+        \type real
+        \minimum> 0.0
+        \default 1.0
+   A7 ; \field Companion Cooling Heat Pump Name
+        \type object-list
+        \object-list WWHPCoolingNames
```
kbenne: Can this be figured out from context? I haven't read up on what this is, but it feels like this is one of those E+ inputs that there is only one obvious choice for based on context, and if that is the case then it should be handled during translation, and relieve the user from worrying about it.

openstudiocore/src/model/HeatPumpWaterToWaterEquationFitCooling.hpp
```diff
@@ -60,16 +62,24 @@ class MODEL_API HeatPumpWaterToWaterEquationFitCooling : public WaterToWaterComp
   //@{
 
   /** In EnergyPlus 8.7.0 and above this field maps to the EnergyPlus field named "Reference Load Side Flow Rate" **/
-  double ratedLoadSideFlowRate() const;
+  boost::optional<double> ratedLoadSideFlowRate() const;
```
kbenne: This is a breaking API change and I don't think that is OK. I'm not sure what the right solution is. Clearly we need to support autosizing, and normally that would mean returning optional.

One idea is to return a sentinel like -999 for autosize, but that would be very non standard for OpenStudio. On the other hand it wouldn't break the API. @macumber I'm confident you will have an opinion, so lets hear it please.

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -11045,16 +11052,31 @@ OS:HeatPump:WaterToWater:EquationFit:Heating,
...
+   N14, \field Heating Compressor Power Coefficient 5
         \required-field
         \type real
+   N15, \field Reference Coefficient of Performance
+        \note This optional field is used to autosize Reference Heating Power Consumption
+        \note COP = Reference Heating Capacity / Reference Heating Power Consumption
+        \type real
+        \minimum> 0.0
+        \default 7.5
+   N16, \field Sizing Factor
```
kbenne: The problem with default values at the end of an object is they show up in that weird gray color in the inspector. I think you should initialize these to the default value in the constructor. I would also prefer them to be a required field and not default in the idd. Both of these changes would require an OS version translation to support existing models. Currently this PR is skirting by with no version translation, but it means old models will always have that IG quirk.

openstudiocore/resources/model/OpenStudio.idd
asparke2: I don't think it can be figured out from context b/c these things are attached to a plant loop on both sides.  So you'd typically have the heating one connected to one plant's supply and another plant's demand, and the cooling one connected to one plant's supply and another plant's demand too.  If there were just one other of the opposite type on the same side of the plant loop then you could infer it, but if there are multiple HPs I don't think you can.  Also, if you are only using cooling or only using heating, there might not even be a companion in the model (which is still valid in E+).

openstudiocore/src/model/HeatPumpWaterToWaterEquationFitCooling.hpp
asparke2: Good point.  The only thing I can think of is that this is a pretty uncommon object, so a breaking change wouldn't affect many people (I'm guessing?)

nrel-bot-3:
@kbenne @asparke2 @lgentile it has been 7 days since this pull request was last updated.

Issue #2192, opened 2016-04-12
CooledBeam is not correctly translated


Anto-F:
When I use this object :
![1](https://cloud.githubusercontent.com/assets/10023510/14458235/1819320e-00b0-11e6-95cd-397a95db2a5b.PNG)
I get the following error :
** Severe  ** IP: IDF line~606 Error detected in Object=AIRTERMINAL:SINGLEDUCT:CONSTANTVOLUME:COOLEDBEAM, name=CHILLED BEAM
   **   ~~~   ** Field [Cooled Beam Type] is required but was blank

I tried to export the .osm in .idf and it confirmed the field 'Cooled Beam Type' is not translated : 
![2](https://cloud.githubusercontent.com/assets/10023510/14458293/80f950b0-00b0-11e6-877a-ca105bd4fbdb.PNG)

jmarrec:
https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateAirTerminalSingleDuctConstantVolumeCooledBeam.cpp#L94

CooledBeamType() returns a std::string, not an optional.

Needs to be changed to:

```c++
  // Field A3 Cooled Beam Type
  idfObject.setString(AirTerminal_SingleDuct_ConstantVolume_CooledBeamFields::CooledBeamType, modelObject.cooledBeamType());
```

jmarrec:
@macumber this issue can be closed, by #2611

PR #2611, opened 2017-05-12
CooledBeamType returns a std::string, not an optional string. See #2192

jmarrec:
@asparke2


nrel-bot-3:

macumber:
I don't see why the previous code would fail but the newer code is more correct

Issue #2613, opened 2017-05-18
OS app Crash if update measures and then quickly go to run tab and start sim


DavidGoldwasser:
@macumber I'm filing this for @brianlball who found this on Windows. He was not going really quick, but was running I think before measure manager finished. I thought we had the GUI locked out until measure mangers was done, but I guess at least in this case it still is not, or something else is happening.

macumber:
@rpg777 also found a crash if you open app, then open OSM before tabs are enabled (app is locked but top menu bar is still active).  Might be different than @brianlball 's crash

jmarrec:
I just had a similar issue on macOS Sierra, OS App 2.1.1. I clciked "Sync project measures with library", one needed updating. The tab refreshed with no measures in it, and OS App crashed immediately.
I just reopened my project and did it a second time, it worked though.

I then tried one more time, and while the page was refreshing I just switched to the Config tab, upon going back to the Measures tab OS Crashed.

macumber:
@DavidGoldwasser can you test this in 2.2.0?

macumber:
@brianlball and @rpg777 can you test as well?

DavidGoldwasser:
@macumber I was not able to get a crash trying to quickly run or opening an OSM. I also tried changing from and back to a directory with large number of measures using "Change my Measures Directory" (This is with 2.1.2 installer)

rpg777:
I'm still waiting for a 2.2.0 Mac package to test this. This wait dialog commit from Kyle is not in 2.1.2.

asparke2:
@dstuder can replicate this repeatably on Windows.

macumber:
@dstuder can you reproduce this in 2.4.0 installer?

dstuder:
Yes, I can replicate on Windows 7.

Specifically, if I sync my measures, and then while OS is repopulating the list of measures on the Measures tab, if I quickly click on the Run Simulation tab, OS crashes.

jmarrec:
So while the sync/list refreshing is happening, we need to lock the app (QMutex::lock), don’t we?

jmarrec:
I can reproduce this on current develop (2.6.0) built in Debug mode on Ubuntu 16.04.
I open a model where I have one outdated measure, go to the measures tab, hit "Sync", then "Update", hit "Ok" for "1 Measure(s) updated", then quickly go to sim tab and hit the run button.

I did it with a debugger to see the root cause of the Segfault, here's the trace back:

```
[...]

PR #2664, opened 2017-06-19
Show wait dialog while measure manager and libs are spinning up

kbenne:
use a thread to keep things going.
ref #2613

### DIFF
--- a/openstudiocore/src/openstudio_app/OpenStudioApp.cpp
+++ b/openstudiocore/src/openstudio_app/OpenStudioApp.cpp
@@ -119,6 +119,7 @@
 #include <QWidget>
 #include <QProcess>
 #include <QTcpServer>
+#include <QTConcurrent>
 
 #include <OpenStudio.hxx>
 #include <utilities/idd/IddEnums.hxx>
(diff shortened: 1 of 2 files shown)


macumber:
Looks good to me, I'll test it out

openstudiocore/src/openstudio_app/OpenStudioApp.cpp
```diff
@@ -1134,16 +1102,16 @@ void OpenStudioApp::startMeasureManagerProcess(){
   }
 
   // find available port
-  QTcpServer* tcpServer = new QTcpServer(this);
-
-  tcpServer->listen(QHostAddress::LocalHost);
-  quint16 port = tcpServer->serverPort();
-  tcpServer->close();
-  delete tcpServer;
```
macumber: I can see a race condition here where one process thinks it has a lock on a port but then another process finds the same port before this one is locked

macumber: Not a deal for now, would affect multiple apps launched right after each other

PR #2699, opened 2017-07-26
Enable chiller heater and central heat pump

jmarrec:
A lot of work has been done last year (around 1.10) but it's still not available in the OS API even though it's listed: https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-2.1.1-doc/model/html/classopenstudio_1_1model_1_1_central_heat_pump_system.html

This was waiting for the tertiary loop to be implemented, which should be ready now.

I wrote a test for the chiller heater because it was lacking one.

The CentralHeatPump also needs a Gtest btw.

@mbadams5 @kbenne


jmarrec:
I'm still working on this right now

PR #2700, opened 2017-07-26
Enable chiller heater and central heat pump

jmarrec:
re https://github.com/NREL/OpenStudio/pull/2699

A lot of work has been done last year (around 1.10) but it's still not available in the OS API even though it's listed: https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-2.1.1-doc/model/html/classopenstudio_1_1model_1_1_central_heat_pump_system.html

This was waiting for the tertiary loop to be implemented, which should be ready now.

I wrote a test for the chiller heater because it was lacking one.

The CentralHeatPump also needs a Gtest btw.

@mbadams5 @kbenne


jmarrec:
I'm fixing some warnings too (override, virtual, etc)

mbadams5:
Still need to look at and update the tests if needed.

jmarrec:
See if without the whitespace diffs: https://github.com/NREL/OpenStudio/pull/2700/files?w=1

jmarrec:
Regarding the issue with the curve type for cool/heat EIRFPLR: https://github.com/NREL/EnergyPlus/issues/6224#issuecomment-318398303

jmarrec:
Where should the "Environment" keyed variable go? Model.cpp? Site.cpp? building.cpp?

see https://github.com/NREL/OpenStudio/pull/2700/commits/332e88e576b263a59f6b2f49f96342f6035c9e0c

macumber:
There is a WeatherFile object, I've thought that variables that truly come from the EPW (e.g. air temp) should go there.

jmarrec:
I'd like these to be listed somewhere so that are available both in the APi and in the OS APP tab "Output Variable". Let me know where thanks.

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

jmarrec:
I still have one issue.

To recap: There's the `CentralHeatPumpSystem`, which can have multiple `CentralHeatPumpSystemModule`s. The `CentralHeatPumpSystemModule` references a single `ChillerHeaterPerformanceElectricEIR` (which is its performance object).

My idea is that several modules can share the same `ChillerHeatPerformanceElectricEIR`. You could do that for example if you have twice the same machine but with different schedules applied to (at the `CentralHeatPumpSystem` level), like one would work constantly and the other would come online during the day or something.

So when I clone a `CentralHeatPumpSystemModule`, I want to clone to point the same `ChillerHeaterPerformanceElectricEIR` as the original one. Maybe this isn't the correct term, but the `ChillerHeaterPerformanceElectricEIR` should be a "resource" just like curves.

@kbenne told me that if I declared the `ChillerHeaterPerformanceElectricEIR` object as children of the  `CentralHeatPumpSystemModule` ([CentralHeatPumpSystemModule.cpp#L119](https://github.com/jmarrec/OpenStudio/blob/e56f1fc2bcb0c8775d38cf8d03334b659d8588bc/openstudiocore/src/model/CentralHeatPumpSystemModule.cpp#L119)), the clone should work like I wanted. But it doesn't. Looking at [ModelObject::Clone](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/ModelObject.cpp#L834) which uses [ModelObject::resources](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/ModelObject.cpp#L504)it seems that it will do that if the object is declared as "ResourceObject"

I have also already created a custom clone function (commented out for now) [CentralHeatPumpSystemModule.cpp#L100](https://github.com/jmarrec/OpenStudio/blob/e56f1fc2bcb0c8775d38cf8d03334b659d8588bc/openstudiocore/src/model/CentralHeatPumpSystemModule.cpp#L100)


**Would you rather I:**

* uncomment the clone function I wrote, or,
* have ChillerHeaterPerformance inherit ResourceObject?

jmarrec:
**Edit: Clone works with children actually...**

I think it may be working now, need to confirm after my build is complete, I'm really not sure why... #confused

In any case, I'm still interested in hearing thoughts about why which is preferred going forward, and whether I should declare the "parent" or not.

https://github.com/jmarrec/OpenStudio/blob/a0185e6c9122f612723fad17ff4535b9ae877917/openstudiocore/src/model/CentralHeatPumpSystemModule.cpp#L100
I have also already created a custom clone function (commented out for now) [CentralHeatPumpSystemModule.cpp#L100](https://github.com/jmarrec/OpenStudio/blob/e56f1fc2bcb0c8775d38cf8d03334b659d8588bc/openstudiocore/src/model/CentralHeatPumpSystemModule.cpp#L100)

openstudiocore/src/model/CentralHeatPumpSystemModule.cpp
```diff
@@ -99,19 +99,17 @@ namespace detail {
...
     CentralHeatPumpSystemModule newCentralHPMod = ModelObject_Impl::clone(model).cast<CentralHeatPumpSystemModule>();
-    // ASSERT that it does have the same Perf component as the original Module
-    OS_ASSERT(newCentralHPMod.chillerHeaterModulesPerformanceComponent() == this->chillerHeaterModulesPerformanceComponent());
 
     // If not using "children", then expliclity do it:
-    // bool ok = true;
+    bool ok = true;
     // We don't want to clone the perf object, set it to the same ChillHeaterPerformanceElectricEIR
-    //ok = newCentralHPMod.setChillerHeaterModulesPerformanceComponent(this->chillerHeaterModulesPerformanceComponent());
+    ok = newCentralHPMod.setChillerHeaterModulesPerformanceComponent(this->chillerHeaterModulesPerformanceComponent());
```
jmarrec: Assertion fails
```
[ RUN      ] ModelFixture.CentralHeatPumpSystem_Clone
[BOOST_ASSERT] <2> Assertion ok failed on line 112 of virtual openstudio::model::ModelObject openstudio::model::detail::CentralHeatPumpSystemModule_Impl::clone(openstudio::model::Model) const in file /Users/julien/Software/Others/OpenStudio/openstudiocore/src/model/CentralHeatPumpSystemModule.cpp.
Assertion failed: (false), function assertion_failed, file /Users/julien/Software/Others/OpenStudio/openstudiocore/src/model/test/../../utilities/core/Assert.hpp, line 60.```

jmarrec: This happens on line 316 of GTest, when I clone into another model... obviously! It's trying to set a chillerHeater that wasn't clone in model2.

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

jmarrec:
Since this was not getting any traction since my question 2 months ago, I decided to fix it my way. Perhaps there was another way, but this works, and I can other examples of checking for (model == this->model()) in other files, so I think it isn't bad practice.

This is ready for review and merge @kbenne @mbadams5

jmarrec:
I merged with latest develop to make it easier on you.
I also stripped all trailing whitespace for all files (with a given extension, see commit message in 85eb764 ) in the last commit, if that's problematic, just discard this commit.

I've checked that it builds and tests still pass.

jmarrec:
FYI @macumber @kbenne 
I remember now what I did "extra" in this PR that makes the number of files/lines changed so high:

* Fixed a bunch of build warnings (that's ok)
* But more importantly, I implemented `outputVariableNames()` for pretty much every object in the SDK (except the keyed "Environment" variables as we didn't reach a consensus on where they should go, see https://github.com/NREL/OpenStudio/pull/2700#issuecomment-319427475).

I expect the OS App "Output Variable" tab to now be pretty unusable for large(r) models. I'm building develop now and will test shortly after and report back.

aaron-boranian:
This pull request is related to a [UserVoice](http://openstudio.uservoice.com/forums/308629/suggestions/12273603) idea.

Issue #2647, opened 2017-06-13
Intersect and Surface Match Regression Between v2.1 and v1.14


mattdoiron:
We've recently tried upgrading to Sketchup 2017 and OpenStudio v2.1 and have been having problems with intersection and surface matching. I've attached a few screenshots and the file in question.

This is a screenshot of before intersection and surface matching (in 2017+v2.1)
![image](https://user-images.githubusercontent.com/3885555/27094946-c6c72fac-5020-11e7-85aa-f03e81f453f5.png)

This is the result of "Intersect in Entire Model" and "Surface match entire model" when performed in 2017 v2.1.
![image](https://user-images.githubusercontent.com/3885555/27094999-f33de3d2-5020-11e7-89ff-56304eed7c39.png)

Below is the expected result as seen in 2016 with v1.14. As can be seen, there are still some areas to be fixed manually, but this is to be expected with a complicated plane of intersection. This is acceptable and can easily be fixed manually.
![image](https://user-images.githubusercontent.com/3885555/27095026-0a8ff16a-5021-11e7-9984-ae4aa3b3e5ff.png)

Here is a link to the osm file in question.
https://www.dropbox.com/s/n213wyuvxwnwrzf/2ndMain-Test-2016-1.14.osm?dl=0

Has any one else experienced this issue? We've had 2-3 people in our group with similar experiences.

Intrinzec-Hackerman:
Try and not use the openstudio plugin surface matching or intersecting tools and instead download a BCL measure called "SurfaceMatching". Once you are done modeling in the plugin, open it up inside of the actual openstudio app and at the very top under "Components and Measures" click on apply measure now. It will be under envelope->Form->SurfaceMatching. Then click apply. There is a check box that allows the model to be intersected before the measure is run. 

DO NOT INTERSECT OR MATCH INSIDE THE OPENSTUDIO PLUG IN.

Any remaining surfaces that did not match you can then match them with the inspector.
Good luck.

mattdoiron:
I tried this method and it appears to work as well as the previous versions. Is there a reason that the tools provided in the plugin are no longer as robust? They have worked well for us for many versions. If they have been changed this should be noted in the release notes along with deprecation warnings, suggested future workflow, etc.

ljbrackney:
[...]

PR #2783, opened 2017-09-14
Adding intersect space geometry measure as a user script for alternat…

macumber:
…e to sketchup intersect method.  Workaround for #2647


macumber:
@DavidGoldwasser can you test by adding the attached user script and modifying the SurfaceMatching.html file is sufficient?  Basically this just says if you experience issues with this dialog try using the user script instead.

DavidGoldwasser:
Yes, this works and is fine for now. At some point it might be nice to just have the surface matching dialog use our code for intersect vs. having them go to user script. One thing for users to be aware if is that if they intersect and have issues, they should close without saving, or at least save as another name, then re-open and use our intersect; but that is kind of long to put on a dialog.

Issue #2778, opened 2017-09-13
Luminaire won't allow input higher than 1.0 for Lighting Power


DavidGoldwasser:
https://unmethours.com/question/26532/unable-to-edit-luminaire-definitions-in-os/

GUI only lets you enter values from 0.0 to 1.0. Should be able to enter any positive number. I haven't confirmed if this is GUI issue or affects model API as well.

jmarrec:
Fixed in #2815.

Here is the default luminaire:

![2778_2710_luminaire_inputs_1_default](https://user-images.githubusercontent.com/5479063/31218481-bf773a96-a9ba-11e7-9f9c-33e5271529db.png)

And now with user input, I can edit all "fraction" fields, and I can also input a Lighting Power that is greater than 1.

![2778_2710_luminaire_inputs_2_user_input](https://user-images.githubusercontent.com/5479063/31218485-c6989ab8-a9ba-11e7-88ad-18a19d804cf6.png)

macumber:
This looks good, noticed that we can have sum of the fractions add up to > 1.  I'll file that as a separate issue https://github.com/NREL/OpenStudio/issues/2867

Issue #2710, opened 2017-08-04
Can't edit fraction radiant, visible, or return air fraction for luminaires and lights


macumber:

jmarrec:
Fixed in #2815, see #2778 for details.

macumber:
Can add up to > 1, filed as https://github.com/NREL/OpenStudio/issues/2867

Issue #2463, opened 2017-02-02
Purge unused not working in Stories subtab


eringold:
OSApp v2.0, Windows. 

In 'Stories' sub-tab of 'Facility' tab, there is a purge unused button. It remains unusable by default, even when un-referenced stories exist in the model. The button _does_ become enabled when a building story row is selected with the check-box. However, when the button is clicked, **all** stories are removed.

jmarrec:
Fixed in #2815.

Here are four spaces, notice that only "Building Story 1" and "Building Story 2" are used.

![2463_purge_stories_1](https://user-images.githubusercontent.com/5479063/31218762-9e78fc2a-a9bb-11e7-8e7e-112cadfdf4f6.png)

And here is the Building Story tab, you can see I have "Building Story 3" too which is unused, and the purge button is active.

![2463_purge_stories_2](https://user-images.githubusercontent.com/5479063/31218878-e67254c2-a9bb-11e7-9895-4c22c0c25daf.png)

After clicking the button, the unused building story (no spaces attached) has been purged:

![2463_purge_stories_3](https://user-images.githubusercontent.com/5479063/31218891-f4034402-a9bb-11e7-90e5-63ca39d14ee7.png)

Issue #2296, opened 2016-07-07
AirLoopHVAC:UnitarySystem Feature Request


dstuder:
I'd like to request the ability to manipulate the “Control Type” parameter in the AirLoopHVAC:UnitarySystem object. While the object itself is already wrapped, that specific parameter doesn’t appear to be. Thanks.

jmarrec:
That shouldn't hard to change in the source code, and yes it's annoying. As a temporary workaround you can call setString... `unitary.setString 2, 'SetPoint'`

kbenne:
If the unitary is wrapping a single coil then everything is pretty straightforward. It is as easy as @jmarrec points out. If there are multiple coils inside the unitary that you want to have on setpoint control then it is tricky because those nodes "inside" the unitary are not created until translation to idf. 

Still. I'm in favor of exposing the controlType parameter.

jmarrec:
in #2815, I exposed the public `controlType()` (and related `setControlType()`, `isControlTypeDefaulted()` and `resetControlType()`), and wrote a Gtest for this.

I exposed the field in the OS App:

![2296_unitarysystem_controltype_1](https://user-images.githubusercontent.com/5479063/31218134-a49b9a1a-a9b9-11e7-9938-e37f002f5bbf.png)


Loading the model I saved in OS App, I can confirm it worked:

![2296_unitarysystem_controltype_2](https://user-images.githubusercontent.com/5479063/31218138-a9b0ae96-a9b9-11e7-976d-03db1dbe94ed.png)

aaron-boranian:
Allow setpoint control within the AirLoopHVAC:UnitarySystem.

This would be helpful for DOAS systems where the airloop is serving multiple zones.

Migrated from [UserVoice](http://openstudio.uservoice.com/forums/308629/suggestions/9565998)

PR #2815, opened 2017-10-05
Fix qt display

jmarrec:
Addressed #2296, #2266 and #2152, #2778 and #2710, #2112, #2175, #86, #2463 and partially #2176.

Proof of functionaly will be provided in the respective issues thread.

### DIFF
--- a/openstudiocore/src/openstudio_lib/FacilityStoriesGridView.cpp
+++ b/openstudiocore/src/openstudio_lib/FacilityStoriesGridView.cpp
@@ -48,6 +48,10 @@
 #include "../model/RenderingColor.hpp"
 #include "../model/RenderingColor_Impl.hpp"
 
+// For purgeObjects
+#include "../model/Space.hpp"
+
+
 #include "../utilities/core/Assert.hpp"
 #include "../utilities/idd/IddEnums.hxx"
 #include "../utilities/idd/OS_BuildingStory_FieldEnums.hxx"
--- a/openstudiocore/src/openstudio_lib/FacilityStoriesGridView.cpp
+++ b/openstudiocore/src/openstudio_lib/FacilityStoriesGridView.cpp
@@ -176,7 +176,7 @@ namespace openstudio {
   void FacilityStoriesGridView::purgeObjects(const IddObjectType& iddObjectType)
   {
     // If no spaces in the story -> remove
-    for (model::BuildingStory story : this->model().getConcreteModelObjects<model::BuildingStory>()){
+    for (model::BuildingStory story : this->m_model.getConcreteModelObjects<model::BuildingStory>()){
       if (story.spaces().empty()){
         story.remove();
       }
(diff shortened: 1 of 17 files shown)


nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

jmarrec:
Also fixed #2839 since it was a very quick fix.

openstudiocore/src/openstudio_lib/FacilityStoriesGridView.cpp
```diff
@@ -233,16 +240,16 @@ namespace openstudio {
   {
     //m_itemSelectorButtons->enableAddButton();
     //m_itemSelectorButtons->enableCopyButton();
-    m_itemSelectorButtons->enableRemoveButton();
```
macumber: @jmarrec this prevents us from deleting stories, is that intended?

macumber: Copying stories is disabled in last release (assume that is on purpose) but delete is enabled if you select something.  Purge should always be enabled.

openstudiocore/src/openstudio_lib/SpacesSubtabGridView.cpp
```diff
@@ -866,8 +867,12 @@ namespace openstudio {
 
   void SpacesSubtabGridView::purgeObjects(const IddObjectType& iddObjectType)
   {
-    for (auto mo : this->m_model.getConcreteModelObjects<model::Space>()){
-      mo.remove();
+    // Purge Spaces. The logic is to remove spaces that don't have a thermalZone nor surfaces
+    for (auto mo : this->m_model.getConcreteModelObjects<model::Space>())
```
macumber: @jmarrec this seems like reasonable logic, but I wonder if there are any other things that should be added?  shading groups?

macumber: Also should thermal zone be considered?  What about building story?

macumber: Seems like it would be a good idea to add an API method to all objects that was "used" or something, that way you could see which objects were going to be purged before you did it.  Oh well, sometime later.

jmarrec: I think the real key here is geometry. The only thing a space must have is geometry.
I'm not sure whether you'd add a space with just shading?
If not, then the space should be used by a ThermalZone, because E+ doesn't have the concept of space, it ends up in a zone, no?

openstudiocore/src/openstudio_lib/FacilityStoriesGridView.cpp
jmarrec: I may have messed it up. The only thing that should toggle is the `RemoveButton` when selected/unselect, and purge should always be enabled yes (all buttons are enabled by default I believe)

Issue #833, opened 2014-02-02
When a surface boundary condition is changed to "Outdoors" the default should be to make it sun and wind exposed


DavidGoldwasser:
Currently it is not. This makes setting up attics with overhangs harder to model.

macumber:
Already done https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/Surface.cpp#L406

DavidGoldwasser:
@macumber they are both set to false now. I think they should both be true.

jmarrec:
https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/Surface.cpp#L434

jmarrec:
David, you're experiencing that when you do it in the Sketchup inspector, right? That is doing only workspace stuff according to @macumber 

If you use the API and do `surface.setOutsideBoundaryCondition("Outdoors")` it does change it to SunExposed and WindExposed (and if you do "Adiabatic", it does change them to NoSun and noWind)

Considering the current plans for the sketchup plugin, I think this one might need to be tagged "Won't fix"  and closed...

DavidGoldwasser:
Looks like API and GUI now change this, so this can be closed, but I made new crash but in the process of testing this. One step forward one step back :)
https://github.com/NREL/OpenStudio/issues/2822

PR #2823, opened 2017-10-16
Fix #2215 make ZoneVentilationDesignFlowRate behave like SpaceInfiltrationDesignFlowRate (switch the flow rate calc method automatically)

### LINKED ISSUE #2215: ZoneVentilation:DesignFlowRate doesn't react as usual when setting flow
If you use any of the `setDesignFlowRate`, `setFlowRateperZoneFloorArea`, `setFlowRateperPerson`, or `setAirChangesperHour`, this will NOT change the `DesignFlowRateCalculationMethod`. You have to explicility do so using `setDesignFlowRateCalculationMethod`.

This not the way `SpaceInfiltrationDesignFlowRate` works for example.

setDesignFlowRateCalculationMethod shouldn't not exist as a method, and using the `setDesignFlowRate` should change the calculation method to `Flow/zone` automatically.

jmarrec:
Fix #2215 (also tested #833 and got it closed by David)

Deprecated the `ZoneVentilationDesignFlowRate::setDesignFlowRateCalculationMethod`, and make it behave like SpaceInfiltrationDesignFlowRate:

if you call setAirChangesperHour(double), it will automatically set the CalculationMethod to "AirChanges/Hour" and put all of other fields as 0.0  (fyi, in SpaceInfiltration, the others are actually optional, and have a default value... )

review assignee: @asparke2 (original #2215 assigned)

### DIFF
--- a/openstudiocore/src/model/ZoneVentilationDesignFlowRate.cpp
+++ b/openstudiocore/src/model/ZoneVentilationDesignFlowRate.cpp
@@ -542,11 +542,9 @@ ZoneVentilationDesignFlowRate::ZoneVentilationDesignFlowRate(const Model& model)
     setSchedule(schedule);
   }
 
-  setDesignFlowRateCalculationMethod("AirChanges/Hour");
-  setDesignFlowRate(0.0);
-  setFlowRateperZoneFloorArea(0.0);
-  setFlowRateperPerson(0.0);
+  // This automatically switches the Calculation Method to "AirChanges/Hour"
   setAirChangesperHour(5.0);
+
   setVentilationType("Natural");
   setFanPressureRise(0.0);
   setFanTotalEfficiency(1.0);
--- a/openstudiocore/src/model/ZoneVentilationDesignFlowRate.cpp
+++ b/openstudiocore/src/model/ZoneVentilationDesignFlowRate.cpp
@@ -270,17 +270,19 @@ namespace detail {
     bool result = true;
     if (designFlowRate < 0){
       result = false;
-    }else{
-      result = setString(OS_ZoneVentilation_DesignFlowRateFields::DesignFlowRateCalculationMethod, "Flow/Zone");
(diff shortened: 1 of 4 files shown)


nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

openstudiocore/src/model/ZoneVentilationDesignFlowRate.cpp
```diff
@@ -259,27 +261,84 @@ namespace detail {
...
-    bool result = setDouble(OS_ZoneVentilation_DesignFlowRateFields::DesignFlowRate, designFlowRate);
+    bool result = true;
+    if (designFlowRate < 0){
+      result = false;
+    }else{
+      result = setString(OS_ZoneVentilation_DesignFlowRateFields::DesignFlowRateCalculationMethod, "Flow/Zone");
+      OS_ASSERT(result);
+      result = setDouble(OS_ZoneVentilation_DesignFlowRateFields::DesignFlowRate, designFlowRate);
+      OS_ASSERT(result);
+      result = setDouble(OS_ZoneVentilation_DesignFlowRateFields::FlowRateperZoneFloorArea, 0.0);
```
macumber: @jmarrec technically you wouldn't want to set these fields if the previous 

`result = setDouble(OS_ZoneVentilation_DesignFlowRateFields::DesignFlowRate, designFlowRate);`

had failed (e.g. because the user passed in NaN, that is just fresh in my brain now).

macumber: Otherwise looks good

jmarrec: Good point. I did the setDouble stuff a month after this. I pushed another commit to address this.

Issue #2757, opened 2017-08-27
Space load instances have non-editable fields in inspector


macumber:
Need to fully create fields so these are editable

DavidGoldwasser:
@macumber Just out of interest, what is not editable? Is this about End-Use Subcategory, or are you wanting to change out definitions in the GUI without making a new load instance?

macumber:
Lots of stuff is not editable, subcategory is one.  Need to create all the fields in model object ctor.

DavidGoldwasser:
Ah, I see for specific instances types like people. Some or not editable and others like work efficiency schedule are missing.

jmarrec:
Fixed in #2826 

I had originally set all fields in Ctor (most often there are 3: Schedule, Multiplier and EndUseSubCategory), and realized later that:
1. Setting only the last is enough to see them
2. Setting some of them, like schedule, is a bad idea because then it won't inherit from the grid.

So I reverted back to only setting the last one (in general EndUseSubCategory, people is Mutliplier for eg).

I also wrote a ruby script in order to populate the same last fields for all the Library OSMs too (I uploaded it with the PR in case...)

![electricequipment](https://user-images.githubusercontent.com/5479063/31820818-4761ab46-b5a3-11e7-9efc-4ac01248966d.png)

![gasequipment](https://user-images.githubusercontent.com/5479063/31820819-47784054-b5a3-11e7-9bfb-ef1d78c217f0.png)

![internalmass](https://user-images.githubusercontent.com/5479063/31820820-478e6f14-b5a3-11e7-85fd-6a0cf1088621.png)

![lights](https://user-images.githubusercontent.com/5479063/31820821-47abcb4a-b5a3-11e7-9e99-0fa43fd917d7.png)

![luminaire](https://user-images.githubusercontent.com/5479063/31820822-47cc04a0-b5a3-11e7-83a5-d77e2519f293.png)

![otherequipment_withfueltypeexposed](https://user-images.githubusercontent.com/5479063/31820823-481fe638-b5a3-11e7-8eff-6489ee753a60.png)

![steamequipment](https://user-images.githubusercontent.com/5479063/31820824-483c8932-b5a3-11e7-9912-a0b824abac67.png)

PR #2826, opened 2017-10-20
Fix 2757 2714

### LINKED ISSUE #2714: Expose fuel type for other equipment in UI
@nmerket FYI

jmarrec:
Fixes #2714 and #2757. Screenshots provided in the issues themselves.

Assignee for review: @macumber

### DIFF
--- a/openstudiocore/src/model/ElectricEquipment.cpp
+++ b/openstudiocore/src/model/ElectricEquipment.cpp
@@ -40,6 +40,7 @@
 #include "DefaultScheduleSet.hpp"
 #include "DefaultScheduleSet_Impl.hpp"
 #include "LifeCycleCost.hpp"
+#include "Model.hpp"
 
 #include <utilities/idd/OS_ElectricEquipment_FieldEnums.hxx>
 #include <utilities/idd/IddEnums.hxx>
@@ -175,9 +176,10 @@ namespace detail {
     OS_ASSERT(result);
   }
 
-  void ElectricEquipment_Impl::setEndUseSubcategory(std::string endUseSubcategory) {
+  bool ElectricEquipment_Impl::setEndUseSubcategory(std::string endUseSubcategory) {
     bool result = setString(OS_ElectricEquipmentFields::EndUseSubcategory, endUseSubcategory);
     OS_ASSERT(result);
+    return result;
   }
 
   void ElectricEquipment_Impl::resetEndUseSubcategory() {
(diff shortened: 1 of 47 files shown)


nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

openstudiocore/src/model/OtherEquipment.cpp
```diff
@@ -315,6 +316,14 @@ OtherEquipment::OtherEquipment(const OtherEquipmentDefinition& definition)
   : SpaceLoadInstance(OtherEquipment::iddObjectType(),definition)
 {
   OS_ASSERT(getImpl<detail::OtherEquipment_Impl>());
+
+  /*
+   *Schedule sch = this->model().alwaysOnDiscreteSchedule();
```
macumber: @jmarrec I think this would have been too much to default in a ctor

openstudiocore/src/model/WaterUseEquipment.cpp
```diff
@@ -227,6 +227,11 @@ WaterUseEquipment::WaterUseEquipment(const WaterUseEquipmentDefinition& waterUse
   : SpaceLoadInstance(WaterUseEquipment::iddObjectType(),waterUseEquipmentDefinition)
 {
   OS_ASSERT(getImpl<detail::WaterUseEquipment_Impl>());
+
+  Schedule sch = this->model().alwaysOnDiscreteSchedule();
```
macumber: @jmarrec why are you doing this here?  seems like this is more than we need to do?  wouldn't a blank here be covered by a default schedule?

openstudiocore/src/openstudio_app/Resources/populate_loads.rb
```diff
@@ -0,0 +1,85 @@
+require 'openstudio'
```
macumber: @jmarrec why was it necessary to update all the osms?  they all defaulted to these values already right?

openstudiocore/src/model/ElectricEquipment.cpp
```diff
@@ -346,6 +348,17 @@ ElectricEquipment::ElectricEquipment(const ElectricEquipmentDefinition& electric
...
   OS_ASSERT(getImpl<detail::ElectricEquipment_Impl>());
+
+  /*
+   *Schedule sch = this->model().alwaysOnDiscreteSchedule();
+   *bool test = this->setSchedule(sch);
+   *OS_ASSERT(test);
+   *test = this->setMultiplier(1.0);
+   *OS_ASSERT(test);
+   */
+  bool test = this->setEndUseSubcategory("General");
```
macumber: @jmarrec this is already the default so don't really need to set this here?

 ```
 A6; \field End-Use Subcategory
       \type alpha
       \retaincase
       \default General
```

openstudiocore/src/model/OtherEquipment.cpp
macumber: this would be covered by a default schedule set so don't want to initialize a schedule

openstudiocore/src/model/ElectricEquipment.cpp
macumber: Oh I get it, this was to force the inspector to have this number of fields.  I guess you could have set it to "" but this is better

openstudiocore/src/model/WaterUseEquipment.cpp
macumber: i see, this was for inspector and this is the default value

jmarrec:
@macumber, in response to all the above comments: like I said on https://github.com/NREL/OpenStudio/issues/2757#issuecomment-338193148, the idea was to populate the last field in the object, so most of the time it's endusesubcategory, sometimes it's the multiplier or the schedule.

openstudiocore/src/openstudio_app/Resources/populate_loads.rb
jmarrec: See comment below: need to hardset the last field.

macumber:
Git just missed closing #2757, I closed it just now

Issue #2749, opened 2017-08-25
Better default for HeatExchangerFluidToFluid ComponentType


jmarrec:
There is a better way to infer whether it's a Heating or Cooling component for the HX:FluidToFluid. Currently it's set to "BOTH" all the time, see [here](https://github.com/NREL/OpenStudio/blob/4bbc02abe49616fefc2952c84cb6dca5edc32284/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslatePlantEquipmentOperationSchemes.cpp#L389)

But based on controlType, we can do better:
```
$ OpenStudio::Model::HeatExchangerFluidToFluid::controlTypeValues
=> ["UncontrolledOn",
 "OperationSchemeModulated",
 "OperationSchemeOnOff",
 "HeatingSetpointModulated",
 "HeatingSetpointOnOff",
 "CoolingSetpointModulated",
 "CoolingSetpointOnOff",
 "DualDeadbandSetpointModulated",
 "DualDeadbandSetpointOnOff",
 "CoolingDifferentialOnOff",
 "CoolingSetpointOnOffWithComponentOverride"]
```

If "HeatingXXX" or "CoolingXXX", should default to `ComponentType::HEATING` and `ComponentType::COOLING` respectively instead of `ComponentType::BOTH`

jmarrec:
This should about do it (replace the [line](https://github.com/NREL/OpenStudio/blob/4bbc02abe49616fefc2952c84cb6dca5edc32284/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslatePlantEquipmentOperationSchemes.cpp#L389) I pointed to with this):

```c++

HeatExchangerFluidToFluid hx = component.cast<HeatExchangerFluidToFluid>;

std::string controlType = hx.controlType();

if ( (controlType == 'HeatingSetpointModulated') ||
     (controlType == 'HeatingSetpointOnOff') ) {
    return ComponentType::HEATING;
} else if ( (controlType == 'CoolingSetpointModulated') ||
            (controlType == 'CoolingSetpointOnOff') ||
            (controlType == 'CoolingDifferentialOnOff') ||
            (controlType == 'CoolingSetpointOnOffWithComponentOverride') ) {
    return ComponentType::COOLING;
} else {
    return ComponentType::BOTH;
} 

```

jmarrec:
@kbenne please close per #2851

Issue #2275, opened 2016-06-14
Develop PlantEquipmentOperationScheme logic for ThermalStorage and develop logic to make unknown objects use uncontrolled scheme.


DavidGoldwasser:
Currently need to use measure to create PlantEquipmentOperationScheme logic to use ThermalStorage. Use case trying now has it on the supply side of one plant loop and the demand side of another.

jmarrec:
Is this still relevant today now that the PlantEquipmentOperationSchemes can be created in the API?

Additionally, if I look at https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslatePlantEquipmentOperationSchemes.cpp#L325

it should get wrapped in a PlantEquipmentOperationHeatingLoad.

I guess better default for CompType could be made by checking what kind of plant loop the tank is on the supply side of instead of defaulting to Heating. Is that what you want @DavidGoldwasser ?

DavidGoldwasser:
@jmarrec I think that was the case. I was trying to help someone add thermal storage at the time, and while thermal storage was in the GUI, there was a way to get the proper controls from the GUI. If defaults can handle it then no changes necessary in the GUI.

jmarrec:
@kbenne please close per #2851

PR #2851, opened 2017-11-17
Fix 2749 2275 plant op logic

jmarrec:
Better PlantEquipmentOperation defaults for `HeatExchangerFluidToFluid`, `WaterHeater:Mixed` and `WaterHeater:Stratified`, with Gtest #2749, #2275.

Here's the logic I have used:

**`HeatExchangerFluidToFluid`**:

* Check the `controlType` first, if it's a heating/cooling/dualsetpoint type then assign `ComponentType::HEATING`, `COOLING`, and `BOTH` respectively
* If it's `OperationSchemeModulated` or `OperationSchemeOnOff`, do as before and return BOTH.
* If it's `Uncontrolled`, check the secondary (source) loop supply side (call `plantLoopType`). If no secondary plant loop (weird for a HX...), then `NONE`.

** `WaterHeater:Mixed` and `WaterHeater:Stratified`:**

* If they have a heating capacity (autosized or hardsized > 0): `HEATING`.
* If they don't have a capacity it's a storage tank.
    * If no secondary loop, it's a buffer tank => `NONE`
    * Otherwise inherit the secondary loop (call `plantLoopType`)

**plantLoopType:** call componentType for each component on the supply side, and check the following:

* If there is only heating, no cooling, no "both": Heating
* If there is only cooling, no heating, no "both": Cooling
* If there is no cooling, no heating, no "both": None
* All other cases: "both"

Gtests have been added.

### DIFF
--- a/openstudiocore/CMakeLists.txt
+++ b/openstudiocore/CMakeLists.txt
@@ -1,10 +1,19 @@
 cmake_minimum_required(VERSION 3.7.0)
 cmake_policy(SET CMP0048 NEW)
 
+# Use ccache is available, has to be before "project()"
+find_program(CCACHE_PROGRAM ccache)
+if(CCACHE_PROGRAM)
+  # Support Unix Makefiles and Ninja
+  set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
+endif()
+
+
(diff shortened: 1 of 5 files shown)


nrel-bot-3:
@kbenne @jmarrec @lgentile it has been 7 days since this pull request was last updated.

Issue #2573, opened 2017-04-14
ScheduleDay accepts nan for values in API.


DavidGoldwasser:
Edge case as measure should not do this, but I did it by mistake, resulting schedule crashes app if the schedule is inspected, and EnergyPlus run fails.

API should be more strict on what it accepts for doubles.

jmarrec:
I can reproduce this, but it turns out this is definitely not limited to `ScheduleDay`, for example it works with `PeopleDefinition` too, and probably wherever the method accepts a double.

It looks like it would take a lot of work for a minor bug, unless someone has a better way of handling that other than (the naive way of) modifying each `setXXX` method that accepts a double to use `std::isnan()` for nan-checking.

```ruby
In [1]:
model = OpenStudio::Model::Model.new
sch_day = OpenStudio::Model::ScheduleDay.new(model)
sch_day.addValue(OpenStudio::Time.new(0,6,0,0), 0/0.0)
puts sch_day

Out[1]: OS:Schedule:Day,
  {4825a6f5-66dd-418f-bd4d-c9887edf24d6}, !- Handle
  Schedule Day 1,                         !- Name
  ,                                       !- Schedule Type Limits Name
  ,                                       !- Interpolate to Timestep
  6,                                      !- Hour 1
  0,                                      !- Minute 1
  nan,                                    !- Value Until Time 1
  24,                                     !- Hour 2
  0,                                      !- Minute 2
  0;                                      !- Value Until Time 2


In [2]: 
p_def = OpenStudio::Model::PeopleDefinition.new(model)
p_def.setNumberofPeople(0/0.0)
puts p_def

Out[2]:
OS:People:Definition,
  {c6a7b86a-e7d9-424a-96a3-fb16474614ed}, !- Handle
  People Definition 1,                    !- Name
  People,                                 !- Number of People Calculation Method
  nan,                                    !- Number of People {people}
  ,                                       !- People per Space Floor Area {person/m2}
  ,                                       !- Space Floor Area per Person {m2/person}
  0.3,                                    !- Fraction Radiant
  ,                                       !- Sensible Heat Fraction
  ,                                       !- Carbon Dioxide Generation Rate {m3/s-W}
  Yes;                                    !- Enable ASHRAE 55 Comfort Warnings
```

jmarrec:
Actually looks easier that I thought, it can just be handled in IdfObject:

[...]

PR #2852, opened 2017-11-17
Fix 2573: setDouble accepts NaNs and Infinity

jmarrec:
Fixed #https://github.com/NREL/OpenStudio/issues/2573

In IdfObject, now check if isnan of isinfinity.

Small Gtest added too.

### DIFF
--- a/openstudiocore/src/model/ScheduleDay.cpp
+++ b/openstudiocore/src/model/ScheduleDay.cpp
@@ -288,6 +288,10 @@ namespace detail {
       return false;
     }
 
+    if (std::isnan(value) || std::isinf(value)){
+      return false;
+    }
+
     int untilHours = untilTime.hours() + 24*untilTime.days();
     int untilMinutes = untilTime.minutes() + floor((untilTime.seconds()/60.0) + 0.5);
 
@@ -313,16 +317,6 @@ namespace detail {
       std::vector<std::string> groupValues;
       groupValues.push_back(boost::lexical_cast<std::string>(untilHours));
       groupValues.push_back(boost::lexical_cast<std::string>(untilMinutes));
-
-      // Check validity, cannot be NaN, Inf, etc
-      if (std::isinf(value)) {
-        LOG(Error, "Cannot setDouble to Infinity for " << this->briefDescription());
-        return false;
-      } else if (std::isnan(value)) {
-        LOG(Error, "Cannot setDouble to a NaN for " << this->briefDescription());
-        return false;
-      }
-
       groupValues.push_back(toString(value));
 
       IdfExtensibleGroup group = insertExtensibleGroup(index, groupValues);
--- a/openstudiocore/src/model/ScheduleDay.cpp
+++ b/openstudiocore/src/model/ScheduleDay.cpp
@@ -288,7 +288,12 @@ namespace detail {
       return false;
     }
 
-    if (std::isnan(value) || std::isinf(value)){
+    // Check validity, cannot be NaN, Inf, etc
+    if (std::isinf(value)) {
+      LOG(Warn, "Cannot setDouble to Infinity for " << this->briefDescription());
+      return false;
+    } else if (std::isnan(value)) {
+      LOG(Warn, "Cannot setDouble to a NaN for " << this->briefDescription());
       return false;
     }
 
--- a/openstudiocore/src/model/ScheduleDay.cpp
+++ b/openstudiocore/src/model/ScheduleDay.cpp
@@ -313,6 +313,16 @@ namespace detail {
       std::vector<std::string> groupValues;
       groupValues.push_back(boost::lexical_cast<std::string>(untilHours));
       groupValues.push_back(boost::lexical_cast<std::string>(untilMinutes));
+
+      // Check validity, cannot be NaN, Inf, etc
+      if (std::isinf(value)) {
+        LOG(Error, "Cannot setDouble to Infinity for " << this->briefDescription());
+        return false;
+      } else if (std::isnan(value)) {
+        LOG(Error, "Cannot setDouble to a NaN for " << this->briefDescription());
+        return false;
+      }
+
       groupValues.push_back(toString(value));
 
       IdfExtensibleGroup group = insertExtensibleGroup(index, groupValues);
--- a/openstudiocore/src/model/ScheduleFixedInterval.cpp
+++ b/openstudiocore/src/model/ScheduleFixedInterval.cpp
@@ -145,7 +145,14 @@ namespace detail {
     // check the values
     openstudio::Vector values = timeSeries.values();
     for (const auto& value : values){
-      if (std::isnan(value) || std::isinf(value)){
+      // Get the position
+      int pos = &value-&values[0];
+      // Check validity, cannot be NaN, Inf, etc
+      if (std::isinf(value)) {
+        LOG(Warn, "There is Infinity on position " << pos <<" in the timeSeries provided for " << this->briefDescription());
+        return false;
+      } else if (std::isnan(value)) {
+        LOG(Warn, "There is a NaN on position " << pos <<" in the timeSeries provided for " << this->briefDescription());
         return false;
       }
     }
--- a/openstudiocore/src/model/ScheduleFixedInterval.cpp
+++ b/openstudiocore/src/model/ScheduleFixedInterval.cpp
@@ -168,6 +168,16 @@ namespace detail {
     openstudio::Vector values = timeSeries.values();
     for (unsigned i = 0; i < values.size(); ++i){
       std::vector<std::string> temp;
+
+      // Check validity, cannot be NaN, Inf, etc
+      if (std::isinf(values[i])) {
+        LOG(Error, "Cannot setDouble to Infinity for " << this->briefDescription());
+        return false;
+      } else if (std::isnan(values[i])) {
+        LOG(Error, "Cannot setDouble to a NaN for " << this->briefDescription());
+        return false;
+      }
+
       temp.push_back(toString(values[i]));
 
       ModelExtensibleGroup group = pushExtensibleGroup(temp, false).cast<ModelExtensibleGroup>();
--- a/openstudiocore/src/model/ScheduleFixedInterval.cpp
+++ b/openstudiocore/src/model/ScheduleFixedInterval.cpp
@@ -142,6 +142,14 @@ namespace detail {
       return false;
     }
 
+    // check the values
+    openstudio::Vector values = timeSeries.values();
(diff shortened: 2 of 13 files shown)


jmarrec:
I realized that some methods, such as ScheduleDay::AddValue don't even call setDouble.

https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/model/ScheduleDay.cpp#L316

definition toString(double v): https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/utilities/core/String.cpp#L69

The Schedule Object are (as far as I know, for a cursory grep of src/model folder) the only ones that dangerously use toString with a double that is user-input. (the other toString calls are for string, or for double that are attributes that have been set in the Ctor or via a regular setDouble call).

toString(double v) now returns empty string if called with nan/inf. **Perhaps it would be good to Log an error to make sure, but unsure how)**

And I added checks in the Schedule objects before the call to toString itself.

openstudiocore/src/model/ScheduleFixedInterval.cpp
```diff
@@ -142,6 +142,14 @@ namespace detail {
       return false;
     }
 
+    // check the values
+    openstudio::Vector values = timeSeries.values();
```
macumber: Moved this check so if there is a failure it exits before the object is changed

openstudiocore/src/utilities/idf/IdfObject.cpp
```diff
@@ -1854,7 +1856,17 @@ void IdfObject_Impl::populateValidityReport(ValidityReport& report, bool checkNa
         else {
           return false;
         }
-      }
+      } else{
+        if (std::isnan(*value)) {
```
macumber: @jmarrec this catches the "Infinity" or "NaN" in the field after it has been set but during validity checking, this will prevent issues where you bypass setDouble using the pushExtensibleGroup code

macumber: The only issue is this code only runs when using it on WorkspaceObjects or ModelObjects, won't work on IdfObjects (but would work if you tried to add an IdfObject to a Workspace)

openstudiocore/src/utilities/idf/Test/IdfObject_GTest.cpp
```diff
@@ -743,3 +742,72 @@ TEST_F(IdfFixture, DoubleDisplayedAsString) {
...
 
+TEST_F(IdfFixture, IdfObject_SetDouble_NaN_and_Inf) {
+
+  // try with an IdfObject
+  // IdfObject does not prevent Infinity and NaN
+  IdfObject object(IddObjectType::OS_People_Definition);
+
+  // Set Number of People
+  // Check for nan
+  EXPECT_TRUE(object.setDouble(3, std::numeric_limits<double>::quiet_NaN()));
```
macumber: @jmarrec check out the difference between EXPECT_ and ASSERT_ in gtest.  EXPECT statements will fail but allow the test to continue, ASSERTS immediately exit on first failure.  It's helpful to see multiple failures at once if possible

macumber:
@jmarrec back to you, if you approve of my changes this is ready to go

openstudiocore/src/model/ScheduleDay.cpp
```diff
@@ -288,6 +288,10 @@ namespace detail {
       return false;
     }
 
+    if (std::isnan(value) || std::isinf(value)){
+      return false;
+    }
+
```
jmarrec: No problem on the position. But this is going to return false without any indication of why. Outputing a message like I did seems like a good thing to me.

openstudiocore/src/utilities/idf/Test/IdfObject_GTest.cpp
jmarrec: I understand the difference between the two, it's a bit of a philosophical question here: I consider that if any of these fails, I must fix it. My "internal" policy was to always use ASSERT (coming from Python, that's what I'm used to anyways), but if EXPECT are preferred I'm fine with it, I just didn't know if was a thing.

I'm pretty sure it's quite inconsistent in the existing GTests though. And you should expect (no pun intended) the same behavior on all the Gtests I've written so far unfortunately :/

If you want to be able to see as much messages as possible, perhaps you could make it a policy to only use EXPECT and no ASSERT (it's easier to remember and apply (because it's binary) than saying "prefer EXPECT")

I understand that ASSERT is for stuff where it's meaningless to continue if it fails, but if you take the following example, I don't think it provides more valuable input:

On line 758 to 760 below, you wrote:
```
EXPECT_TRUE(object.setDouble(3, std::numeric_limits<double>::infinity()));
ASSERT_TRUE(object.getDouble(3));
EXPECT_TRUE(std::isinf(object.getDouble(3).get()));
```
why use an ASSERT here? It'll crash on the next line if the it wasn't ok anyways, so that isn't going to provide more valuable input, is it?

jmarrec:
@macumber I'm getting tons of build errors from the contam stuff about call to toString:

```
/openstudiocore/src/airflow/contam/PrjAirflowElementsImpl.cpp:4758:47: error: call of overloaded ‘toString(std::vector<openstudio::contam::FanDataPoint>::size_type)’ is ambiguous
   string += ANY_TO_STR(m_data.size()) + ' ' + ANY_TO_STR(m_Sarea) + ' ' + ANY_TO_STR(m_u_Sa) + '\n';
                                               ^
In file included from /home/julien/Software/Others/OpenStudio2/openstudiocore/src/qtwinmigrate/../utilities/core/LogMessage.hpp:34:0,
                 from /home/julien/Software/Others/OpenStudio2/openstudiocore/src/qtwinmigrate/../utilities/core/LogSink.hpp:34,
                 from /home/julien/Software/Others/OpenStudio2/openstudiocore/src/qtwinmigrate/../utilities/core/Logger.hpp:37,
                 from /home/julien/Software/Others/OpenStudio2/openstudiocore/src/airflow/contam/PrjReader.hpp:35,
                 from /home/julien/Software/Others/OpenStudio2/openstudiocore/src/airflow/contam/PrjAirflowElementsImpl.hpp:33,
                 from /home/julien/Software/Others/OpenStudio2/openstudiocore/src/airflow/contam/PrjAirflowElementsImpl.cpp:29:
/home/julien/Software/Others/OpenStudio2/openstudiocore/src/qtwinmigrate/../utilities/core/String.hpp:64:15: note: candidate: std::__cxx11::string openstudio::toString(double)
   UTILITIES_API std::string toString(double v);
               ^
/home/julien/Software/Others/OpenStudio2/openstudiocore/src/qtwinmigrate/../utilities/core/String.hpp:67:15: note: candidate: std::__cxx11::string openstudio::toString(int)
   UTILITIES_API std::string toString(int v);
               ^
/home/julien/Software/Others/OpenStudio2/openstudiocore/src/qtwinmigrate/../utilities/core/String.hpp:70:15: note: candidate: std::__cxx11::string openstudio::toString(unsigned int)
   UTILITIES_API std::string toString(unsigned v);

```

jmarrec:
[PrjAirflowElementsImpl.cpp#L4758](https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/airflow/contam/PrjAirflowElementsImpl.cpp#L4758)

    string += ANY_TO_STR(m_data.size()) + ' ' + ANY_TO_STR(m_Sarea) + ' ' + ANY_TO_STR(m_u_Sa) + '\n';

[PrjAirflowElementsImpl.hpp#L1569](https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/airflow/contam/PrjAirflowElementsImpl.hpp#L1569)

    PRJFLOAT m_Sarea;  // shut-off orifice area [m^2] {R4} {W}

[PrjDefines.hpp#L32](https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/airflow/contam/PrjDefines.hpp#L32): 

    #define PRJFLOAT QString

jmarrec:
In utilities/core/String class, you added:

```
  /** Int to std::string at full precision. */ 
  UTILITIES_API std::string toString(int v); 
 
  /** Unsigned to std::string at full precision. */ 
  UTILITIES_API std::string toString(unsigned v); 
```

This is causing the ambiguous call, if I comment it out, it builds again.

jmarrec:
I'm not sure I'm a big fan of moving the check out of setDouble and into the validityReport (not sure what that is).

```
include OpenStudio::Model
m = Model.new
ld = LightsDefinition.new(m)
ld.setLightingLevel(0/0.0)

[BOOST_ASSERT] <2> Assertion result failed on line 198 of bool openstudio::model::detail::LightsDefinition_Impl::setLightingLevel(boost::optional<double>) in file /home/julien/Software/Others/OpenStudio2/openstudiocore/src/model/LightsDefinition.cpp.
[BOOST_ASSERT] <2> Assertion result failed on line 198 of bool openstudio::model::detail::LightsDefinition_Impl::setLightingLevel(boost::optional<double>) in file /home/julien/Software/Others/OpenStudio2/openstudiocore/src/model/LightsDefinition.cpp.
=> true
```
I get no warning whatsoever about what I did wrong there (and if it was compiled in Release rather than Debug, I'd swear nothing happened)

Then again, after refactoring the setLightingLevel to actually check whether the call to setDouble actually worked before doing the rest, it's just going to tell me "false", but no warning message.

I'm a big fan of warnings messages for obvious reasons: anything that can help me quickly get why it didn't work is welcome.

openstudiocore/src/utilities/idf/Test/IdfObject_GTest.cpp
macumber: Crashes and seg faults don't provide as nice of output, so my standard is to use an ASSERT to prevent a crash or seg fault, but EXPECT's otherwise.  It is a matter a preference, i just like to see more output from a failed test if possible

macumber:
Agree about the warn or error level

openstudiocore/src/utilities/idf/IdfObject.cpp
```diff
@@ -1858,15 +1858,15 @@ void IdfObject_Impl::populateValidityReport(ValidityReport& report, bool checkNa
         }
       } else{
         if (std::isnan(*value)) {
-          LOG(Info, "Cannot set field " << index << ", '" << iddField.name() << "', an object of type "
+          LOG(Warn, "Cannot set field " << index << ", '" << iddField.name() << "', an object of type "
```
macumber: Agree with this

openstudiocore/src/model/ScheduleDay.cpp
macumber: This is better, I should be better about this

PR #2853, opened 2017-11-18
Fix 2720 clone plenum space type

### LINKED ISSUE #2720: Crash on clone plenum space type
Code to reproduce:

```
OpenStudio::Logger.instance.standardOutLogger.enable
OpenStudio::Logger.instance.standardOutLogger.setLogLevel(OpenStudio::Debug)

m = OpenStudio::Model::Model.new
s = OpenStudio::Model::Space.new(m)
z = OpenStudio::Model::ThermalZone.new(m)
s.setThermalZone(z)
a = OpenStudio::Model::AirLoopHVACSupplyPlenum.new(m)
a.setThermalZone(z)
st = s.spaceType.get
puts st
s.hardApplySpaceType(true)
```

jmarrec:
Fix #2720.


jmarrec:
space.hardApplySpaceType is trying to clone the plenumSpaceType, which is a unique model object.

For now I added a check in Space::hardApplySpaceType only.

Another location where it might make sense to do a check is SpaceType::clone (need to implement a virtual override in this case). Since I'm not sure it's really required, I haven't done that yet.

jmarrec:
Added a check in `SpaceType::clone` too. If it's the plenum space type that's trying to be cloned, it will Log (currently `Error`, but perhaps can be `Warn` instead), and returns `model.plenumSpaceType()` (that way it works with another target model too!)

jmarrec:
I'll make comments on your new commits on #2720 directly since the discussion appears to be there.

PR #2854, opened 2017-11-20
Fix 2022 air terminal magic

### LINKED ISSUE #2022: Air terminal magic not applied to all terminals
From Elizabeth Gillmor: I had always liked that when you set up one of these on an airloop and then added more zones to your loop, that the air terminals, their settings, and loop connections would be magically reproduced for all the subsequent terminals.  However it does not happen with the 2 or 4-pipe induction unit object or chilled beam object.  One has to go manually click on every object to attach these loops. (or spend the time to go write a measure which is not happening right now) That's not exclusive to 1.10 but it is an unpleasant surprise when you are expecting all of them to act the same and they don't.

jmarrec:
Fix #2022. I realized that the AirTerminal or Coils clone() method isn't called after following the Qt stuff.

Instead AirLoopHVAC::addBranchForZone is called, and in there it's going to rely on the coil 'containingHVACComponent' method. Some missing AirTerminalTypes were uncovered in CoilCoolingWater and CoilHeatingWater, so I implemented that. Also added a containingHVACComponent method to CoilCooledBeam though that was perhaps not necessary due to the need for a check (see next point)

I then realized that ATU:FourPipeInduction is special because it has both a cooling and heating coil.
ATU:CooledBeam is also special because the coilCoolingCooledBeam is a StraightComponent and not a WaterToAirComponent, so I add to implement a check for this in AirLoopHVAC::addBranchForZone.

Wrote Gtests for all terminal types to make sure "magic" is indeed happening.

More info on #2022.

### DIFF
--- a/openstudiocore/CMakeLists.txt
+++ b/openstudiocore/CMakeLists.txt
@@ -13,6 +13,7 @@ project(OpenStudio VERSION 2.3.0)
 
 include(ExternalProject)
 include(CPackComponent)
+include(CheckCXXCompilerFlag)
 
 if( POLICY CMP0022 )
   cmake_policy(SET CMP0022 NEW)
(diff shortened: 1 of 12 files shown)


nrel-bot-3:
@kbenne @jmarrec @lgentile it has been 7 days since this pull request was last updated.

PR #2860, opened 2017-11-21
Fix 2849 and 2850 daylighting control problems (sum of frac > 1.0, and illumanceSP not translated if defaulted)

### LINKED ISSUE #2850: App and Plugin allow invalid zone load fractions on daylighting controls
Like it says; e.g. A single zone can have a primary and secondary daylighting control zone load fraction total > 1.0.

jmarrec:
Fix #2850 , Gtest.

### DIFF
--- a/openstudiocore/src/model/ThermalZone.cpp
+++ b/openstudiocore/src/model/ThermalZone.cpp
@@ -541,12 +541,28 @@ namespace detail {
   }
 
   bool ThermalZone_Impl::setFractionofZoneControlledbyPrimaryDaylightingControl(double fractionofZoneControlledbyPrimaryDaylightingControl) {
+
+    double secondaryFrac = fractionofZoneControlledbySecondaryDaylightingControl();
+    if ( (fractionofZoneControlledbyPrimaryDaylightingControl + secondaryFrac) > 1.0) {
+      LOG(Error, "Fraction of Zone Controlled by Secondary Daylight Control is " << secondaryFrac
+          << " and you supplied a Primary Fraction of " << fractionofZoneControlledbyPrimaryDaylightingControl
+          << " which would result in a sum greater than 1.0");
+      return false;
+    }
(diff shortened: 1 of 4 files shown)


jmarrec:
Ready for review and merge

openstudiocore/src/model/ThermalZone.cpp
```diff
@@ -541,12 +541,28 @@ namespace detail {
   }
 
   bool ThermalZone_Impl::setFractionofZoneControlledbyPrimaryDaylightingControl(double fractionofZoneControlledbyPrimaryDaylightingControl) {
+
+    double secondaryFrac = fractionofZoneControlledbySecondaryDaylightingControl();
```
macumber: @jmarrec these fraction controlled fields are currently set in the inspector which bypasses this logic, so i think this check needs to be implemented in the forward translator as well.  My approach would be if the sum is > 1, set the secondary fraction to 1-primary fraction.  You could do that in the translator as well as in these set fraction controlled methods.  What do you think?  It is a little goofy having one setter change another field, but it also seems like you would want the logic to be consistent between the API and translator

openstudiocore/src/model/test/ThermalZone_GTest.cpp
```diff
@@ -222,12 +222,38 @@ TEST_F(ModelFixture,ThermalZone_FractionofZoneControlledbySecondaryDaylightingCo
...
+  ThermalZone z(m);
+
+  // As Fraction
+  ASSERT_TRUE(z.setFractionofZoneControlledbyPrimaryDaylightingControl(0.5));
+  ASSERT_TRUE(z.setFractionofZoneControlledbySecondaryDaylightingControl(0.5));
+  ASSERT_FALSE(z.setFractionofZoneControlledbySecondaryDaylightingControl(0.75));
+  ASSERT_FALSE(z.setFractionofZoneControlledbyPrimaryDaylightingControl(0.75));
+
+  // As Quantity
+  Unit units = z.getFractionofZoneControlledbySecondaryDaylightingControl(true).units(); // Get IP units.
```
macumber: @jmarrec quantities are deprecated just FYI

PR #2884, opened 2017-12-05
Fix #2883 failing EMS tests

### LINKED ISSUE #2883: Fix failing EMS tests
A few tests are failing due to the EnergyManagementSystemGlobalVariable.

Currently the Ctor is throwing when you supply a name with spaces, whereas the Gtests seem to indicate it should just rename the varaible name to replace the space with an underscore.

https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/model/EnergyManagementSystemGlobalVariable.cpp#L85

https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/utilities/idf/IdfObject.cpp#L433:L454

FYI @brianlball. Am I understanding the expected behavior correctly?

jmarrec:
Fixes #2883: Warn user if a varName with spaces is supplied but do allow it, it'll just replace the space with an underscore.

FYI @brianlball

### DIFF
--- a/openstudiocore/src/model/EnergyManagementSystemGlobalVariable.cpp
+++ b/openstudiocore/src/model/EnergyManagementSystemGlobalVariable.cpp
@@ -82,10 +82,16 @@ EnergyManagementSystemGlobalVariable::EnergyManagementSystemGlobalVariable(const
 {
   OS_ASSERT(getImpl<detail::EnergyManagementSystemGlobalVariable_Impl>());
   bool ok = getImpl<detail::EnergyManagementSystemGlobalVariable_Impl>()->setName(variableName);
-  if ((!ok) || (variableName != this->nameString())) {
+  if (!ok) {
     remove();
     LOG_AND_THROW("Unable to set " << briefDescription() << "'s Name to " << variableName << ".");
   }
+
+  if (variableName != this->nameString()) {
+    LOG(Warn, "Because spaces aren't allowed in ERL variable names, renamed " << briefDescription()
+           << " (requested variableName = '" << variableName << "').");
+  }
+
 }
 
 IddObjectType EnergyManagementSystemGlobalVariable::iddObjectType() {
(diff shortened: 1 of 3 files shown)


brianlball:
No.  Global Variables cannot have spaces or other special characters in their name.  The original code is the correct behavior.  Do not merge this. @kbenne @macumber

jmarrec:
That is not what I’ve done though.
Given your gtest it seems you expected “glob var” to be automatically renamed to “glob_var” which is what I implemented (with a warning message to let the user know)

Otherwise we can just forbid it completely and modify the gtest accordingly 

——
Sent from a mobile device, please excuse the brevity.
Julien Marrec, EBCP, BPI MFBA
Owner


Direct: +33 6 95 14 42 13
Website: www.effibem.com

LinkedIn (en) | (fr)

jmarrec:
This is where the naming convention (no space) is actually enforced by replacing all spaces with an underscore (was existing, I didn't add this): [IdfObject.cpp#L435](https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/utilities/idf/IdfObject.cpp#L435)

@brianlball which behavior do you want to enforce if a name is passed with a space?

1. Automatically rename it replacing all spaces with an underscore, and issue a warning to the user (see [here](https://github.com/NREL/OpenStudio/pull/2884/files#diff-53214e19001ab00c1c2b2640638021eeR90), which is clearly the original intent looking at the existing code + GTEST (see original Gtest [here](https://github.com/NREL/OpenStudio/pull/2884/files#diff-e0db5459839234c9e5350cb893ed3346L51))
2. Refuse to create the object altogether: LOG_AND_THROW

My two cents is that I agree with the original behavior defined in 1.

brianlball:
I addressed this in https://github.com/NREL/OpenStudio/pull/2890 with some other EMS enhancements.
The desired behavior is an error since global variables are used elsewhere in EMS programs and we wont be able to rename everything.  So throw an error and let the user rename their GV everywhere.

Thanks for finding this.

jmarrec:
Cool, thanks @brianlball, I'll close this one then.

PR #2889, opened 2017-12-06
Fix #2888 - Enter fractional fields when locale has a comma separator.

### LINKED ISSUE #2888: Cannot edit any of the Fractions for the loads in current develop in OS App
After some tedious investigation (and a bit of luck probably), I realized it only allows 0 or 1 to be entered, exactly. If you try to put 1.0 or 0.0 it doesn't work. 

The IDD prevents you from inputting something higher than 1 or lower than 0, but it doesn't allow you to input a decimal separator in there for some reason. 

The underlying API calls work perfectly in ruby/c++.

This affects all the OSQuantityEdit fields (and OSDoubleEdit too) (eg for Luminaire Definition, you cannot set Lighting Power to 50.5 either, but 50 or 51 is ok)

@evanweaver any ideas?

jmarrec:
Set the QDoubleValidator local for Double and Quantity fields to accept decimal number with a "." separator.

Even my my LC_NUMERIC=ft_FR.UTF-8 I can now edit double/Quantity fields with a "." in there, eg 0.5 is accepted.

### DIFF
--- a/openstudiocore/src/shared_gui_components/OSDoubleEdit.cpp
+++ b/openstudiocore/src/shared_gui_components/OSDoubleEdit.cpp
@@ -36,6 +36,7 @@
 
 #include <QDoubleValidator>
 #include <QFocusEvent>
+#include <QLocale>
 
 #include <iomanip>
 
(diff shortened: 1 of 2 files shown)


jmarrec:
I think there's a way to globally set the QLocale
https://stackoverflow.com/questions/31105419/how-to-set-global-locale

Issue #2787, opened 2017-09-18
removing airloop removes ptac from inlet port list


joseph-robertson:
```
air_loop.addBranchForZone(control_zone)
ptac.addToThermalZone(control_zone)
air_loop.remove
```
The `ptac` is no longer in the inlet port list for `control_zone`resulting in E+ error (osm [before](https://github.com/NREL/OpenStudio/files/1311180/before.osm.txt) and osm [after](https://github.com/NREL/OpenStudio/files/1311179/after.osm.txt)):

   ** Severe  ** GetPTUnit: ZoneHVAC:PackagedTerminalAirConditioner="RESIDENTIAL ROOM AC ZONE PTAC"
   **   ~~~   ** ..Air-conditioners air outlet node name must be the same as a zone inlet node name.
   **   ~~~   ** ..Zone inlet node name is specified in ZoneHVAC:EquipmentConnections object.
   **   ~~~   ** ..Air-conditioners outlet node name = NODE 25

@kbenne @shorowit 

Extra resources: 
[resstock.osm.txt](https://github.com/NREL/OpenStudio/files/1311165/resstock.osm.txt)
[measures.osw.txt](https://github.com/NREL/OpenStudio/files/1311164/measures.osw.txt)

jmarrec:
Very probably linked to https://github.com/NREL/OpenStudio/issues/2785

shorowit:
I don't think they are linked. This bug has existed across multiple versions whereas #2785 broke between OS 2.2.1 and 2.2.2.

kbenne:
I'm working on this. I think we will know soon.

shorowit:
This bug is still present in the 2.3 pre-release. Same error message as above.

shorowit:
How to more easily reproduce:
1. Run [resstock.osm.txt](https://github.com/NREL/OpenStudio/files/1403445/resstock.osm.txt) via OpenStudio 2.3. It runs successfully.
2. Run [resstock_without_airloop.osm.txt](https://github.com/NREL/OpenStudio/files/1403446/resstock_without_airloop.osm.txt) via OpenStudio 2.3. It fails with the above EnergyPlus error. The only change made was to load resstock.osm and call `model.getAirLoopHVACs[0].remove`.

If you look at the IDF created from resstock.osm, it has:
```
NodeList,
  living zone Inlet Node List,            !- Name
  Node 37,                                !- Node Name 1
  Node 40,                                !- Node Name 2
[...]

PR #2894, opened 2017-12-08
Potential fix for #1286 to retain zero when clicking hardsize

jmarrec:
See comments on #1286 to make sure this doesn't create unwanted problems.


jmarrec:
@kbenne you saw/tested the fix for #2787 too? I didn’t rename the PR but I did push two completely different things, just FYI to make sure you saw that.

Issue #2885, opened 2017-12-05
Fix Failing FT tests for ScheduleInterval


jmarrec:
I'm not sure what is going on there because I'm don't know the intent/conventions as far as start dates for timeSeries versus ScheduleInterval, but there are fishy things happening here.

I tried to figure it out (see the referenced commit https://github.com/NREL/OpenStudio/commit/638fd35ae0a2b38c49be90906389c321d20a458d?w=1), but I need to defer to whoever implemented this first, I think it might have been @jasondegraw ?!

jmarrec:
@jasonsegraw where did you fix it?

PR #2901, opened 2017-12-13
Should fix #2900 I think

### LINKED ISSUE #2900: Crash when clicking on a pump in OS App
Using current develop, built in Debug Mode. This is probably related to the new pump fields added recently by @kbenne.

Happens everytime I click on a Pump:VariableSpeed for eg:

```
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.units.UnitFactory] <1> W*s)*Pa/(m^3 is not properly formatted.
[openstudio.Unit] <2> W*min)*ftH^20/(gal is not a UnitString, per openstudio::isUnit.
terminate called after throwing an instance of 'openstudio::Exception'
  what():  /home/julien/Software/Others/OpenStudio/openstudiocore/src/utilities/units/Unit.cpp@653 : W*min)*ftH^20/(gal is not a UnitString, per openstudio::isUnit.
Aborted (core dumped)
```

jmarrec:
ftH20 (zero) instead of ftH2O (capital letter) for the new pump fields

### DIFF
--- a/openstudiocore/src/model/ModelCore.i
+++ b/openstudiocore/src/model/ModelCore.i
@@ -168,7 +168,7 @@ namespace model {
     return os.str();
   }
 
-  // This really shouldn't be necessary
+  // This really should not be necessary
   IdfObject toIdfObject() const {
     return *self;
   }
--- a/openstudiocore/src/model/People.cpp
+++ b/openstudiocore/src/model/People.cpp
@@ -423,9 +423,9 @@ namespace detail {
   }
 
   boost::optional<double> People_Impl::spaceFloorAreaPerPerson() const {
-    OptionalDouble temp = peoplePerFloorArea();
+    OptionalDouble temp = peopleDefinition().spaceFloorAreaperPerson();
     if (temp) {
-      return 1.0 / temp.get();
+      return temp.get() * multiplier();
     }
     return temp;
   }
--- a/openstudiocore/src/model/People.cpp
+++ b/openstudiocore/src/model/People.cpp
@@ -425,7 +425,10 @@ namespace detail {
   boost::optional<double> People_Impl::spaceFloorAreaPerPerson() const {
     OptionalDouble temp = peopleDefinition().spaceFloorAreaperPerson();
     if (temp) {
-      return temp.get() * multiplier();
+      double mult = multiplier();
+      if (mult > 0) {
+        return temp.get() / mult;
+      }
     }
     return temp;
   }
--- a/openstudiocore/resources/model/OpenStudio.idd
+++ b/openstudiocore/resources/model/OpenStudio.idd
@@ -19806,8 +19806,8 @@ OS:Pump:ConstantSpeed,
        \required-field
        \type real
        \note Used to size Design Power Consumption from design flow rate for head and motor efficiency
-       \units W/((m3/s)-Pa)
-       \ip-units W/((gal/min)-ftH2O)
+       \units W-s/m3-Pa
+       \ip-units W-min/gal-ftH2O
        \minimum> 0
 
 OS:Pump:VariableSpeed,
@@ -19951,8 +19951,8 @@ OS:Pump:VariableSpeed,
        \required-field
         \type real
         \note Used to size Design Power Consumption from design flow rate for head and motor efficiency
-        \units W/((m3/s)-Pa)
-        \ip-units W/((gal/min)-ftH2O)
+        \units W-s/m3-Pa
+        \ip-units W-min/gal-ftH2O
         \minimum> 0
   N15;  \field Design Minimum Flow Rate Fraction
        \required-field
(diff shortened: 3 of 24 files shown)


jmarrec:
It didn't... I'm trying various things on the IDD before I need to resort to debugging the QuantityRegex.cpp > regexUnit() function. Rebuild time in between is long...

jmarrec:
Ok, I no longer get the crash reported on #2900 now that I have modified the IDD.

It still worries me a lot that the "1/ftH_{2}O" is getting parsed as "1/ftH_{2}O^2" though... not sure what to do here. I'll pick up in the AM.

jmarrec:
pasted image at 2017_12_14 06_16 pm

jmarrec:
Oops I committed my changes to QuantityRegex.i, you should discard that.

openstudiocore/src/model/ModelCore.i
```diff
@@ -168,7 +168,7 @@ namespace model {
     return os.str();
   }
 
-  // This really shouldn't be necessary
```
macumber: @jmarrec you the grammar police now? ;-)

macumber:
Thanks for cranking on this @jmarrec !  I'll run the tests and see if anything breaks before merging

openstudiocore/src/model/ModelCore.i
jmarrec: I replaced “shouldn’t” by “should not” because the ‘ was breaking syntax highlighting for the rest of the file on my vim editor :)

jmarrec:
I tested all openstudio_tests, all tests ok (aside from FT ScheduleInterval #2885), the EMS ones (fixed in #2903), and 2 core/path tests that always fail on Linux (see slack)

I launched OpenStudio-resources tests to make sure (it really shouldn't affect it...), and a few failed, but I don't think they have anything to do with this

jmarrec:
Added missing autosize-related methods to fix https://github.com/NREL/OpenStudio-resources/issues/22:

* new terminals: AirTerminal:DualDuct:VAV:OutdoorAir, AirTerminal:DualDuct:ConstantVolume, and,
* CoolingTower:SingleSpeed and TwoSpeed: added for design range temperature and design approach temperature: these aren't being reported to the SQL file, so instead it does what E+ does and goes to the attached loop's Sizing:Plant object (see [CondenserLoopTowers.cc](https://github.com/jmarrec/EnergyPlus/blob/develop/src/EnergyPlus/CondenserLoopTowers.cc#L2480) around this line)

openstudiocore/src/model/People.cpp
```diff
@@ -423,9 +423,9 @@ namespace detail {
   }
 
   boost::optional<double> People_Impl::spaceFloorAreaPerPerson() const {
-    OptionalDouble temp = peoplePerFloorArea();
+    OptionalDouble temp = peopleDefinition().spaceFloorAreaperPerson();
     if (temp) {
-      return 1.0 / temp.get();
+      return temp.get() * multiplier();
```
macumber: This should get divided by the load instance multiplier right?

macumber: @jmarrec if you have 100ft2/person and then you add a multiplier of 5, you want 100ft/(5person) right?  Or do you want 500ft2/person (larger multiplier gets fewer people).  Multipliers are kind of weird with this one

macumber: That might be why this was the way it was?

jmarrec: I agree with you. The space should be 5 times more crowded... We should check if the multiplier is strictly greater than 0 before dividing too.

jmarrec: The way it was is wrong though.

peopleDefintion.setSpaceFloorAreaperPerson(10) 
then from the linked people: 
people.spaceFloorAreaPerPerson => I will get a non initialized currently!

PR #2903, opened 2017-12-15
Fix failing ems tests

jmarrec:
* One test ForwardTranslatorWeatherActuator_EMS was failing because setting an actuator to a WeatherFile which doesn't have a name:  a tiny thing introduced in Brian's PR [here](https://github.com/NREL/OpenStudio/pull/2890/files#diff-d6cc26d9e8d081a74e006bd8ed735047R80). Just need to check whether the actuated component is "Site" or "Weather File" before you check if it has a name in ForwardTranslator

* Behavior at one point was to automatically rename var names with spaces to an underscore, now it throws in the constructor if a name with space is supplied, so need to adjust the FT tests accordingly (passing "glob var" is no longer accepted, so need to pass "glob_var" directly)

* The reverse translator tests were failing because they load the IDF file created by the previous ForwardTranslator test.

All `openstudio_model_tests` are passing, and all `openstudio_energyplus_tests` too except the ScheduleInterval ones (which I briefly tried to fix but need to pass the torch, see #2885)

### DIFF
--- a/openstudiocore/src/energyplus/Test/EMS_GTest.cpp
+++ b/openstudiocore/src/energyplus/Test/EMS_GTest.cpp
@@ -257,7 +257,7 @@ TEST_F(EnergyPlusFixture, ForwardTranslatorSensorRename_EMS) {
   // add sensor
   EnergyManagementSystemSensor sensor(model, "Cooling Coil Runtime Fraction");
   sensor.setKeyName(toString(dx_coil.handle()));
- 
+
   //change name
   dx_coil.setName("New Coil Name");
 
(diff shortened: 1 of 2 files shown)


openstudiocore/src/energyplus/Test/EMS_GTest.cpp
```diff
@@ -1058,16 +1059,16 @@ TEST_F(EnergyPlusFixture, ForwardTranslatorGlobalVariable_EMS) {
   Model model;
 
   // add global variable
-  EnergyManagementSystemGlobalVariable var(model, "glob var");
```
macumber: @jmarrec does `var(model, "glob var")` throw?  If so we should add an expect throw test

macumber: And I usually check that the number of objects did not increase when a ctor throws

brianlball: we do test that here but it could be expanded
https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/test/EnergyManagementSystemGlobalVariable_GTest.cpp#L55

PR #2935, opened 2018-01-15
Add stub for functions to identify ems actuators

kbenne:
Don't merge this yet, but take a look. My idea is that we don't need to do anything for Ems sensors because we can leverage the ::outputVariables api.


openstudiocore/src/model/FanConstantVolume.cpp
```diff
@@ -481,10 +483,15 @@ namespace detail {
     return types;
   }
 
-  std::string HVACComponent_Impl::emsActuatorComponentType() const {
+  std::string FanConstantVolume_Impl::emsActuatorComponentType() const {
     return std::string("FAN");
   }
 
+  std::vector<std::string> FanConstantVolume_Impl::emsInternalVariables() const {
+    std::vector<std::string> types{"Fan Maximum Mass Flow Rate", "Fan Nominal Pressure Rise", "Fan Nominal Total Efficiency"};
```
kbenne: Might consider using these fancy initializer lists instead of push_back. I think it is much faster because it doesn't require reallocation every time you push.

kbenne: oh I see. for output variables, you just followed the code that was there... sorry. no need to update those existing ones.

nrel-bot-3:
@brianlball @kbenne @lgentile it has been 7 days since this pull request was last updated.

openstudiocore/src/model/FanConstantVolume.cpp
jmarrec: Just randomly came accross this. I added most of the output variables for all objects, but I just followed the existing examples with push_back. If we want to replace with an initializer this could be scripted.
Note that currently the overhead wouldn't be too huge because it's static...
Also note that it should ideally be refactored  to not be static... because in some objects the output variables are dependent on the choices you made (eg, [WaterHeaterMixed](https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/model/WaterHeaterMixed.cpp#L117))

brianlball: @jmarrec are your output variables pushed to develop or in a branch still?

jmarrec:
All been pushed. I think i did this in the central heat pump PR (#2700), started implementing a long time ago but got merged in in 2.4.0 or 2.3.1.
There shouldn’t be any more object that has an empty _Impl::outputVariableNames unless there are no output variables (i should have left a comment saying it’s not applicable in this case). I may have missed a couple objects, so If you come across either implement or write down the object name please.


——
Sent from a mobile device, please excuse the brevity.
Julien Marrec, EBCP, BPI MFBA
Owner


Direct: +33 6 95 14 42 13
Website: www.effibem.com

LinkedIn (en) | (fr)

openstudiocore/src/model/FanConstantVolume.cpp
macumber: @jmarrec - @kbenne, @brianlball, and I had an extensive conversation about whether or not our output variables should be static or not.  In the end, the conclusion was that we should standardize on the outputVariableNames method being static and returning all possible output variable names.  If we want to make a dynamic method that could be achieved by creating a new method or by adding an argument to the outputVariableNames method.  I'll try to remember/summarize the logic in our decision but will probably forget some things:

- Making the call dynamic would imply a level of logic and support for output variables and EMS that we currently do not have (e.g. checking them for correctness at translation time), the current approach is to offer a super set to the user and all failures at E+ run time
- Keeping this method static allows us to quickly get all possible output variables without having to query all objects in the model which would be time consuming

jmarrec:
I think your decision is sensible, I agree it’s not worth the effort to make it perfect 

——
Sent from a mobile device, please excuse the brevity.
Julien Marrec, EBCP, BPI MFBA
Owner


Direct: +33 6 95 14 42 13
Website: www.effibem.com

LinkedIn (en) | (fr)

macumber:
@kbenne @brianlball is this going in for 2.4.2 which is the 2.5.0 release candidate?  if so it needs to get merged by tomorrow morning :-/

brianlball:
@macumber doing a merge now

Issue #2946, opened 2018-01-17
Allow rearranging HVAC equipment in Thermal Zones tab


macumber:
https://openstudio.uservoice.com/forums/308629-openstudio/suggestions/9099592-allow-rearranging-material-layers-in-construction

Problem: Currently there's not a quick way to rearrange ZoneHVAC equipment in the Thermal Zones tab without a measure. In order to switch the order of equipment the user must delete ZoneHVAC objects losing any user-specified inputs.

Rationale: Heating and cooling priority is set by the order of the ZoneHVAC objects assigned to a zone in the GUI. Not all users are aware of this and it can significantly affect model results.

Solution: Explicitly show the user heating and cooling priority. Add a Grid View feature to increase or decrease selected equipment priority.

Context: ZoneHVAC:EquipmentList

aaron-boranian:
Migrated from [UserVoice](http://openstudio.uservoice.com/forums/308629/suggestions/9926421)

Issue #1454, opened 2015-02-06
Allow rearranging material layers in construction GUI


macumber:

MatthewSteen:
Maybe overwrite by dropping a new material on top of old layer too.

jmarrec:
+1

MatthewSteen:
https://openstudio.uservoice.com/forums/308629-openstudio/suggestions/9099592-allow-rearranging-material-layers-in-construction

aaron-boranian:
Problem: It is difficult to manipulate/edit material layers in the construction GUI

Solution: Allow rearranging material layers in construction GUI

Context: ported from GitHub: https://github.com/NREL/OpenStudio/issues/1454

Migrated from [UserVoice](http://openstudio.uservoice.com/forums/308629/suggestions/9099592)

jmarrec:
#2964 would be a good reference if someone wants to implement this, since I similarly implemented reordering the AVMs within the AVM list on plant loop "Controls" subtab.

jmarrec:
I opened as PR to fix this at https://github.com/NREL/OpenStudioApplication/pull/47

PR #2964, opened 2018-01-19
Plant loop control

jmarrec:
Here's a little something courtesy of EffiBEM.

Changelog:

* Added AvailabilityManagers to the hvac_library and MainRightColumnController
* Implemented a new PlantLoop "Control": view
    * queries the ForwardTranslatePlantLoopEquipmentOperationSchemes helper functions (moved to a separate header file so they can be called) to tell the plant loop type, and list the component types (cooling, heating, setpoint, uncontrolled). This should give strong hints on what the resulting automatic plantloop equipment operation scheme will look like. 
   * Implemented a drop zone for availability managers
* Implemented a drop zone for Availability Managers for AirLoopHVAC controls view

The drop zone allows reordering the AVMs:

* If you drag and drop from the library, onto the "Drag from library box": it clones the AVM into the model and adds it at the end (already the current behavior)
* If you drag and drop from an existing AVM from the list onto the "Drag from library box": it clones the already existing AVM and adds the clone at the end
* If you drag and drop from the library, onto an existing AVM: it clones it and insert it at the current priority, shifting all existing AVMs down
* If you drag and drop from the list onto another avm: it moves the dragged AVM at the priority of the current avm, shifting all other AVMs down. You can just move AVMs around to reorder the list 

This behavior should be easily adaptable for Constructions and Thermal Zone equipment (see #2946 and #1454). @evanweaver 

Another feature is that if you now click on the Schedules on the AirLoopHVAC tab or any of the AVMs in the dropzone, they are pulled on the Main right column so you can see (and modify) the objects.
One caveat to this is that it'll show you alwaysOnDiscreteSchedule too, perhaps we should protect against editing it.

here's a short gif showing the new behavior:

smaller


jmarrec:
Ideally we would need actual icons for the AVMs now that they are in the OS app, for now I use a placeholder image (SPM:MixedAir). I have ideas about how they should look, but I don't really want to spend time on this.

nrel-bot-3:

evanweaver:
@jmarrec this is going stale and now has conflits. I suggest you choose an assignee to encourage one of us to review it :)

nrel-bot-3:

jmarrec:
@evanweaver I didn't know I had enough juice to assign reviewers myself =)

I actually started implementing a control view for the PlantLoop primary/heating/cooling Plant Equipment Operation Schemes to be able to look at objects there + their corresponding schedules. It isn't nearly finished and I have little time. 

So I guess we could just get this AvailabilityManager part in, and I might do the rest later.

Wanna be the reviewer? I am not one very worried about the merge conflict given it's only the hvac_library.osm... I can solve the conflict for you before you review if you want me to.

nrel-bot-3:

nrel-bot-3:

Issue #2479, opened 2017-02-20
can't remove material layers from ConstructionInternalSource


MatthewSteen:
Occurs in 1.14.0 and 2.0.2. 

To reproduce:

1. Create Internal Source Construction
2. Add material(s) to construction
3. Try to remove using the (X) on the material icon

asparke2:
Confirmed still an issue with OS App 2.1.0

evanweaver:
Weird. Add 3 or more materials and you can delete all but 2.

macumber:
There might be some API stuff going on, internal source means you have a source between layers, so there is a minimum of 2 layers needed.

macumber:
Although it is always possible the API is too strict

evanweaver:
Thanks @macumber. So, @asparke2 and @MatthewSteen, what is the desired behaviour? Right now, you can have 0, or 2 or more. If we reduce strictness, and allow only 1 then we would need to trap the situation where the user tries to move forward (i.e. change fields, change tabs) with only that one defined, perhaps by popping a dialog. Or, if there are only 2 left, and the user wants to delete 1, we could delete both remaining. Or, we can wait for Marj to return so we can determine a way forward :)

evanweaver:
@DavidGoldwasser

DavidGoldwasser:
My view is that any number of layers should be valid for the API and the GUI. You won't always have a model that is ready to run, but we don't stop anybody from creating a standard construction without any layer.

macumber:
I agree with @DavidGoldwasser, just let the object get made however you want and log errors at translation time.  I am not sure if the API will need to be relaxed to allow this.

evanweaver:
@macumber @asparke2 @MatthewSteen I agree with David. He pointed out that we currently allow zero layers, so it seems reasonable that we would also allow one layer.

evanweaver:
Perfect timing Dan. I'll take a look at Model's API and then send you a PR.

MatthewSteen:
Sounds good to me, thanks.

evanweaver:
Fixed via https://www.pivotaltracker.com/n/projects/134010/stories/154537565.

macumber:
@evanweaver pivotal is not public so probably not useful to link to pivotal from here

PR #2973, opened 2018-01-23
[#154537565] can't remove material layers from ConstructionInternalSo…

evanweaver:
…urce #2479

### DIFF
--- a/openstudiocore/src/model/ConstructionWithInternalSource.cpp
+++ b/openstudiocore/src/model/ConstructionWithInternalSource.cpp
@@ -85,7 +85,7 @@ namespace detail {
 
   bool ConstructionWithInternalSource_Impl::eraseLayer(unsigned layerIndex)
   {
-    if (this->numLayers() < 3){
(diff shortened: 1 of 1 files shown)


evanweaver:
@macumber this one is easy with only a single character changed.

macumber:
Ideally we would also issue a warning in forward translator if the source or temp calculations are requested at an index greater than the number of layers (e.g. if number of layers is 2, then source must be after index 1):

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateConstructionWithInternalSource.cpp#L83

https://bigladdersoftware.com/epx/docs/8-8/input-output-reference/group-surface-construction-elements.html#field-source-present-after-layer-number

This error would be a nicety, E+ will most likely fail anyway

nrel-bot-3:
@macumber @evanweaver @lgentile it has been 7 days since this pull request was last updated.

PR #2975, opened 2018-01-24
Relax the gtk linux dependency version

### LINKED ISSUE #2974: version of gtk dependency on linux is not available on ubuntu 16.04
I think we can be a little less specific about the version number and everything will be fine for 14.04 and 16.04.

kbenne:
close #2974


evanweaver:
@kbenne false alarm?

jmarrec:
Why was this deleted?

jmarrec:
Because it wasn't working! I just tested this on Ubuntu 16.04 and it didn't fix it. But I fixed it in #3133

PR #2976, opened 2018-01-26
Fix central heat pump system

### LINKED ISSUE #2916: CentralHeatPumpSystem forward translation problems
There is a variety of things going wrong (which I originally didn't test for), as pointed out by @chriswmackey in a comment on Unmethours [here](https://unmethours.com/question/14888/heat-recovery-chiller-in-openstudio/?comment=28796#post-id-28796). 

* The centralHP doesn't end up on a plant equipment operation scheme for the cooling and the heating loop
* The `ForwardTranslateLoop::populateBranch()` isn't doing the right thing since this object is on the supply side of **two** loops: the cooling loop supply inlet/outlet of the centralhp is being put on the heating loop
* I've noticed a bug in `WaterToWaterComponent::plantLoop()` (theoretically affects `secondaryPlantLoop()` too) due also to the fact that it is on the supply side of two loops. Occasionally it'll return the `tertiaryLoop()` instead (note that this methods is the only one that does do an extra check to ensure the outlet model object (node) is also on the right loop)

A simulation test (regression) should be added to NREL/OpenStudio-resources too.

jmarrec:
This fixes #2916. A regression test was added in https://github.com/NREL/OpenStudio-resources/pull/32.

As part of this pull request:

* Added a special case in the ForwardTranslatePlantEquipmentOperationScheme for the central heat pump system (it's on the supply side of TWO loops...) so that it gets picked correctly in heatingComponents(plantLoop) if the plant loop is the heating plant loop, and similarly for cooling plant loop
* Fixed ForwardTranslator::populateBranch so it does the right thing for this object too
* Added CentralHeatPumpSystem to the OS App:
    * Added to the hvac_library + icons
    * Override CentralHeatPumpSystem::addToNode: if already connected to two plant loops, and trying to add it to the supply side of another plant loop, there it calls addToTertiaryNode. This enables connection in the OS App too. See docstring in CentralHeatPumpSystem_Impl.hpp
* Enabled ability to jump from a tertiary loop to the appropriate plant loop in the OS App: From tertiary, go to either primary or secondary, whichever is on the other side (if tertiary = supply, go to the plant loop that is the demand side)
* Fixed a bug in WaterToWaterComponent::plantLoop that occassionally returns the tertiaryPlantLoop
* Do not translate the CentralHeatPumpSystem if there isn't any CentralHeatPumpSystemModule (and child ChillerHeaterPerformanceElectricEIR) attached to it (note: unable to set these in OS app, and the Ctor doesn't create one, E+ will crash if there isn't at least one ChillerHeaterPerformanceElectricEIR listed...)
* Extended Gtests.

Here's the functionality in action. **Notice how I connect first the cooling loop (supply side) and condenser loop (demand side)** (doesn't matter in which order between the two). **Only then can I connect to the supply side of the heating loop** which will connect as a tertiary loop. Notice how I can jump from the tertiary loop (supply) to the condenser loop (demand)

centralheatpumpsystem

Limitations: there is no way to edit the chiller heater module list (CentralHeatPumpSystemModule + ChillerHeaterPerformanceElectricEIR object) in the OS App. I think they should be children so they show up in the right column inspector... But this would be en enhancement.

----

Please note this comment : https://github.com/NREL/OpenStudio/pull/2976/files#diff-472d78440a0477f9873c08115127ca59R509

**In the future if you implement a WaterToWaterComopnent that has a tertiary loop, you should override the addToNode method to call the addToTertiaryNode when appropriate, so that it works in the OS App**

FYI @chriswmackey


jmarrec:
@chriswmackey

chriswmackey:
This is great @jmarrec !  I'm looking forward to testing this as soon as it's merged and a build is released.

nrel-bot-3:

jmarrec:
@kbenne could you be the reviewer for this please?

nrel-bot-3:

PR #2981, opened 2018-01-29
Avoids crash, not sure that translation is correct

### LINKED ISSUE #2978: Open Studio crashes when importing DOE IDF files
I downloaded Miami model file from DOE website https://www.energycodes.gov/sites/default/files/documents/ASHRAE901_OfficeMedium_STD2013.zip

Next I used IDF Version Updater to update the file to v8.8. When I opened the updated file with IDF Editor, and did a Validity Check I got prompted for an error. The error was the ULC object, the object field was highlighted, so I choose UpperLeftCorner to fix the error. Every time I try to import this IDF file, Open Studio crashes. It happens with other files for other US cities under the above link. Please help.

macumber:
Fixes #2978


brianlball:
this fix seems to work.  the ems in the idf -> RT -> FT is the same

Issue #2099, opened 2016-02-25
TimeStep of 6 per hour in new model isn't respected when simulation is run


DavidGoldwasser:
In a new model there is no time step object in the OSM, but the GUI shows 6 time steps per hour. If you run a simulation you see this error.

   ** Warning ** No Timestep object found.  Number of TimeSteps in Hour defaulted to 4.

Once you go in the GUI and change time step value away from 6 and then back to it, then the timestep is written out and the error goes away.

OS:Timestep,
  {29163c90-40ba-49d0-9dbf-7f439aff1cb7}, !- Handle
  6;                                      !- Number of Timesteps per Hour

@macumber this sounds familiar, but I don't see it filed anywhere, May have been related to other simulation parameters. When I changed the time step it wasn't just that object that got added to the OSM, but also all other parameters. Why don't we always just have the objects in the model?

macumber:
If the GUI is showing a default we might not be translating the default, this is similar to what is reported in this ticket https://github.com/NREL/OpenStudio/issues/2100

macumber:
Hmm, not looks like it should get translated:

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateTimestep.cpp#L36

macumber:
Sounds like the GUI needs to create a TimeStamp object if it does not exist

macumber:
The following objects are not created until you go to the simulation settings tab:

```
OS:RadianceParameters,
OS:SimulationControl,
OS:Sizing:Parameters,
OS:ProgramControl,
OS:Timestep,
OS:OutputControl:ReportingTolerances,
OS:ConvergenceLimits,
OS:ShadowCalculation,
OS:SurfaceConvectionAlgorithm:Inside,
OS:SurfaceConvectionAlgorithm:Outside,
OS:HeatBalanceAlgorithm,
OS:ZoneAirHeatBalanceAlgorithm,
OS:ZoneAirContaminantBalance,
OS:ZoneCapacitanceMultiplier:ResearchSpecial,
OS:RunPeriod,
```

I am thinking that many of these need to be created at Model::new and Model::load time.  Thoughts?

@kbenne @mbadams5 @asparke2 @DavidGoldwasser

macumber:
[...]

PR #2990, opened 2018-02-05
Fix #2989 to force initialize timestep

### LINKED ISSUE #2989: Ensure consistency between default Timestep object and E+
There's something fishy between the model API + GUI and the FT. 

By default, if you don't modify an actual Timestep object in your model, the simulation run tab will show you 6 timesteps per hour. That is, if you do not click/modify in the GUI or call `model.getTimestep` in the API.

Turns out that since you don't have a timestep, it's not forward translated, and as a result the IDF throws this error: `   ** Warning ** No Timestep object found.  Number of TimeSteps in Hour defaulted to 4.`

So the default is 6 timestep, yet the current behavior is that E+ ends up with 4 timesteps.

jmarrec:
Fixes #2989

### DIFF
--- a/openstudiocore/src/energyplus/ForwardTranslator.cpp
+++ b/openstudiocore/src/energyplus/ForwardTranslator.cpp
@@ -179,6 +179,10 @@ Workspace ForwardTranslator::translateModelPrivate( model::Model & model, bool f
   model::Version version = model.getUniqueModelObject<model::Version>();
   translateAndMapModelObject(version);
 
+  // translate Timestep second (this initializes it if need be)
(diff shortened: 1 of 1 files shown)


jmarrec:
@DavidGoldwasser informed me of #2099.

the "minimal fix" by @macumber https://github.com/NREL/OpenStudio/commit/a7b9a396c1f65e219b6f7985989dbf8e2ad0f3b8
 appears to be  triggered in OS App.

In my case, I get the warning in the OpenStudio-resources... the model is created and then run via the CLI, so the OS App doesn't enter the mix.

PR #2999, opened 2018-02-14
Storage Converter isn't translated for FacilityDemandLeveling while required

jmarrec:
I think it was just a mistake on my part, I forgot to uncomment this section, which I think I wrote before I actually implement the StorageConv object. Found this while writing a regression test for storage objects.

Also found a random copy/paste error in the IDD. Note that the copy/paste error has no actual effect aside that it isn't pretty when you look at it, because you need to use addLoadRange anyways (and not setWetBulbTemperature.....etc)

### DIFF
--- a/openstudiocore/resources/model/OpenStudio.idd
+++ b/openstudiocore/resources/model/OpenStudio.idd
@@ -19836,14 +19836,14 @@ OS:PlantEquipmentOperation:OutdoorWetBulb,
    A2 ,\field Name
         \required-field
         \reference ControlSchemeList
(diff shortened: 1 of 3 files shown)


jmarrec:
I modified the code to always translate the storage converter (as long as you have a storage object and a buss type compatible with a storage object), and only check and throw an Error if there isn't a storage converter when one is mandatory. I think that's what E+ does:

https://github.com/jmarrec/EnergyPlus/blob/develop/src/EnergyPlus/ElectricPowerServiceManager.cc#L766:L775

PR #3002, opened 2018-02-18
Feature/add ca update standards gem

asparke2:
Updates the openstudio-standards gem version and replaces the existing libraries with libraries generated from the gem.


asparke2:
@macumber next week if you show me where to upload the .tar.gz of the gems I will do it in the future.  I changed the MD5 to 'TODO' so we wouldn't forget to rebuild the gems.

PR #3021, opened 2018-03-06
Update link to openstudio cli

### LINKED ISSUE #2428: OpenStudio 2.0.0 path installation on mac
System info:

* Sketchup 2016
* macOS Sierra 10.12.2
* OpenStudio 2.0.0 (47ba3cbe19)

----

I’m installing OpenStudio 2.0.0 on mac

I noticed that it’s trying to install to `/usr/local/openstudio-2.0.0`. Before it went to to `/Applications/openstudio-x-x`.

There’s a cli so I guess it makes sense that one goes to `/usr/local`… but not sure if putting all of it to`/usr/loca/l`makes that much sense. Might be offputing to users who aren’t familiar with Unix systems too.

Another problem is that the `OpenStudioApp.app` and `ParametricAnalysisTool.app`aren't at all simlinked to `/Applications` so I can’t easily find the apps (I have to launch them via a terminal or the Finder > Go To Folder).

I would suggest installing everything to `/Applications` and add the cli to the path or to simlink it to /usr/local instead

kbenne:
On Mac, if the /usr/bin/openstudio link already exists,
perhaps from a previous version, then the attempt to
make a new link will fail.

Remove the old stale link before making the new link.

close #2428

### DIFF
--- a/openstudiocore/src/install_utility/main.cpp
+++ b/openstudiocore/src/install_utility/main.cpp
@@ -77,6 +77,9 @@ int main(int argc, char *argv[])
           if( ! openstudio::filesystem::exists(dir) ) {
             openstudio::filesystem::create_directories(dir);
           }
+          try {
+            openstudio::filesystem::remove("/usr/local/bin/openstudio");
+          } catch(...) {}
           fs::create_symlink(cliPath.string(),"/usr/local/bin/openstudio");
         }
       } catch(...) {}


DavidGoldwasser:
@kbenne is the best way for me to test this to install a 2.0.0 installer (or something similar to that) then 2.4.3. Then do I need to try to use the CLI from command line?

I have this code in my bash profile currently? Should I remove that prior to running the installers?
```
# lets command line findd cli
export PATH="/Applications/OpenStudio-2.4.1/bin":$PATH
export PATH="/Applications/OpenStudio-2.4.1/EnergyPlus":$PATH
```

Issue #3023, opened 2018-03-06
2.4.2 & 2.4.3 OS App crashes when going to HVAC tab


asparke2:
1. Open the App  (empty model)
2. Click on the HVAC tab
3. Crash!

Tested on Windows 7 with OS 2.4.2 and 2.4.3.

asparke2:
@kbenne not sure if this is you or @evanweaver

DavidGoldwasser:
I saw this happen once on an empty model, but then could not repeat it. I have also been able to go to HVAC tab and navigate loops on a non-empty model.

brianlball:
This is a real thing.  I reported it to Julien and Kyle the other day.  It has to do with the new availability manager stuff. @jmarrec @kbenne

kbenne:
Yeah. @brianlball brought this to my attention and I am looking into it. I think it is connected to

snlavu:
Has there been a fix to this? I just installed 2.4.3 and opened a version 2.2 file. I am unable to navigate to HVAC. See attached.
[TS-VRF-aircooled.zip](https://github.com/NREL/OpenStudio/files/1794728/TS-VRF-aircooled.zip)

evanweaver:
@snlavu this issue was reported by a user of 2.4.3. A fix should be available with 2.4.4 in another week, or so.

klamoureux:
@evanweaver , any word on the release of 2.4.4? I am running into this issue as well. Maybe there is a way to revert a model back to 2.4.0 in order to continue working. Any ideas or advice? SketchUp 2017 will not open OS 2.4.3 .osm file once 2.4.0 is installed.

evanweaver:
@klamoureux, it's imminent. I anticipate tomorrow or Friday.

asparke2:
@kbenne been building and testing develop today (3/23) and still having this crash happen, can't get into HVAC tab at all.

jmarrec:
I just built develop and I can get to the HVAC tab on Ubuntu. But when I launch OS App, it cannot load the hvac_library.osm etc

![image](https://user-images.githubusercontent.com/5479063/37894539-d49dcaf6-30de-11e8-89e2-e40d20e93759.png)

kbenne:
@jmarrec. Different issue. That is related to not finding E+.

jmarrec:
ok, but anyway, I can get to the HVAC tab on ubuntu with a debug build

jmarrec:
[...]

PR #3031, opened 2018-03-15
Four pipe beam

jmarrec:
I wrapped the object `AirTerminal:SingleDuct:ConstantVolume:FourPipeBeam` in the SDK.

* I implemented too sublcasses for the cooling/heating side: `CoilCoolingFourPipeBeam` and `CoilHeatingFourPipeBeam`
* ForwardTranslator has a couple of sanity checks:
    * must be connected to an AirLoopHVAC and have at least one coil subclass that is connected to a plantLoop
    * If you have a coil (heating or cooling) and this coil is connected to a plantloop, then all three curves are required objects so I check for that too.
* Added model + forward translator Gtest
* Added icons, added to hvac_library.osm, and code added to be able to use it in OS App.
* Regression test in OpenStudio-resources was added in this commit https://github.com/jmarrec/OpenStudio-resources/commit/a1ffa10f9ae0696c21dd71ba2b0810651a541383

The only thing I didn't fully test for is the OS App behavior. I merged latest develop in, in particular so I wouldn't have the crash on hvac tab #3023, but develop is broken so it won't build.
I have doubt about whether I need to override the `remove`method. I'll have to check that once develop is fixed.

Other unplanned/side changes:
* I made all port (eg `inletPort()` methods `const`, hence the high number of files I touched.

Review assignee should probably be @kbenne, time permitting.


jmarrec:
Tested in OS App, and added the actual SQL autosizedXXXX statements.

Ready for review @kbenne @macumber

macumber:
@jmarrec @kbenne says that there are now two separate prs, one that addresses const issues and one that adds this object.  Based on his review we can merge the separate four pipe beam code but not yet the const stuff. Closing this pr, have the other prs already been opened?  The four pipe object does not seem to have been merged yet

kbenne:
well. there are at least two different branches somewhere.

jmarrec:
For future ref, I did split it in two:
* Four Pipe Beam only: https://github.com/jmarrec/OpenStudio/tree/FourPipeBeam_Only, which was already gotten by @macumber in #3136 (and #3137 for CI)
* Making the port (inletPort/outletPort, etc) functions `const`: https://github.com/jmarrec/OpenStudio/tree/PortConst_Only

PR #3072, opened 2018-03-24
Fixes #3071 by not resetting properties that are not editable via the…

### LINKED ISSUE #3071: SpaceType properties lost in merge from Floorspace
1. Open a new model
2. Import a SpaceType from one of the CA DEER libraries
3. Look at the SpaceType on the Space Types tab
  - It should have a default schedule set, standards building type, and standards space type assigned
4. Go into Floorspace and create a new geometry (doesn't need to have anything inside)
5. Merge the geometry
6. Go back to the Space Types tab
  - The default schedule set, standards building type, and standards space type are now missing

@macumber suspects that the issue is in [this code](https://github.com/NREL/OpenStudio/blob/3549a2c2955007d05e8aa55371fcf5ce1e6b2a27/openstudiocore/src/model/ModelMerger.cpp).  I suspect that for properties not editable in Floorspace, calling `resetFoo()` for those properties is incorrect.

asparke2:
… Floorspace interface

### DIFF
--- a/openstudiocore/src/model/ModelMerger.cpp
+++ b/openstudiocore/src/model/ModelMerger.cpp
@@ -390,7 +390,10 @@ namespace openstudio
 
       // multiplier
       if (newThermalZone.isMultiplierDefaulted()){
-        currentThermalZone.resetMultiplier();
+        // DLM: TODO: add option as this might have been intentionally reset on other model
+        //currentThermalZone.resetMultiplier();
+      } else{
+        currentThermalZone.setMultiplier(newThermalZone.multiplier());
       }
 
       // ceilingHeight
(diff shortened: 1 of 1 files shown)


asparke2:
@macumber I removed the resets for any property that cannot be set using Floorspace.  It fixed the issue I was seeing with SpaceTypes and does not appear to have had any adverse effects in my testing.  Can you please review and see if you can think of any unwanted side effects?

macumber:
Your fix was correct @asparke2, I just wanted to preserve the old code so we can add more options for merging later

macumber:
@asparke2 can you try this out now.  if it still looks good to you I suggest you merge it and make sure @evanweaver gets it in the official packages

PR #3073, opened 2018-03-25
Updates openstudio-standards to final release 0.2.0

asparke2:


asparke2:
@macumber @evanweaver I made the final version of openstudio-standards that has all the bug fixes to be in 2.5.0.  Gem packages are on S3 and I built and tested locally with those packages.

PR #3085, opened 2018-04-03
Fix #3084 and NREL/OpenStudio-resources#43

### LINKED ISSUE #3084: Broken version translator for 2.4.3 models in current develop
This was introduced in e3c284b8e6c213842ba9efdf80d712cbc49fa205 [here](https://github.com/NREL/OpenStudio/commit/e3c284b8e6c213842ba9efdf80d712cbc49fa205#diff-3201e82200ae4bf36cd86ea3ba9adf16L261)

2.4.3 was removed from `m_startVersions`. If one was to be removed, it should be 2.4.2, but I'm not sure whether we want to remove it or not @evanweaver 

2.4.2 is listed on the releases pages but only has source and no binaries, and it's not in the [Compatibility Matrix](https://github.com/NREL/OpenStudio/wiki/OpenStudio-Version-Compatibility-Matrix)

Anyways, just re-ading a line with 

```
  m_startVersions.push_back(VersionString("2.4.3")); 
```

make 2.4.3 models load again

jmarrec:
Fix #3084 and NREL/OpenStudio-resources#43 (only the `afn_single_zone_nv.osm` portion of that issue)

Changelog:

* 2.4.3 was removed from `m_startVersions` -> add it back. 
* Handle 2.4.3 to 2.5.0 version translation for `OS:AirflowNetworkZone` which has a newly inserted `Name` field right after the handle. Shift all fields by 1, and for the 'Name', get the attached thermal zone name and name it like `"Airflow Network Zone " + _zone->nameString());`

Could you review @evanweaver please?

----

FYI this new field was added in 18387eb0be4ff41a5d5e0ab3d28060ba817d80cb by @jasondegraw (exactly [here](https://github.com/NREL/OpenStudio/commit/18387eb0be4ff41a5d5e0ab3d28060ba817d80cb#diff-20dc466cd26aed8488ee9b0bb0d5ce9aR28947))

----

Note that the official 2.5.0 build doesn't have this change and therefore will fail to load up a 2.4.3 osm with an `OS:AirflowNetworkZone` object in it. You can find an offending 2.4.3 OSM in the OpenStudio resources repo: https://github.com/NREL/OpenStudio-resources/blob/develop/model/simulationtests/afn_single_zone_nv.osm

### DIFF
--- a/openstudiocore/src/osversion/VersionTranslator.cpp
+++ b/openstudiocore/src/osversion/VersionTranslator.cpp
@@ -121,6 +121,7 @@ VersionTranslator::VersionTranslator()
   m_updateMethods[VersionString("2.1.2")] = &VersionTranslator::update_2_1_1_to_2_1_2;
   m_updateMethods[VersionString("2.3.1")] = &VersionTranslator::update_2_3_0_to_2_3_1;
   m_updateMethods[VersionString("2.4.2")] = &VersionTranslator::update_2_4_1_to_2_4_2;
+  m_updateMethods[VersionString("2.5.0")] = &VersionTranslator::update_2_4_3_to_2_5_0;
   m_updateMethods[VersionString("2.5.1")] = &VersionTranslator::defaultUpdate;
 
   // List of previous versions that may be updated to this one.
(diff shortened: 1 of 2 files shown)


macumber:
FYI @jasondegraw

PR #3086, opened 2018-04-03
Design conditions in OS:EpwFile

joseph-robertson:


macumber:
@jasondegraw FYI

@joseph-robertson and @shorowit what are you planning to do with these methods?  Simply avoid having to read the DDY file?  Is data in the DDY file duplicated in the EPW file?

shorowit:
Yes, we'd like to avoid having to read the DDY (so that our users don't have to supply it alongside the EPW). I don't know if all the data in the DYY is duplicated in the EPW header, but it has everything we need.

macumber:
And you will be making some method to take these values and create SizingPeriod:WeatherFileDays or SizingPeriod:WeatherFileConditionType objects with them?  I have not used these objects very much, I don't know if they are even wrapped.  

Or is there enough info for you to make SizingPeriod:DesignDay objects directly?  Should we add helper methods to the Model API or are you planning to just do this in a measure?

shorowit:
No, we use the values in our own Manual J sizing calculations to hardsize coil objects, etc. We're not using E+ autosizing or sizing objects.

macumber:
Is this done @joseph-robertson ?

joseph-robertson:
Yes @macumber. But I've only functionally tested using one epw file (vs all of them per your suggestion).

openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -748,6 +982,223 @@ namespace openstudio{
...
+  {
+    EpwDesignField id;
+    try {
+      id = EpwDesignField(name);
+    }
+    catch (...) {
+      // Could do a warning message here
+      return boost::none;
+    }
+    return boost::optional<std::string>(getUnits(id));
```
macumber: You can just return getUnits(id).

openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -893,6 +1344,219 @@ namespace openstudio{
     return boost::none;
   }
 
+  boost::optional<double> EpwDesignCondition::getFieldByName(const std::string &name)
```
macumber: Should this be getFieldValueByName?  getFieldByName seems like it would be returning EpwDesignField(name);

openstudiocore/src/utilities/filetypes/EpwFile.hpp
```diff
@@ -404,6 +475,375 @@ class UTILITIES_API EpwDataPoint
...
+  // Data retrieval
+  /** Returns the double value of the named field if possible */
+  boost::optional<double> getFieldByName(const std::string &name);
+  /** Returns the dobule value of the field specified by enumeration value */
+  boost::optional<double> getField(EpwDesignField id);
+  // Conversion
+  /** Create an EpwDesignCondition from an EPW-formatted string */
+  static boost::optional<EpwDesignCondition> fromDesignConditionsString(const std::string &line);
+  /** Create an EpwDesignCondition from a list of EPW designs as strings */
+  static boost::optional<EpwDesignCondition> fromDesignConditionsStrings(const std::vector<std::string> &list);
```
macumber: Need a toDesignConditionsString() method?

openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -2645,6 +4477,48 @@ namespace openstudio{
...
+      return true;
+    }
+
+    int nDesignConditions = std::stoi(split[1]);
+
+    if (split.size() < 70) {
+      LOG(Error, "Expected at least 70 design condition fields rather than the " << split.size() << " fields in the EPW file '" << m_path << "'");
+      return false;
+    }
+    else if (split.size() > 70) {
```
macumber: Shouldn't you check that nDesignConditions  and split.size agree?

macumber: Yes you should check them otherwise you'll get out of index errors below

openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -2202,204 +2866,1371 @@ namespace openstudio{
...
+    }
+    if (values.size()) {
+      DateTime start = dates[1] - Time(0, 0, 0, 3600.0 / m_recordsPerHour);
+      dates[0] = start; // Overwrite the placeholder
+      return boost::optional<TimeSeries>(TimeSeries(dates, openstudio::createVector(values), units));
+    }
+    return boost::none;
+  }
+
+  bool EpwFile::translateToWth(openstudio::path path, std::string description)
```
macumber: What is this?

openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -2202,204 +2866,1371 @@ namespace openstudio{
...
+      return false;
+    }
+
+    openstudio::filesystem::ofstream fp(path, std::ios_base::binary);
+    if(!fp.is_open()) {
+      LOG(Error, "Failed to open file '" + openstudio::toString(path) + "'");
+      return false;
+    }
+
+    fp << "WeatherFile ContamW 2.0\n";
```
macumber: Is this from @joseph-robertson or @jasondegraw ?

openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -2202,204 +2866,1371 @@ namespace openstudio{
...
+      }
+    }
+
+    // Cheat to get data at the start time - this will need to change
+    openstudio::EpwDataPoint lastPt = data()[data().size()-1];
+    std::vector<std::string> epwstrings = lastPt.toEpwStrings();
+    openstudio::DateTime dateTime = data()[0].dateTime();
+    openstudio::Time dt = timeStep();
+    dateTime -= dt;
+    epwstrings[0] = std::to_string(dateTime.date().year());
```
macumber: Need to check size before writing to index

openstudiocore/src/utilities/filetypes/EpwFile.cpp
jasondegraw: That's from me... some years ago.

openstudiocore/src/utilities/filetypes/EpwFile.cpp
joseph-robertson: @macumber I'm confused why this wouldn't be an optional. Why doesn't EpwDataPoint::getUnitsByName(...) return an optional then too?

openstudiocore/src/utilities/filetypes/EpwFile.cpp
joseph-robertson: Again, mimicking what EpwDataPoint does.

openstudiocore/src/utilities/filetypes/EpwFile.hpp
joseph-robertson: No

openstudiocore/src/utilities/filetypes/EpwFile.cpp
macumber: The result is an optional, you just don't need to explicitly construct one like this.  If you return a string, the compiler can implicitly convert that to an optional string.

openstudiocore/src/utilities/filetypes/EpwFile.cpp
macumber: Ok, to me getFieldByName would return a "field" which is basically an index.  This is returning a "field value" so would be better as getFieldValueByName.  If that is the naming convention in place we can stick with it, just trying to give you my feedback.

openstudiocore/src/utilities/filetypes/EpwFile.cpp
joseph-robertson: Will implement a check for making sure that nDesignConditions agrees with the expected length of the split array.

openstudiocore/src/utilities/filetypes/EpwFile.cpp
joseph-robertson: Not me.

PR #3089, opened 2018-04-04
Fix missing condenser type

pepsi7959:


macumber:
Thank you for this PR.  The full PR is quite large and seems to be bigger in scope than the title "Fix missing condenser type" suggests.  Did you mean to suggest merging all of this changes or just a subset related to the condenser types?

macumber:
Closing this PR because the scope is too large.  @pepsi7959 if you can make a PR with just the "Fix missing condenser type" commit we can merge that.

PR #3094, opened 2018-04-10
Fix #2874 - .deb package install to /usr/local/openstudio-x.y.z instead of /usr/

jmarrec:
Fix #2874 (a very annoying long term issue for me).

I built CPACK_BINARY_DEB and checked that .deb installs correctly to `/usr/local/opensudio-x.y.z`

**Still left to do:**

- [x] Name package `openstudio-x.y.z` instead of `openstudio` to allow several versions to be installed on the same system (currently apt/dpkg will just overwrite completely the previous version)

- [x] Add symlink of CLI to `/usr/local/bin/`

PR #3113, opened 2018-04-25
Fix 2874 install path unix

macumber:
Mirror of https://github.com/NREL/OpenStudio/pull/3094 for CI


nllong:
Looks good. I've updated the docker build file to support this as well: https://github.com/NREL/docker-openstudio/pull/13

image
image

PR #3139, opened 2018-05-22
Make port functions const

jmarrec:
This PR makes all the `port`-related functions (`inletPort`, `outletPort`, etc) `const`.

This was done originally as part of the FourPipeBeam implementation since I was getting an error due to non-const, see comment: https://github.com/NREL/OpenStudio/pull/3031#issuecomment-390930380

**Note**: This was moved into its own PR because this creates problems on macOS, the OS App refuses to launch. It does work on Ubuntu 16.04 though.


jmarrec:
@macumber @kbenne I updated this to current develop and fixed a bunch of other constness issue.

Then I built in Debug mode on Ubuntu (worked as expected), but also macOS High Sierra and OSApp launched just fine.

Can one of you two review please?

PR #3178, opened 2018-07-06
Relax limits in EpwFile

shorowit:
Currently EpwFile will not set/get values in some TMY3 EPWs for fields like WindSpeed when the value for a given hour is outside some arbitrary (?) limits. Note that these limits appear to have been set by the EnergyPlus team. In EnergyPlus, the limits are only tested against for the first hour of the year, at which point they would cause an error, and not every hour of the year. This is why the TMY3 EPWs can run successfully in EnergyPlus despite values outside the limits for many hours.

This PR changes the OS behavior so that the value can be set/get even when the value is outside the limits, but gives a warning under these conditions.

TODO:
- [x] Test locally
- [x] Add unit test


shorowit:
This is ready to go from my perspective. The unit test fails prior to these changes.

PR #3186, opened 2018-07-16
Remove large block of if statements in Model_Impl::createObject

kbenne:
Previous REGISTER_CONSTRUCTOR and REGISTER_COPYCONSTRUCTORS macros generated
a very large block of if statements, and for unexplained reasons this lead to
access violations at random points in the createObject process. (on clang)

In place of the if block is a map of IddObjectType to a factory function, which is
statically allocated when the program begins.

Please read [OpenStudio Pull Requests](http://https://github.com/NREL/OpenStudio/wiki/OpenStudio-Pull-Requests) to better understand the OpenStrudio Pull Request protocol.


openstudiocore/src/model/Model_Impl.hpp
```diff
@@ -312,6 +312,29 @@ namespace detail {
     void clearCachedYearDescription(const Handle& handle);
     void clearCachedWeatherFile(const Handle& handle);
 
+    typedef std::function<std::shared_ptr<openstudio::detail::WorkspaceObject_Impl>(Model_Impl *, const std::shared_ptr<openstudio::detail::WorkspaceObject_Impl>&, bool)> CopyConstructorFunction;
+    typedef std::map<IddObjectType, CopyConstructorFunction> CopyConstructorMap;
```
mbadams5: I would consider using `unordered_map`. Since it'll have faster lookup, especially if you use an `int` key.

openstudiocore/src/model/Model.cpp
```diff
@@ -3516,5 +2561,987 @@ void Model::applySizingValues() {
...
+  if ( it != m_copyMap.end() ) {
+    result = it->second(model, obj, keepHandle);
+  }
+  return result;
+}
+
+detail::Model_Impl::ModelObjectCreator::ModelObjectCreator() {
+#define REGISTER_CONSTRUCTOR(_className) \
+  m_newMap[_className::iddObjectType()] = [](openstudio::model::detail::Model_Impl * m, const IdfObject& object, bool keepHandle) { \
+    return std::shared_ptr<_className##_Impl>(new _className##_Impl(object, m, keepHandle)); \
```
mbadams5: I would use `std::make_shared<T>();`

```c++
return std::make_shared<_className##_Impl>(object, m, keepHandle);
```

openstudiocore/src/model/Model.cpp
```diff
@@ -3516,5 +2561,987 @@ void Model::applySizingValues() {
...
+  REGISTER_CONSTRUCTOR(ZoneHVACUnitVentilator);
+  REGISTER_CONSTRUCTOR(ZoneHVACBaseboardRadiantConvectiveWater);
+  REGISTER_CONSTRUCTOR(ZoneHVACBaseboardRadiantConvectiveElectric);
+  REGISTER_CONSTRUCTOR(ZoneMixing);
+  REGISTER_CONSTRUCTOR(ZoneVentilationDesignFlowRate);
+
+#define REGISTER_COPYCONSTRUCTORS(_className) \
+  m_copyMap[_className::iddObjectType()] = [](openstudio::model::detail::Model_Impl * m, const std::shared_ptr<openstudio::detail::WorkspaceObject_Impl>& ptr, bool keepHandle) { \
+    if (dynamic_pointer_cast<_className##_Impl>(ptr)) { \
+      return std::shared_ptr<_className##_Impl>(new _className##_Impl( \
```
mbadams5: Same as above, I would use `std::make_shared<T>();`

```c++
return std::make_shared<_className##_Impl>(*dynamic_pointer_cast<_className##_Impl>(ptr),m,keepHandle);
```

openstudiocore/src/model/Model.cpp
```diff
@@ -3516,5 +2561,987 @@ void Model::applySizingValues() {
...
+          *dynamic_pointer_cast<_className##_Impl>(ptr),m,keepHandle)); \
+    } \
+    else { \
+      OS_ASSERT(!dynamic_pointer_cast<openstudio::model::detail::ModelObject_Impl>(ptr)); \
+      return std::shared_ptr<_className##_Impl>(new _className##_Impl( \
```
mbadams5: Same as above, I would use `std::make_shared<T>();`

openstudiocore/src/model/Model_Impl.hpp
macumber: @kbenne does all this stuff need to be in the Model_Impl header?  Can it just be in the cpp file?  It's not anything we want exported.

macumber: I guess it is private but still wondering

macumber: I'm all for @mbadams5 's ideas if they are faster

Issue #3190, opened 2018-07-20
OS App looks for outdated library paths when installing a new version


jmarrec:
Happened today on Windows 10, I uninstalled 2.5.1 then installed 2.6.0.

![2018-07-20 14_05_13-loading model](https://user-images.githubusercontent.com/5479063/43010046-a358a97c-8c3f-11e8-89af-699a539a4924.png)

I went to Perferences > Change Default Libraries to fix this, but it shouldn't have happened in the first place.

The installer from 2.6.0 should have placed the correct librairies.

macumber:
I think the location of the default libraries are stored in a registry key or something like that, @kbenne could point you to them.  I am not 100% sure what the expected behavior is.  It seems that if you are pointing to old default libraries, it might be ok to just update that to the new default.  But if you set the libraries to custom files then we should use those.  Maybe we offer to reset the libraries to the current default if the libraries are not found?  Also have to be careful about moving backwards, e.g. install 2.6.0, set 2.6.0 default libraries, open 2.5.0, libraries fail to load because version translator can't open newer file. 

@asparke2 might have thoughts on behavior

jmarrec:
This is stored in NREL/OpenStudio, on Unix that's located at `~/.config/NREL/OpenStudio.conf`, windows that's in the registry, and Mac in a .plist file. Search for "Platform specific notes" [here](https://doc.qt.io/archives/qt-4.7/qsettings.html#fallback-mechanism)

Note that this conf file will have a `library` entry ONLY if you have set something custom. Otherwise it'll just look for `Resources/default/office.osm` and `Resources/default/hvac_library.osm`, and that will work with any installed version (see [OpenStudioApp::defaultLibraryPaths](https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/openstudio_app/OpenStudioApp.cpp#L1274))

Once you add something custom, then the library settings are stored and with absolute paths. Here's my OpenStudio.conf for eg:

```
[General]
lastPath=/home/julien
m_lastDdyPathOpened=
m_lastEpwPathOpened=

[library]
1\path=/usr/local/openstudio-2.6.0/bin/../Resources/default/hvac_library.osm
2\path=/usr/local/openstudio-2.6.0/bin/../Resources/default/office_default.osm
3\path=/usr/local/openstudio-2.6.0/Resources/90_1_2013.osm
size=3
```


[...]

PR #3191, opened 2018-07-24
Fix for #3190 - OS App looks for outdated library paths when new version is installed

jmarrec:
This happens when the user has custom settings (eg: other than `default/office_default.osm` and `default/hvac_library.osm`), because they are stored as absolute path and the absolute path in case of Resources/90_1_2013.osm will include the openstudio version.

Short story of the work done:

* Store library paths as relative if they are within the 'Resources' folder
* If failing to load library path, ask user if they want to just restore to default or open the library dialog to set them manually.

I need a reviewer to be assigned to this one since there was no one assigned yet on #3190


jmarrec:
Merged develop in, and resolved conflict.

macumber:
@jmarrec is this ready to merge?  @kbenne doesnt have bandwidth to review, seems like there are a lot of couts in there, seems like its not ready?

jmarrec:
@macumber Sorry I forgot to remove the three cout statements

PR #3195, opened 2018-07-24
Fix #3165

### LINKED ISSUE #3165: E+ example file 1ZoneDataCenterCRAC_wPumpedDXCoolingCoil.idf fails to import
User report this in 2.1 and 2.5, I confirmed it in pre 2.6 release (although I don't get a crash, it doesn't import). 
https://unmethours.com/question/32137/openstudio-crashing-with-idf-import-versions-21-25/

I don't get the same detail as the user

> The collection is INVALID at strictness level 'Draft', because of the errors:
Object     level data error of type NoIdd             .
Error pertains to an object of type 'Catchall', named ''.
Additional information about the error type: input data definition is missing or empty.

jmarrec:
Fix #3165. It was a small C++ loop problem in the Reverse Translator for EMS Program when it starts to replace the equations by the actual EMS Sensors and Actuators and Variables. 

After identifying the cause with a debugger, just needed a break to avoid continuing doing replacements after they have been done and leading to a crash.

@brianlball Would you mind reviewing this (small) one please?


jmarrec:
Yes, the outer loop iterates on each "word" of the actual program line, then tries to find a model object that has that name.

But your question tells me that I need to write Unit tests for this anyways.

Another thing might be that currently it looks for the name in ALL model objects. Shouldn't it do that only for EMS:XXX objects (InternalVariable, GlobalVariable, Sensors, Actuators)?

brianlball:
I think youre right.  After looking back at the book of ERL it looks like it should be limited to EMS:
Sensor
Actuator
GlobalVariable
InternalVariable
TrendVariable
CurveOrTableIndexVariable
ConstructionIndexVariable

jmarrec:
@brianlball I realized I needed to also handle the split on all possible operators, remove parenthesis, ignore reserved keywords (statement keywords(eg:IF), built-in constant (eg:PI) and dynamic (eg: Hour) variables)

So I moved that into a function called `splitEMSLineToTokens` and stored it in `utilities/core/StringHelpers.cpp` (next so `splitString`) so it can be used by both the EMS Program and Subroutine ones.

I also added a GTest for  `splitEMSLineToTokens` and for the Reverse translation of EMS Programs and Subroutines.

Could you please take another pass at reviewing now?

----

On a side note, I think I found a typo in `splitString`, see commit https://github.com/NREL/OpenStudio/pull/3195/commits/b825450cf8e517c950fe6be6c01b851163319228 to ensure that was the intended behavior.

jmarrec:
@brianlball do you want me to run the regression tests and report back?

brianlball:
@jmarrec  yeah that would be a good idea, thanks!  I've been doing that with all the ems related changes to make sure scotts stuff still works properly

jmarrec:
`ems.rb` fails with this branch. These lines https://github.com/jmarrec/OpenStudio-resources/blob/develop/model/simulationtests/ems.rb#L86:L90

You end up with this IDF:
```
EnergyManagementSystem:Program,
  Fan_Constant_Volume_1_Pressure_Rise_Program_by_Line, !- Name
  SET mult = OATdb_Sensor / 15.0,         ! - This is nonsense
  SET Fan_Constant_Volume_1_Press_Actuator = 250 * mult, ! - More nonsense
  ;                                       !- Program Line 3
```

E+ fatals: `   ** Severe  ** <root>[EnergyManagementSystem:Program][Fan_Constant_Volume_1_Pressure_Rise_Program_by_Line][lines][2] - Missing required property 'program_line'.`


I haven't tried current develop yet to see if it's my changes or others

brianlball:
hmm, this is the result from 2.6.0 back on 06/26/18

EnergyManagementSystem:Program,
  Fan_Constant_Volume_1_Pressure_Rise_Program_by_Line, !- Name
  SET mult = OATdb_Sensor / 15.0,         ! - This is nonsense
  SET Fan_Constant_Volume_1_Press_Actuator = 250 * mult; ! - More nonsense

EnergyManagementSystem:Program,
  Fan_Constant_Volume_1_Pressure_Rise_Program_by_Line_1, !- Name
  SET mult = OATdb_Sensor / 15.0,         ! - This is nonsense
  SET Fan_Constant_Volume_1_Press_Actuator = 250 * mult; ! - More nonsense

jmarrec:
@brianlball fails with current develop too.

brianlball:
ems.rb and ems.osm ran successfully with the 2.6.1 installer

jmarrec:
Confirmed that ems.rb runs fine with https://github.com/NREL/OpenStudio/pull/3195/commits/2ab98dbda1928710dd2d2bef5ce69550881c3dfa applied

brianlball:
looks good to me.  Those are the right changes to make after the changes to the .splitString function

brianlball:
@macumber @kbenne this looks good to me.  @jmarrec used the Gtests and the regression tests.

macumber:
@brianlball @jmarrec should this be merged for 2.7.0?

PR #3197, opened 2018-07-24
Fix #2555 - vrf font size too large

### LINKED ISSUE #2555: Larger font on Windows in vrf system view
https://unmethours.com/question/23627/openstudio-vrf-font-size/?answer=23688#post-id-23688

jmarrec:
Fix for #2555.

I change the thermalzone font size from 25 to 16. I don't know if that's enough to close #2555 or not, or if we should consider making the whole view larger.

Before:
before-font25

After
after-font16

Original issue assignee: @kbenne, though isn't really HVAC related, so perhaps @macumber should take it?
Can one of you two review this one please?


jmarrec:
I made the whole view larger, allows for around 50 characters which should be more than enough.

image

PR #3200, opened 2018-07-26
Fix 1465: Can't remove some SPMs via GUI because icons don't show

### LINKED ISSUE #1465: Can't remove some SPMs via GUI because icons don't show
When you have the following SPMs on certain nodes, you can't remove them via the GUI because the icons don't show up:

On any node:
- SetpointManager:MultiZone:Humidity:Minimum
- SetpointManager:MultiZone:MinimumHumidity:Average

On the OA inlet node of the OA system:
- Every SPM type (yes, I know it makes no sense to put these there in the first place, but if you mistakenly drop one you should be able to get rid of it without deleting the whole OA system)

My guess is that the humidity ones don't show because they don't have icons (at least in the library), and that none of them show on the OA system OA inlet node b/c of something in the drawing logic.

Here is a screenshot illustrating the humidity case:
image

No icons in the library:
image

image

And the OA system OA inlet node case:
image

### LINKED ISSUE #2220: air_loop.outdoorAirNode is not implemented
This works:

```
[1] (main)> puts loop.reliefAirNode.get
OS:Node,
{c3199b9a-ea85-4560-b112-ead02eb69a58}, !- Handle
Node 272,                               !- Name
{fa25ef58-bdf7-4517-8022-a75be31852d9}, !- Inlet Port
;                                       !- Outlet Port
```

This doesn't:

```
[1] (main)> loop.outdoorAirNode
[openstudio.model.AirLoopHVAC] <2> Not implemented.
RuntimeError: E:\Git\OS1\openstudiocore\src\model\AirLoopHVAC.cpp@1643 : Not implemented.
```

I know I can correctly get it by doing `oasys = loop.airLoopHVACOutdoorAirSystem.get` and then `oasys.outboardOANode.get`, but this seems weird.

jmarrec:
Fix #1465 and Fix #2220 

Changelog:

* Previous behavior was to only draw (and make removable) the SPMs with a `controlVariable`  that is "Temperature", now allows the other too (=the humidity ones basically)
* Implemented the `AirLoopHVAC::outdoorAirNode` and `AirLoopHVAC::returnAirNode`: since 2011 they were just throwing a "Not implemented" currently.
* Made  `AirLoopHVAC::mixedAirNode` and `AirLoopHVAC::reliefAirNode` const (const-correctness)
* `SetpointManager_Impl::addToNode` now doesn't allow SPMs to be placed on the `reliefAirNode` nor the `outdoorAirNode` of an AirLoopHVACOutdoorAirSYstem
* All SPMs that are allowed on a plantLoop now can only be placed on the supply side (I disallowed placing them on the demand side of a PlantLoop)

Original issue assignee: @kbenne. Could you review this one please?

### DIFF
--- a/openstudiocore/src/model/AirLoopHVAC.cpp
+++ b/openstudiocore/src/model/AirLoopHVAC.cpp
@@ -2156,22 +2156,22 @@ std::vector<ModelObject> AirLoopHVAC::oaComponents(openstudio::IddObjectType typ
   return getImpl<detail::AirLoopHVAC_Impl>()->oaComponents( type );
 }
 
-boost::optional<Node> AirLoopHVAC::outdoorAirNode()
+boost::optional<Node> AirLoopHVAC::outdoorAirNode() const
 {
   return getImpl<detail::AirLoopHVAC_Impl>()->outdoorAirNode();
 }
 
-boost::optional<Node> AirLoopHVAC::reliefAirNode()
+boost::optional<Node> AirLoopHVAC::reliefAirNode() const
 {
   return getImpl<detail::AirLoopHVAC_Impl>()->reliefAirNode();
 }
 
-boost::optional<Node> AirLoopHVAC::mixedAirNode()
+boost::optional<Node> AirLoopHVAC::mixedAirNode() const
 {
   return getImpl<detail::AirLoopHVAC_Impl>()->mixedAirNode();
 }
 
-boost::optional<Node> AirLoopHVAC::returnAirNode()
+boost::optional<Node> AirLoopHVAC::returnAirNode() const
 {
   return getImpl<detail::AirLoopHVAC_Impl>()->returnAirNode();
 }
--- a/openstudiocore/src/model/AirLoopHVAC.cpp
+++ b/openstudiocore/src/model/AirLoopHVAC.cpp
@@ -840,6 +840,18 @@ namespace detail {
     }
   }
 
+  boost::optional<Node> AirLoopHVAC_Impl::outdoorAirNode() const
+  {
+    if( airLoopHVACOutdoorAirSystem() )
+    {
+      return airLoopHVACOutdoorAirSystem()->outboardOANode();
+    }
+    else
+    {
+      return boost::optional<Node>();
+    }
+  }
+
   boost::optional<Node> AirLoopHVAC_Impl::reliefAirNode() const
   {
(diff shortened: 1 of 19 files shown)


kbenne:
@jmarrec I like all of this except one thing that I think might be an issue.

You changed, "All SPMs that are allowed on a plantLoop now can only be placed on the supply side (I disallowed placing them on the demand side of a PlantLoop)"

I'm pretty sure some primary / secondary pumping arrangements require a SPM on the demand side of plant and that is why the app allows them over there. I think this part needs to be reverted.

jmarrec:
Hum okay. This primary secondary pumping arrangement is strongly advised against lately anyway (use two plant loops with an ideal HX in between instead) but ok.
I still don’t think it should be allowed on a demand branch (between splitter and mixer).

Feel free to revert that commit though.

jmarrec:
@kbenne: I allow SPMs anywhere on a plantloop except on demand branches like you asked.

Could you please review again?

----

TL;DR: 

**Design decision**

I made a design decision here to leverage polymorphic behavior in order to keep code more centralized at the base class (`SetpointManager`) level. Instead of overriding the `addToNode` method for derived classes (SPMs) that **are** allowed on PlantLoop to call the `SetpointManager_Impl::addToNode` and if that fails check if you can connect it to a plantLoop, I created a virtual `isAllowedOnPlantLoop` Implementation method, and whether that returns `true` or `false`, the `SetpointManager_Impl::addToNode` will do the connection to the plantLoop. That way the only `override` that is needed for SPMs allowed on plantloops is to make `isAllowableOnPlantLoop` return `true` instead of the default `false`.

The connection logic stays in one place, so if in the future we want to restrict connection to only the supply side of a plant loop, then there's only one place where the change would be needed instead of multiple files.

**Exposing `isAllowableOnPlantLoop` publicly?**

 Because I'm not sure whether I should expose publicly the `isAllowableOnPlantLoop` for now, I committed that separately in the last commit, **so it'll be easy to revert if that's not something you do want.**
I personally think there's some value in exposing it, but I'll let you be the judge.

MatthewSteen:
I'd use an `isAllowedOnPlantLoop` (and AirLoop) method.

* https://unmethours.slack.com/archives/C0DEA30KY/p1522260122000058

I wrote a function to do this using the `hvac_library.osm`.

* https://unmethours.slack.com/archives/C0DEA30KY/p1522349497000669

Issue #2567, opened 2017-04-12
SketchUp plugin can't open files with non a-z letters in filepath 


mdahlhausen:
[From this unmethours question](https://unmethours.com/question/23709/problem-to-save-osm-file/).
User had an **é** in their windows username, which prevented them from opening or saving files to their desktop/documents directories.  Using Sketchup 2017 Make and OpenStudio 2.1.0 (Window 10 64 bits).

I was able to re-create the issue on my machine with an é in a directory name. I can save, but not open the .osm file.

asparke2:
**Workaround**
Don't put special characters in the file path.

macumber:
Also affects OpenStudio Application

jmarrec:
agreed that one should always avoid special characters AND spaces as much as possible (why create some cause for trouble if you can get around it). 

IMHO, what's really problematic is when your username on windows already is setup like this... Save for creating a brand new user, it's a nightmare to change. I've seen this problem pop up on UH a few times already so it's not uncommon.

This may be related to #2832 too (can't double click to open a file with a space in the name)

jmarrec:
I think I fixed the OpenStudio Application problem - NOT the Sketchup Plugin one - in #3207.

jmarrec:
Will test and try to reproduce this bug in current develop, see where we're at nowadays after a few path/encoding issues have been addressed.

jmarrec:
Tried in OpenStudio 2.7.1, opened sketchup, click "Load Model", and selected a model located at `C:\Users\julien\AfolderwithspécialCHar#%ù\test271.osm` and it worked fine.

jmarrec:
Interestingly, OpenStudio::Path.new fails with the example above though (but that's unrelated to this issue)

```
[1] 2832_CannotLoadSpaces(main)> path = "C:/Users/julien/Afolderwithsp\xE9cialCHar#%\xF9/test_model.osm"
=> "C:/Users/julien/Afolderwithsp\xE9cialCHar#%\xF9/test_model.osm"
[2] 2832_CannotLoadSpaces(main)> OpenStudio::Path.new(path)
RuntimeError: bad conversion
from (pry):4:in `initialize'
```

macumber:
The path error is not good and should be captured with ruby and c++ unit tests and an issue

jmarrec:
Opened a specific issue at #3436

PR #3207, opened 2018-07-27
Fix #2832  (and #2567) - OS App Cannot Open File Path with spaces or special characters on windows

### LINKED ISSUE #2832: Double-clicking osm does not load file if file path contains white space
Tested in Windows 7 and 10, OS v2.2.0 and v2.3.0.

1. Save osm to file path that contains white space (e.g. "C:/Project Name/project.osm")
2. Double-click on osm
3. OS app should launch with an empty 'Untitled' file.

Credit to @macumber for diagnosing.

jmarrec:
Fix #2832 - Cannot Open File Path with spaces on windows

Basically just quote the filepath so that if there are spaces, it's still captured as a single argument and not several.

If I double click on `C:\Users\julien\2832_CannotLoadSpaces\a file with spaces.osm`, here are the content of argv:

**Before**:

```
arg 0: C:\Users\julien\Software\Others\OS-build\Products\Debug\OpenStudioApp.exe
arg 1: C:\Users\julien\2832_CannotLoadSpaces\a
arg 2: file
arg 3: with
arg 4: spaces.osm
```

**After:** It is correctly identified as a single argument.

```
arg 0: C:\openstudio-2.6.1/bin/OpenStudioApp.exe
arg 1: C:\Users\julien\2832_CannotLoadSpaces\a file with spaces.osm
```

Review assignee: @macumber

### DIFF
--- a/openstudiocore/src/openstudio_app/install_operations.qs
+++ b/openstudiocore/src/openstudio_app/install_operations.qs
@@ -27,7 +27,8 @@ function Component()
 
     if( kernel == "winnt" ) {
       component.addElevatedOperation("CreateShortcut", "@TargetDir@/bin/OpenStudioApp.exe", "@StartMenuDir@/OpenStudio.lnk");
-      component.addElevatedOperation("RegisterFileType", "osm", "@TargetDir@/bin/OpenStudioApp.exe %1", "OpenStudio Model File", "text/plain", "@TargetDir@/bin/OpenStudioApp.exe,1");
+      // Note JM: you have to quote the %1 which represents the file path, otherwise any space in the path will think there are multiple args
+      component.addElevatedOperation("RegisterFileType", "osm", "@TargetDir@/bin/OpenStudioApp.exe \"%1\"", "OpenStudio Model File", "text/plain", "@TargetDir@/bin/OpenStudioApp.exe,1");
     }
   }
 }
--- a/openstudiocore/src/openstudio_app/OpenStudioApp.cpp
+++ b/openstudiocore/src/openstudio_app/OpenStudioApp.cpp
@@ -227,11 +227,13 @@ void OpenStudioApp::onMeasureManagerAndLibraryReady() {
       osversion::VersionTranslator versionTranslator;
       versionTranslator.setAllowNewerVersions(false);
 
-      boost::optional<openstudio::model::Model> model = versionTranslator.loadModel(toPath(fileName));
+      // Handle special chars by creating the path from a wstring
+      openstudio::path filePath(fileName.toStdWString());
+      boost::optional<openstudio::model::Model> model = versionTranslator.loadModel(filePath);
       if( model ){
 
-        m_osDocument = std::shared_ptr<OSDocument>( new OSDocument(componentLibrary(), 
-                                                                   resourcesPath(), 
+        m_osDocument = std::shared_ptr<OSDocument>( new OSDocument(componentLibrary(),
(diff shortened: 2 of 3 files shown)


jmarrec:
Fix for#2567, **talking specifically about the OS Application problem, NOT the sketchup plugin here **

Trying to load `C:\Users\julien\AfolderwithspécialCHar#%ù\test_model.osm` will fail indeed.

In [OpenStudioApp.cpp#L224:L225](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/openstudio_app/OpenStudioApp.cpp#L224:L225):

```
QFileInfo info(args.at(0)); // handles windows links and "\"
QString fileName = info.absoluteFilePath();
```

Until then, everything is fine, the QString stores stuff as UTF-16, so no problems.

On Line [230](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/openstudio_app/OpenStudioApp.cpp#L230), ``toPath(fileName)` is called. That's where things go wrong.

[toPath](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/core/Path.cpp#L71:75) calls `toString(fileName)` which is in [String.cpp#L69:74](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/core/String.cpp#L69:74)

```
/** QString to UTF-8 encoded std::string. */
std::string toString(const QString& q)
{
  const QByteArray& qb = q.toUtf8();
  return std::string(qb.data());
}
```

so this is similar to `openstudio::path filePath1(fileName.toUtf8().data());` and doesn't work.

I found two ways of dealing with this to construct a valid `openstudio::path` (`boost::filesystem::path`) from a `QString`:

```
// 1
openstudio::path filePath1(fileName.toStdWString());
// 2
openstudio::path filePath2(fileName.toLocal8Bit().constData()));
```

I'm not sure where I should make the modification though: should I do it globally in `String.cpp` (or in Path.cpp directly...), or just in `OpenStudioApp.cpp`. For now I'll just do it in OpenStudioApp.cpp unless you tell otherwise.


To construct a path directly from a QString I like 1/ better. If I were to modify the String.cpp `toString()` method, I'd use 2/ though.

FYI @macumber

openstudiocore/src/openstudio_app/install_operations.qs
```diff
@@ -27,7 +27,8 @@ function Component()
 
     if( kernel == "winnt" ) {
       component.addElevatedOperation("CreateShortcut", "@TargetDir@/bin/OpenStudioApp.exe", "@StartMenuDir@/OpenStudio.lnk");
-      component.addElevatedOperation("RegisterFileType", "osm", "@TargetDir@/bin/OpenStudioApp.exe %1", "OpenStudio Model File", "text/plain", "@TargetDir@/bin/OpenStudioApp.exe,1");
+      // Note JM: you have to quote the %1 which represents the file path, otherwise any space in the path will think there are multiple args
+      component.addElevatedOperation("RegisterFileType", "osm", "@TargetDir@/bin/OpenStudioApp.exe \"%1\"", "OpenStudio Model File", "text/plain", "@TargetDir@/bin/OpenStudioApp.exe,1");
```
macumber: +1 +1 +1 amazing :-)

jmarrec: Btw there’s a registerFileType example on Qt’s dock that uses single quotes around %1. That actually doesn’t work on windows though.

PR #3208, opened 2018-07-28
Fix 2832 cannot load file with spaces

macumber:
Extension of https://github.com/NREL/OpenStudio/pull/3207


macumber:
@jmarrec I took the bait and followed you into this rabbit hole again.....

I am attempting to standardize on the following encodings for each of these types

  | Class       | Internal      | Platform  |
------------ | ------------- | ------------
  | std::string | UTF-8, char   | All       |
  | path        | UTF-8, char   | Unix, Mac |
  | path        | UTF-16, wchar | Windows   |
  | QString     | UTF-16, wchar | All       |

I think the next step would be to replace our typedef openstudio::path = boost;:filesystem::path with an actual wrapper class, this would better support conversion of Ruby strings to paths.

macumber:
@kbenne if you are rebuilding boost I think it would be useful to build the locale library

macumber:
These types of changes scare me quite a bit, we will need a lot of testing around this.  I tried to make some unit tests and add files with special characters, but CMake couldn't copy them :-/  

As I understand things, opening files with special characters in them should have always worked on Mac or Linux.  It was only opening files on Windows that was an issue. Is that true @kbenne, @jmarrec ?

macumber:
Last commit allows this script to run:

```
require 'openstudio'

puts ''.encoding
puts 1
m = OpenStudio::Model::Model::load("日本語.osm")
puts 2.1
puts m.get.getBuilding.nameString
puts 2.2
m = OpenStudio::Model::Model::load(OpenStudio::Path.new("日本語.osm"))
puts 3.1
puts m.get.getBuilding.nameString
puts 3.2
```

jmarrec:
@macumber this line [Path_GTest.cpp#L39](https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/utilities/core/test/Path_GTest.cpp#L39) makes the build fail. I replaced with `#include "../Filesystem.hpp"`

macumber:
K, I'll commit that

PR #3213, opened 2018-08-02
Fix C# bindings after unicode path work - DO NOT MERGE YET

macumber:


jmarrec:
I rebase onto current develop and force-pushed. I ensure I didn't loose anything in between.

PR #3223, opened 2018-08-17
Fix #3222 wrong outputVariableNames

### LINKED ISSUE #3222: outputVariableNames contain "Average," or "Sum,"
eg: https://github.com/jmarrec/OpenStudio/blob/8ec66853fd06ab792ae6352a9e2bd6af1e270284/openstudiocore/src/model/SurfacePropertyOtherSideConditionsModel.cpp#L69

`"Average, Surface Other Side Conditions Modeled Convection Air Temperature",`

should be:

`" Surface Other Side Conditions Modeled Convection Air Temperature",`

jmarrec:
Fix #3222 wrong outputVariableNames.

Also added a small LOG message when calling `TableMultiVariableLookup::evaluate` which isn't implemented (the docstring says so, but I was very confused when it was just returning 1). Made it returns -999 too, that should be explicit...

All that stuff I found while working on other PRs, I put it in another branch for clarity...

Could you review please @macumber? This is really trivial...


openstudiocore/src/model/TableMultiVariableLookup.cpp
```diff
@@ -727,6 +728,7 @@ namespace detail {
       OS_ASSERT(d);
       double yValue = d.get();
 
+      // TODO: this should be SWIG'ed properly
```
macumber: To me this would mean creating and exporting a class for this type, is that what you mean?

openstudiocore/src/model/TableMultiVariableLookup.cpp
```diff
@@ -672,7 +672,8 @@ namespace detail {
 
   double TableMultiVariableLookup_Impl::evaluate(const std::vector<double>& x) const
   {
-    return 1.0;
+    LOG(Info, "Curve evaluation isn't implemented for TableMultiVariableLookup");
```
macumber: I'm going to make this a warning

openstudiocore/src/model/TableMultiVariableLookup.cpp
jmarrec: I’m not sure, but I mean getting something that you can interact with in ruby

jmarrec: I haven’t found a way to looks at points in ruby, the ‘std::vector<std::pair<std::vector<double>,double> >’ doesn’t allow you to interact with it really

jmarrec: Try this:

```ruby
m = OpenStudio::Model::Model.new
capModFuncOfWaterFlow = OpenStudio::Model::TableMultiVariableLookup.new(m, 1)
capModFuncOfWaterFlow.setName("CapModFuncOfWaterFlow")
capModFuncOfWaterFlow.setCurveType("Quadratic")
capModFuncOfWaterFlow.setInterpolationMethod("EvaluateCurveToLimits")
capModFuncOfWaterFlow.setMinimumValueofX1(0)
capModFuncOfWaterFlow.setMaximumValueofX1(1.33)
capModFuncOfWaterFlow.setMinimumTableOutput(0.0)
capModFuncOfWaterFlow.setMaximumTableOutput(1.04)
capModFuncOfWaterFlow.setInputUnitTypeforX1("Dimensionless")
capModFuncOfWaterFlow.setOutputUnitType("Dimensionless")

capModFuncOfWaterFlow.addPoint(0.0, 0.0)
capModFuncOfWaterFlow.addPoint(0.05, 0.001)
capModFuncOfWaterFlow.addPoint(0.33333, 0.71)
capModFuncOfWaterFlow.addPoint(0.5, 0.85)
capModFuncOfWaterFlow.addPoint(0.666667, 0.92)
capModFuncOfWaterFlow.addPoint(0.833333, 0.97)
capModFuncOfWaterFlow.addPoint(1.0, 1.0)
capModFuncOfWaterFlow.addPoint(1.333333, 1.04)

points = capModFuncOfWaterFlow.points
```

```ruby
points
=> #<SWIG::TYPE_p_std__vectorT_std__pairT_std__vectorT_double_std__allocatorT_double_t_t_double_t_std__allocatorT_std__pairT_std__vectorT_double_std__allocatorT_double_t_t_double_t_t_t:0x007f9688250ac8
 @__swigtype__=
  "_p_std__vectorT_std__pairT_std__vectorT_double_std__allocatorT_double_t_t_double_t_std__allocatorT_std__pairT_std__vectorT_double_std__allocatorT_double_t_t_double_t_t_t">
```

Can't do anything with it.

macumber: Yes, I think the right thing to do is make some class like TableMultiVariableLookupPoint, then refactor the API of TableMultiVariableLookup to have getters and setters that take that class.  The TableMultiVariableLookupPoint class should not be a ModelObject or ModelExtensibleGroup, just a dumb object with some plain old data fields.  Feel free to take that on if you'd like

jmarrec: I think there's a way to SWIG it properly using `%template` like in https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/model/ModelCore.i#L124:L125

I'll give that a shot.

jmarrec: The LOG and TODO I had added in TableMultiVariableLookUp actually has nothing to do with what this PR is all about, so I think I should open a new one

jmarrec:
@macumber Removed changes to TableMultiVariableLookUp.

macumber:
Approved

PR #3224, opened 2018-08-17
Fix #3019 - Crash happens when going to results tab when you do NOT have a companion folder

### LINKED ISSUE #3019: OS App mac will crash if you go
If you open an OSM file that doesn't have corresponding directly (like what you get off of downloaded PAT project) OpenStudio will crash if you go the results tab, with a message similar to this. Not finding the directory that it expects.

I re-created this with the following steps.

1. Launch OpenStudio
2 .Add few objects to model (like empty space types)
3. Save the model
4. Quit OS app
5. Throw away the directly that goes with the model you made, keeping only the OSM file
6. Launch OpenStudio and load the OSM
7. Go to results tab (should crash)

screen shot 2018-03-05 at 8 44 50 am

I tested this in 2.4.2, but then also confirmed it in 2.4.0, so it isn't a new bug. I don't think many people hit this, so I don't think it is critical to fix this for 2.5.0 unless it is a low risk fix.

jmarrec:
Fix #3019 - Crash happens when going to results tab when you do NOT have a companion folder

Just need to check if the directory actually exists first before passing it to directory_iterator. I confirmed that it now works on macOS High Sierra with that fix.

Review assignee: @macumber

### DIFF
--- a/openstudiocore/src/openstudio_lib/ResultsTabView.cpp
+++ b/openstudiocore/src/openstudio_lib/ResultsTabView.cpp
@@ -223,25 +223,31 @@ void ResultsView::searchForExistingResults(const openstudio::path &t_runDir, con
   std::vector<openstudio::path> radout;
   std::vector<openstudio::path> reports;
 
-  for ( openstudio::filesystem::recursive_directory_iterator end, dir(t_runDir);
-        dir != end;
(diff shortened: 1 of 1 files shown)


macumber:
Approved

Issue #2357, opened 2016-09-26
OS app crashes under edge case on schedule tab


DavidGoldwasser:
If you are dragging the profile (mouse button still held down) and try to type a value at the same time, you will get a crash sometimes ( in my two tests took about 4 and 15 attempts).

I tested this on 1.13.0 beta and latest 2.x installer. I was able to get each of them to crash. I don't propose to do anything to fix this for 1.x but we may want to investigate for 2.0. @kbenne what are your thought on this. 

As a right handed user I would never normally do this as I let go of the mouse to move my hand to the number keypad. I have not head any reports of anyone else seeing this other than one user (see end of comments on this question.
https://unmethours.com/question/11098/openstudio-error-unhandled-exception-in-schedules-tab/?answer=20136#post-id-20136

cc: @evanweaver

DavidGoldwasser:
Getting crashes on 2.5.1 when dragging vertical elements of schedules profiles. Not all of the time, just some times. Sometimes doesn't crash when I move it, but crashes on the next operation, such as saving, or switching tabs.

jmarrec:
Took me like 5 and 8 attempts to produce two crashes with current develop (9edf6e6, 2.6.0 basically) on Ubuntu 16.04, by following @DavidGoldwasser's steps.

Backtrace:

```
Process 19916 stopped
* thread #1, name = 'OpenStudioApp', stop reason = signal SIGSEGV: invalid address (fault address: 0xfffffffffffffff0)
    frame #0: 0x00007ffff66e3f93 libstdc++.so.6`__dynamic_cast + 35
libstdc++.so.6`__dynamic_cast:
->  0x7ffff66e3f93 <+35>: addq   -0x10(%rax), %r8
    0x7ffff66e3f97 <+39>: movq   -0x8(%rax), %rdi
    0x7ffff66e3f9b <+43>: movq   $0x0, (%rsp)
    0x7ffff66e3fa3 <+51>: movl   $0x0, 0x8(%rsp)


(lldb) bt
* thread #1, name = 'OpenStudioApp', stop reason = signal SIGSEGV: invalid address (fault address: 0xfffffffffffffff0)
  * frame #0: 0x00007ffff66e3f93 libstdc++.so.6`__dynamic_cast + 35
[...]

PR #3225, opened 2018-08-20
Fix #3119 and #2357: Schedules tab crashes

### LINKED ISSUE #3119: Schedules tab crashes when removing Run Period Profile from schedule
The app crashes when removing a Run Period Profile from a model created with the `Create Baseline Building` measure. I've been able to remove Run Period Profiles from schedules with fewer (e.g. 2), but the app does pause longer than normal and the removed schedule stays on the screen grayed out until clicking on another Run Period Profile. I suspect it has something to do with the number of Run Period Profiles in the baseline model, which has one for each day of the week. I'll send the model in for testing.

image

jmarrec:
Fix #3119: Schedules tab crashes when removing Run Period Profile from schedule

I'm not 100% sure why it was crashing only when removing the first run period and not all the time, but this seems to fix the issue. 

**The logic I implemented is as follows:**

* Get the ruleIndex (=priority) of the scheduleRule being deleted and its scheduleRuleset
* Delete it
* After deletion, count the number of scheduleRules and:
    * If none left, display the defaultDaySchedule instead, otherwise
    * Try to display the next lower priority one (the one that has the same priority as the one we deleted now...)
     * If can't, display the next higher priority one

**Couple of notable points for review**:

* I'm not sure I put that in the right place: I put it in `ScheduleTabController::removeScheduleRule(ScheduleRule&)` and not `SchedulesView::onScheduleRuleRemoved(const Handle)` because I want to get model stuff before the deletion happens.
* I left two `TODO` in there because I'm confused about why it's calling `getImpl` without doing anything with it?

**Review assignee:**

@macumber: could you review please?


jmarrec:
Fix #2357 too by disallowing entering number on keypad when dragging the schedule segments.

macumber:
Approved

PR #3226, opened 2018-08-21
Fix #1797 - Don't show 'Space' shading on the Facility 'Shading' subtab

### LINKED ISSUE #1797: "Shading" Sub-tab on building tab, pull down list shouldn't show "Space" as option
Space shading groups are shown in the space tab under the appropriate space. This tab looks like it is intended only for building and site shading. The pulldown list under the "Type" menu shows Building, Site, and Space. It doesn't let you choose space, which is correct, but it should be hidden.

By the way, interesting that this sub-tab lets you make a surface group with the green "+" button and also has a drop zone. I'm wondering if that was a mistake? We don't have that enabled for any other kind of surface group (spaces, space shading, interior partitions). We also don't let you delete the groups once they have been made.

The story tab and exterior equipment tab (also on building) does have "+" enabled and it should, I'm wondering if "Shading was just copied from this.

jmarrec:
Fix #1797 - Don't show 'Space' shading on the Facility 'Shading' subtab

Ensure that we don't display Space-level shading on the Facility 'Shading' subtab. Made it disappear on three locations:
* Shading groups whcih shadingSurfaceType() == 'Space' are not display
* Shading Surface Type Filter only proposes 'Site' and 'Building'
* The 'Type' dropdown only proposes 'Site' and 'Building'

Review assignee: @macumber

### DIFF
--- a/openstudiocore/src/openstudio_lib/FacilityShadingGridView.cpp
+++ b/openstudiocore/src/openstudio_lib/FacilityShadingGridView.cpp
@@ -95,7 +95,15 @@ namespace openstudio {
   FacilityShadingGridView::FacilityShadingGridView(bool isIP, const model::Model & model, QWidget * parent)
     : GridViewSubTab(isIP, model, parent)
   {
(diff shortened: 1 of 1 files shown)


macumber:
Approved

PR #3227, opened 2018-08-23
Fix #3120 - update standards

### LINKED ISSUE #3120: Space Type Measure Tags stability and functionality
screen shot 2018-05-04 at 2 26 10 pm

There are a number of bugs with this sub-tab, for simplicity, I'm going to file them here, with separate check boxes. 

- [ ] If you edit the "Standards Building Type" or "Standards Space Type" field, and then leave the space type tabs, without fist activating the name field, or another sub-tab within space type OpenStudio will crash. I tested this in both 2.5.1 and 2.4.0 on mac.
- [ ] When you change the "Standards Building Type" the "Standard Space Type" choices don't update until you leave the sub-tab and come back. Tested on 2.5.1. and 2.4.0 on mac
- [ ] When you type a string in the "Standards Building Type" or "Standards Space Type" fields it only lets you enter a single character, which you have to leave the sub-tab and come back to see (you can actually now enter a second character, then leave and come back to enter third). Tested on 2.5.1. and 2.4.0 in mac.
- [ ] 2.5.1 seems to have some but not all of the new "Standards Building Types" and "Standards Space Types" from the latest openstudio-standards gem. On the attached screenshot for example you can see what I assume are DEER building types, but you also see "Deli/Bakery" as a single space type, but in the gem there is now a "Deli" space type and a "Bakery" space type. Similar to this the office space types in the gem are now "SmallOffice - OpenOffice" instead of "OpenOffice" but the app in 2.5.1 doesn't reflect this. I ran across this in updating the Space Type and Construction Set Wizard, but this is also relevant to our model articulation measures used elsewhere.

jmarrec:
Fix for #3120 - Space Type Measure Tags stability and functionality

Task list:

* [x] (1) Update OpenStudio to use the latest openstudio-standards
* [x] (2) Properly allow use to enter custom standards building type / space type under the SpaceType tab (used to not work in Release and crash in Debug)
* [x] (3) Update dependent dropdown lists: when changing building type dropdown, the space type dropdown should be repopulated
* [x] (4) Because there are now a lot of building types: categorize by "Template"


jmarrec:
For (1), there is a specific thing that I need to mention (thanks to @asparke2 for bringing this to my attention):
* The regular (non NECB) standards data is here: https://github.com/NREL/openstudio-standards/tree/master/data/standards
* NECB has made their own files in https://github.com/NREL/openstudio-standards/tree/master/lib/openstudio-standards/standards/necb

So I wrote a ruby script to load the regular json file, delete NECB stuff from it (for the space_types.json only), and then load new NECB jsons and merge both datasets into a json file.

The C++  JSON parsers were updated to use the new openstudio-standards json structure

jmarrec:
For (4): because there are now many more building types etc, we decided to add a "Template" attribute.

So I ended up doing this:
* Inserted an IDD field 'Standards Template' in both `OS:Building` and `OS:SpaceType`
* Wrote a version translator to handle these inserted fields (and tested that it works)
* in Building and SpaceType classes, added `standardsTemplate` and  `suggestedStandardsTemplates` methods. Refactored `suggestedStandardsBuildingType` to populate based on `standardsTemplate` (which means it's empty if `standardsTemplate` is not set). In SpaceType, for `suggestedStandardsSpaceTypes` it will only return space types (from standards JSON) that match both Template and Building Type.

----

**I have one lingering question: if the Building has `standardsTemplate` set, when calling SpaceType::standardsTemplate and if not explicitly set, should it inherit from the Building?** Same question for `standardsBuildingType`?

----

Screenshots:

building

spacetype

jmarrec:
@macumber rebased onto develop, and I moved the `VersionTranslation` stuff into 2_6_2 to 2_7_0 as appropriate.

macumber:
Thanks @jmarrec, I am going to build a package for @DavidGoldwasser and @asparke2 to test

macumber:
Package for testing https://openstudio-builds.s3.amazonaws.com/2.7.0/OpenStudio-2.7.0.a237736488-Windows.exe

macumber:
@jmarrec there is interest in moving to standards 0.2.4 https://github.com/NREL/openstudio-standards/releases/tag/v0.2.4 for 2.7.0 release.  Can you see if that change would impact this PR?

macumber:
@jmarrec According to @asparke2 no updates are needed on this branch to support standards gem 0.2.3->0.2.4

jmarrec:
@macumber Actually, the C++ code parses the JSON in `model/resources/standards/`. I just regenerated them, forgot to push yesterday but I just did

PR #3228, opened 2018-08-24
Fix #78 - Add rendering color picker on ThermalZones tab

### LINKED ISSUE #78: Zones Tab - Minor issue but why don't we display color for thermal zone like we do elsewhere (Bugzilla #546)
On 2012-03-04 23:47:11, @DavidGoldwasser wrote:

> Look at Space Type and Story tab for color picker.

On 2012-03-05 16:10:10, @DavidGoldwasser wrote:

> <Batch change milestone from 0.7.0 to 0.7.1>

jmarrec:
Fix #78 - Add rendering color picker on ThermalZones tab

I put it in the first column to match the location on the 'SpaceType' tab:

[image](https://user-images.githubusercontent.com/5479063/44572462-2b5f7d00-a784-11e8-8e1d-813a181b390a.png)

Review assignee: @macumber  (it's really trivial)

### DIFF
--- a/openstudiocore/src/openstudio_lib/ThermalZonesGridView.cpp
+++ b/openstudiocore/src/openstudio_lib/ThermalZonesGridView.cpp
@@ -44,6 +44,8 @@
 #include "../model/ModelObject.hpp"
 #include "../model/ModelObject_Impl.hpp"
 #include "../model/Model_Impl.hpp"
+#include "../model/RenderingColor.hpp"
+#include "../model/RenderingColor_Impl.hpp"
 #include "../model/Schedule.hpp"
 #include "../model/Schedule_Impl.hpp"
 #include "../model/SizingZone.hpp"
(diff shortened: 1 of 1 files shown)


openstudiocore/src/openstudio_lib/ThermalZonesGridView.cpp
```diff
@@ -244,6 +248,12 @@ struct ModelObjectNameSorter{
 
         addSelectColumn(Heading(QString(SELECTED), false, false, checkbox), "Check to select this row");
       }
+      else if (field == RENDERINGCOLOR){
+        addRenderingColorColumn(Heading(QString(RENDERINGCOLOR), true, false),
```
macumber: Does this mean that apply to selected does not work on this column?

jmarrec: I didn't add an Apply to Selected for a few reasons:

* SpaceType tab didn't have it
* Clicking on the color brings up the Color picker, so it's hard to "selected" a row
* My use case for picking colors is usually to specifically make them look different for rendering purposes, so I didn't think this was necessary
* I hadn't done PR #3256 where I added "Apply to selected" to checkboxes yet, so I thought about it even less

Do you see a use case for having an "Apply to Selected" functionality for color pickers?

macumber: No, I just wanted to make sure we considered it, plus I had just looked at the other PR

macumber:
Approved

PR #3229, opened 2018-08-24
Fix #1554 - Always enable Revert to Saved 

### LINKED ISSUE #1554: Always enable revert to saved
Supports the use case where user edits OSM in notepad or SketchUp then wants to reload in App.

jmarrec:
Fix #1554 - Always enable Revert to Saved 

Changelog:
* Always Enable Revert to Saved: if we think the model is dirty, I just display an asterisk next to "Revert to Saved" (which may not be needed since the MainWindow title also does)

revert_to_saved

* Per @macumber's comment on the issue, Revert to Saved now will create a brand new model when it was never saved.

image

Question here is whether after this dialog box, we should also show the standard one (I think there's some value there).

image

Review assignee: @macumber


macumber:
Approve with minor changes

Issue #2165, opened 2016-03-20
ScheduleInterval unit test failures


macumber:
The following failing unit tests have been commented out, but they should be fixed

jmarrec:
Which unit tests? 

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/test/ScheduleInterval_GTest.cpp#L29 I don't see anything commented out.

In the forward trnaslate Gtest there is one test case that is commented out:
https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/energyplus/Test/ScheduleInterval_GTest.cpp#L397

jmarrec:
```
(py38)julien@OpenStudio (3311_skinny_bundle_tests *)$ grep -R "DISABLED_"
src/energyplus/Test/ScheduleInterval_GTest.cpp:408:TEST_F(EnergyPlusFixture, DISABLED_ForwardTranslator_ScheduleFixedInterval_TwoPoint)
src/energyplus/Test/ScheduleInterval_GTest.cpp:655:TEST_F(EnergyPlusFixture, DISABLED_ForwardTranslator_ScheduleVariableInterval_Hourly_Shifted)
src/energyplus/Test/ScheduleInterval_GTest.cpp:926:TEST_F(EnergyPlusFixture, DISABLED_ForwardTranslator_ScheduleVariableInterval_500_Shifted)
src/energyplus/Test/ScheduleInterval_GTest.cpp:1073:TEST_F(EnergyPlusFixture, DISABLED_ForwardTranslator_ScheduleVariableInterval_DaysTimeSeries)
src/isomodel/Test/ForwardTranslator_GTest.cpp:47:TEST_F(ISOModelFixture, DISABLED_ForwardTranslator)

```

jmarrec:
The `DISABLED_ForwardTranslator_ScheduleFixedInterval_TwoPoint` fails here https://github.com/jmarrec/OpenStudio/blob/2700a16d28ce7f8bf10cbdf16bfd1b1ac4263b31/src/energyplus/Test/ScheduleInterval_GTest.cpp#L425

Because of this:

https://github.com/NREL/OpenStudio/blob/17912d4d6be39e999fdd174908189b8a61b2ada5/src/utilities/data/TimeSeries.cpp#L339-L341


So here https://github.com/NREL/OpenStudio/blob/develop/src/model/ScheduleInterval.cpp#L104-L121

It's returning a ScheduleVariableInterval instead.

PR #3231, opened 2018-08-27
Fix failing model tests and disable failing FT tests

jmarrec:
Relates to #2165

Changelog:

* Fix two model tests that were failing
* Make all model tests inherit from ModelFixture, so that the StandardLogger configuration is effective (which will avoid having long useless output when running tests)
* Add `DISABLED_` in front of failing ForwardTranslator ScheduleInterval tests until they get fixed.

Note you can explicitly tell Google Test to run disabled tests by passing `--gtest_also_run_disabled_tests`

Review assignee: @macumber


macumber:
Approve

PR #3232, opened 2018-08-27
Fix #2000 - Allow dropping BCL construction on Surface subtab

### LINKED ISSUE #2000: Constructions from BCL can't be dropped on drop zones in spaces tab
I confirmed this for surfaces, but assume same for sub-surfaces.

Steps to reproduce
1. Open a model with a space and surfaces in it
2. Download a component form BCL using the GUI ( I downloaded "189.1-2009 Nonres 1A Ext Wall Mass")
3. Go to surfaces view on spaces tab, and then go to "Library" find your construction and try to drag it onto the drop zone. While ti should work it doesn't. If you try a non BCL construction it will work.

Bonus steps.
4. Go to the constructions tab and then either drop the BCL construction onto a construction drop zone for a construction set, or go to the "Constructions" sub tab and drop it in the far left drop zone. Either way it will add it to your model fine.
5. Now go back to the surfaces view of the spaces tab. This time go to "My Model" instead of "Library" and try to drag it onto a surface. Now it works.

So it doesn't seem to be issue with the construction object from the BCL component, maybe more related to getting the object from local BCL when dropping it onto a DropZone2?

@macumber I'll assign to you but feel free to re-assign to @evanweaver if needed.

No reason to put this in 1.10, can wait until later. Was found by user, but I'll let him know about workaround.
https://unmethours.com/question/13303/openstudio-cant-apply-constructions/

jmarrec:
Fix #2000 - Allow dropping BCL construction on Surface subtab

I implemented the fix in `OSDropZone2::dropEvent` where I check if it's a BCL thing and then load it appropriately, else continue to call `OSDocument::getModelObject `.

I'm not sure if this fix shouldn't be in `OSDocument::getModelObject` instead...

Review assignee: @macumber

### DIFF
--- a/openstudiocore/src/openstudio_lib/OSDocument.cpp
+++ b/openstudiocore/src/openstudio_lib/OSDocument.cpp
@@ -1598,6 +1598,7 @@ namespace openstudio {
         return m_compLibrary.getModelObject<model::ModelObject>(handle);
       }
     }
+    // TODO: should we handle BCL objects here?
 
     return boost::none;
   }
--- a/openstudiocore/src/openstudio_lib/OSDropZone.cpp
+++ b/openstudiocore/src/openstudio_lib/OSDropZone.cpp
@@ -42,6 +42,9 @@
 #include "../model/ModelObject_Impl.hpp"
 #include "../model/Model_Impl.hpp"
 
+#include "../model/Component.hpp"
+#include "../model/ComponentData.hpp"
+
 #include "../utilities/core/Assert.hpp"
 
 #include <QApplication>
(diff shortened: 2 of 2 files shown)


openstudiocore/src/openstudio_lib/OSDocument.cpp
```diff
@@ -1598,6 +1598,7 @@ namespace openstudio {
         return m_compLibrary.getModelObject<model::ModelObject>(handle);
       }
     }
+    // TODO: should we handle BCL objects here?
```
macumber: You mean like a reference to an online object that needs to be downloaded?

openstudiocore/src/openstudio_lib/OSDropZone.cpp
```diff
@@ -679,10 +682,31 @@ void OSDropZone2::dropEvent(QDropEvent *event)
...
   if((event->proposedAction() == Qt::CopyAction) && m_modelObject && m_set){
 
     OSItemId itemId(event->mimeData());
-    boost::optional<model::ModelObject> modelObject = OSAppBase::instance()->currentDocument()->getModelObject(itemId);
+
+    boost::optional<model::ModelObject> modelObject;
+
+    // If what you dragged is from the BCL, then VT it and insert it in model
+    // TODO: should we modify OSDocument::getModelObject instead?
+    if(itemId.sourceId() == OSItemId::BCL_SOURCE_ID)
```
macumber: @jmarrec I see now, yes please update OSDocument::getModelObject instead

macumber:
Waiting on @jmarrec to move code to OSDocument::getModelObject instead

openstudiocore/src/openstudio_lib/OSDropZone.cpp
jmarrec: Hum, now that I look into it, `OSDocument::getModelObject` is actually marked `const`. So if I modify it to insert the component into model if fromBCL, I have to remove the `const`. What do you think @macumber ?

jmarrec:
Hum, now that I look into it, `OSDocument::getModelObject` is actually marked `const`. So if I modify it to insert the component into model if fromBCL, I have to remove the `const`. What do you think @macumber ?

macumber:
I see, it could change the document by adding the item to the m_compLibrary.  That is fine with me, I'm not sure if making OSDocument::getModelObject non-const would snowball elsewhere.  I would be fine to make m_compLibrary mutable instead and leaving OSDocument::getModelObject const.  I think you can make the case that m_compLibrary  is a helper to the document, m_model should not be mutable though

macumber:
Alright, let me just handle this

macumber:
Approved with small changes

Issue #1697, opened 2015-06-23
Re-wrap Thermochromic window model properly


evanweaver:
Edit by @jmarrec on 2025-06-10 (empty message before):

The ThermochromicGlazing class we have is useless. because it tries to wrap this (emphasis mine)

```diff
OS:WindowMaterial:GlazingGroup:Thermochromic,
       \memo thermochromic glass at different temperatures
       \extensible:2
       \min-fields 4
  A1, \field Handle
       \type handle
       \required-field
  A2, \field Name
       \type alpha
       \required-field
       \reference MaterialNames
       \reference GlazingMaterialNames
  N1, \field Optical Data Temperature
       \type real
       \required-field
+      \begin-extensible
       \units C
       \ip-units F
  A3; \field Window Material Glazing Name
       \type object-list
       \required-field
       \object-list GlazingMaterialNames
```

So there are two extensible groups, but the class itself only has a single getter/setter pair for Optional Data Temperature and does not handle extensible groups at all, nor any Window Material Glazing Name

evanweaver:
https://docs.google.com/document/d/1FZaHTw_I_qkzKQmtsfK0c6CylMvtR-5x4V4sCU9SjXQ/edit#heading=h.24cfythh6riu

Widget to display extensible groups name and temp. GetExtensibleGroup(0).getFoo()

tijcolem:
Going to close this issue out due to inactivity. We can re-open if this is still relevant.

jmarrec:
Time to address this, it's super weird to have a broken class in the model namespace.

Issue #1329, opened 2015-01-09
OS:WindowMaterial:GlazingGroup:Thermochromic is wrong


macumber:
Should have an extensible list of pairs of Optical Data Temperature (double) and Window Material Glazing Name (drop down list)

jmarrec:
Should be addressed in #1697

evanweaver:
@macumber can you verify that this was fixed by #1697? Thanks.

asparke2:
Here's what I see in 2.4.0, not sure if I'm missing something.

![image](https://user-images.githubusercontent.com/2009290/34890495-5e72be8c-f78f-11e7-90b1-429c520dfb21.png)

macumber:
#1697 is another issue so it doesn't fix anything, I think Julien just meant it is a duplicate issue?  This object is not showing up correctly in the GUI, it's also weird that the default names are "80", "80 1".  Basic weirdness going on

jmarrec:
I think it did mean it was a duplicate (rewrapping is actually what’s needed probably).

——
Sent from a mobile device, please excuse the brevity.
Julien Marrec, EBCP, BPI MFBA
Owner


Direct: +33 6 95 14 42 13
Website: www.effibem.com

LinkedIn (en) | (fr)

jmarrec:
Yeah duplicate of #1697  which involves properly wrapping the object, I suggest closing this one for clarity.

#1697 is labeled as an Enhancement Request, which is partly true, but I'd also consider it a bug because we have an unusable thing in the OS App currently...

jmarrec:
Rendered moot by #3233. Can you close please @macumber?

PR #3233, opened 2018-08-27
#1329 - Comment out Thermochromic glazing in OS App

jmarrec:
Relates to #1329 and #1697.

Problem is that the ThermochromicGlazing isn't wrapped correctly (or at least fully) in the SDK. Until it is, we comment it out of the OS App.

What I chose to do is to just not display Thermochromic in MainRightColumn nor under the Materials tab.

That was the minimal amount of commenting out that would just not display it to the user, and it would be really easy to re-add it.

**If you want me to take it out completely** (including removing `WindowMaterialGlazingGroupThermochromicInspectorView` from CMakeLists.txt), **please let me know and I will**

Review assignee: @macumber


macumber:
Approved

jmarrec:
You can't. The object isn't properly wrapped yet. You should use an EnergyPlus Measure to inject it in your model.

PR #3236, opened 2018-08-28
Fix #2934 - Fix qt warnings in app

### LINKED ISSUE #2934: Fix qt warnings in app
Currently a large number of warnings are output to the console on Mac, on Windows these can be seen by running the app with `C:\openstudio-version\bin\OpenStudioApp.exe > out.txt 2>&1`

jmarrec:
Fix #2934 - Fix qt warnings in app

Removed all std::cout statements I could find, and fixed the Qt warning about no placeholder in window title (see Qt docs about [windowModified](https://doc.qt.io/qt-5/qwidget.html#windowModified-prop)).
There aren't any warnings in Release mode anymore.

 In Debug mode, there's this Qt warning that will keep poping up 16 times each time you switch to a tab that has an OSGridController in it:
```
Warning: QLayout: Attempting to add QLayout "" to openstudio::OSGridView "", which already has a layout ((null):0, (null))
```

I haven't been able to pinpoint the root cause of this one. Refer to #2934 for more info.

Review assignee: @macumber

### DIFF
--- a/openstudiocore/src/openstudio_lib/EMSInspectorView.cpp
+++ b/openstudiocore/src/openstudio_lib/EMSInspectorView.cpp
@@ -69,7 +69,7 @@ class EMSInspectorActuator : public QWidget {
 
   public:
 
-  EMSInspectorActuator(const model::EMSActuatorNames & names, const model::ModelObject & modelObject) : 
+  EMSInspectorActuator(const model::EMSActuatorNames & names, const model::ModelObject & modelObject) :
     QWidget(),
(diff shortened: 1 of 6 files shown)


jmarrec:
I added more debugging in a separate branch than the PR see https://github.com/jmarrec/OpenStudio/commit/df22d2408d224610255664a5a93302b4450d19fe

I didn't add it to the PR because I'm not sure we do want to set a name to all the layouts or not

PR #3238, opened 2018-08-29
Fix #1406 - Connect AirLoopHVAC components to PlantLoop like original + Implement PlantLoop::clone

### LINKED ISSUE #1406: AirLoop::Clone
AirLoop::clone needs to follow connections in the supply and demand side and clone those objects foud there and then attach them to the newly cloned AirLoop

jmarrec:
Fix #1406 

Changelog:
* Connect AirLoopHVAC components to PlantLoop like original 
* Implemented `PlantLoop::clone` as well.

Side Notes:
* I made Loop_Impl::demandMixer and demandSplitter const
* I made Impl `createTopology` (function was already defined in AirLoopHVAC) a pure virtual method of `Loop_Impl` and override in both `AirLoopHVAC_Impl` and `PlantLoop_Impl`

Review assignee: @kbenne

### DIFF
--- a/openstudiocore/src/model/AirLoopHVAC.cpp
+++ b/openstudiocore/src/model/AirLoopHVAC.cpp
@@ -907,12 +907,12 @@ namespace detail {
     return result;
   }
 
-  Splitter AirLoopHVAC_Impl::demandSplitter()
+  Splitter AirLoopHVAC_Impl::demandSplitter() const
   {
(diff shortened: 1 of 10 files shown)


kbenne:
@jmarrec I like this and I'm merging it, but can you make sure something gets in the release notes about this change in behavior?

PR #3243, opened 2018-08-30
Wrap Schedule:File

jasondegraw:
## Wrap Schedule:File

Wrap the EnergyPlus Schedule:File object. Two new objects are created for this: ExternalFile and ScheduleFile. ExternalFile manages the file connection, and at some point should do most of the validation. ScheduleFile manages the column of the file that makes up the schedule. Very little validation is done currently, only existence of the file is checked.

### DIFF
--- a/openstudiocore/src/model/ScheduleFile.hpp
+++ b/openstudiocore/src/model/ScheduleFile.hpp
@@ -129,7 +129,6 @@ class MODEL_API ScheduleFile : public ScheduleInterval {
 
   explicit ScheduleFile(std::shared_ptr<detail::ScheduleFile_Impl> impl);
 
-  friend class detail::ScheduleFile_Impl;
   friend class Model;
   friend class IdfObject;
   friend class openstudio::detail::IdfObject_Impl;
--- a/openstudiocore/src/model/ExternalFile.hpp
+++ b/openstudiocore/src/model/ExternalFile.hpp
@@ -95,7 +95,6 @@ class MODEL_API ExternalFile : public ResourceObject {
 
   explicit ExternalFile(std::shared_ptr<detail::ExternalFile_Impl> impl);
 
-  friend class detail::ExternalFile_Impl;
   friend class Model;
   friend class IdfObject;
   friend class openstudio::detail::IdfObject_Impl;
--- a/openstudiocore/src/model/ExternalFile_Impl.hpp
+++ b/openstudiocore/src/model/ExternalFile_Impl.hpp
@@ -104,7 +104,7 @@ namespace detail {
     //@}
    protected:
      bool setFileName(const std::string& fileName);
-     friend class ExternalFile;
+     friend class openstudio::model::ExternalFile;
 
    private:
      REGISTER_LOGGER("openstudio.model.ExternalFile");


jasondegraw:
@macumber So here you go, I was able to put together initial implementations of both forward and reverse translation. There is unit testing, but not much else in terms of testing.

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -4726,6 +4726,87 @@ OS:ScheduleTypeLimits,
...
+       \type handle
+       \required-field
+  A2,  \field Name
+       \required-field
+       \type alpha
+       \reference ScheduleNames
+  A3,  \field Schedule Type Limits Name
+       \type object-list
+       \object-list ScheduleTypeLimitsNames
+  A4,  \field File Name
```
macumber: External File Name?

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -4726,6 +4726,87 @@ OS:ScheduleTypeLimits,
...
+       \default Comma
+  A6,  \field Interpolate to Timestep
+       \note when the interval does not match the user specified timestep a "Yes" choice will average between the intervals request (to
+       \note timestep resolution.  a "No" choice will use the interval value at the simulation timestep without regard to if it matches
+       \note the boundary or not.
+       \type choice
+       \key Yes
+       \key No
+       \default No
+  N4;  \field Minutes per Item
```
macumber: Based on \min-fields 6 this field should have a default value, is the default 60?

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -4726,6 +4726,87 @@ OS:ScheduleTypeLimits,
...
+       \required-field
+       \type integer
+       \minimum 0
+  N3,  \field Number of Hours of Data
+       \note 8760 hours does not account for leap years, 8784 does.
+       \note should be either 8760 or 8784
+       \default 8760
+       \minimum 8760
+       \maximum 8784
+  A5,  \field Column Separator
```
macumber: Hmm, this field is in both objects, seems like it should only be in one or the other.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateScheduleFile.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+  boost::optional<ScheduleTypeLimits> scheduleTypeLimits = modelObject.scheduleTypeLimits();
+  if (scheduleTypeLimits){
+    boost::optional<IdfObject> idfScheduleTypeLimits = translateAndMapModelObject( *scheduleTypeLimits );
+    if (idfScheduleTypeLimits){
+      idfObject.setString( openstudio::Schedule_FileFields::ScheduleTypeLimitsName, idfScheduleTypeLimits->name().get() );
+    }
+  }
+
+  idfObject.setString( openstudio::Schedule_FileFields::FileName, modelObject.externalFile().name().get() );
```
macumber: The way I read the IDD, externalFile().name() would be 'ExternalFile 1' and externalFile().fileName() would be the actual file path right?

openstudiocore/src/energyplus/ReverseTranslator.cpp
```diff
@@ -791,6 +791,11 @@ boost::optional<ModelObject> ReverseTranslator::translateAndMapWorkspaceObject(c
       modelObject = translateScheduleDayHourly(workspaceObject);
       break;
     }
+  case openstudio::IddObjectType::Schedule_File:
```
macumber: I guess we should have a no-op translation for the ExternalFile.  Possibly that could be implemented some day to do some file copying magic?

openstudiocore/src/energyplus/ReverseTranslator/ReverseTranslateScheduleFile.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+static ExternalFile externalFileFromFileName(Model &model, const std::string &filename, const boost::optional<std::string> &separator)
+{
+  if ( separator ) {
+    for ( auto & extfile : model.getConcreteModelObjects<model::ExternalFile>() ) {
+      if ( extfile.fileName() == filename ) {
+        if ( extfile.columnSeparator() == separator.get() ) {
+          return extfile;
+        }
+        // If the separator doesn't match, maybe should thow here
```
macumber: Yeah, this would be a bad IDF file

macumber: Maybe this suggests that we should put the column separator on ScheduleFile instead of ExternalFile?  Then bad IDF objects just get turned into bad OSM objects and fail at E+ like they would before.

openstudiocore/src/model/ExternalFile.cpp
```diff
@@ -0,0 +1,188 @@
...
+    // Failed to open file
+    LOG_AND_THROW("Failed to open file \"" << filename << "\" for " << this->briefDescription());
+  }
+  ifs.close();
+
+  bool ok;
+  ok = setFileName(filename);
+  OS_ASSERT(ok);
+
+  // Moving the file etc. should go here
```
macumber: Here is where you could get the model's WorkflowJSON:
https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/Model.hpp#L108

From that you can get the absoluteRootDir:
https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/utilities/filetypes/WorkflowJSON.hpp#L164

If you copy the file to that directory you would store only the fileName (not full path) in the fileName field

openstudiocore/src/model/ExternalFile.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+std::vector<std::string> ExternalFile::columnSeparatorValues() {
+  return getIddKeyNames(IddFactory::instance().getObject(iddObjectType()).get(),
+                        OS_External_FileFields::ColumnSeparator);
+}
+
+std::string ExternalFile::fileName() const {
+  return getImpl<detail::ExternalFile_Impl>()->fileName();
+}
+
```
macumber: Could add a helper method here to get the full path to the file, something like:

model().workflowJSON().findFile(fileName());

openstudiocore/src/model/ScheduleFile.hpp
```diff
@@ -0,0 +1,143 @@
...
+  //@{
+
+  // TODO: Check return type. From object lists, some candidates are: ScheduleTypeLimits.
+  boost::optional<ScheduleTypeLimits> scheduleTypeLimits() const;
+
+  ExternalFile externalFile() const;
+
+  int columnNumber() const;
+
+  int rowstoSkipatTop() const;
```
macumber: This is a poster child for why snake case is better.  To me rowsToSkipAtTop looks better

openstudiocore/src/model/ScheduleFile.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+  boost::optional<int> numberofHoursofData() const;
+
+  bool isNumberofHoursofDataDefaulted() const;
+
+  bool interpolatetoTimestep() const;
+
+  bool isInterpolatetoTimestepDefaulted() const;
+
+  boost::optional<std::string> minutesperItem() const;
```
macumber: Should this return an int?

openstudiocore/src/model/ScheduleFile.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+  bool setRowstoSkipatTop(int rowstoSkipatTop);
+
+  bool setNumberofHoursofData(int numberofHours);
+
+  bool setInterpolatetoTimestep(bool interpolatetoTimestep);
+
+  void resetInterpolatetoTimestep();
+
+  bool setMinutesperItem(const std::string& minutesperItem);
```
macumber: Add a method taking an int?

openstudiocore/src/model/ExternalFile.cpp
```diff
@@ -0,0 +1,188 @@
...
+    return isEmpty(OS_External_FileFields::ColumnSeparator);
+  }
+
+  bool ExternalFile_Impl::setFileName(const std::string& fileName) {
+    bool result = setString(OS_External_FileFields::FileName, fileName);
+    OS_ASSERT(result);
+    return result;
+  }
+
+  bool ExternalFile_Impl::setColumnSeparator(const std::string& columnSeparator) {
```
macumber: Add support to translate ',' to 'Comma' here?

openstudiocore/src/model/test/ScheduleInterval_GTest.cpp
```diff
@@ -242,3 +246,28 @@ TEST_F(ModelFixture, Schedule_VariableInterval3)
...
 }
+
+TEST_F(ModelFixture, Schedule_File)
+{
+  Model model;
+
+  path p = resourcesPath() / toPath("model/schedulefile.csv");
+
+  ExternalFile externalfile(model, openstudio::toString(p));
+  EXPECT_EQ(openstudio::toString(p), externalfile.fileName());
```
macumber: I think this should equal "schedulefile.csv" and we should also test that the file was copied.

openstudiocore/resources/model/OpenStudio.idd
macumber: After reviewing the PR I think we should keep Column Separator in Schedule:File and remove it from External:File

openstudiocore/src/model/ExternalFile.cpp
macumber: I'll give this a shot and see if it works.

openstudiocore/src/energyplus/ForwardTranslator.cpp
```diff
@@ -2463,6 +2463,12 @@ boost::optional<IdfObject> ForwardTranslator::translateAndMapModelObject(ModelOb
       retVal = translateScheduleFixedInterval(schedule);
       break;
     }
+  case  openstudio::IddObjectType::OS_Schedule_File:
+   {
+    model::ScheduleFile schedule = modelObject.cast<ScheduleFile>();
+    retVal = translateScheduleFile(schedule);
+    break;
+   }
   case  openstudio::IddObjectType::OS_Schedule_Rule :
```
macumber: Add a no-op translation for ExternalFile to avoid warnings, in future an implementation could do file copy magic

openstudiocore/src/model/test/ScheduleInterval_GTest.cpp
```diff
@@ -242,3 +246,28 @@ TEST_F(ModelFixture, Schedule_VariableInterval3)
 
   EXPECT_TRUE(schedule.optionalCast<ScheduleVariableInterval>());
 }
+
+TEST_F(ModelFixture, Schedule_File)
```
macumber: Should be in ScheduleFile_GTest.cpp

macumber: Meh, I see what you did here, you win this time!

openstudiocore/src/model/ScheduleFile.hpp
MatthewSteen: @jasondegraw and @macumber should the `setMinutesperItem` method take an `int` argument rather than a string? I wrote a measure in 3.1.0 and it only takes a string, which differs from the field in EP. If so, I can open an issue.

macumber: Yeah that seems weird, it should be easy to add an override that takes an `int`

jasondegraw: I think this may be an unintended outcome of using a choice in the OpenStudio IDD here rather than what E+ does. But I agree, at the very least just adding an int version makes sense.

openstudiocore/src/model/ScheduleFile.hpp
jasondegraw: Yep, looks like this comment just got missed in review.

PR #3246, opened 2018-09-03
Fix #190 #1445 #2252 - Clean up HVAC Library 

### LINKED ISSUE #190: Seems like library definition for water use equipment should have target temp (Bugzilla #916)
On 2012-09-18 08:49:03, @DavidGoldwasser wrote:

> if this is a guide for user of what they should do, I would think a target temperature would be ideal. This way the user can alter the temperature setting but still supply same water temp to fixture.

### LINKED ISSUE #1445: Add schedule type limits to schedules in HVAC libraries
Purpose is to keep other things from stealing HVAC schedules

### LINKED ISSUE #2252: Clean up HVAC Library
There are multiple duplicate objects in the HVAC library, especially for stuff like schedules.

Example:
- `OS:Schedule:Constant`: `Always On Discrete`, `Always On Discrete 1`, `Always On Discrete 2`
- `OS:ScheduleTypeLimits`: Fractional`,`Fractional 1`,`Fractional 2`

jmarrec:
Changelog:

* Fix #190 : add target temp and fraction latent for water use equipment definition
* Fix #1445: cleanup and assign schedule type limits
* Fix #2252: Clean up HVAC Library
    * Renamed a few object names
    * Mostly I tracked down why some coils and fans that were actually contained in other HVACComponent/ZoneHVACComponent were showing the the library tab  while `ModelObjectListController::makeVector` is properly checking whether comp is contained in HVACComp or ZoneHVACComp. Turns out these objects containingZoneHVACComponent and containingHVACComponent where missing some uses cases. There might still be other types that are missing, but at least I checked all collapsible library menus to ensure that only the right objects show up. See example below:

Before:

image

After:

image

Review assignee: @kbenne


jmarrec:
@macumber I think I have warned that this PR was bound to get stale real quick (pretty much anytime someone adds to or modifies the hvac_library) and conflicts could suck to resolve. 

I’m happy to resolve any conflicts provided that we know for sure someone will be reviewing right after with intent to merge it in, otherwise there’s a good chance it’ll be all for nothing.

And provided someone has bandwidth to review I don’t see why this shouldn’t get in 2.7.1

PR #3248, opened 2018-09-04
Fix #2109 - Allow setting the ZoneAirContaminant Outdoor Air Carbon Dioxide Schedule

### LINKED ISSUE #2109: OS:ZoneAirContaminantBalance not translated

jmarrec:
Fix #2109 - Allow setting the ZoneAirContaminant Outdoor Air Carbon Dioxide Schedule

I made the field an OSComboBox2, and I had to implement a new template `bind` method that creates an underlying `OptionalChoiceSaveDataSourceConceptImpl`. I say I had because I tried for a while to use the existing template at [OSComboBox.hpp#L144:L161](https://github.com/NREL/OpenStudio/blob/f6f0bf1767985a6a84ac2c2b005ea0b421db8411/openstudiocore/src/shared_gui_components/OSComboBox.hpp#L144:L161) but the problem is the setter (you can see my try on a separate branch if you're curious, see [here](https://github.com/NREL/OpenStudio/compare/develop...jmarrec:Not_used/2109_OutdoorCarbon?expand=1#diff-e16e7db4e151dc81ffe7b0b47fe2167eR1589))

image

Not sure if @kbenne (whom I talked to about the bind stuff)  or @macumber should review this one


macumber:
Approve

PR #3253, opened 2018-09-05
Fix #3249 - path problems

### LINKED ISSUE #3249: Measure in user's directory that has a dot in path refuses to run in Apply Now
I helped someone on slack today debug a problem. He was trying to run the [ASHRAE 90.1-2007 Sys 7 Baseline Measure](https://bcl.nrel.gov/node/82782) on a model in OS App 2.3.0, using Windows.

He downloaded the measure on BCL via his browser, and extracted it to his "My measures" folder. His username has no space, but the measure itself has:

`C:\Users\username\OpenStudio\Measures\ASHRAE 90.1-2007 Sys 7 Baseline Measure\`

Running the measure produced a crash for him:

After a little while I had him rename the measure folder to `ASHRAE_90-1-2007_Sys_7_Baseline_Measure` and relaunch the OS App, and it worked just fine.

I was actually able to reproduce this on Ubuntu using 2.6.0 official.

jmarrec:
Fix #3249 - path problems

More info on the issue.

Changelog:
* Do not use `.stem()` when trying to get the last level directory name because it won't work properly if you have a `.` in there, instead use `.filename()`
* Created two functions in PathHelpers for clarity (and easier to change later if need be / add checks): `getCompanionFolder(path& osmPath)` and `getLastLevelDirectoryName(path& directory)`
   * Note that we could include some tests in there, but I didn't because of speed (we assume we know when we pass an OSM file or a directory to the appropriate functions)
* Made various locations where file stuff is happening use these two functions

Warning: Please pay close attention to 8c6b8ed95a366fc7ca654f2bdb2891348b01fafc where I think I caught a problem and fixed it, thanks!

Review assignee: @macumber

### DIFF
--- a/openstudiocore/src/model/FileOperations.cpp
+++ b/openstudiocore/src/model/FileOperations.cpp
@@ -338,8 +338,7 @@ QString longPathName(const QString& path)
       }
 
       // Copy all files from existing resources dir into temp dir when opening
-      // osmPath.stem() removes the file extension
-      openstudio::path sourceDir = osmPath.parent_path() / osmPath.stem();
+      openstudio::path sourceDir = getCompanionFolder(osmPath);
       openstudio::path destDir = modelTempDir / toPath("resources");
       if (openstudio::filesystem::exists(sourceDir)){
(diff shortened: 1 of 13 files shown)


macumber:
Approved

PR #3255, opened 2018-09-06
Fix #1877 non functional link subtab for HVACComponents

### LINKED ISSUE #1877: Edit tab for many zone objects have an extra non-functional iink sub-tab
I confirmed this on Water to Air HP, Unit Ventilator and Unit Heater.

2015-09-17 09_49_24-os-1 9 0-3 links
screenshot from @jmarrec

jmarrec:
Fix #1877 non functional link subtab for HVACComponents

Changelog:
* Added `LibraryTabWidget::hideTab` function to hide the tab by hiding the link button and handle the currentIndex. Adapted from [HorizontalTabWidget::hideTab](https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/openstudio_lib/HorizontalTabWidget.cpp#L225)
* For each component, instead of calling `LoopChooserView::layoutModelObject` with an empty optional, I call `hideTab`

eg: PTAC with an electric coil:

image

PTAC with a HW coil:

image

One thing I need to mention is that if you have a cooling coil (3rd tab) and no heating coil (2nd tab) it'll look like this, which I don't see as a problem:

image

Review assignee: @macumber

### DIFF
--- a/openstudiocore/src/openstudio_lib/InspectorView.cpp
+++ b/openstudiocore/src/openstudio_lib/InspectorView.cpp
@@ -1503,9 +1503,8 @@ void AirTerminalInspectorView::layoutModelObject( model::ModelObject & modelObje
 
   if( ! waterCoil )
   {
-    boost::optional<model::ModelObject> mo;
-
-    m_loopChooserView->layoutModelObject(mo);
+    // Hide the tab (by hiding the button)
+    m_libraryTabWidget->hideTab( m_loopChooserView, true);
   }
 }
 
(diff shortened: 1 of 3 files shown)


jmarrec:
**Note:** During testing I noticed that when you click on a ZoneHVACComp on the ThermalZone tab, for eg`ZoneHVAC:PackagedTerminal:AirConditionner`, the CTOR for `ZoneHVACPackagedTerminalAirConditionerInspectorView` is called twice in a row... In fact, there's a debug message that shows you that removing/setting is done twice in a row.

```
[openstudio.openstudio_lib.MainRightColumnController] <-2> Removing old edit widget: 0x6929380
[openstudio.openstudio_lib.MainRightColumnController] <-2> Setting new edit widget: 0x6929380
[openstudio.openstudio_lib.MainRightColumnController] <-2> Removing old edit widget: 0x6929380
[openstudio.openstudio_lib.MainRightColumnController] <-2> Setting new edit widget: 0x6929380
```

PR #3258, opened 2018-09-07
Fix #535 - update system combo box on loop name change

### LINKED ISSUE #535: Renamed loop doesn't get updated in pull down until you switch to "control tab" or change loops
I noticed that if I change a loop name that the pull down list doesn't immediately update. In fact even if I move away from the HVAC and come back it still doesn't update.

It will update if I change from the the "Layout" to "Controls" view and back, or if I change to another loop in the "Layout" view either from the pulldown menu or from a linked component.

jmarrec:
Fix #535 - update system combo box on loop name change

I used debug message to make sure that I don't end up triggering the same slots repeatedly (it proved useful: I actually was initially, forgot to add `!` to a bool check, but I fixed it!)

Demo:

demo

Review Assignee: @macumber


kbenne:
slick.

PR #3263, opened 2018-09-12
Fix #2776 - Spaces tab Filters

### LINKED ISSUE #2776: Spaces tab Filters do not work for all sub-tabs
The Filters in the Spaces tab work in the Properties sub tab but do nothing in the Loads, Surfaces, Subsurfaces, Interior Partitions, and Shading sub-tabs. Tested with a 1.14.0 model updated to 2.2.1.

jmarrec:
Fix #2776 - Spaces tab Filters

There are two types of filters:
* Row-level filters: these apply to the Space itself
* Subrow-level filters: these apply to whatever is displayed (eg: Loads, etc). The current approach was to check the parent of the subrow-level object, and if that doesn't match, the parent's parent.
    * If the subrow-level object is a subSurface, it looks for Surface, then the Surface's parent = Space: that's OK
    * If the subrow-level object is a load (eg: People), it looks for the parent. If the parent is a Space => works kinda ok. If it's the Space Type => Never matches.

filters

I had to implement a workaround that consist in distinguishing between the "Properties" subtab (`SpacesSpacesGridView`) which doesn't have subrow level objects, and the rest of the subtabs that do. It's perhaps not the most elegant way, but I couldn't figure out how to do it without.

Demonstration of UX:

2776_ux

Review assignee: @macumber

### DIFF
--- a/openstudiocore/src/openstudio_lib/SpacesSpacesGridView.hpp
+++ b/openstudiocore/src/openstudio_lib/SpacesSpacesGridView.hpp
@@ -83,6 +83,9 @@ namespace openstudio{
 
     REGISTER_LOGGER("openstudio.SpacesSpacesGridView");
 
+    // Overrides SpacesSubtabGrid view to trigger appropriate filtering
+    virtual bool hasSubRows() override { return false; };
+
   protected slots:
 
    virtual void onSelectItem() override;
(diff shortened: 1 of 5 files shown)


macumber:
Approved

Issue #3121, opened 2018-05-05
gbXML Reverse Translator does not import UndergroundWall from gbXML


chriswmackey:
Hello,

I am 99% positive that this is a small bug that should be relatively easy to fix.  

The OpenStudio gbXML translators have generally been a big help with preserving the geometry and boundary conditions across translation.  However, there is one boundary condition that does not seem to be working correctly in the gbXML Reverse Translator and this is the UndergroundWall.  The UndergroundWall is working correctly in the Forward Translator and I have noticed that other underground constructions like UndergroundFloors are being imported correctly in the Reverse Translator.

Here is a simple single-zone gbXML file with UndergroundWalls that you can run through the gbXML Reverse Translator to recreate the issue:
https://drive.google.com/file/d/1tsmCshs2PbNcbmg_ZN9JiE_TmzvzMHyw/view?usp=sharing

Thanks, as always, for the wonderful software!

PR #3285, opened 2018-09-25
Set parameters for underground walls

jasondegraw:
Addresses #3121 (and partially at least one other issue). Adds support for handling underground walls. This PR is currently unit test free, a condition that may persist.

### DIFF
--- a/openstudiocore/src/gbxml/ReverseTranslator.cpp
+++ b/openstudiocore/src/gbxml/ReverseTranslator.cpp
@@ -623,6 +623,9 @@ namespace gbxml {
         surface.setSurfaceType("Wall");
       }else if (surfaceType.contains("UndergroundWall")){
         surface.setSurfaceType("Wall");
+        surface.setOutsideBoundaryCondition("Ground");
+        surface.setSunExposure("NoSun");
+        surface.setWindExposure("NoWind");
       // roof types
       }else if (surfaceType.contains("Roof")){
         surface.setSurfaceType("RoofCeiling");


openstudiocore/resources/gbxml/TestCube.xml
```diff
@@ -0,0 +1,6335 @@
...
+                <Coordinate>0</Coordinate>
+              </CartesianPoint>
+              <CartesianPoint>
+                <Coordinate>1</Coordinate>
+                <Coordinate>-0.000000000000003224601</Coordinate>
+                <Coordinate>1</Coordinate>
+              </CartesianPoint>
+              <CartesianPoint>
+                <Coordinate>0.000000000000001766908</Coordinate>
+                <Coordinate>0</C
```
macumber: @jasondegraw did you add these surfaceType attributes by hand?  Are there programs writing these?

macumber: @jasondegraw did you get this from Noah?

openstudiocore/src/gbxml/ReverseTranslator.cpp
```diff
@@ -86,7 +86,7 @@ namespace gbxml {
   }
 
   ReverseTranslator::ReverseTranslator()
-    : m_lengthMultiplier(1.0)
+    : m_nonBaseMultiplier(1.0), m_lengthMultiplier(1.0)
```
macumber: @jasondegraw what is difference between m_nonBaseMultiplier and m_lengthMultiplier?

openstudiocore/src/gbxml/ReverseTranslator.cpp
```diff
@@ -192,17 +192,21 @@ namespace gbxml {
...
+      m_nonBaseMultiplier = 1.0e-3;
+      m_lengthUnit = UnitFactory::instance().createUnit("K").get();
     }else if (lengthUnit.contains("Meters", Qt::CaseInsensitive)){
       m_lengthUnit = UnitFactory::instance().createUnit("m").get();
     }else if (lengthUnit.contains("Miles", Qt::CaseInsensitive)){
       m_lengthUnit = UnitFactory::instance().createUnit("mi").get();
     }else if (lengthUnit.contains("Yards", Qt::CaseInsensitive)){
-      //m_lengthUnit = UnitFactory::instance().createUnit("m").get();
+      m_nonBaseMultiplier = 3.0;
+      m_lengthUnit = UnitFactory::instance().createUnit("ft").get();
```
macumber: @jasondegraw is this because os converter can't handle yards?

openstudiocore/src/gbxml/ReverseTranslator.cpp
```diff
@@ -725,26 +732,95 @@ namespace gbxml {
...
-          if (adjacentSpace){
-            // DLM: we have issues if interior ceilings/floors are mislabeled, override surface type for adjacent surfaces
-            // http://code.google.com/p/cbecc/issues/detail?id=471
+          if (adjacentSpace) {
+
             std::string currentSurfaceType = surface.surfaceType();
-            surface.assignDefaultSurfaceType();
-            if (currentSurfaceType != surface.surfaceType()){
-              LOG(Warn, "Changing surface type from '" << currentSurfaceType << "' to '" << surface.surfaceType() << "' for surface '" << surface.name().get() << "'");
+            if (currentSurfaceType == "RoofCeiling" || currentSurfaceType == "Floor") {
```
macumber: @jasondegraw are we writing these attributes when we convert from OSM to gbXML?

jasondegraw: No, we're not. I'll fix that.

openstudiocore/src/gbxml/ReverseTranslator.cpp
jasondegraw: It accounts for the fact that yards apparently aren't covered in the unit factory, we probably could do something special only for the ones that aren't and use the ones that are, but to be consistent I went this way. It probably doesn't really need to be kept around, but I did so just in case.

openstudiocore/src/gbxml/ReverseTranslator.cpp
jasondegraw: Yes.

Issue #2841, opened 2017-11-09
gbXML import mistakes


ngerasim:
Please see the detailed report [here](https://unmethours.com/question/27690/os-doesnt-assign-construction-to-reversed-surfaces/). The test gbXML in the attachment
[TestCube.zip](https://github.com/NREL/OpenStudio/files/1457877/TestCube.zip)

macumber:
Additional things to check https://unmethours.com/question/27690/os-doesnt-assign-construction-to-reversed-surfaces/ https://unmethours.com/question/28251/revit-gbxml-to-openstudio/ https://unmethours.com/question/18258/problems-converting-gbxml-to-idf/

jasondegraw:
The reported issues here are:

 - [ ] OS does not recognize underground walls. (#3121)
 - [ ] In SketchUp there are no constructions assigned to generated reversed surfaces
 - [ ] Some of surfaces have constructions with unreadable names. This one is essentially expected behavior, the gbXML file has multiple constructions with the same name.
 - [ ] In OS application horizontal surfaces are recognized incorrectly, and as a consequence, constructions for horizontal surfaces are also wrong. (#2992)
 - [ ] For adjacent spaces, constructions for some reversed surfaces are not assigned

Issue #2423, opened 2016-12-29
Import gbXML doesn't import construction for subsurfaces


mostaphaRoudsari:
Both fenestration surfaces and shading surfaces are imported with no construction.

![image](https://cloud.githubusercontent.com/assets/2915573/21536123/6b6daf0c-cd4c-11e6-8b05-37fa668a20db.png)

I used OpenStudio gbXML exporter to create the file, and it might be an issue with the exporter not writing the constructions to xml correctly.

macumber:
Thanks @mostaphaRoudsari can you post a gbXML file here?  I had done some work with window constructions but there are several different ways they can be represented in gbXML

mostaphaRoudsari:
Here it is: [seb.txt](https://github.com/NREL/OpenStudio/files/678116/seb.txt). I renamed it to txt to be able to upload it.

DavidGoldwasser:
@mostaphaRoudsari I was able to import gbXML with sub-surfaces with constructions and materials. When I tried your gbXML there was a construction associated with the sub-surface, but it doesn't have any materials. Looking them side by side below, maybe the missing visible Transmittance value on your window is creating the issue?

As a note, about mine on the right, it comes in as simple glazing, using the first of the SolarHeatGainCoeff values.

<img width="1622" alt="screen shot 2018-08-16 at 2 22 52 pm" src="https://user-images.githubusercontent.com/4857055/44233238-854abc00-a160-11e8-8f54-cd8bc64050dd.png">

jasondegraw:
I was not able to reproduce this issue, and there is now a passing unit test for the ReverseTranslator:

https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/gbxml/Test/ReverseTranslator_GTest.cpp#L154

It's named `ReverseTranslator_SubSurfaceConstructions` for when the link goes stale.

mostaphaRoudsari:
Thanks @jasondegraw and @DavidGoldwasser. I haven't checked gbXML importer recently. @chriswmackey do we have any issues related to subsurfaces?

PR #3290, opened 2018-10-04
gbXML Construction Imports

jasondegraw:
Add two unit tests that indicated that the constructions are being imported from gbXML. The error reports were complicated by some naming issues of the materials, which may have cause issues. Addresses #2423 and the rest of #2841.


openstudiocore/src/gbxml/ReverseTranslator.cpp
```diff
@@ -725,26 +732,95 @@ namespace gbxml {
...
+                  } else if (spaceSurfaceType == "InteriorFloor") {
+                    // Swap roles of space and adjacentSpace
+                    surface.setSpace(*adjacentSpace);
+                    auto temp = space;
+                    space = adjacentSpace;
+                    adjacentSpace = temp;
+                    figuredOut = true;
+                  }
+                }
+              } else if (!adjacentSpaceSurfaceType.isEmpty()) {
```
macumber: @jasondegraw is this block ever going to be hit?  the first attribute is empty but the second is set?  if so, should this logic be reversed?  e.g. the current surface is a floor which is ok if the adjacent surface is a ceiling?

openstudiocore/src/gbxml/ReverseTranslator.cpp
```diff
@@ -623,6 +627,9 @@ namespace gbxml {
         surface.setSurfaceType("Wall");
       }else if (surfaceType.contains("UndergroundWall")){
         surface.setSurfaceType("Wall");
+        surface.setOutsideBoundaryCondition("Ground");
+        surface.setSunExposure("NoSun");
+        surface.setWindExposure("NoWind");
```
macumber: @jasondegraw should we set boundary conditions for all cases?  how about for the other underground surfaces 'UndergroundCeiling', 'UndergroundSlab', 'SlabOnGrade', 'RaisedFloor', 'ExposedFloor'?  It looks like explicit boundary conditions from the xml would override these defaults on line 655 which is correct

openstudiocore/src/gbxml/Test/ReverseTranslator_GTest.cpp
```diff
@@ -93,6 +100,302 @@ TEST_F(gbXMLFixture, ReverseTranslator_ZNETH)
...
+
+  auto surfs = model->getModelObjects<Surface>();
+
+  OptionalSurface osurf = model->getModelObjectByName<Surface>("T-1-5-I-F-6");
+  ASSERT_TRUE(osurf);
+  auto points = osurf->vertices();
+  ASSERT_EQ(4, points.size());
+  EXPECT_EQ(0.9144, points[1].y());
+
+  //auto omat = model->getModelObjectByName<StandardOpaqueMaterial>("Concrete: 100 [mm]");
```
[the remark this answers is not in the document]
jasondegraw: @macumber Once someone can get the unit conversion stuff to work, then the "alternate units" test file would allow these to be uncommented (if I did the manual conversions correctly). Right now these will all fail because the none of the material properties get converted and SI is assumed. So it is waiting on more fixes.

openstudiocore/src/gbxml/ReverseTranslator.cpp
jasondegraw: @macumber I'm not sure why this is showing here as a diff, because it's in develop already. In any case, it could be hit if the surface type isn't specified for both spaces, and I think we've agreed we'd prefer to see that specified for both spaces. I think you're right, the logic needs to be reversed (due the space role swap if currentSurfaceType and adjacentSurfaceType agree).

openstudiocore/src/gbxml/ReverseTranslator.cpp
```diff
@@ -725,26 +732,95 @@ namespace gbxml {
...
-          if (adjacentSpace){
-            // DLM: we have issues if interior ceilings/floors are mislabeled, override surface type for adjacent surfaces
-            // http://code.google.com/p/cbecc/issues/detail?id=471
+          if (adjacentSpace) {
+
             std::string currentSurfaceType = surface.surfaceType();
-            surface.assignDefaultSurfaceType();
-            if (currentSurfaceType != surface.surfaceType()){
-              LOG(Warn, "Changing surface type from '" << currentSurfaceType << "' to '" << surface.surfaceType() << "' for surface '" << surface.name().get() << "'");
+            if (currentSurfaceType == "RoofCeiling" || currentSurfaceType == "Floor") {
```
macumber: @jasondegraw If the surfaceType attributes are set those should override the default surface type even if it is not a roof or a floor right?

PR #3174, opened 2018-07-05
Fuel cell fix

brianlball:
Adds ElectricLoadCenterDistribution to all Generator Constructors. Without, causes crash in E+ (https://github.com/NREL/EnergyPlus/issues/6789)

Blocker for FuelCell project.

All relevant tests pass (including resource tests)

these unrelated tests failed:
model tests:
[ FAILED ] ModelFixture.CoilHeatingDesuperheater_addToNode
Energyplus tests:
[ FAILED ] EnergyPlusFixture.ErrorFile_WarningsAndSevere
[ FAILED ] EnergyPlusFixture.ErrorFile_WarningsAndCrash
[ FAILED ] EnergyPlusFixture.ForwardTranslator_ScheduleFixedInterval_Hourly_Shifted
[ FAILED ] EnergyPlusFixture.ForwardTranslator_ScheduleVariableInterval_Hourly_Shifted
[ FAILED ] EnergyPlusFixture.ForwardTranslator_ScheduleVariableInterval_500_Shifted
[ FAILED ] EnergyPlusFixture.ForwardTranslator_ScheduleVariableInterval_DaysTimeSeries

PR #3306, opened 2018-10-10
Fix generator fuel cell and child objects APIs

jmarrec:
Because `GeneratorFuelSupply` doesn't seem to really support `LiquidGeneric` option (cf: https://github.com/NREL/EnergyPlus/issues/6998), I started touching the Ctor to modify it. 

I realized that the API for `GeneratorFuelSupply` had some problems. So I started fixing it, but realized that it's also the same for pretty much all child objects of `GeneratorFuelCell`

**Changelog:**

* Remove some publicly exposed methods in FuelSupply and AirSupply (the set/reset Number of constituents, which is automatically handled by add/removeConstituent)
    * I didn't want to induce an IDD change, but really the field "Number of constituents" should even be present in the IDD to begin with. We can calculate that no problem with the number of current extensible groups
* For  `GeneratorFuellCellAirSupply` and `GeneratorFuelSupply`: 
    * Created two simple classes, `AirSupplyConstituent` and `FuelSupplyConstituent`, so they are exposed in SWIG correctly. Their Ctor will throw is the ConstituentName isn't valid by IDD choice field. addConstituent will now create an actual constituent first even if you call the `addConstituent(std::string, double)` convenience method, so that you know if the constituent name is wrong.
    * Also added a `double sumofConstituentsMolarFractions()`, and use that in respective Forward Translator to issue a warning if  you have some constituents, but the sum isn't equal to 1 (which E+ logically expects)
* I ran into https://github.com/NREL/OpenStudio/pull/3174 again, only this time I realized the `clone` method for Fuel Cell child objects is also very weird (if you clone a `GeneratorFuelSupply`, it tries to clone the parent Fuel Cell first then return the clone's fuel supply).
     * For OS App related reasons, the only object that should clone its parent FuelCell is the heat recovery module (a `StraightComponent`) so it can be dragged from the library onto a plant loop. So I removed the custom clone methods for all other child objects of FuelCell
* All of the fuel cell child objects have this method:
```
  // Return optional parent generator
  GeneratorFuelCell fuelCell() const;
```
Obviously that should really be returning a boost::optional<GeneratorFuelCell>, as the comment also indicates, so I did that. `FuelCellElectricalStorage` was missing this method so I added it. And for all, I use `getModelObjectSources` now, and issue a Warn if more than one fuel cell references the child object (and return the first one)
* Updated Forward Translator and google tests to match

**Minor API breaking change:**
* constituents now returns a std::vector<AirSupplyConstituent> (or FuelSupplyConstituent) instead of `std::vector< std::pair<std::string, double> >`

**Not done (yet) because it would break API / not sure it's needed:**

* In `GeneratorFuellCellAirSupply` and `GeneratorFuelSupply`, Public methods: `removeConstituent` should return bool, the number of components one should be a non optional unsigned int (I did do that for the Impl side of things)
* The child objects of FuelCell that have curves etc are `ModelObjects` but do implement the `allowableChildTypes` and `children` methods, I wonder if they shouldn't be `ParentObjects` though.

Review assignee: @kbenne probably


jmarrec:
Ran the regression tests with https://github.com/NREL/OpenStudio-resources/pull/59, all tests that usually pass do pass (lifecyclecostparameters ones have been broken since E+ 8.9.0, and intersect_136.osm will timeout as usual)

brianlball:
This PR really scares me.  The clone methods were originally done in such a way so that the HX would work in the GUI and then pull in all the other FuelCell objects.

jmarrec:
Like I said, the HX is the only one that needs to clone the parent FuelCell, the other do not.
The FuelCell is responsible for cloning its children, which I didn't touch.

macumber:
@brianlball first thing is this PR will not be merged for 2.7.0.  However, after that, I think my general preference is to the make the model API clean and consistent with other objects while adding any ugly workarounds to the Application code directly.  That seems like a better solution than adding workarounds to support the GUI directly in the model objects.

jmarrec:
@brianlball thanks for reminding me to test in OS App, I found a typo that I made in the clone method for the heat exchanger. Now I can confirm that it still works exactly the same way it used to in OS app.

Here's ruby code to test that it works too:
```
[1] OS-build(main)> m = Model.new
=> #<OpenStudio::Model::Model:0x00000004099ad0 @__swigtype__="_p_openstudio__model__Model">
[2] OS-build(main)> lib = osload('/home/julien/Software/Others/OpenStudio/openstudiocore/src/openstudio_app/Resources/default/hvac_library.osm')
=> #<OpenStudio::Model::Model:0x00000004071dc8 @__swigtype__="_p_openstudio__model__Model">
[3] OS-build(main)> hx = lib.getGeneratorFuelCellExhaustGasToWaterHeatExchangers[0]
=> #<OpenStudio::Model::GeneratorFuelCellExhaustGasToWaterHeatExchanger:0x00000003b64ab0
 @__swigtype__="_p_openstudio__model__GeneratorFuelCellExhaustGasToWaterHeatExchanger">
[4] OS-build(main)> hxClone = hx.clone(m).to_GeneratorFuelCellExhaustGasToWaterHeatExchanger.get
=> #<OpenStudio::Model::ModelObject:0x000000037393a0 @__swigtype__="_p_openstudio__model__ModelObject">

[5] OS-build(main)> puts m.getGeneratorFuelCells[0]
OS:Generator:FuelCell,
  {2622bcf3-5663-4123-b49f-806829cd4aa6}, !- Handle
  Generator Fuel Cell 1,                  !- Name
  {66ffe50e-11c1-487b-b50b-bd87d25b5b50}, !- Power Module Name
  {b9832514-24a2-41f7-bedb-283fc3de1d87}, !- Air Supply Name
  {7e17efda-b1bf-4ce4-a9b7-00f0e5036b03}, !- Fuel Supply Name
  {9b08136c-2da8-4b3d-9e54-e1b3f6dd4e26}, !- Water Supply Name
  {f1acb947-24d9-4b6d-ba4b-a48dd3d43426}, !- Auxiliary Heater Name
  {7669bc99-1845-4cee-bf32-47777ff620ae}, !- Heat Exchanger Name
  {8a82fc27-4513-4b60-8df9-7447b4d51a1a}, !- Electrical Storage N
```

brianlball:
After our email convo I think this is good to go.
@macumber @jmarrec

PR #3310, opened 2018-10-12
Update libraries to 270

jmarrec:
I've parallelized the `UpdateHVACLibrary.rb`

You'll need
```
gem install parallel
gem install ruby-progressbar
```

And I've run it. 

But I'm in the process of re-regenerating the openstudio-standards OSM files using https://github.com/NREL/openstudio-standards/blob/master/data/standards/export_OpenStudio_libraries.rb because I think they are way outdated

@macumber


jmarrec:
Running the script took 19 hours, even parallelized. DEER 2014, 2015 and 2017, and DEER_pre_1975, and the two DOE_REF (DOE_Ref_1980_2004 and DOE Ref Pr 1980) didn't get generated, not sure why/if normal.

I committed the the ones that did get generated

PR #3319, opened 2018-10-19
Adding licenses for Spider gbXML Tools and ThreeJS which are distribu…

macumber:
…ted with the OpenStudio Application


theo-armour:
@macumber 

Thank you.

PR #3323, opened 2018-10-25
Encoding tests

macumber:
Updates SWIG to 3.0.12 and applies the following patch:

https://github.com/macumber/swig/commit/45ffaf776c06d7f3bb64ac75e6dab4a9f733b414

This assumes strings coming from C++ to SWIG are UTF-8 encoded

### DIFF
--- a/openstudiocore/src/cli/test/test_encodings.rb
+++ b/openstudiocore/src/cli/test/test_encodings.rb
@@ -27,4 +27,27 @@ class Encoding_Test < Minitest::Test
     assert_equal("UTF-8", path.to_s.encoding.to_s)
   end   
   
+  def test_encoding2
+    test_string = "模型"
+    assert_equal("UTF-8", test_string.encoding.to_s)
+    
+    model = OpenStudio::Model::Model.new
+    space = OpenStudio::Model::Space.new(model)
+    
+    space.setName(test_string)
+    
+    name_string = space.nameString
+    assert_equal(test_string, name_string)
+    assert_equal("UTF-8", name_string.encoding.to_s)
+    
+    optional_string = space.name
+    assert(optional_string.is_initialized)
+    assert_equal(test_string, optional_string.get)
+    assert_equal("UTF-8", optional_string.get.encoding.to_s)
+    
+    path = OpenStudio::toPath(test_string)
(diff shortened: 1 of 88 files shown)


macumber:
@kbenne @jmarrec would be great if you guys can test this out on Mac/Unix (I have not tried that yet)

openstudiocore/src/cli/test/test_encodings.rb
```diff
@@ -0,0 +1,53 @@
...
+    assert(optional_string.is_initialized)
+    assert_equal(test_string, optional_string.get)
+    assert_equal("UTF-8", optional_string.get.encoding.to_s)
+    
+    path = OpenStudio::toPath(test_string)
+    assert_equal(test_string, path.to_s)
+    assert_equal("UTF-8", path.to_s.encoding.to_s)
+  end   
+  
+  def test_encoding2
```
macumber: @pflaumingo tests pass after these changes!

jmarrec:
I'm building and will test shortly after. Would be nice to add the cli tests to Cmake too.

jmarrec:
I got a build error, relaunching appears to make it disappear. But if I head to either `OS-build/SWIG-prefix/src/SWIG` or `OS-build/SWIG-prefix/src/SWIG-install/bin` and type `swig -v` I get 

```
SWIG Version 3.0.8

Compiled with g++ [x86_64-pc-linux-gnu]

Configured options: +pcre
```

I'll try a clean build now.

jmarrec:
Ok, test pass on Unix too.

I got SWIG Version 3.0.8 from the command line because that's my system installed one... If I correctly do

```
cd OS-build/SWIG-prefix/src/SWIG-install/bin
./swig -version
=> SWIG Version 4.0.0
```

Though I do have only OS-build/SWIG-prefix/src/swig-3.0.12.tar.gz

macumber:
Thanks julien I think I know what I did wrong, I’ll fix it

macumber:
@jmarrec this is fixed now, probably have to do a clean build to test since the generated swig files aren't dependent on swig

jmarrec:
I didn’t follow one bit what’s going on, but if swig 4.0 appears to be working then why not go with that? (A very naive question probably)

PR #3340, opened 2018-11-06
Remove QRegExp and QRegularExpression

jmarrec:
## Remove QRegularExpression and QRegExp dependencies

Replaced with `boost::regex`.

I think I caught every location that isn't OS App related
```bash
$ cd OpenStudio/openstudiocore
$ grep -Rl "QReg[uE]" 
model_editor/InspectorGadget.cpp
model_editor/IGLineEdit.cpp
model_editor/IGPrecisionDialog.cpp
openstudio_lib/FacilityShadingGridView.cpp
openstudio_lib/FacilityStoriesGridView.cpp
openstudio_lib/ResultsTabView.cpp
openstudio_lib/SpacesSubtabGridView.cpp
shared_gui_components/EditView.cpp
```

Note: I left two "TODO" inside BCLMeasure.cpp because boost::replace_all isn't working in my current setting (I use a much newer GCC than the one that compiled boost), but it should work fine after update_depends is merged in (on my system, I have a test project with boost 1.68 compiled with GCC-7 and this works)

### DIFF
--- a/openstudiocore/src/utilities/bcl/BCLMeasure.cpp
+++ b/openstudiocore/src/utilities/bcl/BCLMeasure.cpp
@@ -40,8 +40,8 @@
 
 #include <QDir>
 #include <QSettings>
-#include <QRegularExpression>
-
+#include <boost/regex.hpp>
+#include <boost/algorithm/string/replace.hpp>
 
 #include <src/utilities/embedded_files.hxx>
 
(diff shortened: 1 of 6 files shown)


macumber:
Changed base branch to update_depends because there are C++/boost updates necessary to review

jmarrec:
Just an FYI, technically speaking it doesn't _really_ need update depends. 

`boost::replace_all` should work just fine in current too as long as you're using the same GCC as boost was compiled with. In other terms, currently in develop it would work with ubuntu 16.04, but for 18.04 I think default gcc is version 7, in which case it wouldn't work. But just need a check on LSB_RELEASE and supplying boost 1.55 compiled with gcc 7 would work too.

jmarrec:
Better yet:
`if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 7)`

PR #3343, opened 2018-11-08
Add ITE object

KaiyuSun:
## Add the ElectricEquipmentITEAirCooled object.

Updates include:
(1) header file, impl-header file, source file of the main object and definition
(2) Unit test on basic functions of the main object and definition
(3) Forward translator
(4) Unit test on basic functions and constraints of forward translator
(5) Changes on related objects such as space, spacetype, thermal zone, etc.


openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipmentITEAirCooled.cpp
```diff
@@ -0,0 +1,265 @@
...
+        if (terminal) {
+          if (!terminal->optionalCast<AirTerminalSingleDuctVAVReheat>() && !terminal->optionalCast<AirTerminalSingleDuctVAVNoReheat>()) {
+            LOG(Warn, modelObject.briefDescription() << " The FlowControlWithApproachTemperatures only applies to ITE zones with single duct VAV terminal unit.");
+          }
+        }
+      }
+    }
+
+  } else {
+    LOG(Warn, modelObject.briefDescription() << " not assigned to a valid space.");
```
macumber: Again is it best to not translate the object or let the simulation fail?  People may ignore a warning if the simulation does not error out......  @kbenne you have thoughts on this?

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipmentITEAirCooled.cpp
```diff
@@ -0,0 +1,265 @@
...
+  if (d) {
+    idfObject.setDouble(ElectricEquipment_ITE_AirCooledFields::DesignFanAirFlowRateperPowerInput, *d);
+  }
+
+  idfObject.setString(ElectricEquipment_ITE_AirCooledFields::AirFlowFunctionofLoadingandAirTemperatureCurveName, definition.airFlowFunctionofLoadingandAirTemperatureCurve().name().get());
+
+  idfObject.setString(ElectricEquipment_ITE_AirCooledFields::FanPowerInputFunctionofFlowCurveName, definition.fanPowerInputFunctionofFlowCurve().name().get());
+
+  if (!definition.isDesignEnteringAirTemperatureDefaulted()) {
+    idfObject.setDouble(ElectricEquipment_ITE_AirCooledFields::DesignEnteringAirTemperature, definition.designEnteringAirTemperature());
```
macumber: Our new standard is to write the OS value to IDF even if it is defaulted.  This covers the case where OS and E+ have different defaults.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateSpaceType.cpp
```diff
@@ -155,6 +159,13 @@ boost::optional<IdfObject> ForwardTranslator::translateSpaceType( SpaceType & mo
     translateAndMapModelObject(equipment);
   }
 
+  // translate IT electric equipment
+  ElectricEquipmentITEAirCooledVector electricEquipmentITEAirCooled = modelObject.electricEquipmentITEAirCooled();
```
macumber: There won't be any assigned to the space type right?

openstudiocore/src/energyplus/Test/ElectricEquipmentITEAirCooled_GTest.cpp
```diff
@@ -0,0 +1,477 @@
...
+  Model model;
+  ElectricEquipmentITEAirCooledDefinition definition(model);
+  ElectricEquipmentITEAirCooled electricEquipmentITEAirCooled(definition);
+
+  ForwardTranslator forwardTranslator;
+  Workspace workspace = forwardTranslator.translateModel(model);
+
+  EXPECT_EQ(0u, workspace.getObjectsByType(IddObjectType::ElectricEquipment_ITE_AirCooled).size());
+
+  model.save(toPath("./ITE_translator_NoSpace.osm"), true);
```
macumber: @KaiyuSun comment this out after your are done using it

openstudiocore/src/model/ElectricEquipmentITEAirCooled.cpp
```diff
@@ -0,0 +1,560 @@
...
+  }
+
+  boost::optional<Schedule> ElectricEquipmentITEAirCooled_Impl::designPowerInputSchedule() const {
+    boost::optional<Schedule> result = getObject<ModelObject>().getModelObjectTarget<Schedule>(OS_ElectricEquipment_ITE_AirCooledFields::DesignPowerInputScheduleName);
+    if (!result) {
+      // search upwards
+      OptionalSpace space = this->space();
+      OptionalSpaceType spaceType = this->spaceType();
+      if (space) {
+        result = space->getDefaultSchedule(DefaultScheduleType::DesignPowerInputScheduleName);
```
macumber: Was this added to the DefaultScheduleType object?  I didn't see that change in the IDD.

macumber: This schedule represents adding new servers over time right?  Seems kind of detailed....

macumber: The default is a schedule that is always 1?

openstudiocore/src/model/ElectricEquipmentITEAirCooled.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+  std::string electricPowerSupplyEndUseSubcategory() const;
+
+  bool isElectricPowerSupplyEndUseSubcategoryDefaulted() const;
+
+  //@}
+  /** @name Setters */
+  //@{
+
+  // TODO: Check argument type. From object lists, some candidates are: ElectricEquipmentITEAirCooledDefinition.
```
macumber: Might want to clean up these auto generated comments

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipmentITEAirCooled.cpp
```diff
@@ -0,0 +1,265 @@
...
+        if (auto node = mo->optionalCast<Node>()) {
+          idfObject.setString(ElectricEquipment_ITE_AirCooledFields::SupplyAirNodeName, node->name().get());
+        }
+      } else {
+        // if no airloop, just get a supply node of the thermal zone (could be zoneHVAC)
+        if (thermalZone->inletPortList().modelObjects().size() > 0) {
+          std::vector<ModelObject> objects = thermalZone->inletPortList().modelObjects();
+          for (const auto & elem : objects) {
+            if (auto node = elem.optionalCast<Node>()) {
+              idfObject.setString(ElectricEquipment_ITE_AirCooledFields::SupplyAirNodeName, node->name().get());
```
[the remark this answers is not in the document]
KaiyuSun: There is supposed to be only one single node for this field. I double checked with my colleague Xuan, who implemented part of the ITE object in EnergyPlus. The source code is:
           ZoneITEq(Loop).SupplyAirNodeNum = GetOnlySingleNode(
                        AlphaName(14), ErrorsFound, CurrentModuleObject, AlphaName(1), 
                        NodeType_Air, NodeConnectionType_Sensor, 1, ObjectIsNotParent);
Where only single node is allowed in this field, node list is not allowed.

openstudiocore/src/energyplus/ForwardTranslator.cpp
```diff
@@ -266,6 +266,26 @@ Workspace ForwardTranslator::translateModelPrivate( model::Model & model, bool f
     }
   }
 
+  // Energyplus only allows single zone input for ITE object. If space type is assigned in OS, 
+  // will translate to multiple ITE objects assigned to each zone under the same space type.
+  // then delete the one that pointed to a spacetype.
+  // By doing this, we can solve the potential problem that if this load is applied to a space type, 
+  // the load gets copied to each space of the space type, which may cause conflict of supply air node.
+  std::vector<ElectricEquipmentITEAirCooled> iTEAirCooledEquipments = model.getConcreteModelObjects<ElectricEquipmentITEAirCooled>();
```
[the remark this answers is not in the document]
KaiyuSun: Not really. I've tried using the existing code but it doesn't include what this part does. CombineSpace will merge ITE object of each space under the same **ThermalZone** into single ITE object, while this part distribute ITE object associated with **SpaceType** to each space under it, which will generate multiple ITE objects. They function differently.

It is mainly because ITE object can be assigned to a spacetype in OpenStudio but can't be assigned to a zonelist in EnergyPlus.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipmentITEAirCooled.cpp
```diff
@@ -0,0 +1,265 @@
...
+  // ITE object that is assigned to space type in OS has been switched to separate spaces in ForwardTranslator
+  boost::optional<Space> space = modelObject.space();
+  if (space) {
+    boost::optional<ThermalZone> thermalZone = space->thermalZone();
+    if (thermalZone) {
+      // Assign object to zone
+      idfObject.setString(ElectricEquipment_ITE_AirCooledFields::ZoneName, thermalZone->name().get());
+      // attach the supply air node to zone if there is an available supply air node
+      // search airloop first
+      if (auto mo = thermalZone->inletPortList().airLoopHVACModelObject()) {
```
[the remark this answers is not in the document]
KaiyuSun: Good point. I'll remove the field from definition.

macumber:
@KaiyuSun the next step will be to add integration tests at https://github.com/NREL/OpenStudio-resources

You will need to add a ruby test to test the ruby bindings, https://github.com/NREL/OpenStudio-resources/blob/develop/model/simulationtests/air_terminals.rb

And an osm test to test future version translation:
https://github.com/NREL/OpenStudio-resources/blob/develop/model/simulationtests/air_terminals.osm

The tests are added to this file:
https://github.com/NREL/OpenStudio-resources/blob/develop/model_tests.rb#L891

And can be run as described in:
https://github.com/NREL/OpenStudio-resources/blob/develop/README.md

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipmentITEAirCooled.cpp
```diff
@@ -0,0 +1,265 @@
...
+      }
+    }
+
+    // Contraint #2:The FlowControlWithApproachTemperatures only applies to ITE zones with single duct VAV terminal unit.
+    if (istringEqual(thisMethod, "FlowControlWithApproachTemperatures")) {
+      if (thermalZone) {
+        auto terminal = thermalZone->airLoopHVACTerminal();
+        if (terminal) {
+          if (!terminal->optionalCast<AirTerminalSingleDuctVAVReheat>() && !terminal->optionalCast<AirTerminalSingleDuctVAVNoReheat>()) {
+            LOG(Warn, modelObject.briefDescription() << " The FlowControlWithApproachTemperatures only applies to ITE zones with single duct VAV terminal unit.");
```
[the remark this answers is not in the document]
KaiyuSun: Yes, this will crash in E+. To not translate the object, should I first Log Warn and then return boost::none?

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipmentITEAirCooled.cpp
```diff
@@ -0,0 +1,265 @@
...
+  for (LifeCycleCost lifeCycleCost : modelObject.lifeCycleCosts()){
+    translateAndMapModelObject(lifeCycleCost);
+  }
+
+  ElectricEquipmentITEAirCooledDefinition definition = modelObject.electricEquipmentITEAirCooledDefinition();
+
+  // Pre-process: 
+  // (1) assign to zone; (2) connect the supply air node (3) apply contraints
+  // ITE object that is assigned to space type in OS has been switched to separate spaces in ForwardTranslator
+  boost::optional<Space> space = modelObject.space();
```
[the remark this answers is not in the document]
KaiyuSun: I thought about this too. Let alone connecting the node, E+ will crash when no space is assigned to the object. However, I checked ForwardTranslateLights for reference, which didn't assert that either a space or spacetype has to be initialized and associated.
[https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateLights.cpp#L65](url)  So I assume you let this pass to E+ error? Not sure if I misunderstand, please correct me if I'm wrong.

Now about connecting the node, a space is needed for node extraction ONLY if the Calculation Method = "FlowControlWithApproachTemperatures" or if the "Air Inlet Connection Type" is "AdjustedSupply". Otherwise, however, the object can proceed without a space. But eventually E+ will still crash if no space is assigned to the object.

If it is true that you are letting the "no space assignment" pass to E+ error, I guess I should not add assert here.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipmentITEAirCooled.cpp
KaiyuSun: It seems the lights translator let the simulation fail when neither a space nor a spacetype is associated.
[https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateLights.cpp#L65](url)

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateSpaceType.cpp
KaiyuSun: ITE object **can** be assigned to space type in OS, but will be separately assigned to each space when being translated to E+.

openstudiocore/src/energyplus/Test/ElectricEquipmentITEAirCooled_GTest.cpp
KaiyuSun: Do you mean all model.save and workspace.save?

openstudiocore/src/model/ElectricEquipmentITEAirCooled.cpp
KaiyuSun: Basically there are two layers of schedules originally designed for this object: first layer is the fraction of equipment that is running among all equipment (DesignPowerInputSchedule), second layer is the CPU load fraction of the running equipment (CPULoadingSchedule). Their default are both always 1.

I changed in DefaultScheduleSet.hpp [https://github.com/NREL/OpenStudio/pull/3343/files#diff-03f5b453c4044b5abceabd9f90492cf4](url)   Not sure if this is enough. It works for now.

openstudiocore/src/energyplus/Test/ElectricEquipmentITEAirCooled_GTest.cpp
KaiyuSun: I commented the model.save and workspace.save in the test file.

KaiyuSun:
@macumber Could you share access to the OpenStudio-resources repository? Thanks!

macumber:
@KaiyuSun added you to https://github.com/NREL/OpenStudio-resources/

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipmentITEAirCooled.cpp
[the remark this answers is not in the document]
KaiyuSun: @macumber Any thoughts?

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipmentITEAirCooled.cpp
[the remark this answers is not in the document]
KaiyuSun: @macumber To not translate the object, should I first Log Warn and then return boost::none?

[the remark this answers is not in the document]
macumber: If you don't want to translate the object you should return boost::none AND not call `m_idfObjects.push_back(idfObject);`.   You should only add the object to the m_idfObjects list after you have decided you want to translate it.

PR #3346, opened 2018-11-08
 Remove QJson and related

jmarrec:
## Remove QJson

Replace with `jsoncpp`.

A good side effect is that I embedded the Standards json files instead of relying on the Qt QRC file.

There are some files that are somewhat JSON-related in the sense that they had functions inside Json.hpp (some QVariant to JSON string ones), but that need a much deeper rewrite to I moved these into a `QJson.hpp` for now.

```
src/measure/OSArgument.cpp
src/utilities/data/Attribute.cpp
src/utilities/data/Test/Attribute_GTest.cpp
```


openstudiocore/src/utilities/filetypes/Filetypes.i
```diff
@@ -11,6 +11,7 @@
   #include <utilities/filetypes/WorkflowStep.hpp>
   #include <utilities/filetypes/WorkflowStep_Impl.hpp>
   #include <utilities/filetypes/WorkflowJSON.hpp>
+  #include <utilities/filetypes/StandardsJSON.hpp>
 
   // this is all garbage, I don't know why it is needed, probably comes from quantity
```
jmarrec: Lol

jmarrec:
With the update_depends branch, there are a bunch of deprecated declarations for writer and reader (`Json::StyledWriter`, `Json::FastWriter`, and `Json::Reader`). 

Should we update these @macumber?

openstudiocore/src/measure/OSArgument.cpp
```diff
@@ -38,7 +38,7 @@
 #include "../utilities/core/Assert.hpp"
 #include "../utilities/core/Containers.hpp"
 #include "../utilities/core/Compare.hpp"
-#include "../utilities/core/Json.hpp"
+#include "../utilities/core/QJson.hpp"
```
macumber: @jmarrec is the idea that this file will be removed or will it hang around?  I don't think we should have a file named QJson hanging around for a long time.

openstudiocore/src/model/CMakeLists.txt
```diff
@@ -1603,28 +1606,43 @@ set(${target_name}_swig_src
...
+set(${target_name}_depends
+  openstudio_utilities
 )
 
-# generate rules for building source files from the resources
-# DLM: TODO, replace with embed_files
-qt5_add_resources(${target_name}_qrcs ${${target_name}_qrc})
+# Embed resource files
+# Glob all files inside model/Resources
+file(GLOB_RECURSE embedded_files "${CMAKE_CURRENT_SOURCE_DIR}/Resources/**/*.*")
```
macumber: @jmarrec "resources" is lower case right?

openstudiocore/src/model/SpaceType.cpp
```diff
@@ -282,6 +280,26 @@ namespace detail {
...
+      std::string embedded_path = ":/Resources/standards/OpenStudio_Standards_space_types_merged.json";
+      std::string fileContent = ::openstudio::embedded_files::getFileAsString(embedded_path);
+
+      // Create a StandardsJSON
+      StandardsJSON standard(fileContent);
+
+      // Now try to get the primaryKey
+      std::string primaryKey = "space_types";
+
+      if (boost::optional<Json::Value> _standardsArr = standard.getPrimaryKey(primaryKey)) {
```
macumber: should print a warning or error if not found right?

openstudiocore/src/utilities/core/QJson.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include "FilesystemHelpers.hpp"
+
+#include <OpenStudio.hxx>
+
+#include <QJsonDocument>
+#include <QJsonObject>
+
+namespace openstudio {
+
+QVariant jsonMetadata() {
```
macumber: @jmarrec we will eventually replace QVariant with std::variant (on update_depends because requires c++17).  What happens to this file then?  Should it just stay in Json.hpp?

macumber: If we even need any of these methods anymore.... first thing we should try to do is figure out if any of these can be removed.  I dont think we are using these methods in Ruby

openstudiocore/src/utilities/core/QJson.hpp
```diff
@@ -0,0 +1,184 @@
...
+  for (const QVariant& listItem : list) {
+    QVariant listItemCopy(listItem);
+    T value = typeConverter(&listItemCopy);
+    result.push_back(value);
+  }
+  return result;
+}
+
+template<typename T>
+std::vector<T> deserializeUnorderedVector(const QVariantList& list,
```
macumber: @jmarrec are we using this stuff?  Would be great to just remove

openstudiocore/src/utilities/filetypes/StandardsJSON.hpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+namespace openstudio{
+
+namespace detail{
+  class StandardsJSON_Impl;
+}
+
+/** Class for accessing the OpenStudio Standards (OSW) JSON format. */
+class UTILITIES_API StandardsJSON
```
macumber: @jmarrec Awesome thanks for adding this!

openstudiocore/src/model/CMakeLists.txt
jmarrec: I renamed it to "Resources" to match what's used in other places (eg: src/energyplus/Resources, src/openstudio_app/Resources), so that all "Resources" files currently end up in the same "embedded" folder.

openstudiocore/src/model/SpaceType.cpp
jmarrec: That's not necessary until we implement the ability to use a supplied standards JSON. The primaryKey is harcoded there, and it'll be caught by the GTests if that stops working.

macumber:
I added a comment in the google doc, let's mark code that should be moved out of utilities, etc with `// TODO: QT-Separation-Move`

macumber:
I tagged you with removal of QVariant so you can finish this one off

openstudiocore/src/model/CMakeLists.txt
macumber: Ok, just as an FYI people may need to manually correct this as I don't think git will re-case directories

macumber:
@jmarrec This looks good to me however there are two test failures:

```
1>  The following tests FAILED:
1>  	2616 - OpenStudioAppFixture.Resources_Templates (Failed)
1>  	2617 - OpenStudioAppFixture.Resources_HVACLibrary (Failed)
```

Can you look at these?  It says there is some issue with the 2.7.0 IDD.

jmarrec:
I found the problem @macumber, it's because I embedded the model/Resources files into namespace "openstudio", same as the embedded files of utilities. And it was "hiding" the utilities files, so it couldn't find the idd ones.

macumber:
Nice!

PR #3353, opened 2018-11-17
Remove Qt Threading Dependencies

lefticus:
## Remove Qt Threading Related Dependencies

Because of the interrelated nature of these dependencies, this one PR removes:

* QThread
* QMutex
* QReadWriteMutex
* QWriteLocker
* QMutexLocker
* QThreadPool

Some of them existed only in tests. Some required updates (notably the Logger updates) to libraries outside of the target set that we wanted to eliminate Qt dependencies from.

Each commit should stand independent if it's desired to cherry-pick them for some reason.

## Removal of Mutexes from IddFactory

The necessity for QMutex in IddFactory is a historical remnant from before we had thread safe initialization of `static`s and tools to `const`ant initialize our statics.

The mutexes in this code has been completely removed in favor of more modern constructs. This should result in a considerable runtime improvement on the IddFactory. This change is isolated to 95ff7d6


macumber:
@lefticus this all looks good.  I am going to try to get more tests passing on update_depends.  Then this will be one of the first PRs I merge and test

lefticus:
> 
> 
> @lefticus this all looks good. I am going to try to get more tests passing on update_depends. Then this will be one of the first PRs I merge and test

Great, I've been making sure to maintain parity with existing test success/failures to make sure I'm changing existing behavior as little as possible.

macumber:
@lefticus I am about ready to merge this, only issue I see is that this test is failing 

`EnergyPlusFixture.ForwardTranslatorTest_MultiThreadedLogMessages (Failed)`

I'm not sure if the test is wrong now or what, do you see this failure?  I am on Windows in Release mode

PR #3354, opened 2018-11-19
Remove QSQL

jmarrec:
Remove `QSql`, basically in LocalBCL.cpp.

I did test in OS App and I'm able to drag BCL components and measures from the library.

Last commit is the one that actually removes any trace of Qt5Sql in the various CMakeLists.txt etc.

**Note:** as stated in the 'Qt Dependency Replacement Design' document, I couldn't find a version of OpenStudio that would have an outdated BCL database version (OpenStudio v1.0.0 has db version 1.3 like current) to test that the `updateLocalDb` works correctly.


openstudiocore/CMakeLists.txt
```diff
@@ -963,7 +963,6 @@ endif()
 
 find_package(Qt5Core 5.6.1 REQUIRED PATHS ${QT_INSTALL_DIR} NO_DEFAULT_PATH)
 find_package(Qt5Widgets 5.6.1 REQUIRED PATHS ${QT_INSTALL_DIR} NO_DEFAULT_PATH)
-find_package(Qt5Sql 5.6.1 REQUIRED PATHS ${QT_INSTALL_DIR} NO_DEFAULT_PATH)
```
macumber: Oh yeah!

openstudiocore/resources/model/OpenStudio.idd
```diff
@@ -21801,14 +21801,14 @@ OS:Sizing:System,
        \object-list AirPrimaryLoops
   A3, \field Type of Load to Size On
        \note Specifies the basis for sizing the system supply air flow rate
-       \note Sensible and VentilationRequirement are the only available options
```
macumber: @jmarrec I merged develop into develop3 but this is still showing as a diff.  Are there some other branches mixed in here?  I saw some more port/const stuff below as well.  All that should get merged to develop and develop3.  I wonder if there are some other PRs I need to merge to develop?

openstudiocore/src/model/AirLoopHVAC.cpp
```diff
@@ -1383,6 +1401,21 @@ namespace detail {
     return result;
   }
 
+  boost::optional<Node> AirLoopHVAC_Impl::returnAirNode() const
```
macumber: @jmarrec here is some more stuff, what branch is this from?  is it not already on develop?

jmarrec:
@macumber I rebase the branch onto develop/develop3 and force-pushed. The diff should be clean now.

PR #3355, opened 2018-11-19
Update depends

jmarrec:
The only substantial change here is that on ubuntu 16.04 I have to explicitly locate YAML, the rest is whitespace noise and removing a few build warnings.


openstudiocore/src/model/Space.cpp
```diff
@@ -953,7 +953,7 @@ namespace detail {
       }
     }
 
-    if ((numRoof > 0) * (numFloor > 0)){
```
macumber: Wow, wft was this?

PR #3356, opened 2018-11-19
Remove QTemporaryFile and QDir

lefticus:
## Remove QTemporaryFile and QDir

While removing these two dependencies I also removed dead code that relied on them (as opposed to spending the time updating code that isn't used) and cleaned up some redundant code.

 * `QTemporaryFile` is a clean extraction
 * `QDir` mostly effected BCL.
    * It's possible we'll get merge conflicts with code that is removing `QSql` because I took the opportunity to clean up some redundant code. If this happens I can hopefully sort out the merge conflicts easily.

### DIFF
--- a/openstudiocore/src/model/FileOperations.cpp
+++ b/openstudiocore/src/model/FileOperations.cpp
@@ -79,14 +79,12 @@
 #include "../utilities/core/Logger.hpp"
 #include "../utilities/core/PathHelpers.hpp"
 #include "../utilities/core/Assert.hpp"
+#include "../utilities/core/FilesystemHelpers.hpp"
 
 #include <OpenStudio.hxx>
 
 #include <QDir>
 #include <QMessageBox>
-#include <QTemporaryFile>
-
-
 
 
 namespace openstudio {
@@ -290,31 +288,10 @@ QString longPathName(const QString& path)
 
   openstudio::path createModelTempDir()
   {
-    openstudio::path result;
-    auto tempFile = new QTemporaryFile();
-
-    if (tempFile->open())
-    {
-      QString path = longPathName(tempFile->fileName());
-
-      // must close tempFile so you can create directory with same name on Windows
-      delete tempFile;
-
-      QDir modelTempDir(path);
-      LOG_FREE(Info, "createModelTempDir", "Creating directory '" << toString(modelTempDir.path()) << "'");
-      bool test = modelTempDir.mkpath(modelTempDir.path());
-      OS_ASSERT(test);
-
-      result = toPath(path);
-
-      return result;
-    }
-
-    delete tempFile;
-
-    OS_ASSERT(false);
-
-    return openstudio::path();
+    const auto result = openstudio::filesystem::create_temporary_directory("osmodel");
+    LOG_FREE(Info, "createModelTempDir", "Created directory '" << toString(result) << "'");
+    OS_ASSERT(!result.empty());
+    return result;
   }
 
   bool initializeModelTempDir(const openstudio::path& osmPath, const openstudio::path& modelTempDir)
--- a/openstudiocore/src/model/FileOperations.cpp
+++ b/openstudiocore/src/model/FileOperations.cpp
@@ -92,48 +92,61 @@ namespace openstudio {
 namespace model {
   bool copyDir(const openstudio::path &sourceDir, const openstudio::path &destinationDir)
   {
-    if (!openstudio::filesystem::exists(sourceDir) || !openstudio::filesystem::is_directory(sourceDir))
-    {
-      throw std::runtime_error("Source directory " + sourceDir.string() + " does not exist or is not a directory");
-    }
+    const auto src = openstudio::filesystem::canonical(sourceDir);
+    const auto dest = openstudio::filesystem::canonical(destinationDir);
 
-    if (openstudio::filesystem::exists(destinationDir))
-    {
-      throw std::runtime_error("Destination directory " + destinationDir.string() + " already exists");
-    }
-
-    if (!openstudio::filesystem::create_directory(destinationDir))
-    {
-      throw std::runtime_error("Cannot create destination directory " + destinationDir.string());
+    if (src == dest) {
+      LOG_FREE(Warning, "Refusing to copy directory '" << toString(sourceDir) << "' only itself '" << toString(destinationDir) 
+          << "' canonical path: '" << toString(src) << "'");
+      return true; // there is no error, just nothing to do
     }
 
 
+    if (!openstudio::filesystem::exists(src) || !openstudio::filesystem::is_directory(src))
+    {
+      throw std::runtime_error("Source directory " + src.string() + " does not exist or is not a directory");
+    }
 
-    for (const auto& dirEnt : openstudio::filesystem::recursive_directory_iterator{sourceDir})
+    if (openstudio::filesystem::exists(dest))
+    {
+      throw std::runtime_error("Destination directory " + dest.string() + " already exists");
+    }
+
+    if (!openstudio::filesystem::create_directory(dest))
+    {
+      throw std::runtime_error("Cannot create destination directory " + dest.string());
+    }
+
+
+
+    for (const auto& dirEnt : openstudio::filesystem::recursive_directory_iterator{src})
     {
       const auto& path = dirEnt.path();
       auto relativePathStr = path.string();
-      boost::replace_first(relativePathStr, sourceDir.string(), "");
+      boost::replace_first(relativePathStr, src.string(), "");
 
       try {
-        openstudio::filesystem::copy(path, destinationDir / relativePathStr);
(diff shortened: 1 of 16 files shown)


openstudiocore/src/utilities/bcl/LocalBCL.cpp
```diff
@@ -103,8 +111,8 @@ namespace openstudio{
     if (!ptr) {
       QSettings settings("OpenStudio", "LocalBCL");
```
macumber: @lefticus got any ideas for what to replace QSettings with?

openstudiocore/src/utilities/bcl/LocalBCL.cpp
```diff
@@ -368,12 +380,12 @@ namespace openstudio{
   std::vector<BCLMeasure> LocalBCL::measures() const
   {
     std::vector<BCLMeasure> allMeasures;
-    QSqlDatabase database = QSqlDatabase::database(m_libraryPath+m_dbName);
+    auto database = getDatabase();
```
macumber: @lefticus will this make it harder to find/replace instances of QSqlDatabase?  I guess the intent is to write code that can remain after the refactor?

openstudiocore/src/utilities/core/Filesystem.hpp
```diff
@@ -55,6 +55,7 @@ namespace filesystem {
   using boost::filesystem::recursive_directory_iterator;
   using boost::filesystem::filesystem_error;
   using boost::filesystem::copy_option;
+  using copy_options = boost::filesystem::copy_option;
```
macumber: @lefticus seems duplicate of line above?

openstudiocore/src/utilities/core/FilesystemHelpers.cpp
```diff
@@ -172,5 +187,80 @@ namespace openstudio {
...
+      constexpr auto allowed_attempts = 1000;
+
+      const auto temp_dir = openstudio::filesystem::temp_directory_path();
+
+      int attempts{0};
+
+      while (attempts < allowed_attempts) {
+        // concat number to path basename, without adding a new path element
+        auto filename = basename;
+        filename += openstudio::toPath("-" + std::to_string(count++));
```
macumber: @lefticus if these never get cleaned up then this will start to always fail right?

openstudiocore/src/utilities/core/FilesystemHelpers.cpp
```diff
@@ -172,5 +187,80 @@ namespace openstudio {
...
+        return home;
+      }
+
+      if (const auto home = build_path("HOME"); !home.empty()) {
+        LOG_FREE(Info, "FilesystemHelpers", "home_path HOME: " << toString(home));
+        return home;
+      }
+
+      LOG_FREE(Warn, "FilesystemHelpers", "home_path No Home Found");
+      return toPath("/");
```
macumber: @lefticus is "~/" a better guess, at least on unix?

openstudiocore/src/model/FileOperations.cpp
```diff
@@ -79,242 +79,69 @@
...
-        test = QFile::remove(dstItemInfo.absoluteFilePath());
-        if (!test){
-          LOG_FREE(Error, "synchDirStructures", "Failed to remove file '" << toString(dstItemInfo.absoluteFilePath()) << "'");
-          result = false;
-
-          // DLM: do we really want to give up here?
-          //return result;
-        }
-      }
+      throw std::runtime_error("Destination directory " + destinationDir.string() + " already exists");
```
macumber: @lefticus this is obviously much simpler than what was there.  It will fail if the destination already exists, so we'll need to ensure that we are deleting the destination before we call this.  Can you do a text search to see where these are being called and check that we are doing that?  It is also throwing when previously would return false.  Can we replace these throws with logging and error and returning false?

openstudiocore/src/model/FileOperations.cpp
```diff
@@ -79,242 +79,69 @@
...
-
-  bool synchDirStructures(const QString &srcPath, const QString &dstPath)
-  {
-    bool result = true;
-    bool test = true;
-
-    QDir srcDir(srcPath);
-    QDir dstDir(dstPath);
-
-    if (srcDir.canonicalPath() == dstDir.canonicalPath()){
```
macumber: We're not handling this case anymore?

openstudiocore/src/model/FileOperations.cpp
macumber: Maybe we should delete the destination in this call?

openstudiocore/src/model/FileOperations.cpp
```diff
@@ -79,242 +79,69 @@
...
     }
 
-    return result;
   }
 
-#ifdef Q_OS_WIN
-
-#include <windows.h>
-
-QString longPathName(const QString& path)
```
macumber: Do we not need this longPathName anymore?

openstudiocore/src/utilities/core/FilesystemHelpers.cpp
lefticus: I maintained the behavior with the existing `QDir::homeDir()`. We cannot use `~` because that expansion won't be done, as far as I understand it. This is probably why Qt doesn't use `~`

openstudiocore/src/utilities/bcl/LocalBCL.cpp
lefticus: I found myself changing the exact same line of code 20 times and figured it would make it easier for the next phase of refactor if you only had 1 line to change here.

openstudiocore/src/utilities/core/FilesystemHelpers.cpp
lefticus: Yeah, there might be a better way to do this, I'll spend some more time on it.

openstudiocore/src/utilities/core/FilesystemHelpers.cpp
macumber: Ok great, if you matched Qt's implementation that will be perfect.

openstudiocore/src/model/FileOperations.cpp
macumber: In general the old behavior was too complicated for my tastes, but I'm not sure how we evolved to get there.

lefticus: Well, then do we want simplified behavior for today and see if the old behavior *needs* to exist, or do we go back and carefully re-create the old behavior? (I can at the very list make sure we're not copying onto ourselves as this code is checking for.)

openstudiocore/src/utilities/bcl/LocalBCL.cpp
lefticus: We could try to roll something around boost::property_tree or maybe one of the existing JSON libraries used at NREL? I don't have any clear idea.

openstudiocore/src/utilities/core/FilesystemHelpers.cpp
lefticus: Updated to include a timestamp to further reduce possible issues. e19550

openstudiocore/src/model/FileOperations.cpp
lefticus: I think this depends on if we want to stick with this more simplified behavior or recreate the older more complex sync behavior? It would seem that a viable option is to delete the existing dir and copy this one in to it? This should negate the need for the sync? (but I admittedly don't understand the need for the complexity of the old version)

openstudiocore/src/model/FileOperations.cpp
lefticus: Not with the code as it stands, if we go back to the older more complex synch behavior this will get re-added.

openstudiocore/src/model/FileOperations.cpp
lefticus: @macumber I just confirmed that the previous code would "synchronize" by simply deleting any files and folders that were not in the destination. It would seem much simpler to just delete the existing folder and copy over the new one.

macumber: Cool that seems ok to me.  I think the things I would ask for are:

1. Trap for case where src and dest are the same (have to get the lexically correct path to compare)
1. Delete the destination if it exists (don't require this outside the call)
1. Use log error and return false instead of throw

lefticus: Updated in 5d54887cf1990b6230771ac05e76f56ebfd60286

openstudiocore/src/model/FileOperations.cpp
lefticus: This is also resolved with 5d54887cf1990b6230771ac05e76f56ebfd60286

lefticus:
I think this is actually merged already, right? And this PR could be closed?

PR #3361, opened 2018-11-29
Change daylighting error to a warning

shorowit:
The Daylighting control object is not required for the WindowShadingControl object. See AirflowWindowsAndBetweenGlassBlinds.idf example file, for example. We get lots of errors because of this but simulations run fine.

cc @joseph-robertson


shorowit:
Alternatively, we could remove any error/warning altogether. Or we could get nuanced and throw an error depending on the Shading Control Type.

PR #3364, opened 2018-11-30
Fix #3363 - FT the primary air design flow rate field

### LINKED ISSUE #3363: EvaporativeCoolerDirectResearchSpecial primary design flow rate isn't Forward Translated
Like the title said. Found via @shorowit 

Should be present here: https://github.com/jmarrec/OpenStudio/blob/develop/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateEvaporativeCoolerDirectResearchSpecial.cpp#L82:L84

jmarrec:

Fix #3363 - FT the primary air design flow rate field


shorowit:
Made a build and confirmed that this fixes the bug.

PR #3374, opened 2018-12-08
Change XML library for utilities/bcl

jasondegraw:
Replace the use of Qt's XML library with pugixml. This PR adds the pugixml library and changes the code for utilities/bcl.

PR #3382, opened 2018-12-17
Remove QVariant, Move QMetaTypes, and remove Quantity from model API

jmarrec:
Changelog:

* Ported utilities/data/Attribute and measure/OSArgument to use std::variant rather than QVariant
* Removed all getters/setters that involve `Quantity` in `model/`
    * For classes where removing Quantity-related Gtest would produce an empty Gtest, I wrote a brand new fresh one testing all getters and setters.
* Removed Q_PROPERTY in model/
* Move all QMetaType stuff into `model_editor` (eg `Q_DECLARE_METATYPE`)
* For Attribute, ported XML stuff from `QXml` to `pugixml`

Misc:
* Updated utilities/data/Variant to use std::variant instead of boost::variant (a minor change.)

Notes:
* I merged @jasondegraw 's PR #3374 in this one so I could do the XML stuff so I suggest to merge his PR into develop3 before touching this one.

@macumber


openstudiocore/src/model/AirLoopHVAC_Impl.hpp
```diff
@@ -308,4 +304,4 @@ class MODEL_API AirLoopHVAC_Impl : public Loop_Impl {
 
 } // openstudio
 
-#endif
+#endif
```
macumber: @jmarrec let's avoid these warnings about no new line at end of file

openstudiocore/src/model/Blind.hpp
```diff
@@ -78,142 +76,92 @@ class MODEL_API Blind : public ShadingMaterial {
...
 
   double slatBeamSolarTransmittance() const;
 
-  Quantity getSlatBeamSolarTransmittance(bool returnIP=false) const;
-
   bool isSlatBeamSolarTransmittanceDefaulted() const;
 
   double frontSideSlatBeamSolarReflectance() const;
 
-  Quantity getFrontSideSlatBeamSolarReflectance(bool returnIP=false) const;
```
macumber: So much deleting, great!

openstudiocore/src/utilities/core/QJson.cpp
```diff
@@ -1,125 +0,0 @@
-/***********************************************************************************************************************
```
macumber: @jmarrec is this entire file removed?

openstudiocore/src/model_editor/CMakeLists.txt
```diff
@@ -34,9 +34,8 @@ set(${target_name}_src
   OSProgressBar.cpp
   PathWatcher.hpp
   PathWatcher.cpp
-  #  TODO: Unsure if needed? 
   QMetaTypes.hpp
-  QMetaTypes.cpp
+  # QMetaTypes.cpp # Not needed
```
macumber: @jmarrec how did you determine that this source file was not needed, just that there are no linker errors?  this code gets called at run time, I think the symptoms would be parts of the app not working

openstudiocore/src/model/AirLoopHVAC_Impl.hpp
jmarrec: Sorry, for what it's worth I didn't do that on purpose: I removed the `Quantity`-related methods using python, basically `file.read().splitlines()` then `file.write("\".join(new_lines))`, which has the effect of removing the last `\n`...

openstudiocore/src/utilities/core/QJson.cpp
jmarrec: Yes sir. I removed the Json-portion of Attribute, which isn't being used anywhere (it uses XML for I/O), so this could just go*, and removed from `utilties/CMakeList.txt`  https://github.com/NREL/OpenStudio/pull/3382/files#diff-6f721ba6b2eb2c989d633da86dfddb30

* (even if it couldn't I wouldn't have kept a file named `QJson` here anyways, would have replaced that with `jsoncpp`)

openstudiocore/src/utilities/units/Quantity.hpp
```diff
@@ -34,8 +34,6 @@
 
 #include "Unit.hpp"
 
-#include <QMetaType>
```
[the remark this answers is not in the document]
jmarrec: An interesting question. Yes I think we could rely on `convert(double original, const std::string& originalUnits, const std::string& finalUnits)` (which uses `Unit` internally) throughout instead. I would do it on another dedicated PR though, this one is messy enough. I could handle it if you want me to

openstudiocore/src/model_editor/CMakeLists.txt
jmarrec: I tried again this morning the OS App, and I couldn't see any problems without the Cpp, though I do think we *should* need them indeed.

From the [Qt documentation](https://doc.qt.io/qt-5/qmetatype.html#Q_DECLARE_METATYPE):

> Adding a Q_DECLARE_METATYPE() makes the type known to all template based functions, including QVariant. Note that if you intend to use the type in **queued** signal and slot connections or in QObject's **property** system, you also have to call qRegisterMetaType() since the names are resolved at runtime.

so `Q_PROPERTY` is gone now, so not a problem. The question stays whether we have our custom types using as **arguments** inside a **queued** signal and slot connection.

I think we do... cf `ModelObjectListsView.hpp`:


```c++
class ModelObjectListController : public OSVectorController
[...]
private slots:
 void objectAdded(std::shared_ptr<openstudio::detail::WorkspaceObject_Impl>, const openstudio::IddObjectType&, const openstudio::UUID&);
```

`ModelObjectListsView.cpp`:


```c++
 connect(OSAppBase::instance(), &OSAppBase::workspaceObjectAddedPtr, this, &ModelObjectListController::objectAdded, Qt::QueuedConnection)

```

Anyways,I re-added the qRegisterMetaType calls inside a static initializer (and put the cpp file into CMakeLists again)

jmarrec: Last commit is an alternate way, which doesn't use the initializer Struct, I think you prefer this one @macumber but if not you can revert the commit

PR #3385, opened 2018-12-18
Remove qbytearray

lefticus:
## Remove QByteArray

This PR removes QByteArray from develop3. However, by removing it from the core libraries, there are many interactions with QJson and QDomDocument that have to be pushed back into `QByteArray`s.

I think this will ultimately end up in too many merge conflicts. I think ultimately that we should wait until QJson / QDomDocument have been removed, then revisit this branch for merging, after I resolve the issues.


lefticus:
@macumber see notes on the PR. I'm not sure this is actually the right time to merge it, and should wait for more removal of QDomDocument, then I'll sort out conflicts.

openstudiocore/src/utilities/core/System.hpp
```diff
@@ -49,6 +49,12 @@ namespace openstudio {
     /// Returns the number of processors on this computer
     static unsigned numberOfProcessors();
 
+    /// \note not using string_view because we need null terminated strings
+    static boost::optional<std::string> getenv(const std::string &name) noexcept;
```
macumber: @lefticus cool thanks!

PR #3396, opened 2019-01-04
Fix warnings and enable -Werror on Unix

jmarrec:
**Changelog:**

* Started by replacing the `jsoncpp` deprecrated `Json::StyledWriter`, `Json::FastWriter` and `Json::Reader` with `Json::StreamWriterBuilder` (different options to mimic Fast/Styled) and `Json::CharReaderBuilder`.
* Enabled `-Werror` on Unix, and fixed all warnings. For some of the `unused-variable`, I added a C++17 flag `[[maybe_unused]]` when I didn't know if I should just take them out (and a `TODO`), a second pair of eyes there would be helpful.

**Todo:**

* Ensure similar settings for MSVC to avoid more permissive settings: we want to avoid windows users compiling just fine but develop3 breaking on Unix machines as soon as a PR is merged. I'm not familiar with MSVC settings though, so someone else should probably do it (@macumber ?)


openstudiocore/ruby/CMakeLists.txt
```diff
@@ -19,7 +19,13 @@ add_library( init_openstudio STATIC
 )
 
 if(MSVC)
+  # TODO: JM 2019-01-04: @macumber it seems like all three targets use the same COMPILE_FLAGS,
```
jmarrec: @macumber thoughts?

openstudiocore/src/model/EnergyManagementSystemProgram.cpp
```diff
@@ -114,7 +114,8 @@ namespace detail {
   bool EnergyManagementSystemProgram_Impl::setBody(const std::string& body) {
     //set body of program to input string
     bool result = false;
-    bool comment_result = false;
+    // TODO: JM 2019-01-03: This is unused
+    [[maybe_unused]] bool comment_result = false;
```
jmarrec: Hey @brianlball, would `comment_result` be used in the future for EMS program/subroutine? I'd like to avoid using `[[maybe_unused]]` flags as possible

openstudiocore/ruby/CMakeLists.txt
jmarrec: So I've tried the other day  to modify CMAKE_CXX_FLAGS inside a subdirectory, then in the root CMakeList print the CMAKE_CXX_FLAGS jsut after the add_subdirectory, and it was like it was unchanged.

macumber: I'm not crazy about changing it just because I don't know all the side effects, I'd rather set a variable and use that three times

openstudiocore/src/utilities/filetypes/EpwFile.cpp
```diff
@@ -98,17 +98,16 @@ namespace openstudio{
     return fp*psat;
   }
 
-  // TODO: Warning -Wunused-function
-  static double enthalpy(double T, double p, double phi)
+  // Note JM 2019-01-03: suppress -Wunused-function warning for this API function
+  [[maybe_unused]] static double enthalpy(double T, double p, double phi)
```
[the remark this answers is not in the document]
jmarrec: It's not in header. It's a free standing function, that's not being called in AirState ctor unlike the other freestanding functions, though AirState does the same as these functions so it could call them (search for 'eqn 22' for eg)

PR #3397, opened 2019-01-05
Remove more qstring

lefticus:
## Remove More QString

This removes QString from utilities/core, which includes String.* and Path.* files.

It would potentially have lots of impact on the parts of OpenStudio that is using QDomDocument, so highly localized helper functions were added to prevent too much merge conflict impact for those working on it.

### DIFF
--- a/openstudiocore/src/bimserver/BIMserverConnection.cpp
+++ b/openstudiocore/src/bimserver/BIMserverConnection.cpp
@@ -49,6 +49,13 @@
 #include <iostream>
 #include <boost/none.hpp>
 
+///
+/// TODO: Remove this helper when Qt is fully removed
+///
+static auto toQString(const std::string &s) {
+  return QString::fromUtf8(s.data(), s.size());
+};
+
 
 namespace openstudio {
 namespace bimserver {
@@ -572,7 +579,7 @@ namespace bimserver {
   }
 
   void BIMserverConnection::sendCheckInIFCRequest(QString IFCFilePath) {
-    const auto path = openstudio::toPath(IFCFilePath);
+    const auto path = openstudio::toPath(IFCFilePath.toStdString());
     openstudio::filesystem::ifstream file(path);
     if (!file.is_open()) {
       emit errorOccured(QString("Cannot open file, please verify and try again"));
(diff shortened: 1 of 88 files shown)


lefticus:
@Myoldmopar @axelstudios see notes on PR for impact on libraries that use XML

macumber:
@jasondegraw this pr is going to cause some conflicts with your work.  However, they should be single line, really easy to resolve (just use your code).  I'm going to merge this because I think its best to get it over with sooner than later.

openstudiocore/src/bimserver/BIMserverConnection.cpp
```diff
@@ -49,6 +49,13 @@
 #include <iostream>
 #include <boost/none.hpp>
 
+///
+/// TODO: Remove this helper when Qt is fully removed
+///
+static auto toQString(const std::string &s) {
```
macumber: @lefticus I didn't mention but bimserver project is not being de-qt-ized.  It is going with the other graphical components.  I'll change this to use the "../model_editor/Utilities.hpp" file

openstudiocore/src/model_editor/Qt.i
```diff
@@ -161,12 +161,12 @@ class QComboBox : public QWidget
 
 //class QString{};
 
-//%extend QString{
-//  // to std::string
-//  std::string __str__() const{
-//    return toString(*self);
-//  }
-//}
+%extend QString{
```
macumber: @kbenne this just reminded me that ModelEditor.so (used by SU Plugin) will still need to link to static Qt (unless we want to install Qt binaries which is a pain because SU is the host application).  So the app repo can still build against both static and shared Qt, or we have to mess with the SU installer.  Probably start by building both and then fix installer later

openstudiocore/src/bimserver/BIMserverConnection.cpp
lefticus: I wondered about it, but didn't see it mentioned in any of the previous emails, I might have missed it (t's also not a project I'm familiar with)

Well, at least I did this as a tiny patch here, making the fixes much easier!

PR #3400, opened 2019-01-08
Remove qt utilities

macumber:
Removing QObject and other Qt includes from headers and swig files in utilities


openstudiocore/src/utilities/core/CommonInclude.i
```diff
@@ -6,8 +6,18 @@
...
-#include <QString>
+//#include <QByteArray>
+//#include <QString>
+
+// xkeycheck.h emits warning 4505 if any C++ keyword is redefined, ruby.h does redefine many keywords
+// for some reason our undefs below are not sufficient to avoid this warning
+// Qt headers QByteArray and QString (now removed) included some workaround for this which I did not find
+// we can use #define _ALLOW_KEYWORD_MACROS 1
+// other option is to patch ruby.h and config.h to remove keyword redefinitions (e.g. #define inline __inline)
+#ifdef _MSC_VER
```
macumber: @lefticus I wonder if this rings a bell to you?  ruby.h and config.h define a whole bunch of things, including protected C++ keywords like inline.  There is a file xkeycheck.h which ensures these keywords are not redefined. I tried to undef these in the SWIG file.  However, the pre-processor didn't seem to care that I had undef'd the keywords.  The only workaround I found was to define _ALLOW_KEYWORD_MACROS to ignore all of these in SWIG CXX files which I don't like.

lefticus: The issue of Ruby redefining a bunch of things (which the more I learn about C++ the more I understand just how terrible this is and the source of uncountably many possible issues) is in the front of my mind, but I don't recall anything about the `xkeycheck.h` or `_ALLOW_KEYWORD_MACROS` things. That might have come in during my absence from the project?

macumber: The odd thing is that something in one of the Qt headers avoided this error.  Removing the Qt headers is what raised this error.

lefticus: I wonder if you're seeing the same issue I brushed against: it wasn't a SWIG issue, it was a Qt defining `signal` and `slot` macros, which allowed SWIG to parse the headers?

PR #3401, opened 2019-01-09
Update URL/Path to not include 'file://'

jmarrec:
Update URL/Path fields to not include the `file://` prefix which is something related to QUrl and that's causing potential problems expetially in `WorkflowJSON::findFile` to find the epw file via `OS:WeatherFile` for eg.

Changelog:

* Added VersionTranslation 271 to 272 to remove `file://` in the OSM url fields (there are 3 of them, though the WeatherFile one is the only that's actually used)
* 6dea9dd + 27933f4 : Added an extra check in `WorkflowJSON::findFile(path&)`: if it fails to do it, check if path starts with `file://`, remove that portion, and try again. This is optional, but in some nrel_published measure such as ChangeBuildingLocation, WeatherFile url is actually set directly using `setString(10, 'file://#{filename})` so if the user is using an older version of the measure they could get into trouble.

Related PRs: https://github.com/NREL/OpenStudio-measures/pull/203 and https://github.com/NREL/OpenStudio-workflow-gem/pull/85

### DIFF
--- a/openstudiocore/src/osversion/VersionTranslator.cpp
+++ b/openstudiocore/src/osversion/VersionTranslator.cpp
@@ -4401,6 +4401,57 @@ std::string VersionTranslator::update_2_7_0_to_2_7_1(const IdfFile& idf_2_7_0, c
 
 }
 
+std::string VersionTranslator::update_2_7_1_to_2_7_2(const IdfFile& idf_2_7_1, const IddFileAndFactoryWrapper& idd_2_7_2) {
+  std::stringstream ss;
+  boost::optional<std::string> value;
+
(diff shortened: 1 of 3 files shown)


macumber:
I'm thinking this change should go on develop too, we want to keep the osm versions consistent between the two branches

jmarrec:
Want me to rebase on `develop` rather than `develop3`?

macumber:
I cherry picked these commits into develop and merged them.  I was looking to see what other documentation I might want to update too before I push it.

macumber:
Cherry picked and merged as of https://github.com/NREL/OpenStudio/commit/e5177df46caa6bdb0f8e45e2bba07178cf98ecc4

PR #3402, opened 2019-01-09
Fix OS App Issues with develop3

jmarrec:
**Status: IN PROGRESS** (so that I can track progress)

**Changelog:**

* [x] Menu Bar wasn't working on Ubuntu 16.04 (worked on 18.04 though). Use `QCoreApplication::setAttribute(Qt::AA_DontUseNativeMenuBar, true);` to make it work again:
    * Before:
Before
    * After: 
image

* [x] Impossible to save a new model, or a model that just has the .osm but no companion folder, OS App crashes. 
    * Problem is that unlike `QDir` equivalent, `boost::filesystem::canonical()` used at [FileOperations.cpp#L92](https://github.com/NREL/OpenStudio/blob/df94a2183ddde851fad7e43f2dc0ef219d97e16a/openstudiocore/src/model/FileOperations.cpp#L92) actually imposes that the path exists, when the destination directory (=companion folder) could actually not exist until created. 
    * Replace this with `boost::filesystem::weakly_canonical()`

* [x] Top portion of WebEngine tabs (Geometry & Result tabs) had a lot of blank space, use setPolicy to expanding for these
image
image

* [x] All measures appear with a "broken" icon, though they can be dragged and do work.
image
    * This is due to new parsing of BCXML: whether the `error` key is present or not `m_error` is initialized (to an empty string if the key was missing). Adjsuted to check for presence of key before

* [x] DialogButtons suddenly have icons on Ubuntu (and they are ugly) (tested both 16.04 and 18.04)
image

* [x] Change a Debug message in FilesystemHelper::home_path() to a Trace: left a **TODO** to determine whether we shouldn't completely comment it out (seems it's useful for debugging but not often...) @macumber

### DIFF
--- a/openstudiocore/src/openstudio_app/OpenStudioApp.cpp
+++ b/openstudiocore/src/openstudio_app/OpenStudioApp.cpp
@@ -144,6 +144,10 @@ OpenStudioApp::OpenStudioApp( int & argc, char ** argv)
   QCoreApplication::setOrganizationDomain("nrel.gov");
   setApplicationName("OpenStudioApp");
 
+  // Don't use native menu bar, necessary on Ubuntu 16.04
+  // TODO: check for adverse side effects on other OSes
+  QCoreApplication::setAttribute(Qt::AA_DontUseNativeMenuBar, true);
+
   readSettings();
 
   QFile f(":/library/OpenStudioPolicy.xml");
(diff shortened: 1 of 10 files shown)


openstudiocore/src/openstudio_app/OpenStudioApp.cpp
```diff
@@ -144,6 +144,10 @@ OpenStudioApp::OpenStudioApp( int & argc, char ** argv)
   QCoreApplication::setOrganizationDomain("nrel.gov");
   setApplicationName("OpenStudioApp");
 
+  // Don't use native menu bar, necessary on Ubuntu 16.04
+  // TODO: check for adverse side effects on other OSes
+  QCoreApplication::setAttribute(Qt::AA_DontUseNativeMenuBar, true);
```
macumber: @kbenne is this going to be an issue on Mac?

PR #3403, opened 2019-01-10
Ruby patches

macumber:


macumber:
@lefticus I rebuilt libFFI and Ruby (made some minor changes to CMake for building libffi because Ruby's extension checker assumes certain name for libffi, otherwise your changes were all big improvements) and posted my builds to s3.  I then made core download these new builds.  I built in debug and ran the cli calling a simple ruby script `openstudio test.rb` (I don't think the contents matter).  7/10 runs succeeded, 3/10 failed with the following output:

[error.txt](https://github.com/NREL/OpenStudio/files/2751300/error.txt)

This is definitely an improvement (used to fail 100% of the time) but there is still something weird going on.

PR #3404, opened 2019-01-11
Change XML library in gbXML code

jasondegraw:
Replace the Qt XML reading and writing code in the gbXML code with pugixml. There are a couple of additions over in the SDD code, I'll leave those in to avoid potential clashes, but those changes can be removed as needed.


jmarrec:
Might want to pull cfb4b67dd97c3da239e3bc6a4b96b7d89d59f856 into this especially if you're reviewing on Ubuntu (where `-Werror` is enabled and has different warnings) as it won't build otherwise currently.

PR #3409, opened 2019-01-21
Cpprestsdk

axelstudios:
QNetwork replaced with cpprestsdk


openstudiocore/src/cli/openstudio_cli.rb
```diff
@@ -294,17 +293,6 @@ def parse_main_args(main_args)
         #$logger.warn "'#{dir}' passed to #{main_args[i]} is not a directory"
       end
       new_path << dir
-    elsif main_args[i] == '--no-ssl'
```
macumber: @axelstudios should probably still handle this argument and just spit out a message that it is no longer needed for backwards compatibility.  I know residential was using this.

openstudiocore/src/utilities/bcl/RemoteBCL.cpp
```diff
@@ -255,84 +219,86 @@ namespace openstudio{
...
-/* 
-      //request.setRawHeader("User-Agent", "Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.17 (KHTML, like Gecko) Chrome/24.0.1312.56 Safari/537.17");
-      //m_networkManager->get(request);
-      //std::vector<BCLSearchResult> result = waitForSearch();
-      //if (result.size() > 0 && result[0].versionId() != component.versionId()){
-       // m_componentsWithUpdates.push_back(result[0]);
-      //}
+      m_lastSearch.clear();
+
+      web::http::client::http_client client(toWString(remoteUrl()));
```
macumber: @axelstudios does this client use wstring on all platforms?

macumber: If so, does it assume UTF-16 encoding?  Which byte order?  We need to make sure toWString does the right conversion.

openstudiocore/src/cli/openstudio_cli.rb
shorowit: No big deal, we can update our code to not pass this argument.

openstudiocore/CMakeLists.txt
```diff
@@ -751,21 +760,27 @@ if(BUILD_WITH_OPENSSL)
     file(COPY "${LIBEAY32_PATH}" "${SSLEAY32_PATH}" DESTINATION "${CMAKE_BINARY_DIR}/Products/MinSizeRel/")
     install(FILES "${LIBEAY32_PATH}" "${SSLEAY32_PATH}" DESTINATION "bin")
     #install(FILES "${LIBEAY32_PATH}" "${SSLEAY32_PATH}" DESTINATION ${RUBY_MODULE_ARCH_DIR})
+
+#    find_package(OpenSSL REQUIRED PATHS "${OPENSSL_ROOT_DIR}")
```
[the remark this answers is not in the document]
axelstudios: Cruft - this was when I was building cpprestsdk via our CMake and pointing it to all the correct dependencies.  I'll remove it

PR #3417, opened 2019-01-23
Fix #3415 - getOpenStudioCLI in Debug on Unix

### LINKED ISSUE #3415: OpenStudio.getOpenStudioCLI is empty for develop3 Debug build on ubuntu when using ruby
When using ruby rather than CLI, this raises:

**test.rb**

```ruby
require 'openstudio'
# The above is the same as
# => require '/home/julien/Software/Others/OS-build/Products/ruby/openstudio.rb'
# because that's what I have in /usr/share/rvm/rubies/ruby-2.5.1/lib/ruby/site_ruby/openstudio.rb

cli = OpenStudio.getOpenStudioCLI.to_s

if cli.empty?
  raise "CLI has no path!"
else
  puts "OK, CLI is at #{cli}"
end
```
**Shell**:

```shell
$ ruby test.rb 
Traceback (most recent call last):
test.rb:6:in `<main>': CLI has no path! (RuntimeError)
```

This works though:

```bash
$os_build/Products/openstudio -e "puts OpenStudio.getOpenStudioCLI"
=> /home/julien/Software/Others/OS-build/Products/openstudio-2.7.2
```

If I do the same with my installed openstudio-2.7.1 it works in both cases.

@macumber

jmarrec:
Fix #3415 - getOpenStudioCLI in Debug on Unix

### DIFF
--- a/openstudiocore/src/utilities/core/ApplicationPathHelpers.cxx.in
+++ b/openstudiocore/src/utilities/core/ApplicationPathHelpers.cxx.in
@@ -204,7 +204,8 @@ namespace openstudio {
       }
     #else
       cliPath = getOpenStudioModuleDirectory() / toPath("openstudio");
-      if (!openstudio::filesystem::exists(cliPath)){
+      // If it doesn't exist or it's a directory, try one level above
+      if (!openstudio::filesystem::exists(cliPath) || openstudio::filesystem::is_directory(cliPath)) {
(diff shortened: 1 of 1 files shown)


macumber:
Thanks @jmarrec, can you rebase so this PR is to develop?

jmarrec:
@macumber Done.

PR #3427, opened 2019-02-25
Update OpenStudio to E+ 9.1.0

jmarrec:
Based off on https://github.com/NREL/EnergyPlus/releases/tag/v9.1.0-IOFreeze
I have locally modified and rebuilt the Transition Utility.


openstudiocore/CMake/FindEnergyPlus.cmake
```diff
@@ -42,8 +43,9 @@
 # (To distributed this file outside of CMake, substitute the full
 #  License text for the above reference.)
 
-file(GLOB ENERGYPLUS_POSSIBLE_PATHS "${CMAKE_BINARY_DIR}/../../../[eE]nergy[pP]lus?*")
-
+# TODO: JM 2019-02-26 to @kbenne
+# not sure why it's looking three folders above CMAKE_BINARY_DIR (=the build folder).
+file(GLOB ENERGYPLUS_POSSIBLE_PATHS "${CMAKE_BINARY_DIR}/[eE]nergy[pP]lus-*" "${CMAKE_BINARY_DIR}/../../../[eE]nergy[pP]lus?*")
```
jmarrec: @kbenne I'm fine leaving this as is, just want to know if it makes sense in which case I'll remove the TODO

jmarrec:
cf: NREL/EnergPlus#7148 for the CMake changes

shorowit:
Thanks for adding the new ZoneHVACEquipmentList fields!

jmarrec:
It was a mistake on my part @shorowit, I didn't know you weren't supposed to touch Model API during the update ;)

jmarrec:
ok @kbenne I think this is ready for review. 

I had originally started to allow ZoneEquipmentList Heating/cooling priorities to be zero after noticing that the minimum is zero in the E+.idd, but after testing turns out it isn't respected by E+ (https://github.com/NREL/EnergyPlus/issues/7193). I left the IDD minimum at zero, and the entire logic I wrote for future use, I just put back an `if (priority < 1) priority =1` in setHeatingPriority/setCoolingPriority (https://github.com/NREL/OpenStudio/pull/3427/files#diff-281e7c3043de7b736304b4a8a44adb85R137)

All unit tests pass (I commented out sections in the one I added because of this temporary reset to priority = 1).

I ran regression tests, found some broken tests but due to E+ (https://github.com/NREL/EnergyPlus/pull/7192), the rest pass and don't exhibit any significant difference in EUI (I didn't expect this...). No difference greater than 1%, and if I lower the threshold to 0.1% I get:

image

openstudiocore/CMake/FindEnergyPlus.cmake
macumber: That does not make sense to me

jmarrec: It doesn't make sense to me either, so I guess we should remove it.

Really this part isn't going to be used anyways in the vast majority of cases (unless something goes wrong)

What this does is to glob possible folders in which E+ might be located, but then it inserts in first position the `ENV{ENERGYPLUSDIR}` which is set when you downloaded and extracting the zip, so that's the one it'll find, and it should go just fine in the foor loop so it'll hit the break statement and will never try the other paths.

PR #3437, opened 2019-03-02
Test conan 

lefticus:
## Test Conan

@kbenne this is an example of using conan to manage OpenSSL and zlib/minizip dependencies.

### Setup

To utilize conan you need to install python and `pip install --user conan`, and make sure that your python bin dir is in your path. After that the rest of it *should* be automatic.

### Notes

 * This is tested for both Linux/GCC and Windows/MSVC builds.
 * IMO this shows a lot of promise, but it did take a learning curve to get there.
 * I have not gotten around yet to testing the integrated cmake/openssl build option.
    * The main solutions I've seen either require perl and do it with the official mechanism 
    * or: https://www.valvers.com/open-software/projects/openssl-cmake/ which seems to be unmaintained
 * For completeness dead code has been removed

#### Bugs Found

It *appears* that reducing some of these dependencies uncovered incorrect `#include` paths, so there were some changes to get those building properly.


openstudiocore/src/cli/CMakeLists.txt
```diff
@@ -281,12 +281,10 @@ target_link_libraries(openstudio
 )
 
 if (UNIX AND NOT APPLE)
-  target_link_libraries(openstudio ${RUBY_STATIC_LIBRARY} ${EXTENSION_LIBS} dl crypt ssl)
+  target_link_libraries(openstudio ${RUBY_STATIC_LIBRARY} ${EXTENSION_LIBS} dl crypt)
```
macumber: We will want the CLI to have access to ssl libraries as we want to use the embedded ruby to connect to websites over https

openstudiocore/CMakeLists.txt
```diff
@@ -12,6 +12,34 @@ endif()
...
+
+include(${CMAKE_BINARY_DIR}/conan.cmake)
+
+message("RUNNING CONAN")
+conan_check(VERSION 1.0.0 REQUIRED)
+
+conan_add_remote(NAME bincrafters
+  URL https://api.bintray.com/conan/bincrafters/public-conan)
+
+conan_cmake_run(REQUIRES OpenSSL/1.1.0i@conan/stable
```
macumber: What is the best place to find this recipe?  I searched around on bintray, but didn't find the recipe to see what options there are etc.

macumber: It seems like there is a lot going on in this command, it is replacing three of our dependencies.....

macumber: What does the `BUILD missing` option mean?  Where are we specifying static libs?

openstudiocore/CMakeLists.txt
```diff
@@ -714,67 +757,6 @@ if(NOT OPENSTUDIO_GEMS_ZIP_MD5 STREQUAL OPENSTUDIO_GEMS_ZIP_EXPECTED_MD5)
   file(REMOVE_RECURSE "${CMAKE_BINARY_DIR}/src/cli/embedded_files")
 endif()
 
-# OpenSSL / libssh support
```
macumber: I like this....

openstudiocore/CMakeLists.txt
```diff
@@ -1032,44 +1014,13 @@ add_definitions(-DBOOST_CHRONO_HEADER_ONLY)
...
-  #find_package(cpprestsdk REQUIRED PATHS "${CMAKE_BINARY_DIR}/${CPPRESTSDK_DIR}/lib/cpprestsdk")
-  set(CPPRESTSDK_INCLUDE_DIR "${CMAKE_BINARY_DIR}/${CPPRESTSDK_DIR}/include")
-  include_directories(${CPPRESTSDK_INCLUDE_DIR})
-  set(CPPRESTSDK_LIB_RELEASE "${CMAKE_BINARY_DIR}/${CPPRESTSDK_DIR}/lib/cpprest141_2_10.lib")
-  set(CPPRESTSDK_LIB_DEBUG "${CMAKE_BINARY_DIR}/${CPPRESTSDK_DIR}/lib/cpprest141_2_10d.lib")
+if(WIN32)
+  # cpprest defines its own NOMINMAX and this is causes a macro redefinition "warning as error"
+  remove_definitions(-DNOMINMAX)
+endif()
+add_subdirectory(../cpprestsdk cpprestsdk EXCLUDE_FROM_ALL)
```
macumber: Does this build correctly with the OpenSSL dependency from Conan?  Why is it excluded from all?  Isn't cpprestsdk a dependency of openstudio_utilities?

macumber:
Biggest question is just how do i find out what packages are available and what their options are?  Can we use Qt and Boost from Conan?

openstudiocore/CMakeLists.txt
kbenne: I think @lefticus has it right here. just because it is excluded from the "ALL" target doesn't prevent someone like openstudio_utilities from putting a dependency on it.

PR #3438, opened 2019-03-04
Fix #2798 - Show Uneditable measure name in addition to Display Name

### LINKED ISSUE #2798: OS App only shows the custom editable measure, name and not the one defined in measure.rb and measure.xml
OS app used to show both the un-editable measure name set by the measure writer and an editable name which allows multiple instances of the measure to have the same name.

I wanted to edit one of the measures on a project, but it wasn't clear at first what measure it was. I had to either recognize the arguments or modeler description, or dig around in the OSW.

Not something that needs to be addressed for 2.3.0. May have been this way since 2.x?

jmarrec:
Fix #2798 - Show Uneditable measure name in addition to Display Name

Upon dragging a new measure:

* `name` goes to "Original Name" and cannot be modified
* `display_name` will serve as default for the (Editable) 'Name' in the Edit Pane (if same display name is found, then add an integer counter to display name). Whatever is set there will be shown in the Measure List in the main pane, and will be saved in workflow.osw as well.

rationale

### DIFF
--- a/openstudiocore/CMake/FindEnergyPlus.cmake
+++ b/openstudiocore/CMake/FindEnergyPlus.cmake
@@ -81,7 +81,7 @@ foreach(PATH ${ENERGYPLUS_POSSIBLE_PATHS})
     endif()
 
     # we just need to read the first part of this large file
-    file(READ "${ENERGYPLUS_IDD}" IDD_TEXT LIMIT 1000) 
+    file(READ "${ENERGYPLUS_IDD}" IDD_TEXT LIMIT 1000)
     string(REGEX MATCH "!IDD_BUILD [0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z]" BUILD_SHA_LINE "${IDD_TEXT}")
     string(REGEX MATCH "[0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z]" BUILD_SHA "${BUILD_SHA_LINE}")
     set(ENERGYPLUS_GE_8_2_0 TRUE)
--- a/openstudiocore/src/openstudio_lib/OSDocument.cpp
+++ b/openstudiocore/src/openstudio_lib/OSDocument.cpp
@@ -1397,7 +1397,8 @@ namespace openstudio {
           //if (!tags.empty()){
           //  srmStep.setTaxonomy(tags[0]);
           //}
-          srmStep.setName(srm->displayName());
+          srmStep.setName(srm->name());
+          srmStep.setDisplayName(srm->displayName());
           srmStep.setDescription(srm->description());
           srmStep.setModelerDescription(srm->modelerDescription());
           steps.push_back(srmStep);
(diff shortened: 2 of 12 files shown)


jmarrec:
@DavidGoldwasser Let me know if that solves your issue, and whether you'd like to request a different terminology for 'Original Name' if need be.

macumber:
There are actually three names here:

Measure machine name - name from measure.xml
Measure human readable name - display_name from measure.xml
Workflow step human readable name - name from workflow step in OSW (https://github.com/NREL/OpenStudio-workflow-gem/blob/develop/spec/schema/osw.json#L345)

DavidGoldwasser:
@macumber I don't think the "Name" field in the "Edit" tab can come from the measure.xml, seems like it has to be from OSW steps so that multiple instances of the same measure can have unique name. the non editable "Original Name" could come from the measure.xml

jmarrec:
If I follow correctly I **think** I got it right:

* if you drag from library, you pull info from measure.xml to create a new workflow step. The original name is created from measure.xml tag 'name' (uneditable), the Editable name is generated from tag 'display_name' (with an integer if already exists among the workflow steps that are already there)
* in the edit pane, what you change is the workflow step (inside workflow.osw in companion directory), which stores the original name in 'name' (won't change since not editable, but you have to be able to reload it don't you?) And the one you can edit in 'display_name'

This appears to work I think. I'm not sure if I'm clarifying here or adding confusion. I can take a gif of it in action tonorrow or you can build my branch and see yourself (no Idd change, should rebuild fast)

macumber:
Yeah I think it is confusing.  I lobbied for a more literal display on this tab but got shot down., I would prefer to have the following:

"Measure Step Name" - editable, initialized from measure.xml display name, writes to OSW measure step name
"Measure Step Description" - editable, initialized from measure.xml description, writes to OSW measure step description
--- Section Break --- (collapsible section, initially collapsed)
"Measure Name" - read only, from measure.xml
"Measure Directory" - read only, from measure.xml
"Measure Display Name" - read only, from measure.xml
"Measure Class Name" - read only, from measure.xml
"Measure Description" - read only, from measure.xml
"Measure Modeler Description" - read only, from measure.xml
--- Section Break ---
Inputs

macumber:
As far as I know, nothing in the app (except for creating measures) will alter the measure.rb file.  The measure.rb file is the source for measure name, display name, class name, description, and modeler description when check for updates runs

jmarrec:
Ok. So whenever you clicked on a measure step, the edit pane gets brought up. Currently information is all coming from the workflow.osw,.

With your proposed scenario, do you:
* Store all of a that uneditable stuff from measure.xml into the workflow.osw when you drag, or,
* Read the measure.xml every time? (Or even query the measure.rb via measure manager?)

I find your proposed scenario to be sensible, I'm just wondering if this extra work is worth it in comparison of the (IMHO) little gain in functionality, and I'm afraid of the added complexity if you need to re-read measure.xml every time (potential path issues, extra operations to be done).

macumber:
I feel that the term "Original Measure Name" is confusing things even more.  If this edit pane is restricted to items in the OSW only then there is no "Original Measure Name" in the OSW, the closest thing would be "Measure Dir Name".  IMHO "Measure Dir Name" is useful, is not user editable, and is what WorkflowJSON actually uses to locate the measure.

For now, let's constrain the UI to only show content in the OSW and keep the keys aligned with what is in the OSW:

"Name" - editable step name
"Measure Dir Name" - non-editable measure dir name <-- new field
"Description" - editable step description
"Modeler Description" - non-editable step modeler description

jmarrec:
The terminology could be changed to Name and display Name to match the workflow OSW tag if that's the problem.

So, bottom line, I added "display_name" to OSW, should I just undo all of that work and just use the measure_dir_name then?

macumber:
Yes please, no changes should be made to the OSW.  Let's restart and just add "Measure Dir Name", the "Original Name" is a red herring and we don't want to pursue it.

PR #3439, opened 2019-03-05
Fix #974 - Add 'Adiabatic Surface Construction' to DefaultConstructionSet

### LINKED ISSUE #974: No default construction for adiabatic surfaces
Capturing an email thread between @DavidGoldwasser, Phylroy, and @macumber
1.  Extend the construction set to have fields for adiabatic just like we do for ground, exterior, etc.
2.  Leave construction set as is but have translation treat adiabatic surfaces as interior surfaces (or exterior). Either way at least there is a default, and the user can override the default if they want to.

I think all the proposed approaches are similar in that they only come into play in the E+ translator when we need to pick a construction for adiabatic surfaces.  They differ in their ability to let the user choose what type of construction is used for adiabatic surfaces.  I am not sure if adiabatic surfaces can store heat or not, Edwin do you know?  Does the construction assigned to an adiabatic surface matter at all?  If we do want to let the user choose what construction to use for adiabatic surfaces, I feel it is not intuitive to assign adiabatic surfaces an interior wall construction.  If we want the user to be able to pick what construction is used for adiabatic surfaces then I would prefer we extend the construction set object so it is explicit.

Unsure if adiabatic surfaces can store heat, they do have visible reflectance

jmarrec:
Fix #974 - Add 'Adiabatic Surface Construction' to DefaultConstructionSet

**Changelog:**

* `IDDChange`: add the 'Adiabatic Surface Construction' field to the IDD for `OS:DefaultConstructionSet` as last field (takes in a Construction directly)
* Added to Model API (+Gtest), and OS App. Tested that inheritance in gridview works too. Demo in gif

974_demo

* Added regression test in https://github.com/NREL/OpenStudio-resources/pull/67
*  FT didn't need any work since it relies on `Surface::construction`. VT isn't needed because it's added as a last field in IDD (tested too).
* Added a warning in `Surface::setOutsideBoundaryCondition('Adiabatic')` to tell user that this deletes the `SubSurfaces` as it wasn't clear to me and I did end up having a crash when testing.
    * Before:
```ruby
ss = m.getSubSurfaces[0]
s = ss.surface.get
s.setOutsideBoundaryCondition("Adiabatic")
puts ss
[utilities.idf.WorkspaceObject] <2> Attempt to write a disconnected WorkspaceObject out to Idf.
RuntimeError: /home/bldadmin/openstudio/openstudiocore/src/utilities/idf/WorkspaceObject.cpp@827 : Attempt to write a disconnected WorkspaceObject out to Idf.
```
    * Now:
```ruby
s.setOutsideBoundaryCondition("Adiabatic")
[openstudio.model.Surface] <0> When setting the Outside Boundary Condition for Surface 'Surface 10' to 'Adiabatic', we removed 1 SubSurfaces.
```

### DIFF
--- a/openstudiocore/src/model/DefaultConstructionSet.cpp
+++ b/openstudiocore/src/model/DefaultConstructionSet.cpp
@@ -250,6 +250,10 @@ namespace detail {
         defaultSurfaceConstructions = this->defaultGroundContactSurfaceConstructions();
       }else if (istringEqual("Surface", outsideBoundaryCondition)){
         defaultSurfaceConstructions = this->defaultInteriorSurfaceConstructions();
+      } else if (istringEqual("Adiabatic", outsideBoundaryCondition)) {
+        // Adiabatic is special - doesn't has multiple choices by orientation - so return directly
+        result = this->adiabaticSurfaceConstruction();
+        return result;
       }else{
        //Adiabatic
        //GroundFCfactorMethod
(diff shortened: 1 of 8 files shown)


openstudiocore/src/model/DefaultConstructionSet.cpp
```diff
@@ -250,6 +250,10 @@ namespace detail {
         defaultSurfaceConstructions = this->defaultGroundContactSurfaceConstructions();
       }else if (istringEqual("Surface", outsideBoundaryCondition)){
         defaultSurfaceConstructions = this->defaultInteriorSurfaceConstructions();
+      } else if (istringEqual("Adiabatic", outsideBoundaryCondition)) {
+        // Adiabatic is special - doesn't has multiple choices by orientation - so return directly
+        result = this->adiabaticSurfaceConstruction();
+        return result;
       }else{
        //Adiabatic
```
macumber: You could remove this comment now

openstudiocore/src/model/Surface.cpp
```diff
@@ -473,9 +473,18 @@ namespace detail {
...
+          int n_subsurfaces = 0;
           for (auto subSurface : subSurfaces()){
             subSurface.remove();
+            ++n_subsurfaces;
           }
+          if (n_subsurfaces > 0) {
+            // Note JM 2019-03-05: Warn user, it's not obvious that this is happening and they might try to access again
+            // one of these subsurfaces which are now disconnected objects
+            LOG(Warn, "When setting the Outside Boundary Condition for Surface '" << this->nameString()
+                   << "' to 'Adiabatic', we removed " << n_subsurfaces << " SubSurfaces.");
```
macumber: Can you reword to:

Setting the Outside Boundary Condition for Surface '" << this->nameString()
                   << "' to 'Adiabatic', removed " << n_subsurfaces << " SubSurfaces

PR #3440, opened 2019-03-05
Fix #3430 - Four pipe beam crash in remove

### LINKED ISSUE #3430: error removing four pipe active beams in os v26 and v27
Cannot remove four pipe active beam system in openstudio v2.6 and v2.7.

Detailed description and screenshots here:
https://unmethours.com/question/37348/error-removing-four-pipe-active-beams-in-os-v26-and-v27/

jmarrec:
Fix #3430: add AirTerminalSingleDuctConstantVolumeFourPipeBeam::remove() method

Started by adding the unit test, then fixed it (TDD). You can checkout first commit to see it fail before fix.

Later realized I also needed to change `AirLoopHVAC::addBranchForZone(ThermalZone&)` to handle reconnection of the cloned last ATU's cooling/heating coils for use in OS App in particular.


jmarrec:
Force pushed because I did an interactive rebase to reorder the commits so that all tests (including for addBranchForZone, which I added later) would come first before fixes.

PR #3449, opened 2019-03-12
Fix #1722 - CoilSystemCooling Water/DX HeatExchangerAssisted problems

### LINKED ISSUE #1722: CoilSystemCoolingWaterHeatExchangerAssisted is not translated correctly
For one there needs to be an inner SPM on a node which is not described until translation time.  For two it seems the forward translator is just not handling this type correctly.  It seems to show up on the branch but the object itself is not translated.

jmarrec:
Fix #1722 - CoilSystemCooling Water/DX HeatExchangerAssisted problems

**Changelog:**

* Caught a mistake in `\object-list` for the `CoilSystem:Cooling:DX:HeatExchangerAssisted` cooling coil: should allow both `CoilCoolingDXSingleSpeed` (already the case) AND `CoilCoolingDXVariableSpeed`
* Add `SPM:MixedAir` in FT for the `CoilSystemCoolingWaterHeatExchangerAssisted`'s cooling coil outlet
* Implement `children()` and `clone()` for `CoilSystemCoolingWaterHeatExchangerAssisted`
* Implement `containingHVACComponent()` for all potential children for these two `CoilSystemCoolingXXX` objects, and modify their `addToNode` to  disallow connection if part of a `CoilSystemCoolingXXX`
* Disable addToNode altogether for CoilSystem DX, since it CANNOT be placed directly on an airloop hvac
* Implemented containingHVACComponent and containingZoneHVACComponent for both CoilSystem objects.
* Updated relevant unitary/zoneHVAC `setCoolingCoil` methods to allow these two CoilSystem as appropriate
* Added a bunch of unit testing for getters, setters, addToNode and clone, and containingHVAC/ZoneHVAC COmponent
* Added a test in OpenStudio-resources: https://github.com/NREL/OpenStudio-resources/pull/68


openstudiocore/src/model/AirLoopHVACUnitaryHeatPumpAirToAir.cpp
```diff
@@ -698,16 +704,43 @@ AirLoopHVACUnitaryHeatPumpAirToAir::AirLoopHVACUnitaryHeatPumpAirToAir( const Mo
...
+  }
+
   setCoolingCoil( coolingCoil );
+  if (!ok) {
+    remove();
+    LOG_AND_THROW("Unable to set " << briefDescription() << "'s cooling coil to "
+                  << coolingCoil.briefDescription() << ".");
+  }
+
   setSupplementalHeatingCoil( supplementalHeatingCoil );
```
macumber: should be `ok = setSupplementalHeatingCoil( supplementalHeatingCoil )` ?

openstudiocore/src/model/AirLoopHVACUnitaryHeatPumpAirToAirMultiSpeed.cpp
```diff
@@ -754,13 +754,38 @@ AirLoopHVACUnitaryHeatPumpAirToAirMultiSpeed::AirLoopHVACUnitaryHeatPumpAirToAir
...
 
-  setSupplyAirFan(fan);
+  bool ok = setSupplyAirFan(fan);
+  if (!ok) {
+    remove();
+    LOG_AND_THROW("Unable to set " << briefDescription() << "'s supply fan to "
+                  << fan.briefDescription() << ".");
+  }
+
+  setHeatingCoil( heatingCoil );
```
macumber: More missing oks?

openstudiocore/src/model/AirLoopHVACUnitaryHeatPumpAirToAir.cpp
```diff
@@ -698,16 +704,43 @@ AirLoopHVACUnitaryHeatPumpAirToAir::AirLoopHVACUnitaryHeatPumpAirToAir( const Mo
...
+  }
+
+  ok = setSupplyAirFan(supplyFan);
+  if (!ok) {
+    remove();
+    LOG_AND_THROW("Unable to set " << briefDescription() << "'s supply fan to "
+                  << supplyFan.briefDescription() << ".");
+  }
+
   setHeatingCoil( heatingCoil );
```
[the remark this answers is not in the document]
jmarrec: Yes, some copy paste errors. Thanks for the catch.

jmarrec:
> Added some minor comments. Is there a regression test that covers the configurations of equipment in this PR? This seems like we need to test the specific configurations of equipment to get at interactions between them, basically we'd like to test all the control paths with a regression test

As linked above, I added one test for the **water** version and another for the **dx** version in https://github.com/NREL/OpenStudio-resources/pull/68. Isn't that enough? I'm a bit unclear about what you mean by control path here, does this mean I should also test the DX version with a Cooling:Cooling:DX:VariableSpeed in addition to the SingleSpeed one? Do I also need to test the water one inside of a unitary sys? I don't see any potential other configuration but I might be missing the point 

(For what it's worth, I don't think the vast majority of the openstudio-resources test currently check for every possible combination there is)

PR #3452, opened 2019-03-14
Fix #3421 - Fix forward translator issues in CoilHeatingDesuperheater and CoilWaterHeatingDesuperheater

### LINKED ISSUE #3421: Coil Heating Desuperheater not translating correctly
The forward translator no longer assigns the correct values to the .idf fields.

https://github.com/NREL/OpenStudio/blob/9cae34291ddb630f8a7a9d0207c85d8bf5ab4aed/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilHeatingDesuperheater.cpp#L102-L114

The assigments are correct in the .osm file but the following code will result in an .idf file referencing a `CoilSystem` object instead of a valid `CoilCoolingDXSingleSpeed` or any other [valid type](https://bigladdersoftware.com/epx/docs/9-0/input-output-reference/group-heating-and-cooling-coils.html#field-heating-source-object-type) and an empty name.

Code below from one of my in.idf files.

`  CoilSystem:Cooling:DX,                  !- Heating Source Object Type`
 ` ,                                       !- Heating Source Name`

jmarrec:
Fix #3421 - Fix forward translator issues in CoilHeatingDesuperheater and CoilWaterHeatingDesuperheater

If the Heating Source is a DX cooling coil, and if this DX coil is not already inside another component, the FT will wrap it into a CoilSystem:Cooling:DX wrapper and return that. Cf: [ForwardTranslator::isHVACComponentWithinUnitary](https://github.com/NREL/OpenStudio/blob/v2.7.1/openstudiocore/src/energyplus/ForwardTranslator.cpp#L4434:L4452) and [translateModelObject for CoilCoolingDXSingleSPeed](https://github.com/NREL/OpenStudio/blob/v2.7.1/openstudiocore/src/energyplus/ForwardTranslator.cpp#L951:L960)

Except that the desuper heater IDD will not accept this object type, so it gets remove. I added code to retrieve the actually DX Cooling coil from the wrapper if it exists.

Added unit tests


openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilHeatingDesuperheater.cpp
```diff
@@ -104,13 +101,36 @@ boost::optional<IdfObject> ForwardTranslator::translateCoilHeatingDesuperheater(
   boost::optional<ModelObject> heatingSource = modelObject.heatingSource();
 
   if ( heatingSource ) {
-      boost::optional<IdfObject> _heatingSource = translateAndMapModelObject(heatingSource.get());
+    if (boost::optional<IdfObject> _heatingSource = translateAndMapModelObject(*heatingSource)) {
```
macumber: I thought that we decided against assigning optional in conditional tests.  It might be fine, but just the possibility that it isn't scares me off

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilHeatingDesuperheater.cpp
```diff
@@ -104,13 +101,36 @@ boost::optional<IdfObject> ForwardTranslator::translateCoilHeatingDesuperheater(
...
+        objectType = _heatingSource->iddObject().name();
+        objectName = _heatingSource->name().get();
+      }
 
-      if ( _heatingSource && _heatingSource->name() )
-      {
-        object.setString(Coil_Heating_DesuperheaterFields::HeatingSourceObjectType,_heatingSource->iddObject().name());
-        object.setString(Coil_Heating_DesuperheaterFields::HeatingSourceName,_heatingSource->name().get());
+      if (objectType.empty() || objectName.empty()) {
+        LOG(Error, "Something went wrong in the translation of " << modelObject.briefDescription() << ", couldn't retrieve the coil");
```
macumber: Why not issue this log message, but assign the empty strings to the new object and let it fail at E+ run time?

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilHeatingDesuperheater.cpp
jmarrec: I never knew this was frown upon until now (and that there was a possibility it wouldn't work?). I'll remember that going forward then, and modify this line and the equivalent in the other file.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilHeatingDesuperheater.cpp
jmarrec: Isn't it already what will happen unless you have a debug build? I don't assign the string in question so they come up empty, and if in release you don't hit the OS_ASSERT, so E+ will end up failing.

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilHeatingDesuperheater.cpp
jmarrec: 94ea298a0be22770faebccbcfe5eb988c8ddb0d2

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilHeatingDesuperheater.cpp
jmarrec: Commented out in a03e5f42b71df2536194af85578f5f95d262285e

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilHeatingDesuperheater.cpp
```diff
@@ -104,13 +101,40 @@ boost::optional<IdfObject> ForwardTranslator::translateCoilHeatingDesuperheater(
...
+      // and this DX coil isn't already wrapped in a Unitary, then the FT will wrap it into a CoilSystem:Cooling:DX object and return that, but we
+      // need the DX coil here and not the wrapper.
+      //
+      // Note: Other accepted types are Refrigeration objects and don't suffer the same problem
+      // (Refrigeration:Condenser:AirCooled, Refrigeration:Condenser:EvaporativeCooled, Refrigeration:Condenser:WaterCooled)
+
+      std::string objectType;
+      std::string objectName;
+
+      if (_heatingSource->iddObject().type() == IddObjectType::CoilSystem_Cooling_DX) {
```
rraustad: @jmarrec  I don't know this syntax nor what this translator is trying to accomplish, just a question here. What if the desuperheater coil is in a UnitarySystem?

jmarrec: The coil is translated and returned, and it's the nodes of the coil (and not the parent unitary) that end up there. Does that make sense?

Source:
* https://github.com/NREL/OpenStudio/blob/b32c3d6fe31f7401a0be860ed301039b99ea98d9/openstudiocore/src/energyplus/ForwardTranslator.cpp#L973
* [translateCoilDXSingleSpeedWithoutUnitary](https://github.com/NREL/OpenStudio/blob/b32c3d6fe31f7401a0be860ed301039b99ea98d9/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateCoilCoolingDXSingleSpeed.cpp#L53) you'll see it's the coil that's returned. The naming of the method can be confusing because it's this one that's called when the coil is already inside a unitary, and if not then another method is called where a CoilSystemDX is added ("With unitary")

Issue #69, opened 2013-07-19
Operation slower on repeated task even after making new SketchUp file (Bugzilla #484)


axelstudios:
On 2012-01-19 13:59:30, @DavidGoldwasser wrote:

> TestCase:
> 1. Launch SketchUp
> 2. Create a 100'x50' rectangle.
> 3. Divide it into 4 rectangles
> 4. Use Space diagram to make 5 story building (time this)
> 5. Use script to assign thermal zones to untagged spaces (time this)
> 6. Make a new SketchUp file (but don't quite SketchUp
> 7. Repeat steps 2-6
> 
> In the past I noticed that run a task over and over got slower within the same model, but I assumed it was because the model was smaller. (see bug 252).
> 
> But in this test it happens even making a new model and a NewSketchUp file. It keeps getting slower and slower until you quit and re-start SketchUp.
> 
> Oddly, if I skip step 5 I don't notice the slowdown, not sure why.
> 
> Here are some results (similar results in 0.6.0, 0.6.1 or 0.6.2)
> 
> Loop 1 Step 4 - 2 sec
> Loop 1 Step 5 - 15 sec
> 
> Loop 2 Step 4 - 5 sec
> Loop 2 Step 5 - 30 sec
> 
> Loop 3 Step 4 - 12 sec
> Loop 3 Step 5 - 80 sec
> 
> Loop 4 step 4 - 24 sec
> * didn't run Loop 4 step 5

On 2012-08-15 11:12:31, @macumber wrote:

> Created an attachment (id=192)
> Tests slow down with repeated actions

On 2012-08-15 11:13:29, @macumber wrote:

> Looks like observers are part of the problem, was able to reproduce slow down outside of OpenStudio using attached TestPlugin.rb script
> 
> use_groups = true, use_observers = true, use_operations = true
> 3.385, 4.087, 4.103, 4.336, 4.743, 5.082, 5.603, 5.912, 6.521, 6.833, 7.191, 7.395, 7.816, 8.196, 8.845, 9.438, 9.609, 10.047, 9.921, 10.811
> 
> use_groups = true, use_observers = false, use_operations = true
> 1.622, 1.534, 1.82, 1.7, 1.701, 1.419, 1.441, 1.45, 1.388, 1.372, 1.343, 1.389, 1.373, 1.685, 2.075, 2.199, 1.904, 1.928, 2.229, 1.997

On 2012-09-10 00:55:50, @DavidGoldwasser wrote:

[...]

PR #3453, opened 2019-03-15
Fix #745 - Use the 'Year' field from Sql file

### LINKED ISSUE #745: General problems with leap year simulations

jmarrec:
Fix #745 - Use the 'Year' field from Sql file

Fix https://github.com/NREL/OpenStudio-measures/issues/138

### DIFF
--- a/openstudiocore/src/model/UtilityBill.cpp
+++ b/openstudiocore/src/model/UtilityBill.cpp
@@ -1199,19 +1199,10 @@ Vector BillingPeriod::modelConsumptionValues() const
     Date endDate = this->endDate();
     while (date <= endDate){
 
-      boost::optional<DateTime> dateTime;
+      // Since E+ v8.9.0, sql has year field
+      DateTime dateTime(date, Time(1));
 
-      // Do not include year in date for query, trap for case of leap year
-      if ((date.monthOfYear() == MonthOfYear::Feb) && (date.dayOfMonth()==29)){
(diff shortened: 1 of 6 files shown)


macumber:
Wouldn't hurt to include daylighting reports or any other timeseries data and check that those are reported correctly too.

jmarrec:
> Wouldn't hurt to include daylighting reports or any other timeseries data and check that those are reported correctly too.

Daylight maps don't have the year field in E+ SQL yet (cf issue https://github.com/NREL/EnergyPlus/issues/7225 and PR I opened to add it https://github.com/NREL/EnergyPlus/pull/7235)

jmarrec:
Added tests in #69. All 8 pass with this branch, versus only 1 with 2.7.1.

There's one thing that could be problematic for us actually: currently if you specify a leap year, say 2012, run period, if you pass a weather file that doesn't include Feb 29 you only get 365 days of data (and no warning in eplusout.err). 

When reading the time series back, everything would be shifted by one day for us (and it'd be probably very weird for the user to have a Feb 29 but No December 31...) ... Because we expect for it to be a leap year and we construct the timeseries given only the firstReportDateTime and incrementing intervals.

macumber:
Cool thanks for the tests @jmarrec, seems like the leap year / epw issue should be filed with the EnergyPlus team.  And it looks like it already does https://github.com/NREL/EnergyPlus/issues/7240

PR #3259, opened 2018-09-07
Fix #2718 #2625 fields that became autosizable

jmarrec:
**Warning:** this PR breaks backward compatibility in the API, because fields that weren't autosizable used to have a getter that returned a double, and now it must return boost::optional<double>.

----

## Changelog:

* Fix #2718 : `Sizing:System` "Minimum System Air Flow Ratio" is renamed to "Central Heating Maximum System Air Flow Ratio" to match E+, and becomes autosizable. 
* Fix #2625:  `AirTerminalSingleDuctVAVReheat` and `AirTerminalSingleDuctVAVNoReheat` have two fields - Constant Minimum Air Flow Fraction and Fixed Minimum Air Flow Rate that become autosizable
* All new autosizable fields go their related SQL querying methods (tested with the OpenStudio-resources autosizing_rb test) and added to the `autosize()` and `applySizingValues()`

## API Breaking changes:

* `SizingSystem::minimumSystemAirFlowRatio()` is now deprecated - please use `centralHeatingMaximumSystemAirFlowRatio()` instead -, and it used to return a `double` but now returns a `boost::optional<double>`.
* `AirTerminalSingleDuctVAVReheat::constantMinimumAirFlowFraction()` and `AirTerminalSingleDuctVAVReheat::fixedMinimumAirFlowRate()` used to return a `double` but now returns a `boost::optional<double>`.

Note: For `AirTerminalSingleDuctVAVReheat` and `AirTerminalSingleDuctVAVNoReheat`, I would have removed `isConstantMinimumAirFlowFractionDefaulted()` and `resetConstantMinimumAirFlowFraction()`, but in order to preserve the API, I kept them. Same for `fixedMinimumAirFlowRate`

----

## Other IDD change:

Fix #2458 - Implement the 'Heat Pump Coil Water Flow Mode' field for `ZoneHVAC:WaterToAirHeatPump`

I added 'Design Specification ZoneHVAC Sizing Object Name' to the IDD too, but since that class isn't wrapped in OS SDK I did do anything but add a "TODO" for a reminder.  Similarly 'Availability Manager List Name' was in the IDD but isn't in the SDK (though here the 'AvailabiltiyManagerList' class was added when I worked on #2844)

Review assignee: @kbenne

PR #3455, opened 2019-03-15
Remove 2.6.2 api change warnings

### LINKED ISSUE #3454: Remove 2.6.2 API-change warnings
These API-change warnings are clogging up openstudio-standards logging, making it difficult to find the actual issues.  Propose removing them.

For example, 4,500 of 6,500 lines of a recent openstudio-standards circleci file were:
`[openstudio.model.AirTerminalSingleDuctVAVReheat] <0> Prior to OpenStudio 2.6.2, this field was returning a double, it now returns an Optional double`

asparke2:

Please read [OpenStudio Pull Requests](https://github.com/NREL/OpenStudio/wiki/OpenStudio-Pull-Requests) to better understand the OpenStrudio Pull Request protocol.

fixes #3454

### DIFF
--- a/openstudiocore/src/model/AirTerminalSingleDuctVAVReheat.cpp
+++ b/openstudiocore/src/model/AirTerminalSingleDuctVAVReheat.cpp
@@ -53,10 +53,6 @@
 #include "../utilities/core/Compare.hpp"
 #include "../utilities/core/Assert.hpp"
 
-// TODO: only needed for API warning
-#include <OpenStudio.hxx>
-
-
 namespace openstudio {
 
 namespace model {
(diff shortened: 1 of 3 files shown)


asparke2:
@jmarrec if you want to remove the deprecated methods this is probably a good time/place

jmarrec:
Thanks for the reminder @asparke2!

I will remove the warnings, but actually I had left a note to remove a method altogether:  `LOG(Warn, "SizingSystem::minimumSystemAirFlowRatio has been deprecated and will be removed in a future release, please use SizingSystem::centralHeatingMaximumSystemAirFlowRatio")`

Should I remove `SizingSystem::minimumSystemAirFlowRatio`? @kbenne @macumber

macumber:
@jmarrec I don't think you should remove the method right now as @asparke2 is still using it (and getting bugged by the log messages).  The right thing to do would be for Andrew to start using the new methods rather than remove log messages in deprecated ones.  Can you point him to the methods he should use instead?

jmarrec:
The other method that has a warning is just to tell you that it used to return a double and now an optional. If you had upgraded your code base already you'd still see this message and it would  be annoying. but you wouldn't see the deprecated one since you would be using the right method.
I'm betting that's more what's happening here. @asparke2 are you still using the deprecated method?

macumber:
If @asparke2 is using the right method then lets remove the old method in 2.8.0 (as the warning says) and remove the warning from the right method.  Just so we are all clear, @jmarrec which one is the "right' method and which one is the "deprecated" method?

macumber:
I don't see any log messages in the right method `centralHeatingMaximumSystemAirFlowRatio` so i am confused

jmarrec:
The right commit to look at is this one: https://github.com/NREL/OpenStudio/pull/3259/commits/ec56c6a5dbdb772ffffe2ed53d7750e0b3ac6f6e (#3259)

And quoting the PR description #3259:

> API Breaking changes:

>    SizingSystem::minimumSystemAirFlowRatio() is now deprecated - please use centralHeatingMaximumSystemAirFlowRatio() instead -, and it used to return a double but now returns a boost::optional<double>.
>    AirTerminalSingleDuctVAVReheat::constantMinimumAirFlowFraction() and AirTerminalSingleDuctVAVReheat::fixedMinimumAirFlowRate() used to return a double but now returns a boost::optional<double>.

PR #3465, opened 2019-03-21
Fix #2798 - Show Uneditable measure name in addition to Display Name

### LINKED ISSUE #2798: OS App only shows the custom editable measure, name and not the one defined in measure.rb and measure.xml

jmarrec:
(Initially tried another approach in #3438, after discussion I am starting from a clean slate)

Fix #2798 - Show Uneditable measure name in addition to Display Name

Upon dragging a new measure:

* New: OSW 'measure_dir_name" goes to "Measure Directory Name" and cannot be modified
* OSW 'name' goes to "Name" as usual

image

image

### DIFF
--- a/openstudiocore/src/shared_gui_components/EditView.cpp
+++ b/openstudiocore/src/shared_gui_components/EditView.cpp
@@ -134,10 +134,12 @@ EditRubyMeasureView::EditRubyMeasureView(bool applyMeasureNow)
 
   if(applyMeasureNow){
     nameLineEdit->setReadOnly(true);
-    nameNonEditableLineEdit->setVisible(false);
     descriptionTextEdit->setReadOnly(true);
+
     nameLineEdit->setDisabled(true);
     descriptionTextEdit->setDisabled(true);
+    nameNonEditableLineEdit->setDisabled(true);
+    modelerDescriptionTextEdit->setDisabled(true);
   }
 }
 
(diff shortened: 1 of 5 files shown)


openstudiocore/src/shared_gui_components/WorkflowController.cpp
```diff
@@ -403,15 +402,17 @@ QString MeasureStepItem::name() const
...
-//    return result = QString::fromStdString(*name);
-//  }
-//  return result;
-//}
+QString MeasureStepItem::measureDirectory() const {
+  QString result;
+  // TODO: JM 2019-03-21 Should I ensure that I get just the measure directory NAME and not a path?
+  // Within OS App there's no risk since the measure_dir_name doesn't include "../../" or absolute path
+  // (QFileInfo.fileName)
+  if (boost::optional<std::string> name = m_step.measureDirName()){
```
jmarrec: Need to test that out...

openstudiocore/src/shared_gui_components/EditView.cpp
```diff
@@ -121,6 +134,7 @@ EditRubyMeasureView::EditRubyMeasureView(bool applyMeasureNow)
 
   if(applyMeasureNow){
     nameLineEdit->setReadOnly(true);
+    nameNonEditableLineEdit->setVisible(false);
```
macumber: @jmarrec just curious, why not show this as read only in apply measure now?

jmarrec: Yeah I don't know either, let me change that real quick. (plus, I would actually need to hide the title too if hide the input box)

jmarrec:
One thing that I discovered is that "Modeler Description" is always empty in Apply Now for some reason.

PR #3481, opened 2019-03-28
CMake adjustements for E+

jmarrec:
Revamp E+ section to allow building on windows, I had adjusted the 9.1.0 `ENV{ENERGYPLUSDIR}` in the UNIX block only.

Additional:
* Allow to place E+ zip/tar.gz in build folder and let Cmake deal with it (useful when you have several build folders, you can copy the new E+ tar.gz from one build folder to the others to avoid downloading 150 MB each time)
* Add status message to explain what's downloading too (I see a lot of "Downloading XX%" but you don't know whether it's radiance, Qt, Ruby, etc)

### DIFF
--- a/openstudiocore/CMakeLists.txt
+++ b/openstudiocore/CMakeLists.txt
@@ -397,25 +397,16 @@ if(UNIX)
   endif()
   if(NOT EXISTS "${CMAKE_BINARY_DIR}/${ENERGYPLUS_PATH}.tar.gz" OR NOT "${ENERGYPLUS_HASH}" MATCHES "${ENERGYPLUS_EXPECTED_HASH}")
     message(STATUS "Downloading EnergyPlus ${ENERGYPLUS_VERSION} (${ENERGYPLUS_PLATFORM})")
-    unset(ENERGYPLUS_FOUND CACHE)
-    unset(ENERGYPLUS_EXE CACHE)
-    unset(ENERGYPLUS_IDD CACHE)
-    unset(ENERGYPLUS_WEATHER_DIR CACHE)
+
     if(EXISTS "/etc/redhat-release")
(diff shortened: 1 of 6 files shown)


jmarrec:
I ran the Gtest and the OpenStudio-resources regression suite. Everything seems to pass.

There is a concerning difference for the bundle test though:

image

openstudiocore/CMakeLists.txt
```diff
@@ -397,25 +397,16 @@ if(UNIX)
...
-    unset(ENERGYPLUS_WEATHER_DIR CACHE)
+
     if(EXISTS "/etc/redhat-release")
       file(DOWNLOAD "http://openstudio-resources.s3.amazonaws.com/dependencies/${ENERGYPLUS_PATH}.tar.gz" "${CMAKE_BINARY_DIR}/${ENERGYPLUS_PATH}.tar.gz" INACTIVITY_TIMEOUT 320 SHOW_PROGRESS EXPECTED_MD5 ${ENERGYPLUS_EXPECTED_HASH})
     else()
       file(DOWNLOAD "https://github.com/NREL/EnergyPlus/releases/download/${ENERGYPLUS_RELEASE_NAME}/${ENERGYPLUS_PATH}.tar.gz" "${CMAKE_BINARY_DIR}/${ENERGYPLUS_PATH}.tar.gz" INACTIVITY_TIMEOUT 320 SHOW_PROGRESS EXPECTED_MD5 ${ENERGYPLUS_EXPECTED_HASH})
     endif()
+
+    # Remove the old extracted dir if exists
     execute_process(COMMAND ${CMAKE_COMMAND} -E remove_directory "${CMAKE_BINARY_DIR}/${ENERGYPLUS_PATH}")
```
macumber: @jmarrec i guess we could also delete "${CMAKE_BINARY_DIR}/resources/energyplus/" here too?

jmarrec: I guess we could indeed. Do you think the E+ simulation aren't being update when a replacement E+ package is downloaded?

PR #3489, opened 2019-04-08
Fix problem in sql values(start, end)

jmarrec:


macumber:
@jmarrec let's put this back as a PR to develop, we'll fix this after 2.8.0 and will require a large number of tests to ensure correct behavior (can be in this repo or in regression tests)

For now @DavidGoldwasser can fix measures on the measure side

PR #3494, opened 2019-04-10
SWIG cleanup

jmarrec:
SWIG cleanup with a special look at csharp.


jmarrec:
@macumber I think we probably need to create a `ModelEnergyMangementSystem.i` file to organize things a little better and be able to include stuff before it. Currently all EMS stuff is in `ModelCore.i` except that the EMS Ctors (in particular) of the different classes depend on a few things:

* EnergyManagementSystemSensor: depends on OutputMeter and OutputVariable: It's OK they are in ModelCore.i
* EnergyManagementSystemOutputVariable: depends on Curves (ModelResources.i), Constructions (Geometry.i),
* EnergyManagementSystemActuator: depends on Space (ModelGeometry.i), and  ThermalZone (ModelHVAC.i)

As a result, and because for Python bindings in particular, we might want to put them in a single file/module instead of distributing them throughout the files. 
Exception: `EMSActuatorNames` which is a class defined in ModelObject should stay in `ModelCore.i`

Thoughts?

macumber:
I am generally supportive of this, especially if it enables currently unavailable functionality in C#.  However, each new module is kind of a pain since we have to track module dependencies.  I had a start at that in the https://github.com/NREL/OpenStudio/blob/develop/developer/ruby/MakeDependencyGraph.rb script, this is now out of date with the latest changes.  @brianlball is the person working most closely with EMS objects so he should be involved here too.

I am getting more and more interested in regression tests for C#, we should consider that for a future work.

jmarrec:
I managed to build Ironbug but didn't test it in Grasshopper (in action) since I don't have Rhino. I sent the DLLs to @MingboPeng for him to make sure it still works.

MingboPeng:
@jmarrec  I have just tested 10 ASHRAE baseline systems and others, like VRF, chilled beam, FCU systems for both Honeybee and Ironbug, they all works fine from what I can see. I will do more tests tonight.
Thanks so much for putting efforts to fix this CSharp issue. I think we finally can move from 2.5 to 2.8 now.

Thanks,
Mingbo

jmarrec:
@MingboPeng Note that these changes aren't in 2.8.0 that was released before this PR was merged, but it'll be in the next release. Only https://github.com/NREL/OpenStudio/pull/3493 is in.

PR #3509, opened 2019-04-25
OS_App_killswitch: real time stdout/stderr for conan & Mac Support

jmarrec:
I was very puzzled by seeing conan churn for like 30minutes on Windows and getting ZERO output. I didn't know if it was still running or not, so I created a small fork for cmake-conan that allows real time output in cmake, and also allowed running with MSVC 2019 in prevision.


ConanInstall.cmake
```diff
@@ -4,11 +4,11 @@ if(NOT CONAN_OPENSTUDIO_ALREADY_RUN)
   # Download automatically, you can also just copy the conan.cmake file
   # Put it in CMAKE_BINARY_DIR so we don't end up with two when building OpenStudioApplication
   if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
+    # TODO: temporarily use my fork of conan.cmake (Output in stdout/stderr in real time + support MSVC 2019 in prevision)
     # message(STATUS "openstudio: Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
     # file(DOWNLOAD "https://github.com/conan-io/cmake-conan/raw/v0.13/conan.cmake"
```
macumber: @jmarrec @lefticus what version of conan should we be using, I downgraded to 1.12.3 based on instructions in the wiki

jmarrec: I use the one released a few hours ago, 1.14.4

jmarrec:
@lefticus: the MSVC 2019 portion already has two PRs (included one that's quite old): https://github.com/conan-io/cmake-conan/pull/117 and https://github.com/conan-io/cmake-conan/pull/139.

The output thing is covered in an issue already https://github.com/conan-io/cmake-conan/issues/129#issuecomment-486540226

jmarrec:
I updated this to track cmake-conan v0.14 that now includes these changes!

Note that I personally set my conan.conf to always print run commands. `print_run_commands = True  # environment CONAN_PRINT_RUN_COMMANDS`

jmarrec:
I actually had more problems on Mac. Some stuff was specifically related to 10.13 (std::visit and std::get for variants are only wokring on 10.14), but I needed to go around that so I could do further tests, which highlighted that I had undefined references to some stuff, so I modified the conan stuff to build boost::asio and add websocketpp

jmarrec:
Tandem PR: https://github.com/NREL/conan-openstudio-ruby/pull/3

As noted over there, I did see that gdbm.a was missing on Mac compared to Linux when building openstudio ruby.

Then during building of this PR on mac I have an unresolved symbol error for gdbm too..

```
Undefined symbols for architecture x86_64:
 “_Init_gdbm”, referenced from:
     _main in main.cpp.o
ld: symbol(s) not found for architecture x86_64
```

jmarrec:
Ok, I think I know why on OpenStudio's side: the wiki does mention you should install them from brew, so I updated the search paths for GDBM.

ConanInstall.cmake
```diff
@@ -5,8 +5,8 @@ if(NOT CONAN_OPENSTUDIO_ALREADY_RUN)
   # Put it in CMAKE_BINARY_DIR so we don't end up with two when building OpenStudioApplication
   if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
```
macumber: @jmarrec shouldn't this have our usual MD5 hash check?

jmarrec: Done

PR #3515, opened 2019-05-03
Fix #3436 more path issues

### LINKED ISSUE #3436: More path issues with special characters
Found in #2567.

```
path = "C:/Users/julien/Afolderwithsp\xE9cialCHar#%\xF9/test_model.osm"
# Or
path = 'C:\Users\julien\AfolderwithspécialCHar#%ù\test_model.osm'
OpenStudio::Path.new(path)

RuntimeError: bad conversion
from (pry):4:in `initialize'
```

The path error is not good and should be captured with ruby and c++ unit tests and a new issue

jmarrec:
Fix #3436 more path issues. I added the fix on C++ side, falling back to decoding a string as is if failed as UTF8. Another option would be to do it in SWIG's side, by calling `string.encode('UTF-8')` before trying to make the path.

I added C++ Gtest, and ruby tests in https://github.com/NREL/OpenStudio-resources/pull/70


openstudiocore/src/utilities/core/Path.cpp
```diff
@@ -129,9 +130,27 @@ path toPath(const char* s)
...
-    return path(wstr);
+    std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
+	try {
+		std::wstring wstr = converter.from_bytes(s);
+		return path(wstr);
+	} catch (std::range_error& e) {
+		// Fallback to basically decoding latin-1
+		// Masking with 0xFF reduces any negative values into the range 0-255.
+		// TODO: another option is to write a codecvt_latin1 class
+		// cf: https://stackoverflow.com/questions/49669048/why-mask-a-char-with-0xff-when-converting-narrow-string-to-wide-string#49669048
```
jmarrec: @macumber FYI

macumber: @jmarrec what was happening? Was the previous code throwing an error?

jmarrec: Yes, was throwing a `std::range_error` in `converter.from_bytes`

openstudiocore/src/utilities/core/test/Path_GTest.cpp
```diff
@@ -481,3 +481,49 @@ TEST_F(CoreFixture, LastLevelDirectoryWithDot) {
   EXPECT_EQ("A measure with 90.1 dots", toString(lastLevelDir));
   EXPECT_EQ(measure_directory, measure_directory.parent_path() / lastLevelDir);
 }
+
+
+TEST_F(CoreFixture, Path_WithSpecialChars) {
+  // Use a wide string to make sure it's ok when creating and for ulterior comparison with the narrow one
+  std::wstring ws(L"Afolderwithsp�cialCHar#%�/test.osm");
```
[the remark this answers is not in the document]
jmarrec: You're right.

openstudiocore/src/utilities/core/test/Path_GTest.cpp
```diff
@@ -481,3 +481,49 @@ TEST_F(CoreFixture, LastLevelDirectoryWithDot) {
...
+  openstudio::path pwide_full = outfolder / pwide;
+  makeParentFolder(pwide_full, path(), true);
+  boost::filesystem::ofstream outfile(pwide_full);
+  outfile.close();
+  // Ensure that worked
+  EXPECT_TRUE(openstudio::filesystem::exists(pwide_full));
+
+
+  // Alright, now we try with a regular string with special chars
+  std::string s("Afolderwithsp�cialCHar#%�/test.osm");
```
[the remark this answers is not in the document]
jmarrec: Yeah ruby is definitely produces a different encoding on Windows at least ("Windows-1252"), on Unix it has changed recently to UTF-8 instead of US-ASCII. Cf output results from my docker runs in https://github.com/NREL/OpenStudio-resources/pull/70/files

jmarrec:
@macumber The fallback to latin-1 / Windows-1252 only happens for Windows. Do we actually want to allow that on Unix too?

This fails on Unix for eg: https://github.com/NREL/OpenStudio/pull/3515/files#diff-57da90b27b0038e42d9287e9c2bdba68R538

Issue #2449, opened 2017-01-23
building.clone removes 


DavidGoldwasser:
Hits\ this on net zero school work. which clones building in first measure to pick primary or secondary custom seed. VRF's in those seeds are removed when this happens. I worked around this by writing a downstream measure to add VRF's vs. relying on them being in the seed model.

This happens to both AirConditionerVariableRefrigerantFlow and ZoneHVACTerminalUnitVariableRefrigerantFlow objects

A related issue to this is #1571 which shows building.clone breaking surface matching.

jmarrec:
Here's what I tried:

```
require 'openstudio'
include OpenStudio::Model

m = Model.new
z = ThermalZone.new(m)

outdoor_unit = AirConditionerVariableRefrigerantFlow.new(m)
vrf = ZoneHVACTerminalUnitVariableRefrigerantFlow.new(m)
outdoor_unit.addTerminal(vrf)
vrf.addToThermalZone(z)

baseboard = ZoneHVACBaseboardConvectiveElectric.new(m)
baseboard.addToThermalZone(z)

b = m.getBuilding

puts "Before clone, Original model has #{m.getAirConditionerVariableRefrigerantFlows.size} Outdoor VRF Units, " +
     "#{m.getZoneHVACTerminalUnitVariableRefrigerantFlows.size} VRF terminals, " +
     "#{m.getZoneHVACBaseboardConvectiveElectrics.size} baseboards"
   
m2 = Model.new
b.clone(m2)

puts "After clone, Original model has #{m.getAirConditionerVariableRefrigerantFlows.size} Outdoor VRF Units, " +
     "#{m.getZoneHVACTerminalUnitVariableRefrigerantFlows.size} VRF terminals, " +
     "#{m.getZoneHVACBaseboardConvectiveElectrics.size} baseboards"

puts "After clone, Target model has #{m2.getAirConditionerVariableRefrigerantFlows.size} Outdoor VRF Units, " +
     "#{m2.getZoneHVACTerminalUnitVariableRefrigerantFlows.size} VRF terminals, " +
     "#{m2.getZoneHVACBaseboardConvectiveElectrics.size} baseboards"
```

Result:
```
Before clone, Original model has 1 Outdoor VRF Units, 1 VRF terminals, 1 baseboards
After clone, Original model has 1 Outdoor VRF Units, 1 VRF terminals, 1 baseboards
[...]

PR #3534, opened 2019-05-13
Fix 3524: errors in Building::remove

jmarrec:
Implement a full `Building::remove` method to delete objects rather than default to ParentObject::remove which will not call individual ModelObject::remove methods (and therefore in particular will delete a thermalZone but not properly disconnect it first leading to a crash)

### DIFF
--- a/openstudiocore/src/model/Building.cpp
+++ b/openstudiocore/src/model/Building.cpp
@@ -126,6 +126,12 @@ namespace detail {
     SpaceVector spaces = this->spaces();
     result.insert(result.end(), spaces.begin(), spaces.end());
 
+    // TODO: JM 2019-05-13: handle HVAC (#2449)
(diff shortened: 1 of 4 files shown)


openstudiocore/src/model/Building.cpp
```diff
@@ -126,6 +126,57 @@ namespace detail {
     SpaceVector spaces = this->spaces();
     result.insert(result.end(), spaces.begin(), spaces.end());
 
+    // TODO: JM 2019-05-13: handle HVAC (#2449)
```
macumber: @jmarrec is there an issue for this todo?  Also, how does this compare to clone?  I thought building.clone handled airloops?

jmarrec: No it's a Todo for https://github.com/NREL/OpenStudio/issues/2449#issuecomment-491901154

PR #3548, opened 2019-05-23
Fix crash on mac related to os edit

### LINKED ISSUE #3434: OS app repeatable crash on Mac
Many internal users have seen stability issues on mac, but have been seemingly random. I came up with a repeatable case from a clean model following steps below.

1. Launch OpenStudioApp
2. Go to Space Types tab
3. Bring in first space type from library into model.
4. Go to "Loads" sub-tab
5. Click on text cell in Space Type Name, Load Name, or Multiplier Column.
6. Without doing anything else after clicking in cells described in step 5, switch to "Spaces" tab.

For me this always crashes. Definition Column or drop zone columns do not crash. I expect this kind of behavior on other tabs is also cause of many crashes. I have not tested this on windows to see if it is unique to mac. I tested on 2.7.0 and 2.7.2 (internal installer). Screenshot below is just before I switch to "Spaces" tab.

screen shot 2019-02-27 at 8 09 09 am

CC: @macumber

I wanted to put a list of checkbox with specific scenarios beyond the one  above so that someone can get overview without having to look through the comments. (>name = tab/sub-tab, |name = column). Unless stated otherwise these crash after entering text in this tab and then changing the major tab.

Summary: 

- Any grid view text on Site, SpaceType, and ThermalZone crash app. 
- On Facility Tab all building story text fields crash, some exterior equipment fields crash (all except for End Use Subcategory) and on Shading I haven't gotten a crash yet
- I haven't gotten a crash yet on > Space > Properties, but need to test more in sub tabs. There was reported lockup of space name on UH on Windows (maybe a different issue)

Detail:

- [ ] > Site > Weather File & Design Days > Date | Design Day Name 
- [ ] > Site > Weather File & Design Days > Date | Day of Month 
- [ ] > Site > Weather File & Design Days > Date | Month 
- [ ] > Site > Weather File & Design Days > Temperature | Maximum Dry Bulb Temperature 
- [ ] > Site > Weather File & Design Days > ** spot checked a few others and didn't find text entry that didn't crash**
- [ ] > SpaceType > Loads | Space Type Name
- [ ] > SpaceType > Loads | Load Name
- [ ] > SpaceType > Loads | Multiplier 
- [ ] > SpaceType > Measure Tags | Standards Template
- [ ] > SpaceType > Measure Tags | Standards Building Type
- [ ] > SpaceType>Measure Tags | Standards Space Type
- [ ] > Facility > Stories|Story Name
- [ ] > Facility > Stories | Nominal Z Coordinate
- [ ] > Facility > Stories | Nominal Floor to Floor Height
- [ ] > Facility > Stories | Nominal Floor to Ceiling Height
- [ ] > Facility > Exterior Equipment | Exterior Lights Name
- [ ] > Facility > Exterior Equipment | Multiplier
- [ ] > ThermalZones > HVAC Systems | Name
- [ ] > ThermalZones  > HVAC Systems | Multiplier
- [ ] > ThermalZones  > ** spot checked a few others and didn't find text entry that didn't crash**

- [ ] > Space > Loads | Multiplier

kbenne:
The issue stems from the destructor of OSEdit classes
calling unbind. The unbind method does three things.

1. resets the member variables related to the ModelObject that is bound
2. removes some qt signal connections
3. calls setEnabled(false)

You can see these steps here https://github.com/NREL/OpenStudio/blob/1ae1cec65b25421bd2dabcb7501618a00338645b/openstudiocore/src/shared_gui_components/OSLineEdit.cpp#L152

During destruction of the widget it is item 3 that causes the issue,
because as the widget calls the setEnabled method, the focus event is triggered
and that causes a bunch of refreshing logic that we don't want because the overall
view is in a half destroyed state.

The solution is simply to not call unbind when the OSEdit classes are destroyed.
This is defensbile because

1. member state does not need to be reset because it is soon to be destroyed.
2. qt connections will be removed by QObject destructor https://doc.qt.io/archives/qt-4.8/qobject.html#dtor.QObject
3. setEnable (the source of the issue) does not need to be called because the widget is soon to be
nuked anyway.

close #3434

Please read [OpenStudio Pull Requests](https://github.com/NREL/OpenStudio/wiki/OpenStudio-Pull-Requests) to better understand the OpenStrudio Pull Request protocol.

### DIFF
--- a/openstudiocore/src/shared_gui_components/OSDoubleEdit.cpp
+++ b/openstudiocore/src/shared_gui_components/OSDoubleEdit.cpp
@@ -62,7 +62,6 @@ OSDoubleEdit2::OSDoubleEdit2( QWidget * parent )
 
 OSDoubleEdit2::~OSDoubleEdit2()
 {
-  unbind();
 }
 
 void OSDoubleEdit2::bind(model::ModelObject& modelObject,
--- a/openstudiocore/src/shared_gui_components/OSIntegerEdit.cpp
+++ b/openstudiocore/src/shared_gui_components/OSIntegerEdit.cpp
@@ -57,7 +57,6 @@ OSIntegerEdit2::OSIntegerEdit2( QWidget * parent )
 
 OSIntegerEdit2::~OSIntegerEdit2()
 {
-  unbind();
 }
 
 void OSIntegerEdit2::bind(model::ModelObject& modelObject,
--- a/openstudiocore/src/shared_gui_components/OSLineEdit.cpp
+++ b/openstudiocore/src/shared_gui_components/OSLineEdit.cpp
@@ -65,7 +65,6 @@ OSLineEdit2::OSLineEdit2( QWidget * parent )
 
 OSLineEdit2::~OSLineEdit2()
 {
-  unbind();
 }
 
 void OSLineEdit2::bind(model::ModelObject& modelObject,
--- a/openstudiocore/src/shared_gui_components/OSQuantityEdit.cpp
+++ b/openstudiocore/src/shared_gui_components/OSQuantityEdit.cpp
@@ -93,7 +93,6 @@ OSQuantityEdit2::OSQuantityEdit2(const std::string& modelUnits, const std::strin
 
 OSQuantityEdit2::~OSQuantityEdit2()
 {
-  unbind();
 }
 
 void OSQuantityEdit2::bind(bool isIP,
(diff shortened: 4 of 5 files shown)


macumber:
@kbenne nice find!  I'm a little concerned that calling unbind is needed to disconnect the model object's nano signals from the widget.  If the object is later deleted, the nano signal might fire and try to access methods on the deleted object.  Can you do a test to see if that happens?

macumber:
Alternate solution would be an unbindAtDelete method that does everything the same but does not call setEnabled

kbenne:
I did consider a new method along the lines of what you say, however I think this is safe. It is hard for me to make a guarantee though. "works for me"

macumber:
I think you can test by go to a tab, use one of these widgets, go to another tab, come back and delete the object?

PR #3549, opened 2019-05-23
Update OSM Libraries to 2.8.1

jmarrec:
Update OSM Libraries to 2.8.1

### DIFF
--- a/developer/ruby/UpdateHVACLibrary.rb
+++ b/developer/ruby/UpdateHVACLibrary.rb
@@ -2,6 +2,13 @@
 #
 # openstudio.exe UpdateHVACLibrary.rb
 
+# TODO: Julien Marrec 2019-05-23
+# Note: this takes a rather long time to run (due to the size of the standards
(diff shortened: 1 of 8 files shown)


developer/ruby/UpdateHVACLibrary.rb
```diff
@@ -2,6 +2,13 @@
 #
 # openstudio.exe UpdateHVACLibrary.rb
 
+# TODO: Julien Marrec 2019-05-23
```
macumber: I don't like shortcutting the version translator, the real issue is why on earth this takes 2 hrs!

jmarrec: The standards OSM are humongous. `212,055` lines for `DEER_2014.osm` for eg. `grep "OS:" DEER_2014.osm | wc -l` gives out 18041 objects in the file.

It would probably be a good idea to do some profiling and see if we couldn't cut the time it takes to load these models though...

PR #3552, opened 2019-05-24
Fix 281 SSL error + Enable HighDpiScaling

jmarrec:
As the title says

### DIFF
--- a/openstudiocore/src/openstudio_app/OpenStudioApp.cpp
+++ b/openstudiocore/src/openstudio_app/OpenStudioApp.cpp
@@ -139,11 +139,7 @@ OpenStudioApp::OpenStudioApp( int & argc, char ** argv)
     m_measureManagerProcess(nullptr)
 {
   setOrganizationName("NREL");
-  QCoreApplication::setOrganizationDomain("nrel.gov");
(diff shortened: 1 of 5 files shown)


macumber:
If the ssl error handler avoids the crash, then we will want to hoist OSWebEnginePage up to its own file and reuse it in these other classes which currently use QWebEngineView directly

https://github.com/NREL/OpenStudio/blob/e077c7edfa6c94bb37aeb0e31f40b69ed2740223/openstudiocore/src/openstudio_lib/GeometryPreviewView.hpp

https://github.com/NREL/OpenStudio/blob/e077c7edfa6c94bb37aeb0e31f40b69ed2740223/openstudiocore/src/openstudio_lib/ResultsTabView.hpp

tijcolem:
just built and tested this on my mac and, unfortunately, it doesn't fix the floor space crash issue.  
It sounds like this is a good fix for unhandled exceptions in windows and Linux though so a good idea to include this but it doesn't solve the immediate mac crash issue.

macumber:
superseded by https://github.com/NREL/OpenStudio/pull/3557

Issue #3502, opened 2019-04-22
OpenStudio Application crashes when making new plan in geometry editor tab


DavidGoldwasser:
This happens as soon as I click on the "New" button in the Editor sub-tab in the Geometry tab

DavidGoldwasser:
I tried opening a model made in 270 with FloorSpaceJS. When I went to editor it didn't crash right away, but then crashed a few seconds later.

macumber:
Is this Mac only?

DavidGoldwasser:
yes

jmarrec:
I can reproduce with current develop on macOS 10.13.6.

jmarrec:
Hum I think I finally found a hint after much trial and error.

in `openstudiocore/src/openstudio_lib/library/embeddable_geometry_editor.html`

I started by deleting the custom <script> around line 257. That would still crash. I then had to figure out what else was making it crash, and finally found that in the style section arouhd line 7 you have a big minified line in which there's basically:

```
@import url(https://fonts.googleapis.com/css?family=Roboto);

body,
html {
    color: #bbc3c7;
    font-size: 16px;
    font-family: Roboto, sans-serif;
    height: 100vh;
    -webkit-font-smoothing: antialiased;
    -moz-osx-font-smoothing: grayscale;
    margin: 0
}
```

If I remove the Roboto stuff, suddenly I don't get the crash anymore.

```
/* @import url(https://fonts.googleapis.com/css?family=Roboto); */

body,
html {
    color: #bbc3c7;
    font-size: 16px;
    /*font-family: Roboto, sans-serif; */
    height: 100vh;
    -webkit-font-smoothing: antialiased;
    -moz-osx-font-smoothing: grayscale;
    margin: 0
}
```

I then re-enabled the custom <script> block. Added breakpoint on FloorspaceEditor::loadEditor. Going line by line I see it go until the end. But crash happens anyways. Looks like some kind of race condition somewhere... I tried increasing the processEvents delay to 1000 instead of 200, didn't help. The Qt remote debugging console is absolutely no use here

jmarrec:
Between the javascript and the debugger breakpoint (and a bit of luck in juddling the two) I have gotten the floorspaceJS editor to load up and render  - then it crashed anyway - so I think it may be a race condition indeed.

I had two breakpoints on FloorspaceEditor::loadEditor and FloorspaceEditor::FloorspaceEditor

I wrapped part of the javascript init function inside a setTimeout.
```
[...]

PR #3561, opened 2019-06-04
Pr ext gem apple

tijcolem:
Three things in this PR. 

1 ) commenting out ruby ext gdbm and readline since these are not building on windows and mac. The goal will be to have these enabled for develop3 branch. 

2 ) reverting back to using the qt shared libs which fixes [issue 3502](https://github.com/NREL/OpenStudio/issues/3502)

3 ) Updating e+ windows hash code


openstudiocore/src/cli/main.cpp
```diff
@@ -165,11 +165,17 @@ extern "C" {
...
     void Init_console(void);
     void Init_dbm(void);
-    void Init_gdbm(void);
     void Init_pty(void);
-    void Init_readline(void);
     void Init_syslog(void);
   #endif
+
+// Currently only work on linux
+ #ifdef __linux__
```
macumber: @tijcolem i thought this worked on windows but not linux?

tijcolem: just changed this and commented gdbm and readline out.

PR #3583, opened 2019-06-24
Changing sequential load fractions to schedules

joseph-robertson:
Design doc: https://docs.google.com/document/d/1fM_CyZW7yad4kM_R2FUaMfnjin47GLHAocDNBof-jng/edit#

https://github.com/NREL/EnergyPlus/pull/7360

- [x] Check forward translation
- [ ] Check simulations
- [x] Write version translator


joseph-robertson:
This appears to be creating the desired idf.

joseph-robertson:
There's some issue with these methods: https://github.com/NREL/OpenStudio/blob/load-frac-sched-ems/openstudiocore/src/model/ZoneHVACEquipmentList.cpp#L488-L517. The model test results in `Assertion failed: this->is_initialized(), file C:\OpenStudio\OpenStudio-load-frac-sched-ems\core-build\Boost-install\include\boost-1_55\boost/optional/optional.hpp, line 641`, and calling the getter in ruby results in the same assertion error.

openstudiocore/src/model/ThermalZone.hpp
```diff
@@ -416,24 +416,47 @@ class MODEL_API ThermalZone : public HVACComponent {
    */
   boost::optional<double> sequentialCoolingFraction(const ModelObject& equipment) const;
```
jmarrec: Mark as OS_DEPRECATED

openstudiocore/src/model/ThermalZone.hpp
```diff
@@ -416,24 +416,47 @@ class MODEL_API ThermalZone : public HVACComponent {
...
+   *  Returns nothing if when equipment is not in the ZoneHVACEquipmentList, its heating priority is zero,
+   *  or the loadDistributionScheme isn't 'Sequential'
+   */
+  boost::optional<Schedule> sequentialCoolingFractionSchedule(const ModelObject& equipment) const;
+
   /** Return the Sequential Heating Fraction of equipment.
    *  Returns nothing if when equipment is not in the ZoneHVACEquipmentList, its cooling priority is zero,
    *  or the loadDistributionScheme isn't 'Sequential'
    */
   boost::optional<double> sequentialHeatingFraction(const ModelObject& equipment) const;
```
jmarrec: Mark as OS_DEPRECATED

openstudiocore/src/model/ThermalZone.hpp
```diff
@@ -416,24 +416,47 @@ class MODEL_API ThermalZone : public HVACComponent {
...
+   *  Returns nothing if when equipment is not in the ZoneHVACEquipmentList, its cooling priority is zero,
+   *  or the loadDistributionScheme isn't 'Sequential'
+   */
+  boost::optional<Schedule> sequentialHeatingFractionSchedule(const ModelObject& equipment) const;
+
   /** Set the Sequential Cooling Fraction of equipment.
    *  Returns false when equipment is not in the ZoneHVACEquipmentList, its cooling priority is zero,
    *  or the loadDistributionScheme isn't 'Sequential'
    */
   bool setSequentialCoolingFraction(const ModelObject& equipment, double fraction);
```
jmarrec: Mark as OS_DEPRECATED

openstudiocore/src/model/ThermalZone.hpp
```diff
@@ -416,24 +416,47 @@ class MODEL_API ThermalZone : public HVACComponent {
...
+   *  Returns false when equipement is not in the ZoneHVACEquipmentList, its cooling priority is zero,
+   *  or the loadDistributionScheme isn't 'Sequential'
+   */
+  bool setSequentialCoolingFraction(const ModelObject& equipment, const Schedule& schedule);
+
   /** Set the Sequential Heating Fraction of equipment.
    *  Returns false when equipment is not in the ZoneHVACEquipmentList, its heating priority is zero,
    *  or the loadDistributionScheme isn't 'Sequential'
    */
   bool setSequentialHeatingFraction(const ModelObject& equipment, double fraction);
```
jmarrec: Mark as OS_DEPRECATED

openstudiocore/src/model/ZoneHVACEquipmentList.cpp
```diff
@@ -438,7 +444,40 @@ boost::optional<double> ZoneHVACEquipmentList_Impl::sequentialCoolingFraction(co
       OS_ASSERT(wo);
 
       if (wo->handle() == equipment.handle()) {
-        result = group.getDouble(OS_ZoneHVAC_EquipmentListExtensibleFields::ZoneEquipmentSequentialCoolingFraction, true);
+        boost::optional<WorkspaceObject> wo = group.cast<WorkspaceExtensibleGroup>().getTarget(OS_ZoneHVAC_EquipmentListExtensibleFields::ZoneEquipmentSequentialCoolingFractionScheduleName);
+        schedule = wo->optionalCast<ScheduleConstant>();
```
jmarrec: Refactor this function to use a bool on line 434, and deal with schedule directly in the block around 447, because Ideally we do the following:
* If a Schedule was found, and it's a ScheduleConstant => return that double
* If a Schedule was found, and it's NOT a ScheduleConstant => issue a LOG(Warn, "message")
* If no schedule was found, return boost::none

Same applies to the `sequentialHeatingFraction` getter.

openstudiocore/src/model/ZoneHVACEquipmentList.hpp
```diff
@@ -115,26 +115,48 @@ class MODEL_API ZoneHVACEquipmentList : public ModelObject {
    *  or the loadDistributionScheme isn't 'Sequential'
    */
   boost::optional<double> sequentialCoolingFraction(const ModelObject& equipment) const;
```
jmarrec: deprecated, and the other getter and two double setters too.

openstudiocore/src/osversion/VersionTranslator.cpp
```diff
@@ -4490,6 +4491,72 @@ std::string VersionTranslator::update_2_7_1_to_2_7_2(const IdfFile& idf_2_7_1, c
 
 }
 
+std::string VersionTranslator::update_2_8_0_to_2_8_1(const IdfFile& idf_2_8_0, const IddFileAndFactoryWrapper& idd_2_8_1) {
```
jmarrec: Needs to be moved (and refs updated) in `update_2_8_1_to_2_9_0` now

openstudiocore/src/osversion/VersionTranslator.hpp
```diff
@@ -241,6 +241,7 @@ class OSVERSION_API VersionTranslator {
   std::string update_2_6_2_to_2_7_0(const IdfFile& idf_2_6_2, const IddFileAndFactoryWrapper& idd_2_7_0);
   std::string update_2_7_0_to_2_7_1(const IdfFile& idf_2_7_0, const IddFileAndFactoryWrapper& idd_2_7_1);
   std::string update_2_7_1_to_2_7_2(const IdfFile& idf_2_7_1, const IddFileAndFactoryWrapper& idd_2_7_2);
+  std::string update_2_8_0_to_2_8_1(const IdfFile& idf_2_8_0, const IddFileAndFactoryWrapper& idd_2_8_1);
```
jmarrec: Update (=remove problably, since you move that to update_2_8_1_to_2_9_0)

openstudiocore/src/osversion/VersionTranslator.cpp
```diff
@@ -4490,6 +4491,72 @@ std::string VersionTranslator::update_2_7_1_to_2_7_2(const IdfFile& idf_2_7_1, c
...
+        for (size_t i = 0; i < 5; ++i) {
+          if( i > 2 ) {            
+            if( fraction = eg.getDouble(i) ) {
+              auto iddObject = idd_2_8_1.getObject("OS:Schedule:Constant");
+              IdfObject scheduleConstant(iddObject.get());
+              std::string uuid = toString(createUUID());
+              scheduleConstant.setString(0, uuid);
+              scheduleConstant.setString(1, "Name");
+              for (const IdfObject& object2 : idf_2_8_0.objects()) {
+                if ( object2.getString(0).get() == eg.getString(0).get() ) {
```
jmarrec: Generally speaking, use `openstudio::istringEqual` to do a case insensitive comparison instead please, like E+ does.

Also, please do **not** loop on all objects here. You can get the object by it's handle directly via `boost::optional<IdfObject> IdfFile:getObject(const Handle& handle) const;`

jmarrec:
@joseph-robertson  I also suggest merging `everybody_merge_now` into this one so you can have the `update_2_8_1_to_2_9_0` directly, and changing the base branch to that one (unless you address about that branch has been merged to develop already)

openstudiocore/src/model/ThermalZone.hpp
jmarrec: Hum, if someone were to advocate that this setter is useful and shouldn't be marked as deprecated, I wouldn't fight back too hard.

jmarrec:
@joseph-robertson I have made the changes I think are warranted for your convenience, please take a look.

macumber:
@jmarrec thanks for the review.  I have already merged all of @joseph-robertson's prs into `everybody_merge_now` and we have fixed several errors in regressions there.  These will all show as merged once we merge that branch.

macumber:
We can remerge your changes after merging everybody_merge_now

jmarrec:
I think the setters should be named setSequentialCoolingFractionSchedule and setSequentialHeatingFractionSchedule to match the getters and IDD. That's what's used constantly throughout our API.

joseph-robertson:
I think we still want to be able to set doubles (i.e., set doubles that are turned into ScheduleConstants).

joseph-robertson:
Then for the getters we have both capabilities: return a double (if it's a ScheduleConstant) or return a schedule (if it's not a ScheduleConstant).

jmarrec:
So you're advocating to not deprecate any of the getters and setters?

Anyways, aside from marking them as OS_DEPRECATED, I've made the changes I think are the best course of action, including renaming the setters to `setSequentialXXXFractionSchedule` instead of `setSequentialXXXFraction`. It compiles, model and energyplus tests pass.

joseph-robertson:
Yes I think renaming the setters that pass in schedule arguments to `setSequentialXXXFractionSchedule` is fine (but keep the other `setSequentialXXXFraction` setters that pass in doubles).

I do advocate that we don't deprecate the getter methods that return doubles.

(Also, I don't think you made the changes to the `ThermalZone` files...)

jmarrec:
@joseph-robertson I "undeprecated" the thermalZone ones as I think there are convenience methods alright. I did keep the ZoneHVACEquipmentList ones as deprecated (for now at least), but the ThermalZone won't call these directly, instead they do the logic to call the Schedule getters/setters. Generally speaking, you're not meant to interact directly with the ZoneHVACEquipmentList class anyway (you have to find it youself...)

I have made the changes to the ThermalZone class too, I double checked.

I merged this into everybody_merge_now (by mistake TBH...) but I'll have some time tomorrow morning to fix any kinks left.

nmerket:
So I'm the guy to blame for the sequential load fraction stuff on the `ZoneHVAC:EquipmentList` in EnergyPlus. Changing the single fraction to a schedule was a tough decision because the vast majority of users really want a single number and using schedules adds an extra layer of complexity to give them that. But ultimately we determined the complexity was worth it to support some more advanced use cases. 

Looking at the OpenStudio wrapping of this object, I like that there are setters and getters for both a schedule _and_ a single value. This is a great example of OpenStudio providing a friendlier API that abstracts away some of the complexity of using EnergyPlus. For that reason and because deprecating them unnecessarily breaks backwards compatibility of measures, I strongly encourage you to _not_ deprecate the `sequentialHeatingFraction` or `sequentionCoolingFraction` getters, or the `setSequentialHeatingFraction` and `setSequentialCoolingFraction` setters that accept a double.

openstudiocore/src/model/ThermalZone.hpp
macumber: We don't want to deprecate these, @joseph-robertson will review and provide another PR if needed

jmarrec:
I deprecated only the double getters and setters on the ZoneHVACEquipmentList class.

* Technically speaking, you're not supposed to use this class directly, instead you rely on convenience methods on the ThermalZone (that will forward to its ZoneHVACEquipmentList). 
* There is no getter on ThermalZone to find the corresponding ZoneHVACEquipmentList, so the only option you have is to do model.getZoneHVACEquipmentLists then loop on that one to find the correct one, which is unusual
* The `ThermalZone::setSequentialCoolingFraction(double)` is responsible to create the ScheduleConstant object and then defers to `ZoneHVACEquipmentList::setSequentialCoolingFractionSchedule(schedule)` and therefore will not issue any deprecation warnings.

I'm not that worried about the setters that take a double actually, since these will always work (they will end up creating a ton of unecessary schedules in some cases though...). I'm more worried about the getter that returns a boost::optional<double>, since it'll return nothing in all cases where it's not a ScheduleConstant. I did add a Warning, but that assumes users will see them.


(Honestly, in this specific case, I would probably have advocated to NOT change OpenStudio's side at all, leave everything as doubles, and just create a schedule in ForwardTranslation. At least until someone came knocking saying he needed that advanced feature...).

nmerket:
Okay, looking at the ThermalZone header file I feel _a little_ better about this. I wasn't aware of OS trying to abstract the EnergyPlus objects further than they already are in this case. (As a general philosophical aside, it can sometimes be frustrating as someone who knows what they want to do in EnergyPlus to figure out how manipulate the OpenStudio API to achieve the same results when the object model is messed with.) 

Nevertheless, it looks like the capability is there on the ThermalZone. I still don't see the harm for keeping methods around on the EquipmentList that do the same thing. I don't quite understand why having a `boost::optional<double>` getter is okay on the ThermalZone but not on the EquipmentList. The ThermalZone methods could call the EquipmentList methods to avoid duplicate code.

As for why we're doing this now, we (NREL Residential) are in a position to need the schedules (and their associated EMS actuators) sooner rather than later, which is why @joseph-robertson wrapped them.

jmarrec:
My idea is that if you are modifying the eqlist directly then you should know how to create a schedule and pass it in. It's a weak argument, and I don't care if we remove the OS_DEPRECATED on eqlist as well. Also having OS_DEPRECATED only means you get a warning in the bindings, but also one when developping the C++ API directly and would ensure developpers are aware of the fact that this is a convenience method and not the actual on (method names are very similar..). Some stuff has been marked deprecated for a long long time and not removed yet too.

I'll let you (NREL residential) decide to remove the deprecated mentions on eqlist

PR #3595, opened 2019-07-11
Fix #3591: ForwardTranslate End Use by Subcategory for Fan:VariableVolume and Fan:ConstantVolume

### LINKED ISSUE #3591: End Use by Subcategory does not ForwardTranslate for Fan:VariableVolume or Fan:ConstantVolume
Newer versions of OpenStudio include the end-use by subcategory for `Fan:ConstantVolume` and `Fan:VariableVolume` in the GUI but they do not translate to the [IDF](https://bigladdersoftware.com/epx/docs/9-0/input-output-reference/group-fans.html#field-end-use-subcategory-2-000). 

Please see related issues to add them to the GUI here:
#3001  @jmarrec 
#3260  @matthewsteen 

Forward translation for end-use by subcategory works for the [Fan:ZoneExhaust](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFanZoneExhaust.cpp#L131) object.

Forward translation for end-use by subcategory is not included for the [constant ](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFanConstantVolume.cpp) and [variable ](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFanVariableVolume.cpp) fans.

jmarrec:
Fix #3591

### DIFF
--- a/openstudiocore/src/energyplus/ForwardTranslator.cpp
+++ b/openstudiocore/src/energyplus/ForwardTranslator.cpp
@@ -3494,9 +3494,14 @@ std::vector<IddObjectType> ForwardTranslator::iddObjectsToTranslateInitializer()
   result.push_back(IddObjectType::OS_DistrictCooling);
   result.push_back(IddObjectType::OS_DistrictHeating);
   result.push_back(IddObjectType::OS_EvaporativeCooler_Direct_ResearchSpecial);
-  result.push_back(IddObjectType::OS_Fan_ConstantVolume);
+
+  // Equipments should be responsible for translating their fans
+  // result.push_back(IddObjectType::OS_Fan_Variable);
+  // result.push_back(IddObjectType::OS_Fan_ConstantVolume);
+  // TODO: JM 2019-07-11 These two should also be commented out. Fan_ZoneExhaust will be translated by ZoneHVACEquipmentList
   result.push_back(IddObjectType::OS_Fan_OnOff);
   result.push_back(IddObjectType::OS_Fan_ZoneExhaust);
+
   result.push_back(IddObjectType::OS_Node);
   result.push_back(IddObjectType::OS_PlantLoop);
   result.push_back(IddObjectType::OS_Splitter);
--- a/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFanVariableVolume.cpp
+++ b/openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFanVariableVolume.cpp
@@ -83,7 +83,7 @@ boost::optional<IdfObject> ForwardTranslator::translateFanVariableVolume( FanVar
     idfObject.setString(Fan_VariableVolumeFields::AvailabilityScheduleName,sched.name().get());
   }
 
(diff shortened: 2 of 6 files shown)


jmarrec:
Disabled always FT of Fan:ConstantVolume. Ran the OpenStudio-resources OSM tests to ensure it didn't produce any side effects (no kbtu difference)

openstudiocore/src/energyplus/ForwardTranslator.cpp
```diff
@@ -3494,9 +3494,14 @@ std::vector<IddObjectType> ForwardTranslator::iddObjectsToTranslateInitializer()
   result.push_back(IddObjectType::OS_DistrictCooling);
   result.push_back(IddObjectType::OS_DistrictHeating);
   result.push_back(IddObjectType::OS_EvaporativeCooler_Direct_ResearchSpecial);
-  result.push_back(IddObjectType::OS_Fan_ConstantVolume);
+
+  // Equipments should be responsible for translating their fans
+  // result.push_back(IddObjectType::OS_Fan_Variable);
+  // result.push_back(IddObjectType::OS_Fan_ConstantVolume);
+  // TODO: JM 2019-07-11 These two should also be commented out. Fan_ZoneExhaust will be translated by ZoneHVACEquipmentList
```
macumber: @kbenne do you agree?  is there ever a need to translate these to OSM on their own?

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFanVariableVolume.cpp
```diff
@@ -181,10 +181,12 @@ boost::optional<IdfObject> ForwardTranslator::translateFanVariableVolume( FanVar
     }
   }
 
+  // End Use Subcategory
+  idfObject.setString(Fan_VariableVolumeFields::EndUseSubcategory, modelObject.endUseSubcategory());
```
macumber: @jmarrec should you check that this is set first?  what is the default value?

openstudiocore/src/energyplus/ForwardTranslator.cpp
macumber: @jmarrec, @kbenne gives this a big +1 so I think you can proceed with more of these changes (removing forward translation of individual components) in the future

openstudiocore/src/energyplus/ForwardTranslator/ForwardTranslateFanVariableVolume.cpp
jmarrec: The default value is "General". You merged, but I can use isEndUseSubcategoryDefaulted if you want @macumber

macumber: Nah, I remembered we decided to put os defaults in the idf in case they diverge from e+

PR #3603, opened 2019-07-16
Adding gbXMLId and improving CADObjectId import

### LINKED ISSUE #3554: As a user, I want the gbXML import to store the id attribute of each element so I can retrieve elements imported from the gbXML translator
At times it is desired to be able to append information to elements that are stored in the gbXML, but not translated by the gbXML translator. An example could simply be a different opinion on how the translation occurs for a property. Currently the only way to retrieve elements is by name and there is no guarantee that the name is unique or that it is supported (sometimes OpenStudio translates characters such as commas). In order to be able to consistently retrieve desired elements, it would be good to store the id attribute of each gbXML element.

### LINKED ISSUE #3555: As a user, I want the gbXML import to store the CADObjectId element value of each element so I can map data back to the CAD/BIM environment
This is required in addition to storing IDs because often the mapping from BIM/CAD to gbXML isn't one to one. For example, a single wall in BIM may be triangulated on export to gbXML and therefore be represented by multiple walls. When mapping back to the CAD/BIM environment it may be necessary to fetch all objects of a specific CADObjectId and transform the data (i.e. sum them) to get the desired result.

macumber:
Fixes #3555
Fixes #3554


macumber:
Here's a list of issues to discuss:

- I've imported CADObjectId where it is allowed in the v6.01 schema
- I am importing id attribute from all objects as gbXMLId
- The schema allows for an array of CADObjectId, I am only importing the first one if any exist
    - CADObjectId has an optional programIdRef attribute, I think the idea is that a given element might have CADObjectId in multiple programs.  I am importing this as a separate programIdRef attribute if it exists, but only for the first CADObjectId.  Also, I am not traversing the document to find the Program name.
- I am applying the same CADObjectId and gbXMLId to reversed surfaces and subsurfaces, I am adding an additional gbXMLReversed attribute to reversed surfaces (that way we can tell which one is reversed.
- I am not forward translating the CADObjectId and gbXMLId properties from OSM to gbXML

pflaumingo:
I don't see any issues with the proposed changes or need alteration to the discussion items above.

PR #3628, opened 2019-08-16
Bump version 3 0 0 rc1

macumber:


macumber:
@jmarrec @tijcolem @anyaelena FYI, not sure what all issues this will cause

PR #3642, opened 2019-08-29
Fix #2796 - No rule to make target 'src/utilities/idd/IddFieldEnums.ixx'

### LINKED ISSUE #2796: Build fails with `No rule to make target 'src/utilities/idd/IddFieldEnums.ixx'`
I'm trying to build OpenStudio (v2.2.0) on archlinux, on an x86-64 machine. There seems to be an issue with `cmake` not picking up dependencies correctly. The error in context is:
```
Scanning dependencies of target CreateEmbeddedSource
[  1%] Building CXX object embedded/CMakeFiles/CreateEmbeddedSource.dir/CreateEmbeddedSource.cpp.o
[  1%] Linking CXX executable ../Products/CreateEmbeddedSource
[  1%] Built target CreateEmbeddedSource
Scanning dependencies of target openstudio_modeleditor_static_GeneratedHeaders
[  1%] Generating ui_precisiondialog.h
[  1%] Generating ../../openstudio_modeleditor_static_HeadersGenerated_done.stamp
[  1%] Built target openstudio_modeleditor_static_GeneratedHeaders
Scanning dependencies of target openstudio_utilities_static_GeneratedHeaders
make[2]: *** No rule to make target 'src/utilities/idd/IddFieldEnums.ixx', needed by 'openstudio_utilities_static_HeadersGenerated_done.stamp'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:1437: src/utilities/CMakeFiles/openstudio_utilities_static_GeneratedHeaders.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
```
I have attached the full typescript as well: [typescript.txt](https://github.com/NREL/OpenStudio/files/1334578/typescript.txt)

It might to be a problem with Swig not being used correctly by `cmake`. Even though I have `swig-3.0.12` installed locally, the makefile insists on compiling `swig-3.0.7` from scratch.

jmarrec:
Fix #2796:

I finally figured out why I had so much problems reproducing this issue regardless of the platform I was using (macOS or Ubuntu).

Turns out that if you want to reproduce repeatedly you need to be very literal with the wiki build instructions, and use `make` as in **monothread** make. Using `make -j 16` or just `ninja` (=`ninja -j $(nproc)`) will not exhibit the race condition where **MAKE_SWIG_TARGET** is called before **GenerateIddFactoryRun** actually runs.

### DIFF
--- a/openstudiocore/ProjectMacros.cmake
+++ b/openstudiocore/ProjectMacros.cmake
@@ -60,7 +60,7 @@ macro(ADD_GOOGLE_TESTS executable)
   else()
     set(NEWPATH $ENV{PATH})
   endif()
-  
+
   foreach(source ${ARGN})
     if(NOT "${source}" MATCHES "/moc_.*cxx")
       string(REGEX MATCH .*cpp source "${source}")
@@ -159,7 +159,6 @@ endmacro()
 # add a swig target
 # KEY_I_FILE should include path, see src/utilities/CMakeLists.txt.
 macro(MAKE_SWIG_TARGET NAME SIMPLENAME KEY_I_FILE I_FILES PARENT_TARGET PARENT_SWIG_TARGETS)
-  set(DEPENDS "${PARENT_TARGET}")
   set(SWIG_DEFINES "")
   set(SWIG_COMMON "")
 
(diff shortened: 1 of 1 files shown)


jmarrec:
So there's one thing I'm not certain of yet, is that I need to ensure that my changes won't actually call GenerateIddFactoryRun **each** time you build...

lefticus:
@kbenne I suggest we move forward with this merge and close the 2+ year old bug it references.

PR #3659, opened 2019-09-02
Fix #3654: Add ThermalStorage:Ice:Detailed "(Dis)charging Curve Variable Specifications"

### LINKED ISSUE #3654: Fix ThermalStorageIceDetailed Forward Translator
Need to fix this on the energyplus_9_2_idd_proposed branch https://github.com/NREL/OpenStudio/pull/3640/files#diff-4d0a7e22d316a3aff08b1a18a47dd6d4R95

jmarrec:
Fix #3654 

New fields in E+ 9.2.0, cf https://github.com/NREL/EnergyPlus/pull/7339.

I see no way of really handling them in the FT appropriately, or it would be very, very confusing. So I have added the two new fields, and I have tried to mimic the E+ Translation Rules ([here](https://github.com/NREL/EnergyPlus/pull/7339/commits/941a6703227fe716e9133c5d3c48b06c5f4ce4aa#diff-6bcecd46a03668bc5e9998616e6e8066R476)) in our FT, but I simplified them a bit:

if CubicLinear => LMTDMassFlow, otherwise default to PercentDischargedLMTD/PercentChargedLMTD for discharge/charge respectively


jmarrec:
Merging it integration branch now as well.

PR #3661, opened 2019-09-03
#3649 pump variable speed E+ change

### LINKED ISSUE #3649: E+ Change: Pump:VariableSpeed
There is a change to units so check that out and bring the new units over if they are correct in the context of the OS IDD. The other change is just a case change so i would ignore that because it will only server to mess up OS API

jmarrec:
Fix #3649


jmarrec:
Small enough to merge now. I'll make sure to test it out

PR #3663, opened 2019-09-04
#3584 add Temperature Difference Between Cutout And Setpoint for ZoneControl:Thermostat

### LINKED ISSUE #3584: ZoneControl:Thermostat is missing the Temperature Difference Between Cutout And Setpoint field
In the OpenStudio ProposedEnergy+.idd the `ZoneControl:Thermostat` is missing the `Temperature Difference Between Cutout And Setpoint` field at the end of the object.  This is because OpenStudio  treated this object as extensible when really it is not (it does look like it should be extensible).

jmarrec:
Fix #3584 

* energyplus namespace: In E+, the `ZoneControl:Thermostat` has the numeric field `Temperature Difference Between Cutout And Setpoint` after the "extensible" groups, so until that's addressed (cf NREL/EnergyPlus#7484), I removed the extensible portion and am explicitly defining the four groups Control Object Type / Control Object Name, so I can add this field at the end.
* model namespace: this is added to the `OS:ThermostatSetpoint:DualSetpoint` object, and is handled in the FT. 

I didn't have time to build / test.

### DIFF
--- a/openstudiocore/resources/energyplus/ProposedEnergy+.idd
+++ b/openstudiocore/resources/energyplus/ProposedEnergy+.idd
@@ -456,7 +456,7 @@ PerformancePrecisionTradeoffs,
       \memo This object enables users to choose certain options that speed up EnergyPlus simulation,
       \memo but may lead to small decreases in accuracy of results.
   A1; \field Use Coil Direct Solutions
-      \note If Yes, an analytical or empirical solution will be used to replace iterations in 
+      \note If Yes, an analytical or empirical solution will be used to replace iterations in
       \note the coil performance calculations.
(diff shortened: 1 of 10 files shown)


jmarrec:
I'm going to merge this one is as I have tested locally that it builds, testing the bindings, and ran a subset of the regression tests without any issues.

jmarrec:
@macumber couple of things:
* should I add a reset method?
* should I translate even if defaulted? My idea was that I'd rather not write this last field if no value (given that for now I had to make it non extensible). I think the FT will not write fields after Control 1 Name if not given the min-fields attribute, but not 100% sure.

PR #3681, opened 2019-09-18
Windows long path and ninja

jmarrec:
Add a manifest for CreateEmbeddedSource, that in conjunction with LongPathsEnable regkey should make it work.
I am also testing the Ninja generator on Windows, and finding problems and fixing as I go.


jmarrec:
So long paths will still not work, MSVC (cl.exe) itself doesn't support them, at least for the 2013 version. I think there's a change that MSVC 2019 updated to latest would, but we can't use it.

I have successfully built a package (without PAT, PAT fails on my machine but it never worked...) with ninja though.

```
"C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\vcvarsall.bat" x64


cmake -G Ninja -DBUILD_CSHARP_BINDINGS=ON -DBUILD_TESTING=ON -DBUILD_DVIEW=ON -DBUILD_OS_APP=ON -DBUILD_PACKAGE=ON -DBUILD_PAT=OFF -DCMAKE_BUILD_TYPE=Release -DCPACK_BINARY_DEB=OFF -DCPACK_BINARY_IFW=ON -DCPACK_BINARY_NSIS=OFF -DCPACK_BINARY_RPM=OFF -DCPACK_BINARY_STGZ=OFF -DCPACK_BINARY_TBZ2=OFF -DCPACK_BINARY_TGZ=OFF -DCPACK_BINARY_TXZ=OFF -DCPACK_BINARY_TZ=OFF  ../openstudiocore

ninja
```

PR #3708, opened 2019-10-04
Update to gems that include standards 0.2.10 official

jmarrec:


jmarrec:
Done (with more recent) in v290rc3 already

Issue #3718, opened 2019-10-09
Add a c++ linter for code styling


tijcolem:
Evaluate c++ linter options and add to CI pipeline.  May add a separate CI pipeline vs the build/test pipe and run on every pr.

jmarrec:
Might want to add a static analyzer like cppcheck too.

jmarrec:
This feature was added to develop in #3721. Closing this one, will open a new issue for tweaks

PR #3721, opened 2019-10-10
Add a .clang-format file

jmarrec:
Here's the start of a .clang-format file, tested with clang-format 6.0. It's not bad, but could use a bit of tweaking. In particular it seems our coding style to indent namespace is

```
namespace openstudio {
// no indent
namespace model {
// no indent
namespace detail {
  // 2 space indent
}
}
```

which isn't support by clang-format, can either indent nothing, indent inner, or indent all. See `NamespaceIndentation`

https://releases.llvm.org/6.0.1/tools/clang/docs/ClangFormatStyleOptions.html

Example of usage:
```
cd openstudiocore/src/
clang-format -style=file -verbose -i model/ZonePropertyUserViewFactorsBySurfaceName.cpp
```


jmarrec:
part of #3718

tijcolem:
I think we should use a subset of some .cpp files using your suggested  .clang-format and try some different settings. We can use this pr to to keep track of the changes we make. I need to setup a seperate CI task first though.

kbenne:
@tijcolem can we use a commit hook to run this? Or maybe use ci? I feel like some level of automation would keep this from being ignored by developers.

tijcolem:
I have clang-format and clang-tidy and the new instances for the CI agents and just need to roll the updates.  Let's set up a new CI job that just does the formating.  I can probably get this today/tomorrow once the CI agents are updated.

tijcolem:
Adding in a bash script to run the clang-format utility (ci/clang-format.sh).   This script will detect additions and mods to any files of types (cpp, hpp, c, h) between two branches. So for the purposes of CI, the CI job will compare a pull request branch to the target branch and determine what files have changed and process these using the clang-format utility with our own style guide (.clang-format). The default behavior of clang-format is to auto-correct the files.  If this happens on CI, the job will fail and it'll be up to the user to run this script locally to make the necessary changes, commit them and then CI will re-run and it should pass.

tijcolem:
Okay, this should be ready to go. The CI will key off file 'Jenkins_develop_format' and run the CI job. The CI looks only at file types (c, cpp, h, hpp) that have changed between pr-branch <--> target-branch and then run the clang-format on these files only.

jmarrec:
@tijcolem should probably do the same on develop3 (perhaps you did already) since most code changes are happening there right now

tijcolem:
We can add to develop3 but the final merge from develop to develop3 is imminent ( shortly after 2.9.1 release ) so this will automatically be merged. If we want to add to develop3 now we'll need to add another CI job and another Jenkins file "Jenkins_develop3_format" to reference that CI job.

PR #3724, opened 2019-10-11
Merge develop into develop3

jmarrec:
```
git config --global diff.renames true
git config --global diff.renamelimit 0
git checkout develop3
git merge develop
# Make sure everything is up to date, and that it's clean (no changes)
yes d | git mergetool -- openstudiocore/src/openstudio_lib/
yes d | git mergetool -- openstudiocore/src/openstudio_app/
del openstudiocore/sketchup_plugin/
del openstudiocore/src/openstudio_lib/
cp -R openstudiocore/src/* src/
```

problems are in gbxml (and sdd but won't be next time). Especially gbxml/MapEnveloppe.cpp, then I also merged CMakeLists.txt manually.


tijcolem:
Nice work @jmarrec. The first CI build built on osx and linux so this is encouraging. I'm looking into the windows build issue now. I removed the build dir in case some residuals were causing issues. 

We'll see how this looks after the new run completes.

jmarrec:
Moved to the duplicated and correctly labeled #3725

Issue #3746, opened 2019-10-29
json is broken in develop3


kbenne:
Error: source sequence "{\"foo\":\"bar\"}" is illegal/malformed utf-8 in :/ruby/2.5.0/gems/json_pure-2.2.0/lib/json/pure/generator.rb:243:in `generate'
:/ruby/2.5.0/gems/json_pure-2.2.0/lib/json/common.rb:255:in `fast_generate'

PR #3751, opened 2019-10-31
removing json/ext check as this is statically linked in cli

tijcolem:
This addresses #3746 but is specific to Windows. Current ruby windows build via conan https://github.com/NREL/conan-openstudio-ruby provides msvc native json/ext statically linked in so this check is no longer needed.


tijcolem:
This fixes windows. Checking Linux...

PR #3765, opened 2019-11-18
Use cached YearDescription everywhere if available instead of using getUniqueModelObject<YearDescription>

### LINKED ISSUE #3755: getUniqueModelObjects is very slow for large models
I was seeing some performance issues in a call to the openstudio-standards with the thermal_zone_get_occupancy_schedule and when looking at it further and profiling it, it seems that calls to getUniqueModelObject are very expensive. In my case it is the YearDescription object with as shown in the image below, but it's looking like OS doesn't keep a reference to these Singleton objects but rather refetches them at each call. I could be wrong here, but either way, it's causing that thermal_zone_get_occupancy_schedule call to take minutes when done for each zone on large models. We could change the logic to get a pointer to the YearDescription object and pass that around to get the performance improvement needed, but that is cumbersome as a user.

image

jmarrec:
Pull request overview
---------------------

 - Fixes #3755
 - Use cached `yearDescription()` everywhere if available  instead of using `getUniqueModelObject<YearDescription>` that will always loop through all objects in the model (and therefore incur a huge penalty costs).

The other singleton objects aren't going to be called often so I didn't touch them. (eg SimulationControl isn't going to be accessed more than a couple of times)

### Pull Request Author

 - [x] Any new or modified fields have been implemented in the EnergyPlus ForwardTranslator (and ReverseTranslator as appropriate): N/A
 - [x] Model API methods are tested (in `src/model/test`): tests are already existing, I'm modifying the implemention of existing methods only

**Labels:**

 - [x] If change to an IDD file, add the label `IDDChange`: N/A
 - [x] If breaking existing API, add the label `APIChange`: N/A

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/openstudiocore/src/model/Model.hpp
+++ b/openstudiocore/src/model/Model.hpp
@@ -114,14 +114,14 @@ class MODEL_API Model : public openstudio::Workspace {
   /** Get the Building object if there is one, this implementation uses a cached reference to the Building
    *  object which can be significantly faster than calling getOptionalUniqueModelObject<Building>(). */
   boost::optional<Building> building() const;
-  
+
   /** Get the FoundationKivaSettings object if there is one, this implementation uses a cached reference to the FoundationKivaSettings
    *  object which can be significantly faster than calling getOptionalUniqueModelObject<FoundationKivaSettings>(). */
-  boost::optional<FoundationKivaSettings> foundationKivaSettings() const;  
-  
+  boost::optional<FoundationKivaSettings> foundationKivaSettings() const;
+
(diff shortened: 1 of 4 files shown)


jmarrec:
So thinking about it more, and given that I (logically) found a bunch of occurrences of using the `model.getYearDescription` in standards and openstudio measures which uses getUniqueModelObject<YearDescription>, I think it'd be good to export the caching version to bindings as well.

I tried to create an explicit `Model::getYearDescription` method but SWIG didn't like it one bit not matter my efforts in modifying the .i files.

I think the easiest is probably to provide a template specialization for `Model::getUniqueModelObjects<YearDescription>()` to use the cache. I'll try that.

jmarrec:
Superseded by #3766 where I have gone the template specialization road

PR #3768, opened 2019-11-19
Use template specialization for getUniqueModelObjects<YearDescription> to use cached object

### LINKED ISSUE #3755: getUniqueModelObjects is very slow for large models

jmarrec:
**Supersedes and therefore closes #3765**

**Closed in favor of the develop3 version at #3779**

Pull request overview
---------------------

 - Fixes #3755
 - Use template specialization for `getUniqueModelObject<YearDescription>` that uses the cached YearDescription everywhere if available  instead of using the generic `getUniqueModelObject<>` that will always loop through all objects in the model (and therefore incur a huge penalty cost).
 - The benefit of this is that the getYearDescription that is exposed to ruby bindings in particular will also use caching.
 - I also did it for all other singleton objects that already had caching in Model_Impl:  `Building`, `FoundationKivaSettings`, `LifeCycleCostParameters`, `PerformancePrecisionTradeoffs`, `RunPeriod` and `WeatherFile`

-----

## Timings

I started by creating a model with a bunch of objects (so we see the effect of looping through all objects in the model):

```ruby
require 'openstudio'

z = m.getThermalZones[0]
sch = m.getScheduleRulesets[0]
500.times.each do |i|
  z.clone(m)
  sch.clone(m)
end
m.save('test.osm', true)
assert m.objects.size == 5763   # <= bunch of objects
```

I used the following script for the timings:

```ruby
$VERBOSE = nil

n_times = ARGV[0].to_i
require 'openstudio'

# Helper to load a model in one line
# It will raise if the path (or the model) isn't valid
#
# @param path [String] The path to the osm
# @return [OpenStudio::Model::Model] the resulting model.
def osload(path)
  translator = OpenStudio::OSVersion::VersionTranslator.new
  ospath = OpenStudio::Path.new(path)
  model = translator.loadModel(ospath)
  if model.empty?
      raise "Path '#{path}' is not a valid path to an OpenStudio Model"
  else
      model = model.get
  end
  return model
end


m = osload('test.osm')

n_times.times.each do |i|
 yd = m.getYearDescription
end
```

I use it with `time openstudio test2.rb 100`, or `time /path/to/OS-build-develop-release/Products/openstudio test2.rb 100`

here are the timings at various `n_times`:

| n_times   | Before  | After  | Abs Diff  | % Diff  | Speed Gain (x) |
|-----------|---------|--------|-----------|---------|----------------|
| 1         | 4.211   | 4.740  | 0.529     | 12.56%  | 0.89           |
| 10        | 4.218   | 4.701  | 0.483     | 11.45%  | 0.90           |
| 100       | 4.256   | 4.704  | 0.448     | 10.53%  | 0.90           |
| 1,000     | 4.502   | 4.712  | 0.21      | 4.66%   | 0.96           |
| 10,000    | 7.495   | 4.715  | -2.78     | -37.09% | 1.59           |
| 100,000   | 61.386  | 4.4839 | -56.9021  | -92.70% | 13.69          |
| 1,000,000 | 314.599 | 5.5997 | -308.9993 | -98.22% | 56.18          |

### Pull Request Author

 - [x] Any new or modified fields have been implemented in the EnergyPlus ForwardTranslator (and ReverseTranslator as appropriate): N/A
 - [x] Model API methods are tested (in `src/model/test`): tests are already existing, I'm modifying the implemention of existing methods only

**Labels:**

 - [x] If change to an IDD file, add the label `IDDChange`: N/A
 - [x] If breaking existing API, add the label `APIChange`: N/A

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/openstudiocore/src/model/Model.cpp
+++ b/openstudiocore/src/model/Model.cpp
@@ -2585,6 +2585,26 @@ void Model::applySizingValues() {
   return getImpl<detail::Model_Impl>()->applySizingValues();
 }
 
+
+// Template specilization to use caching for YearDescription
+template<>
+YearDescription Model::getUniqueModelObject<YearDescription>() {
+  if (boost::optional<YearDescription> _yd = yearDescription()) {
+    return _yd.get();
+  } else {
+    return YearDescription(*this);
+  }
+}
+
+template<>
+Building Model::getUniqueModelObject<Building>() {
+  if (boost::optional<Building> _b = building()) {
+    return _b.get();
+  } else {
+    return Building(*this);
+  }
+}
+
 std::shared_ptr<openstudio::detail::WorkspaceObject_Impl> detail::Model_Impl::ModelObjectCreator::getNew(
   Model_Impl * model,
   const IdfObject& obj,
--- a/openstudiocore/src/model/Model.cpp
+++ b/openstudiocore/src/model/Model.cpp
@@ -2586,7 +2586,52 @@ void Model::applySizingValues() {
 }
 
 
-// Template specilization to use caching for YearDescription
+// Template specializations for getUniqueModelObject to use caching
+template<>
+Building Model::getUniqueModelObject<Building>() {
+  if (boost::optional<Building> _b = building()) {
+    return _b.get();
+  } else {
+    return Building(*this);
+  }
+}
+
+template <>
+FoundationKivaSettings Model::getUniqueModelObject<FoundationKivaSettings>() {
+  if (boost::optional<FoundationKivaSettings> _b = foundationKivaSettings()) {
+    return _b.get();
+  } else {
+    return FoundationKivaSettings(*this);
+  }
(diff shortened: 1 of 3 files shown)


jmarrec:
Closing because it was decided to make it go in develop3 instead, cf #3779

PR #3781, opened 2019-11-22
Fix bug of ITE object supply/approach temperature difference schedule

KaiyuSun:
Pull request overview
---------------------

 - Fixes #3793 
 - Fix bug of ITE object supply/approach temperature difference schedule:

The display names of these two schedules were not correct in ScheduleTypeRegistry, which leads to the two functions "setSupplyTemperatureDifferenceSchedule" and "setReturnTemperatureDifferenceSchedule" not working. Bug is now fixed and related test was added into ElectricEquipmentITEAirCooled_GTest as well.

### Pull Request Author

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.

 - [x] Model API Changes / Additions
 - [ ] Any new or modified fields have been implemented in the EnergyPlus ForwardTranslator (and ReverseTranslator as appropriate)
 - [x] Model API methods are tested (in `src/model/test`)
 - [x] EnergyPlus ForwardTranslator Tests (in `src/energyplus/Test`)
 - [ ] If a new object or method, added a test in NREL/OpenStudio-resources: Add Link
 - [ ] If needed, added VersionTranslation rules for the objects (`src/osversion/VersionTranslator.cpp`)
 - [ ] Checked behavior in OpenStudioApplication, adjusted policies as needed (`src/openstudio_lib/library/OpenStudioPolicy.xml`)
 - [ ] Verified that C# bindings built fine on Windows, partial classes used as needed, etc.
 - [ ] All new and existing tests passes
 - [ ] If methods have been deprecated, update rest of code to use the new methods

**Labels:**

 - [x] If change to an IDD file, add the label `IDDChange`
 - [x] If breaking existing API, add the label `APIChange`
 - [x] If deemed ready, add label `Pull Request - Ready for CI` so that CI builds your PR

### Review Checklist

This will not be exhaustively relevant to every PR.
 - [x] Perform a Code Review on GitHub
 - [x] Code Style, strip trailing whitespace, etc.
 - [ ] All related changes have been implemented: model changes, model tests, FT changes, FT tests, VersionTranslation, OS App
 - [x] Labeling is ok
 - [ ] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified

PR #3807, opened 2019-12-09
Use OptionalDouble() instead of +-50°C

jmarrec:
Follow up to https://github.com/NREL/OpenStudio/pull/3781

specifically https://github.com/NREL/OpenStudio/pull/3781#discussion_r355713433


KaiyuSun:
Great! Thank you both @jmarrec @kbenne !

PR #3856, opened 2020-01-21
Add Fan:SystemModel to OpenStudio

### LINKED ISSUE #3044: Add Fan:SystemModel to OpenStudio
Per the EnergyPlus InputOutputReference…“As of version 8.7, a new simple “Fan:SystemModel” input object was added that can be substitued for Fan:ConstantVolume, Fan:OnOff, Fan:VariableVolume, and FanPerformance:NightVentilation. Users are encouraged to migrate their models to use this new Fan in the air loop or zone equipment because the original fan objects may be deprecated and eventually removed in the future.”

Please add Fan:SystemModel to OpenStudio as EnergyPlus is encouraging users to migrate towards using this object for air loop and zone equipment fans.

It also includes a new design power sizing method for ‘Electric Power Per Unit Flow Rate’ which is geared toward the [W/CFM] metric many practitioners are familiar with.

Migrated from [UserVoice](http://openstudio.uservoice.com/forums/308629/suggestions/33056077)

jmarrec:
Pull request overview
---------------------

 - Fixes #3044
 - Add Fan:SystemModel to OpenStudio
 - Companion PRs:
    - https://github.com/NREL/OpenStudioApplication/pull/72
    - https://github.com/NREL/OpenStudio-resources/pull/92

## NOTES

* I have removed the "Number of Speeds" field from the OpenStudio.idd altogether, the number of extensible groups is enough.
* Technically, when Speed Control Method is "Discrete", you can add Speeds (extensible) by specifying only the Speed Flow Fraction and omitting the Speed Electric Power Fraction if you have used an Electric Power Function of Flow Fraction Curve (power fraction is then obtained by evaluating the curve at the given Flow Fraction specified). I have **disabled** this behavior. Speeds can only be added by specifying both flow fraction and power fraction. (_Well, in fact you can still manually use pushExtensibleGroup then use setDouble to only set the flow Fraction, I didn't make Speed Electric Power Fraction a required field.._)
* I am NOT providing extensive logic in the FT such as checking that if a ZoneHVACFourPipeFanCoil is "VARIABLEFANVARIABLEFLOW" and using a FanSystemModel, then the FanSystemModel needs to have Speed Control = Continuous. There are probably too many rules to implement correctly, so I'm letting E+ throw fatals for that.
* There are a bunch of objects that should accept Fan:SystemModel but actually don't due to IDD \object-list and \references not matching. In particular, Fan:SystemModel isn't part of `FanOnOff` while a bunch of ZoneHVAC objects accept Type= Fan:SystemModel but Name only FanOnOff. Cf https://github.com/NREL/EnergyPlus/issues/7697
    * Not sure if I should modify our OpenStudio.idd and ProposedEnergy+.idd or wait for it to be implemented in EnergyPlus to begin with.

### Pull Request Author

 - [ ] If a new object or method, added a test in NREL/OpenStudio-resources: https://github.com/NREL/OpenStudio-resources/pull/92
 - [x] If needed, added VersionTranslation rules for the objects (`src/osversion/VersionTranslator.cpp`): N/A
 - [ ] Checked behavior in OpenStudioApplication, adjusted policies as needed (`src/openstudio_lib/library/OpenStudioPolicy.xml`): https://github.com/NREL/OpenStudioApplication/pull/72
 - [x] If methods have been deprecated, update rest of code to use the new methods: N/A

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


kbenne:
> ## NOTES
> * I have removed the "Number of Speeds" field from the OpenStudio.idd altogether, the number of extensible groups is enough.

KB: Strongly agree with your decision. I also try to reduce this kind of redundancy when I see it. 

> * Technically, when Speed Control Method is "Discrete", you can add Speeds (extensible) by specifying only the Speed Flow Fraction and omitting the Speed Electric Power Fraction if you have used an Electric Power Function of Flow Fraction Curve (power fraction is then obtained by evaluating the curve at the given Flow Fraction specified). I have **disabled** this behavior. Speeds can only be added by specifying both flow fraction and power fraction. (_Well, in fact you can still manually use pushExtensibleGroup then use setDouble to only set the flow Fraction, I didn't make Speed Electric Power Fraction a required field.._)


KB: This seems ok to me. In general I like not needing to have a decision tree diagram to figure out what is required input data. Does this choice avoid any optional fields in the API? If so then I am doubly in favor.

> * I am NOT providing extensive logic in the FT such as checking that if a ZoneHVACFourPipeFanCoil is "VARIABLEFANVARIABLEFLOW" and using a FanSystemModel, then the FanSystemModel needs to have Speed Control = Continuous. There are probably too many rules to implement correctly, so I'm letting E+ throw fatals for that.

KB: Yeah and if we do try to implement rules within OpenStudio we will probably fail in keeping them in sync with EnergyPlus over time. Also if you were to implement these rules in the model API (as opposed to the FT) how would you communicate the dynamically required/optional fields as related field change? I think your choice is the correct one. Let EnergyPlus report any input errors in this case.

> * There are a bunch of objects that should accept Fan:SystemModel but actually don't due to IDD \object-list and \references not matching. In particular, Fan:SystemModel isn't part of `FanOnOff` while a bunch of ZoneHVAC objects accept Type= Fan:SystemModel but Name only FanOnOff. Cf [NREL/EnergyPlus#7697](https://github.com/NREL/EnergyPlus/issues/7697)

KB: My first reaction is follow the EnergyPlus IDD even if it is wrong. Consider a PR to EnergyPlus instead, and then update OpenStudio IDD if EnergyPlus merges. Don't you think that even if you correct the OpenStudio IDD you might still have a failed simulation when EnergyPlus can't validate against its IDD?

PR #3871, opened 2020-02-04
Fix #3870 - Remove deprecated Active Year/Instiution fields in OS:ClimateZones

### LINKED ISSUE #3870: OpenStudio ClimateZones initialization behavior and .clear behavior are inconsistent
When initializing an OpenStudio Model, it creates a blank ClimateZone object as can be seen here: 

https://github.com/NREL/OpenStudio/blob/da39c25c4d5a0a561457f25971513b8f9665dbaa/src/model/ClimateZones.cpp#L538-L545

A call to model.getClimateZones.numClimateZones therefore returns 1. However, if .clear is called, it removes the blank ClimateZone and numClimateZones equals 0. I think these behaviors should be consistent. If it's initialized with a ClimateZone it's suggestive that OpenStudio needs one, but if OpenStudio doesn't need one (I'm not aware of a reason it does) then we should consider removing the addition during initialization.

The issue I'm having is when doing some state verification because if I don't add a climate zone it's length is 1, but if I add only 1 climate zone its length remains as 1.

jmarrec:
Pull request overview
---------------------

 - Fixes #3870
 - Bare minimum changes to fix #3870 without revamping entire object and making it non unique

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/resources/model/OpenStudio.idd
+++ b/resources/model/OpenStudio.idd
@@ -771,27 +771,17 @@ OS:RadianceParameters,
 OS:ClimateZones,
        \unique-object
        \extensible:4
-       \min-fields 7
   A1, \field Handle
        \type handle
        \required-field
-  A2, \field Active Institution
-       \deprecated
-       \type alpha
-       \default ASHRAE
-  N1, \field Active Year
-       \deprecated
-       \type integer
-       \minimum 0
-       \default 2006
-  A3, \field Climate Zone Institution Name
+  A2, \field Climate Zone Institution Name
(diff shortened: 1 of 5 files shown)


kbenne:
@jmarrec this looks like a breaking API change. Can you add a note about this change to the release notes. Check with @tijcolem about where he wants to capture release notes.

jmarrec:
@kbenne This is missing VersionTranslation...

PR #3873, opened 2020-02-04
Sdk merge

kbenne:
This PR adds install commands for an OpenStudio C++ SDK. The key components of the SDK install are the API header files and a single openstudio shared library. In this way consuming projects only need to install openstudio and then consume the API using the supplied headers and link to the openstudio shared library. 

There is a cmake exports file included as part of the install so consuming projects may simply...
1. `export openstudio_DIR='/path/to/openstudio/install/`
2. Within the consuming project CMakeLists.txt find_package(openstudio CONFIG REQUIRED)
3. Link to the openstudio library `target_link_libraries(yourapp openstudio::openstudiolib)

The openstudio::openstudiolib target is provided by the package as an import lib, and by linking to it the necessary include path will be automatically configured. An example of doing this is here https://github.com/NREL/OpenStudioApplication/blob/sdk-update/src/openstudio_app/CMakeLists.txt#L239


ProjectMacros.cmake
```diff
@@ -359,7 +339,7 @@ macro(MAKE_SWIG_TARGET NAME SIMPLENAME KEY_I_FILE I_FILES PARENT_TARGET PARENT_S
   #endif()
   set_target_properties(${swig_target} PROPERTIES LIBRARY_OUTPUT_DIRECTORY "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/ruby/")
   set_target_properties(${swig_target} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/ruby/")
-  target_link_libraries(${swig_target} ${PARENT_TARGET})
+  target_link_libraries(${swig_target} ${${PARENT_TARGET}_depends})
```
jmarrec: This looks fragile. I may have missed it above, but can we use something like
`get_target_properties` to get the _depends list from the PARENT_TARGET (I think there's a change right now it's just relying on us maintaining the same naming conventions in 12 different files)?

ruby/CMakeLists.txt
```diff
@@ -116,19 +87,34 @@ target_link_libraries(openstudio_rb
...
     ${OPENSSL_SSL_LIBRARY}
   )
 endif()
 
-install(TARGETS openstudio_rb DESTINATION Ruby COMPONENT "RubyAPI")
+#export(TARGETS
+#  openstudio_rb
+#  NAMESPACE "openstudio::"
+#  FILE "${PROJECT_BINARY_DIR}/openstudioConfig.cmake"
+#)
```
jmarrec: Remove?

src/energyplus/CMakeLists.txt
```diff
@@ -674,28 +668,30 @@ include("${PROJECT_SOURCE_DIR}/embedded/EmbedFiles.cmake")
 embed_files("${FILES}" "${EMBEDDED_PATHS}" OUTPUT energyplus)
 
 set(${target_name}_depends
-  openstudio_model
+  ${openstudio_model_depends}
 )
+set(${target_name}_depends ${${target_name}_depends} PARENT_SCOPE)
```
jmarrec: Why do we need to export to parent_scope here?

src/lib/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+# openstudio_model
+# openstudio_energyplus
+# openstudio_measure
+# openstudio_osversion
+# openstudio_sdd
+# openstudio_isomodel
+# openstudio_gbxml
+# openstudio_radiance
+#)
+#
```
jmarrec: That part is repeated. Remove the comments?

src/lib/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+ openstudio_osversion
+ openstudio_sdd
+ openstudio_isomodel
+ openstudio_gbxml
+ openstudio_radiance
+)
+
+target_link_libraries(
+ openstudiolib
+ INTERFACE
```
jmarrec: A comment here might be helpful.

src/lib/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+# then we have to export and install them as well, which we probably don't want to do.
+
+# If we try to include directories from the dependent libraries directly, we have to use
+# generator expressions to get the specific configuration's INTERFACE_INCLUDE_DIRECTORIES
+
+# This also does not seem to do what we ultimately want, as it would also require
+# that we export the .a files.
+
+# So here we are, I'm manually exporting the directories, it's probably a mistake
+# TODO reevaluate this.
```
jmarrec: When?

CMakeLists.txt
```diff
@@ -1300,6 +1300,45 @@ if(BUILD_PAT AND APPLE)
...
+        DESTINATION lib/cmake/openstudio
+        NAMESPACE "openstudio::"
+        COMPONENT "CPPAPI"
+        FILE openstudioConfig.cmake)
+
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file(
+  "${PROJECT_BINARY_DIR}/openstudioConfigVersion.cmake"
+  VERSION ${CMAKE_PROJECT_VERSION}
+  COMPATIBILITY AnyNewerVersion
```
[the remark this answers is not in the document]
kbenne: I'm thinking we might need down to the sha resolution, but hey we can start here and see how it goes.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10.2)
+cmake_minimum_required(VERSION 3.14.0)
```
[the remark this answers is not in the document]
kbenne: At least one thing is the behavior of object libraries like this https://github.com/NREL/OpenStudio/blob/3b4d9c0262fc31db0e8cae6620bfb5886680be6e/src/model/CMakeLists.txt#L1653 The API didn't change, but @lefticus found that behavior wasn't exactly correct until more recent versions. (if it even is now completely correct)

CMakeLists.txt
```diff
@@ -342,8 +320,14 @@ if(BUILD_V8_BINDINGS)
   option(BUILD_NODE_MODULES "Build V8 Bindings as Node Modules" ON)
 endif()
 
-# Use PCH
-option(USE_PCH "Use precompiled headers" OFF)
+option(USE_LTO "Use Link-Time-Optimization" OFF)
```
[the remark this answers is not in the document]
kbenne: no clue. Maybe @lefticus has an opinion.

ruby/CMakeLists.txt
kbenne: ah yes. I hate random commented code as much as the next person. https://github.com/NREL/OpenStudio/commit/676874f7e1a13cda2ffdcb697a1d461cc23fddaa

src/energyplus/CMakeLists.txt
kbenne: some other sibling targets are using this variable. for better or worse. I feel like that is questionable design but that is how it is. See for instance https://github.com/NREL/OpenStudio/blob/676874f7e1a13cda2ffdcb697a1d461cc23fddaa/src/sdd/CMakeLists.txt#L31

Yes it would be simpler and easier to link to the target (energyplus) and you would expect energyplus libs dependencies to be transitive, however jason found that is not the case with these object libraries. Certainly surprising behavior from cmake, but @lefticus researched this a lot so i have accepted it.

src/lib/CMakeLists.txt
kbenne: done.

src/lib/CMakeLists.txt
kbenne: Well the only mistake i can detect is that many of the conan library dependencies are still exposed to any consuming projects, meaning they have to provide those same conan libs. Note for example that the only consuming project in existence (OS App) has a conan file that replicates this project's conan file. I think me and @lefticus discussed on the phone and it is a big maybe if this is required. It depends if all of the symbols in the conan dependencies that are exposed in the api headers are included in the openstudio.dll. It is hard to guarantee that so to be safe we expose the conan libs as interface libraries.

tijcolem:
Gonna do a quick run of CI and then let's go ahead and get this PR in. I'll work on companion App [PR-75](https://github.com/NREL/OpenStudioApplication/pull/75) to verify SDK builds with App but those will probably be App related issues and not SDK.

tijcolem:
Seeing a few new ctest failures. This is on Linux. 

```
The following tests FAILED:
	167 - SqlFileFixture.AnnualTotalCosts (Failed)
	174 - ApplicationPathHelpers.Simple_test_forThisModule (Failed)
	175 - ApplicationPathHelpers.PathWhenSymlinkInPath_Setup (Failed)
```

PR #3899, opened 2020-02-28
Updated copyrights for 2020

axelstudios:
- Updated 2020 copyright dates
- Fixed mixed tabs/spaces


jmarrec:
@axelstudios I took the liberty of adding a ".git-blame-ignore-revs" file to try to and ignore all copyright-related changes so far. Whenever you use "git blame", you see a ton of lines blames because of that stuff while it's actually irrelevant to what you're usually trying to get at.

I am not 100% sure whether github will actually use it.

Locally you can do `$ git config blame.ignoreRevsFile .git-blame-ignore-revs` to register it.

axelstudios:
@jmarrec That's awesome, I didn't even know that was an option

PR #3912, opened 2020-03-07
Add consistently failing test for warnings/errorsa and document

lefticus:
- Objects of the same type share warning/error messages
   through the logger, because the logger is global
   and there is no perfect way to filter on object
 - The only known solution would end up missing some log messages
   (Those that come from LOG_FREE)
 - The current implementation always shares messages
   between objects of the same time on the same thread
 - The current implementation has a race condition in mulithreaded
   use cases.

The situation is documented as well as possible with the failing
test casees. This is a real problem that should be solved some
day.

Pull request overview
---------------------

 - DESCRIBE PURPOSE OF THIS PULL REQUEST

New test case fails intentionally as it demonstrates the extent of the problem with shared warnings and errors across objects of the same type.

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


tijcolem:
Cleaning things up and this is old PR.  I have disabled the test by commenting it out. It's well documented and can be enabled if we want.

PR #3929, opened 2020-03-24
Separate and package ruby mingw fetch commands

lefticus:
- needed for later consumption by OpenStudioApplication
   when building ModelEditor ruby bindings


lefticus:
Tied to: https://github.com/NREL/OpenStudioApplication/pull/89

tijcolem:
@lefticus  I built the SDK on this branch and tested with the App and your changes do indeed fix all the windows related build issues.   I also verified that osx and linux still work as expected.    Thanks for the great work!

PR #3946, opened 2020-04-12
Fix ruby windows crash with system ruby

lefticus:
Move to dynamic linking of openstudiolib.dll to ruby openstudio.dll to pretend like we have some control over the Undefined Behavior caused by ruby's definition of `fclose` and other system library functions.

### DIFF
--- a/ruby/openstudio.rb
+++ b/ruby/openstudio.rb
@@ -30,20 +30,19 @@
 # add binary dir to system path
 original_path = ENV['PATH']
 if /mswin/.match(RUBY_PLATFORM) or /mingw/.match(RUBY_PLATFORM)
-  front = []
-  back = []
-  original_path.split(';').each do |p|
-    if /SketchUp/.match(p)
-      front << p
-    else
-      back << p
-    end
+  cur_location = File.dirname(__FILE__)  
+  
+  build_type = File.basename(cur_location)
+
+  if build_type == "Debug" or build_type == "Release" or build_type == "RelWithDebugInfo" or build_type == "RelMinSize"
+    # in build dir execution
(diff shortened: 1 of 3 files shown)


kbenne:
I checked it out on Mac, and everything is working as expected. I tested packaged ruby bindings and cli, and I also tested both from the developer tree. I can confirm Mac builds are functioning exactly like they were before this change.

ruby/openstudio.rb
```diff
@@ -30,20 +30,19 @@
...
+  
+  build_type = File.basename(cur_location)
 
+  if build_type == "Debug" or build_type == "Release" or build_type == "RelWithDebugInfo" or build_type == "RelMinSize"
+    # in build dir execution
+    relative_dll_path = File.join(File.dirname(File.dirname(cur_location)), build_type)  
+    RubyInstaller::Runtime.add_dll_directory(relative_dll_path)
+  else 
+    # install dir execution
+    RubyInstaller::Runtime.add_dll_directory(File.join(File.dirname(cur_location), "lib"))
```
kbenne, line 43: Is RubyInstaller guaranteed to exist? 

Also there is a lib directory that is a sibling to the openstudio.rb file? I thought lib was one directory up.

lefticus: * As far as I can tell RubyInstaller exists on Windows, at least. It worked in all of my tests locally, and worked for ctest use case.
 * `lib` is in two different locations relative to openstudio.rb, depending on installed directory or build directory path.
    * ../Debug/lib ( `RubyInstaller::Runtime.add_dll_directory(relative_dll_path)` )
    * ../lib ( `RubyInstaller::Runtime.add_dll_directory(File.join(File.dirname(cur_location), "lib"))` )

PR #3950, opened 2020-04-18
Reduce mutable global state

lefticus:
Pull request overview
---------------------

Two goals:

 * reduce mutable global state as much as possible
 * make sure all global state is safely initialized in a thread safe way, using C++11's "magic statics" and eliminating multi-stage initialization of statics.

This code was not requested by anyone but was developed while exploring the threading issues in the OpenStudio logger.

It does not fix any particular bug, but does improve the code.


src/model/WaterHeaterStratified.cpp
```diff
@@ -84,102 +84,102 @@ namespace detail {
...
+
+        // This should be based on fuel type
+
+        //"Water Heater <Fuel Type> Rate",
+        //"Water Heater <Fuel Type> Energy",
+        // Fuel type specific
+        // TODO: DLM: the return type of this method needs to change to std::vector<std::string> in ModelObject
+        // until then, make this include all possible outputVariableNames for class regardless of fuelType
+        // std::string fuelType = this->fuelType();
+        // if (fuelType == "Electricity") {
```
jmarrec, line 127: This is an important comment. Some output variables in E+ are added conditionally based on the object's state.
I'm not sure this is worth penalizing all objects were this isn't the case though.

src/model/AirConditionerVariableRefrigerantFlow.cpp
```diff
@@ -91,7 +91,7 @@ namespace detail {
 
   const std::vector<std::string>& AirConditionerVariableRefrigerantFlow_Impl::outputVariableNames() const
   {
-    static std::vector<std::string> result{ "VRF Heat Pump Total Cooling Rate",
+    static const std::vector<std::string> result{ "VRF Heat Pump Total Cooling Rate",
```
jmarrec, line 94: A ton of changes (most of them probably) are similar to this one and deal with all children of ModelObject with regards to `outputVariableNames()`

src/utilities/idf/IdfObject.cpp
```diff
@@ -1325,6 +1325,8 @@ namespace detail {
       // different formatting for vertices
       if ((m_iddObject.properties().format == "vertices") && (m_iddObject.isExtensibleField(index))) {
         ExtensibleIndex eIndex = m_iddObject.extensibleIndex(index);
+        // making this static is a bad idea,
+        // it makes no sense in a threaded environment
```
jmarrec, line 1329: Should we change that then?

src/utilities/core/UUID.cpp
```diff
@@ -108,8 +108,8 @@ UUID toUUID(const std::string& str)
 }
 
 // Finds Version 4 uuid in a string including {}
-boost::regex &uuidInString() {
-  static boost::regex result("(\\{[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89aAbB][a-f0-9]{3}-[a-f0-9]{12}\\})");
+const boost::regex &uuidInString() {
+  static const boost::regex result("(\\{[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89aAbB][a-f0-9]{3}-[a-f0-9]{12}\\})");
```
jmarrec, line 112: :+1:

src/sdd/MapGeometry.cpp
```diff
@@ -119,8 +119,8 @@
 namespace openstudio {
 namespace sdd {
 
-  const double footToMeter =  0.3048;
-  const double meterToFoot = 1.0/0.3048;
+  constexpr double footToMeter =  0.3048;
+  constexpr double meterToFoot = 1.0/0.3048;
```
jmarrec, line 123: :+1:

src/utilities/data/Matrix.cpp
```diff
@@ -285,7 +285,7 @@ namespace openstudio{
     // ETH@20100120. What library does this come from? The user should be able to seed the
     // generator independently of this function.
     // seed random number generator
-    static std::minstd_rand generator(42u);
+    thread_local std::minstd_rand generator(42u);
```
jmarrec, line 288: Interesting.

PR #3958, opened 2020-04-23
Allow embedded gems even with --bundle

### LINKED ISSUE #3933: OpenStudio CLI usage with bundle with native ext
## Issue: 

Running the 3.0.0-rc1 CLI via with --bundle args, the cli does not load bundled gems with native extensions and throws an error. For example, this case below uses jaro_winkler 1.5.4 which has a native component. 

## Steps to Reproduce:

```
$ docker run -d nrel/openstudio:3.0.0-rc1
$ docker exec -it CONTAINER_ID bash
```
Create a simple Gemfile with jaro_winkler. 

```
$ echo -e "source 'https://rubygems.org'\ngem 'jaro_winkler', '1.5.4'" > Gemfile
$ gem install bundler -v '2.1.0'
$ bundle install -path .bundle --without test
```
Create a simple `test.rb` with following contents to test CLI with jaro_winkler. 

```
require 'jaro_winkler'
JaroWinkler.distance "MARTHA", "MARHTA"
JaroWinkler.distance "MARTHA", "marhta", ignore_case: true
JaroWinkler.distance "MARTHA", "MARHTA", weight: 0.2
puts JaroWinkler.jaro_distance "MARTHA", "MARHTA"
```

(using CLI directly with this works fine as jaro_winkler is available in CLI)
`$ openstudio test.rb`
0.9444444444444445

(Using here with bundle fails)
`$ openstudio --bundle Gemfile --bundle_path .bundle/ test.rb `
W, [2020-03-28T17:28:08.229044 #136] WARN -- : Bundle activated but ENV[‘BUNDLE_WITHOUT’] is not set
Warning: the running version of Bundler (2.1.0) is older than the version that created the lockfile (2.1.4). We suggest you to upgrade to the version that created the lockfile by running `gem install bundler:2.1.4`.
Error executing argv: [“--bundle”, “Gemfile”, “--bundle_path”, “.bundle/“, “test.rb”]
Error: Could not find jaro_winkler-1.5.4 in any of the sources in :/ruby/2.5.0/gems/bundler-2.1.0/lib/bundler/spec_set.rb:86:in `block in materialize’
:/ruby/2.5.0/gems/bundler-2.1.0/lib/bundler/spec_set.rb:80:in `map!'
:/ruby/2.5.0/gems/bundler-2.1.0/lib/bundler/spec_set.rb:80:in `materialize’
:/ruby/2.5.0/gems/bundler-2.1.0/lib/bundler/definition.rb:170:in `specs’
:/ruby/2.5.0/gems/bundler-2.1.0/lib/bundler/definition.rb:237:in `specs_for’
:/openstudio_cli.rb:557:in `parse_main_args’
:/openstudio_cli.rb:727:in `execute’
:/openstudio_cli.rb:1757:in `<main>'
eval:137:in `eval’
eval:137:in `require_embedded_absolute’
eval:110:in `block in require’
eval:104:in `each’
eval:104:in `require’
eval:3:in `<main>'

### LINKED ISSUE #3957: native extension gems dont work after bundle is updated
Native extension gems dont work when bundle is called again.
This breaks the server initialization scripts.
ex, pycall gem.

kbenne:
This is significant @tijcolem @nllong @brianlball

Previously, when --bundle is used then even the baked in libraries from openstudio-gems are made unavailable, with this change the baked in openstudio-gems (even native ones like pycall) are still available. Anything defined in --bundle will override the baked in stuff. So for instance it is possible to define a bundle with a custom version of the OpenStudio standards gem and that version of standards will be used instead of the baked in version.

close #3957
close #3933

Also, minor fix to annoying messages shown when NOT using --bundle...
"Ignoring jaro_winkler-1.5.4 because its extensions are not built. Try: gem pristine jaro_winkler --version 1.5.4"
This is not true as jaro_winkler is actually still available from the embedded system. The message has been silenced.

OpenStudio Server might want to change the policy of defining a bundle with all of the openstudio-gems and instead define a bundle such as openstudio-server-gems which may only define what server wants to override or add. That said I think Server can keep doing what it is doing and things will "just work".

This should specifically address @brianlball's issue about not being able to access pycall when using the --bundle option.

### DIFF
--- a/src/cli/embedded_help.rb
+++ b/src/cli/embedded_help.rb
@@ -53,7 +53,6 @@ module Kernel
   $LOAD_PATH << ':/ruby/2.5.0'
   $LOAD_PATH << ':/ruby/2.5.0/x86_64-darwin16'
   $LOAD_PATH << ':/ruby/2.5.0/x64-mswin64_140'
-  $LOAD_PATH << ':/ruby/2.5.0/bundler/gems/pycall.rb-3d4041b8fb9d/lib'
   # DLM: now done in embedded gem initialization section in openstudio_cli.rb
   #$LOAD_PATH << EmbeddedScripting::findFirstFileByName('openstudio-standards.rb').gsub('/openstudio-standards.rb', '')
   #$LOAD_PATH << EmbeddedScripting::findFirstFileByName('openstudio-workflow.rb').gsub('/openstudio-workflow.rb', '')
@@ -71,7 +70,7 @@ module Kernel
     'liboga' => 'init_liboga',\
     'sqlite3/sqlite3_native' => 'init_sqlite3_native',\
     'jaro_winkler_ext' => 'init_jaro_winkler_ext',\
-    'pycall.so' => 'init_pycall', \
+    'pycall.so' => 'init_pycall',\
     'pycall.dll' => 'init_pycall'
   }
 
@@ -86,10 +85,6 @@ module Kernel
   end
 
   def require path
-    if require_embedded_extension path
-      return true
-    end
-
     original_directory = Dir.pwd
     path_with_extension = path
 
@@ -126,12 +121,18 @@ module Kernel
       end
     end
 
-    result = original_require path
+    begin
+      result = original_require path
+    rescue Exception => e 
+      result = require_embedded_extension path
+      if not result
+        raise e
+      end
+    end
 
     current_directory = Dir.pwd
     if original_directory != current_directory
       Dir.chdir(original_directory)
-      puts "Directory changed from '#{original_directory}' to '#{current_directory}' while requiring '#{path}', result = #{result}, restoring original_directory"
       STDOUT.flush
     end
 
--- a/src/cli/embedded_help.rb
+++ b/src/cli/embedded_help.rb
@@ -85,58 +85,84 @@ module Kernel
   end
 
   def require path
+    result = false
     original_directory = Dir.pwd
     path_with_extension = path
 
-    if path.include? 'openstudio/energyplus/find_energyplus'
-      return false
-    end
-
-    # Returns the extension
-    # (the portion of file name in path starting from the last period).
-    extname = File.extname(path)
-
-    if extname.empty? or ! RUBY_FILE_EXTS.include? extname
-      path_with_extension = path + '.rb'
-    end
-
-    if path_with_extension.to_s.chars.first == ':'
-      if $LOADED.include?(path_with_extension)
-         return true
-      else
-        return require_embedded_absolute(path_with_extension)
-      end
-    elsif path_with_extension == 'openstudio.rb'
-      return true
-    else
-      $LOAD_PATH.each do |p|
-        if p.to_s.chars.first == ':'
-          embedded_path = p + '/' + path_with_extension
-          if $LOADED.include?(embedded_path)
-            return true
-          elsif EmbeddedScripting::hasFile(embedded_path)
-            return require_embedded_absolute(embedded_path)
-          end
-        end
-      end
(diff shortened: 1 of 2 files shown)


kbenne:
I tested this with the following ruby program:

```
require 'pycall/import'
include PyCall::Import
pyimport :math
puts math.sin(math.pi / 4) 
```

Using no --bundle
```
kbenne-27067s:build kbenne$ ./Products/openstudio  test.rb 
0.7071067811865475
```

And using --bundle
```
kbenne-27067s:build kbenne$ ./Products/openstudio --bundle openstudio-gems/Gemfile --bundle_path /Users/kbenne/Development/OpenStudio/build/openstudio-gems/ test.rb 
0.7071067811865475
```

You should trust but verify when I say that the --bundle version is using pycall from the bundle and the no --bundle version is using the baked in pycall.

tijcolem:
@kbenne Thanks for working on this! This will be a very useful feature for many users. 

One question I have is when using --bundle with openstudio what should be the behavior when native extension gems are in the Gemfile and bundle path.  For instance, if I use this [Gemfile](https://github.com/NREL/openstudio-gems/blob/develop/Gemfile) that contain a group of [native ext gems](https://github.com/NREL/openstudio-gems/blob/develop/Gemfile#L24-L29) and run the following: 
`$ bundle install --path .gems`
And then using your same pycall `test.rb` file 
```
require 'pycall/import'
include PyCall::Import
pyimport :math
puts math.sin(math.pi / 4) 
```
Then run openstudio via --bundle with all gems in [Gemfile](https://github.com/NREL/openstudio-gems/blob/develop/Gemfile)
`$ openstudio --bundle Gemfile --bundle_path .gems/ test.rb`
```
Error executing argv: ["--bundle", "Gemfile", "--bundle_path", ".gems/", "test.rb"]
Error: Unable to find a spec satisfying pycall (= 1.2.1) in the set. Perhaps the lockfile is corrupted? in :/ruby/2.5.0/gems/bundler-2.1.0/lib/bundler/spec_set.rb:39:in `block in for'
:/ruby/2.5.0/gems/bundler-2.1.0/lib/bundler/spec_set.rb:21:in `loop'
:/ruby/2.5.0/gems/bundler-2.1.0/lib/bundler/spec_set.rb:21:in `for'
:/ruby/2.5.0/gems/bundler-2.1.0/lib/bundler/definition.rb:237:in `specs_for'
:/openstudio_cli.rb:593:in `parse_main_args'
:/openstudio_cli.rb:714:in `execute'
:/openstudio_cli.rb:1744:in `<main>'
eval:174:in `eval'
eval:174:in `require_embedded_absolute'
eval:160:in `block in require_embedded'
eval:154:in `each'
eval:154:in `require_embedded'
eval:113:in `require'
eval:3:in `<main>'
```
Now if I omit these native gems by commenting them out of the Gemfile and then installing again. 
`$ rm -rf .gems`
`$ bundle install --path .gems`
`$ openstudio --bundle Gemfile --bundle_path .gems/ test.rb`
```
E, [2020-05-14T20:27:58.701854 #6646] ERROR -- : Error activating gem :/ruby/2.5.0/specifications/parser-2.7.1.0.gemspec
E, [2020-05-14T20:27:58.702479 #6646] ERROR -- : Error activating gem :/ruby/2.5.0/specifications/openstudio-extension-0.2.1.gemspec
0.7071067811865475
```
This works!  And openstudio can access the native pycall gem whereis previously it couldn't. I'm guessing this is the desired beahvior? Meaning you cannot use native ext gems in your bundle when using openstudio via --bundle.  And if this is the case, is it possible to gracefully handle this and display a warning to the user?

PR #3966, opened 2020-05-04
Addresses #3945, wrap SurfaceControl:MovableInsulation in OS

### LINKED ISSUE #3945: Wrap SurfaceControl:MovableInsulation in OS
Enhancement Request
--------------

Support the EnergyPlus object `SurfaceControl:MovableInsulation` in the OS SDK

## Detailed Description

The movable insulation object can be used to add insulation surface(s) on top of opaque surfaces, such as a transparent coating that affects the radiative gains to a roof surface. Below is an example EP object that would be generated from OS - the `TransparentInsulationMaterial` is a `WindowMaterial:Glazing` object defined elsewhere. 

```
SurfaceControl:MovableInsulation,
  Outside,                          ! Insulation Type
  Surface 14,                       ! Surface Name
  TransparentInsulationMaterial,    ! Material Name
  Always On Discrete;               ! Schedule Name
```

Attached is a text file of an idf that uses this object to apply a transparent insulation material on the roof surfaces:
[coating_idf.txt](https://github.com/NREL/OpenStudio/files/4453222/coating_idf.txt)

@joseph-robertson

joseph-robertson:
Pull request overview
---------------------

 - Closes #3945

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


src/model/SurfaceControlMovableInsulation.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+  bool setSurface(const Surface& surface);
+
+  bool setMaterial(const Material& material);
+
+  void resetMaterial();
+
+  bool setSchedule(Schedule& schedule);
+
+  void resetSchedule();
```
jmarrec, line 95: no reset

src/model/SurfaceControlMovableInsulation.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+  bool setInsulationType(const std::string& insulationType);
+
+  void resetInsulationType();
+
+  bool setSurface(const Surface& surface);
+
+  bool setMaterial(const Material& material);
+
+  void resetMaterial();
```
jmarrec, line 91: If we decide to not allow it to be in a broken state, there should be no reset here either.

src/model/SurfaceControlMovableInsulation.hpp
```diff
@@ -0,0 +1,127 @@
...
+  static IddObjectType iddObjectType();
+
+  /** @name Getters */
+  //@{
+
+  boost::optional<std::string> insulationType() const;
+
+  Surface surface() const;
+
+  boost::optional<std::string> surfaceName() const;
```
jmarrec, line 73: I don't think this is terribly useful, as you can call `surface().nameString()` and you should be certain to get a valid string since `surface()` correctly returns a Surface and not an optional

src/model/SurfaceControlMovableInsulation.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+  boost::optional<Material> material() const;
```
jmarrec, line 75: Perhaps non optional if not allowing a broken state.

src/model/PlanarSurface_Impl.hpp
```diff
@@ -177,6 +178,8 @@ namespace detail {
 
     std::vector<GeneratorPhotovoltaic> generatorPhotovoltaics() const;
 
+    std::vector<SurfaceControlMovableInsulation> surfaceControlMovableInsulation() const;
+
     std::vector<SurfacePropertyConvectionCoefficients> surfacePropertyConvectionCoefficients() const;
```
jmarrec, line 183: More of a side note at this point (didn't think about it for too long), but why are these two different:
* `std::vector<SurfaceControlMovableInsulation> PlanarSurface::surfaceControlMovableInsulation() const`
* `boost::optional<SurfaceControlMovableInsulation> Surface_Impl::surfaceControlMovableInsulation()`

And why isn't the latter calling the base class vector one if it makes sense?

src/model/Surface.cpp
```diff
@@ -886,6 +888,28 @@ namespace detail {
...
+      std::vector<SurfaceControlMovableInsulation> scmis;
+      for (auto& scmi : allscmis) {
+          boost::optional<std::string> surfaceName = scmi.surfaceName();
+          if (surfaceName) {
+            Surface surface = scmi.surface();
+            if (surface.handle() == handle()) {
+              scmis.push_back(scmi);
+            }
+          }
+      }
```
jmarrec, line 902: Cf PlanarSurface:

```
    std::vector<SurfaceControlMovableInsulation> PlanarSurface_Impl::surfaceControlMovableInsulation() const
    {
      return getObject<ModelObject>().getModelObjectSources<SurfaceControlMovableInsulation>();
    }
```

src/energyplus/ReverseTranslator/ReverseTranslateSurfaceControlMovableInsulation.cpp
```diff
@@ -0,0 +1,110 @@
...
+  target = workspaceObject.getTarget(openstudio::SurfaceControl_MovableInsulationFields::SurfaceName);
+  if (target){
+    OptionalModelObject modelObject = translateAndMapWorkspaceObject(*target);
+    if (modelObject){
+      if (modelObject->optionalCast<Surface>()){
+        surface = modelObject->cast<Surface>();
+      }
+    }
+  }
+
```
jmarrec, line 75: if not surface => LOG_AND_THROW or log and return boost::none or something.

You're dereferencing it later without any checks during the Ctor call on line 98, which will make it throw in case the IDF object wasn't valid.

src/energyplus/ReverseTranslator/ReverseTranslateSurfaceControlMovableInsulation.cpp
```diff
@@ -0,0 +1,110 @@
...
+  boost::optional<Schedule> schedule;
+  target = workspaceObject.getTarget(openstudio::SurfaceControl_MovableInsulationFields::ScheduleName);
+  if (target){
+    OptionalModelObject modelObject = translateAndMapWorkspaceObject(*target);
+    if (modelObject){
+      if (modelObject->optionalCast<Schedule>()){
+        schedule = modelObject->cast<Schedule>();
```
jmarrec, line 93: Move the `   surfaceControlMovableInsulation.setSchedule(*schedule); ` here, to avoid dereferencing something that doesn't exist. if you model Ctor defaults it to alwaysOnDiscrete, it'll still have the alwaysOnDiscrete so object will be valid anyways.

src/energyplus/ReverseTranslator/ReverseTranslateSurfaceControlMovableInsulation.cpp
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }
+
+  boost::optional<Material> material;
+  target = workspaceObject.getTarget(openstudio::SurfaceControl_MovableInsulationFields::MaterialName);
+  if (target){
+    OptionalModelObject modelObject = translateAndMapWorkspaceObject(*target);
+    if (modelObject){
+      if (modelObject->optionalCast<Material>()){
+        material = modelObject->cast<Material>();
```
jmarrec, line 82: Either `   surfaceControlMovableInsulation.setMaterial(*material); ` gets set here.

But if not allowing a broken state, then you do want to both check if Surface exists and material too.

Then 

```c++
if (!surface.has_value() || !material.has_value()) {
 // LOG_AND_THROW or return
}

SurfaceControlMovableInsulation surfaceControlMovableInsulation(*surface, *material);
```

src/energyplus/ForwardTranslator/ForwardTranslateSurfaceControlMovableInsulation.cpp
```diff
@@ -0,0 +1,86 @@
...
+  boost::optional<Schedule> schedule = modelObject.schedule();
+
+  if (!(insulationType && material && schedule)) {
+    LOG(Error, "SurfaceControlMovableInsulation for Surface '" << surface.nameString() << "' missing required fields, will not be translated");
+    return boost::none;
+  }
+
+  idfObject.setString(SurfaceControl_MovableInsulationFields::InsulationType, insulationType.get());
+
+  idfObject.setString(SurfaceControl_MovableInsulationFields::SurfaceName, modelObject.surfaceName().get());
```
jmarrec, line 69: `modelObject.surface().nameString()` is probably fine.

src/energyplus/ForwardTranslator/ForwardTranslateSurfaceControlMovableInsulation.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+  boost::optional<std::string> insulationType = modelObject.insulationType();
+  Surface surface = modelObject.surface();
+  boost::optional<Material> material = modelObject.material();
+  boost::optional<Schedule> schedule = modelObject.schedule();
+
+  if (!(insulationType && material && schedule)) {
+    LOG(Error, "SurfaceControlMovableInsulation for Surface '" << surface.nameString() << "' missing required fields, will not be translated");
+    return boost::none;
+  }
```
jmarrec, line 65: insulationType and schedule probably need to be non optional and therefore not checked here. Material depends, but at least you're maintaining only valid objects going to the IDF :+1: !

src/model/PlanarSurface_Impl.hpp
joseph-robertson: This is true for `surfacePropertyConvectionCoefficients` too right? I'm not sure why the two are different. I just used `surfacePropertyConvectionCoefficients` as a guide...

jmarrec: Yeah true for `surfacePropertyConvectionCoefficients()` as well. 

I would assume that one was probably copying `generatorPhotovoltaics()` a couple of lines above in `PlanarSurface`.

Except one object is called `Generator:Photovoltaic` and generatorPhotovoltaics is supposed to really return an array. The other is called `SurfaceProperty:ConvectionCoefficients` (plural) and it's should have returned a `boost::optional<SurfacePropertyConvectionCoefficients> `object

jmarrec: An alternative is that an earlier version might have returned both SurfacePropertyConvectionCoefficients and SurfacePropertyConvectionCoefficientsMultipleSurface

https://github.com/NREL/OpenStudio/blob/8ae4d883ed1a6ba704aa8b60003646de69a00f39/src/model/PlanarSurface.hpp#L235-L236

Anyways, I don't think it's right.

joseph-robertson:
@jmarrec A couple of things:
* Every time this object is reverse translated, it will now create (and orphan) a schedule. Is this behavior ok? Should we be deleting the old schedule every time `setSchedule` is called and the argument passed in EQUALS the old schedule?
* What does the `ScheduleTypeRegistry` actually do? I have an entry for the surface control schedule, but it doesn't appear to have any effect on the model.

src/model/SurfaceControlMovableInsulation.cpp
```diff
@@ -0,0 +1,239 @@
...
+    bool result = setString(OS_SurfaceControl_MovableInsulationFields::MaterialName, "");
+    OS_ASSERT(result);
+  }
+
+  bool SurfaceControlMovableInsulation_Impl::setSchedule(Schedule& schedule) {
+    bool result = ModelObject_Impl::setSchedule(OS_SurfaceControl_MovableInsulationFields::ScheduleName,
+                                                "SurfaceControlMovableInsulation",
+                                                "Resistance Modifier Fraction",
+                                                schedule);
+    return result;
```
jmarrec, line 152: Missing getScheduleTypeKeys for the ScheduleRegistry to work. @joseph-robertson

jmarrec:
> @jmarrec A couple of things:
> 
>     * Every time this object is reverse translated, it will now create (and orphan) a schedule. Is this behavior ok? Should we be deleting the old schedule every time `setSchedule` is called and the argument passed in EQUALS the old schedule?

The Reverse Translator do this:
* If the "Schedule Name" field has a string
* Try to translate the IDF object referenced in the "Schedule Name"
* Check that resulting modelObject is a schedule (which if translation worked, it IS)
 * If so, assign it.

If the `SurfaceControlMoveableInsulation` is defaulting to `alwaysOffContinuousSchedule` for eg (see below), then before it tries to assign the schedule it's using a default and "unique" schedule, so that's not a problem at all.

>     * What does the `ScheduleTypeRegistry` actually do? I have an entry for the surface control schedule, but it doesn't appear to have any effect on the model.

You're missing `SurfaceControlMoveableInsulation_Impl::getScheduleTypeKeys` for the `ScheduleRegistry` to work correctly.

Looking at the `ScheduleRegistry` modification you've made which is correct per the E+ I/O ref, the model ctor should default to `alwaysOnContinuousSchedule` or `alwaysOffContinuousSchedule` (not the discrete ones) since it's continuous.

PR #3970, opened 2020-05-07
Bump to conan fmt/6.2.0, and use the recipe option to define FMT_HEADER_ONLY

jmarrec:
Pull request overview
---------------------

Bump to conan fmt/6.2.0, and use the recipe option to define FMT_HEADER_ONLY instead of doing it in all files.

@lefticus You added conan fmt to these files when replacing QString, so a quick look from you would make sense please.

If we do want to use `FMT_HEADER_ONLY` (perhaps we don't), then we should use this option to the conan recipe.

https://github.com/conan-io/conan-center-index/blob/f65f7e668d39431a2d1330b49646f648220fcb04/recipes/fmt/all/conanfile.py#L87-L88

### DIFF
--- a/ConanInstall.cmake
+++ b/ConanInstall.cmake
@@ -38,6 +38,8 @@ if(NOT CONAN_OPENSTUDIO_ALREADY_RUN)
   # set(BOOST_VERSION "1.71.0")
 
   list(APPEND CONAN_OPTIONS "zlib:minizip=True")
+  list(APPEND CONAN_OPTIONS "fmt:header_only=True")
+
   # You do want to rebuild packages if there's a newer recipe in the remote (which applies mostly to our own openstudio_ruby where we don't
   # bump the actual package version when we make changes) than the binaries were built with
   # 'outdated' also acts like 'missing': if no binary, will build them.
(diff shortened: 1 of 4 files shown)


jmarrec:
Linux and mac look pretty happy. Windows choked for another reason: Unable to delete 'D:\jenkins\openstudio\develop\build\Testing-Windows-Server-2019'. Tried 3 times (of a maximum of 3) waiting 0.1 sec between attempts.

lefticus:
Agreed that the original code didn't make sense, no reason to link it and use it in `HEADER_ONLY` mode. 

I think we should just use it as a compiled library. My original commits were arguably just bad. In `HEADER_ONLY` mode it is **considerably** slower to compile.

ConanInstall.cmake
```diff
@@ -38,6 +38,8 @@ if(NOT CONAN_OPENSTUDIO_ALREADY_RUN)
   # set(BOOST_VERSION "1.71.0")
 
   list(APPEND CONAN_OPTIONS "zlib:minizip=True")
+  list(APPEND CONAN_OPTIONS "fmt:header_only=True")
```
lefticus, line 41: Recommend that we remove this line and just use it as a proper library. It's much better compile times.

lefticus: If we cannot get {fmt} linking without this, then keep it in place. It's at least only one thing that has to be changed in the future.

tijcolem:
Windows ran now. I fixed that CI issue. Unit tests look good so I think we can merge this one in.

jmarrec:
I rebased onto develop (there was a conflict) **and force pushed**, and removed the `fmt:header_only=True`, will see if CI can link against {fmt} as a library or not.

jmarrec:
Seems like CI was able to build on all three platforms.

2 Unit tests failing are the usual suspects on mac (bundle tests)

PR #3980, opened 2020-05-14
Fix #2033 - Properly connected the secondary/induced air node for ATUs

### LINKED ISSUE #2033: AirTerminalSingleDuctConstantVolumeFourPipeInduction does not always connect properly
If you have an AirLoopHVAC instance with a single demand branch that contains only an AirTerminalSingleDuctConstantVolumeFourPipeInduction (no zone), then add a zone, the terminal will not be connected properly. The recirculation loop between the zone exhaust node and the terminal's secondary air inlet will not be hooked up. This is almost certainly true for other terminals that have secondary air inlet (i.e. PIUs). The reason for this is that terminals depend on their ::addToNode to make these connections, but adding a thermal zone when the terminal already exists bypasses the terminal addToNode logic. The cleanest solution that does not duplicate code (consolidates on the sometimes complex logic of addToNode), is to remove the terminal from the loop, add the zone, then re-add the terminal.

jmarrec:
Pull request overview
---------------------

 - Fixes #2033
 - Properly connected the secondary/induced air node for ATUs

Three terminals are concerned:

* `AirTerminalSingleParallePIUReheat`: this one already works because of a special case that's existing in `ThermalZone::addToNode` (cf review comment below)

* `AirTerminalSingleDuctSeriesPIUReheat`
* `AirTerminalSingleDuctConstantVolumeFourPipeInduction`

Though #2033 mentioned a way of avoid duplicate code, which is to disconnect the ATU first, I just fixed it following the pattern for `AirTerminalSingleParallePIUReheat`

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/model/ThermalZone.cpp
+++ b/src/model/ThermalZone.cpp
@@ -33,6 +33,10 @@
 #include "AirLoopHVACSupplyPlenum_Impl.hpp"
 #include "AirTerminalSingleDuctParallelPIUReheat.hpp"
 #include "AirTerminalSingleDuctParallelPIUReheat_Impl.hpp"
+#include "AirTerminalSingleDuctSeriesPIUReheat.hpp"
+#include "AirTerminalSingleDuctSeriesPIUReheat_Impl.hpp"
+#include "AirTerminalSingleDuctConstantVolumeFourPipeInduction.hpp"
+#include "AirTerminalSingleDuctConstantVolumeFourPipeInduction_Impl.hpp"
 #include "AirLoopHVACReturnPlenum.hpp"
 #include "AirLoopHVACReturnPlenum_Impl.hpp"
 #include "ZoneHVACEquipmentList.hpp"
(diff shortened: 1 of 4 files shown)


src/model/ThermalZone.cpp
```diff
@@ -2122,6 +2126,38 @@ namespace detail {
                               terminal.get(),
                               terminal->secondaryAirInletPort() );
             }
```
jmarrec, line 2128: Existing bit for `AirTerminalSingleDuctParallePIUReheat`

src/model/ThermalZone.cpp
```diff
@@ -2122,6 +2126,38 @@ namespace detail {
...
+              _model.connect( t_exhaustPortList,
+                              t_exhaustPortList.nextPort(),
+                              inducedAirInletNode,
+                              inducedAirInletNode.inletPort() );
+
+              _model.connect( inducedAirInletNode,
+                              inducedAirInletNode.outletPort(),
+                              terminal.get(),
+                              terminal->inducedAirInletPort() );
+            }
```
jmarrec, line 2160: Fix for SeriesPIU and FourPipeInduction

src/model/ThermalZone.cpp
jmarrec: https://github.com/NREL/OpenStudio/blob/050075639a5e1f0c8118184672d9f2ae7c50b926/src/model/ThermalZone.cpp#L2109-L2125

PR #4008, opened 2020-06-30
#1906 - Use sqlite3_bind* to ensure proper escaping of parameters such as Zone Names etc

### LINKED ISSUE #1906: Standard Report fails on model with zones with apostrophes in name
1. Create a model
2. Put a ' in the name of the zone
3. Fails on performing a sql query

This may be true of other characters as well...probably should escape the queries somehow.

jmarrec:
Pull request overview
---------------------

 - Fixes #1906

It was a though decision trying to decide whether investigating using a modern C++ library for sqlite3 (such as SQLiteCpp) was a good idea or to implement something myself, but given the effort already in place I chose to expand on it.

I've basically extracted the PreparedStatement struct that was defined in `SqlFile_Impl.cpp` into its own header file and expanded on it, and made every query go through it. I'm using variadic templates (...) to provide an additional PreparedStatement constructor that looks like the original one, but also allows to pass any bind parameters (and the correct sqlite3_bind_XXX is deduced like before based on the type of the current parameter).

High level usage is not using PreparedStatement directly anyways, so the difference is this:

Before:

```c++
sqlFile.execAndReturnFirstInt("SELECT Value from Table WHERE RowName = '" + rowName + "' AND ColumnName = '" + rowName + "';"); 
```

Before:

```c++
sqlFile.execAndReturnFirstInt("SELECT Value from Table WHERE RowName = ? AND ColumnName = ?;", rowName, columnName); 
```

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


src/utilities/sql/Test/SqlFile_GTest.cpp
```diff
@@ -656,3 +656,110 @@ TEST_F(SqlFileFixture, SqlFile_LeapYear)
...
+
+    std::vector<std::string> envPeriods = sql.availableEnvPeriods();
+    EXPECT_EQ(1u, envPeriods.size());
+
+    auto _s = sql.execAndReturnFirstString("SELECT name FROM sqlite_master WHERE type='table' AND name='EnvironmentPeriods';");
+    EXPECT_TRUE(_s) << "Table EnvironmentPeriods doesn't exists";
+
+    //  This finishes the 'insert into zones' statement; calls drop table; then adds a comment so the rest is ignored. Classic SQL Injection
+    std::string zoneName {"Little Bobby Tables', 0, 0, 0, 0, 1, 1, 1, 3, 1, 1, 0, 2, 0, 2, 0, 2, 2, 8, 3, 3, 4, 4, 2, 2, 1); DROP TABLE EnvironmentPeriods;--"};
+    EXPECT_NO_THROW(sql.insertZone(zoneName,
```
jmarrec, line 735: Show the #1906 problem

src/model/ModelObject.cpp
```diff
@@ -344,29 +345,30 @@ namespace detail {
...
+      std::string rowCheckQuery = R"(
+        SELECT RowName FROM TabularDataWithStrings
+          WHERE ReportName = 'Initialization Summary'
+          AND ReportForString = 'Entire Facility'
+          AND TableName = 'Component Sizing Information'
+          AND RowName = ?
+          AND Value = ?;)";
+      boost::optional<std::string> rowValueName = model().sqlFile().get().execAndReturnFirstString(rowCheckQuery,
+          // bindArgs
+          rowName, valueNameAndUnits);
```
jmarrec, line 357: typical example of usage now

src/utilities/sql/PreparedStatement.hpp
```diff
@@ -0,0 +1,315 @@
...
+    if (nPlaceholders != size) {
+      throw std::runtime_error("Wrong number of placeholders [" + std::to_string(nPlaceholders) + "] versus bindArgs [" + std::to_string(size) + "].");
+      return false;
+    } else {
+      int i = 1;
+      // Call bind(i++, args[i]) until any returns false
+      bool ok = (bind(i++, args) && ...);
+      return ok;
+    }
+  }
```
jmarrec, line 157: bindAll calls the right "bind" overload based on type.

src/utilities/sql/PreparedStatement.hpp
```diff
@@ -0,0 +1,315 @@
...
+  bool bindAll(Args&&... args) {
+    const std::size_t size = sizeof...(Args);
+    int nPlaceholders = sqlite3_bind_parameter_count(m_statement);
+    if (nPlaceholders != size) {
+      throw std::runtime_error("Wrong number of placeholders [" + std::to_string(nPlaceholders) + "] versus bindArgs [" + std::to_string(size) + "].");
+      return false;
+    } else {
+      int i = 1;
+      // Call bind(i++, args[i]) until any returns false
+      bool ok = (bind(i++, args) && ...);
```
jmarrec, line 154: This is the fold expression used to do it for each arg

src/utilities/sql/PreparedStatement.hpp
```diff
@@ -0,0 +1,315 @@
...
+      throw std::runtime_error("Error executing SQL statement step");
+    }
+  }
+
+  // Executes a **SINGLE** statement
+  int execute() {
+    int code = sqlite3_step(m_statement);
+    sqlite3_reset(m_statement);
+    return code;
+  }
```
jmarrec, line 173: The non variadic functions could be put into a cpp file, but I didn't do it yet.

src/utilities/sql/SqlFile.hpp
```diff
@@ -35,6 +35,7 @@
 #include "SummaryData.hpp"
 #include "SqlFileDataDictionary.hpp"
 #include "SqlFileEnums.hpp"
+#include "SqlFile_Impl.hpp"
```
jmarrec, line 38: That's the thing I dislike the most. Because I use variadic templates (to forward to the Impl variadic functions) in this file, I need to define them in the header file. And therefore I must include the SqlFile_Impl.hpp becase otherwise `m_impl` is not a complete type with just a forward declaration...

src/utilities/sql/SqlFile.hpp
```diff
@@ -59,9 +60,9 @@ class EpwFile;
 class Calendar;
 class SqlFileTimeSeriesQuery;
 
-namespace detail {
-  class SqlFile_Impl;
-}
+//namespace detail {
+  //class SqlFile_Impl;
+//}
```
jmarrec, line 65: previous forward declaration

src/utilities/sql/SqlFile.hpp
```diff
@@ -816,26 +817,79 @@ class UTILITIES_API SqlFile {
...
-  boost::optional<double> execAndReturnFirstDouble(const std::string& statement) const;
+  // Variadic arguments are the bind arguments if any, to replace '?' placeholders in the statement string
+  template<typename... Args>
+  boost::optional<double> execAndReturnFirstDouble(const std::string& statement, Args&& ... args) const {
+    boost::optional<double> result;
+    if (m_impl){
+      result = m_impl->execAndReturnFirstDouble(statement, std::forward<Args>(args)...);
+    }
+    return result;
+  }
```
jmarrec, line 830: m_impl is used here... hence why the include SqlFile_Impl is needed

src/utilities/sql/SqlFile_Impl.cpp
```diff
@@ -234,78 +233,22 @@ namespace openstudio{
       close();
     }
 
-    struct PreparedStatement
+    void SqlFile_Impl::execAndThrowOnError(const std::string &t_stmt)
```
jmarrec, line 236: execAndThrowOnError is different, relies on sqlite3_exec to execute several statement, so no binding parameters possible

src/utilities/sql/SqlFile_Impl.cpp
```diff
@@ -378,31 +334,17 @@ namespace openstudio{
...
       return m_connectionOpen;
     }
 
     int SqlFile_Impl::getNextIndex(const std::string &t_tableName, const std::string &t_columnName)
     {
-      boost::optional<int> maxindex = execAndReturnFirstInt("select max(" + t_columnName + ") from " + t_tableName);
+      // Interestingly, you CANNOT bind any database identifier (such as the table name / column name) but only litteral values...
+      // boost::optional<int> maxindex = execAndReturnFirstInt("SELECT MAX( ? ) FROM ?", t_columnName, t_tableName);
+      std::string query = "SELECT MAX(" + t_columnName + ") FROM " + t_tableName + ";";
+      boost::optional<int> maxindex = execAndReturnFirstInt(query);
```
jmarrec, line 347: One where I couldn't use parameters, so no escaping, but it's actually safe here, since table/column names are E+ defined and safe

src/utilities/sql/SqlFile_Impl.cpp
```diff
@@ -542,38 +484,9 @@ namespace openstudio{
...
-        << t_extGrossWallArea << ", "
-        << t_extNetWallArea << ", "
-        << t_extWindowArea << ", "
-        << t_isPartOfTotalArea << ");";
-
-      execAndThrowOnError(insertZone.str());
-
+      execAndThrowOnError("insert into zones (ZoneIndex, ZoneName, RelNorth, OriginX, OriginY, OriginZ, CentroidX, CentroidY, CentroidZ, OfType, Multiplier, ListMultiplier, MinimumX, MaximumX, MinimumY, MaximumY, MinimumZ, MaximumZ, CeilingHeight, Volume, InsideConvectionAlgo, OutsideConvectionAlgo, FloorArea, ExtGrossWallArea, ExtNetWallArea, ExtWindowArea, IsPartOfTotalArea) values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)",
+          // Bind Args
+          zoneIndex, t_name, t_relNorth, t_originX, t_originY, t_originZ, t_centroidX, t_centroidY, t_centroidZ, t_ofType, t_multiplier, t_listMultiplier, t_minimumX, t_maximumX, t_minimumY, t_maximumY, t_minimumZ, t_maximumZ, t_ceilingHeight, t_volume, t_insideConvectionAlgo, t_outsideConvectionAlgo, t_floorArea, t_extGrossWallArea, t_extNetWallArea, t_extWindowArea, t_isPartOfTotalArea);
```
jmarrec, line 489: That one is extremely readable, but the original was no better.

src/utilities/sql/SqlFile_Impl.cpp
```diff
@@ -659,7 +572,7 @@ namespace openstudio{
         stmt1->bind(++b, dayOfMonth);
         stmt1->bind(++b, hours);
 
-        stmt1->execute();
+        stmt1->execAndThrowOnError();
```
jmarrec, line 575: Right now I'm not sure why I did this.

src/utilities/sql/SqlFile_Impl.cpp
```diff
@@ -3769,22 +3438,13 @@ namespace openstudio{
...
-      sqlite3_finalize(sqlStmtPtr);
-
-      return index;
+      // TODO: haven't figured out how to bind properly to the LIKE statement, tried
+      // LIKE ?
+      // LIKE %?%  => sqlite3_prepare_v2 fails
+      // LIKE %%?%%
+      // LIKE '%?%' (not recognized as a placeholder)
+      const std::string& s = "SELECT MapNumber FROM DaylightMaps WHERE MapName LIKE '%" + name + "%'";
+      return execAndReturnFirstInt(s);
```
jmarrec, line 3447: One case where I didn't figure out how to use binding...

src/utilities/sql/SqlFile_Impl.hpp
```diff
@@ -738,26 +739,84 @@ namespace openstudio{
...
       void illuminanceMap(const int& hourlyReportIndex, std::vector<double>& x, std::vector<double>& y, std::vector<double>& illuminance) const  ;
 
-      // execute a statement and return the first (if any) value as a double
-      boost::optional<double> execAndReturnFirstDouble(const std::string& statement) const;
+      // execute a statement and return the first (if any) value as a double.
+      // Variadic arguments are the bind arguments if any, to replace '?' placeholders in the statement string
+      template<typename... Args>
+      boost::optional<double> execAndReturnFirstDouble(const std::string& statement, Args&& ... args) const {
+        if (m_db) {
+          PreparedStatement stmt(statement, m_db, false, args...);
```
jmarrec, line 747: Not sure whether I should std::forward<Args> here actually.

src/utilities/sql/SqlFile.hpp
```diff
@@ -816,26 +817,79 @@ class UTILITIES_API SqlFile {
...
 
+
   /// execute a statement and return the first (if any) value as a double
-  boost::optional<double> execAndReturnFirstDouble(const std::string& statement) const;
+  // Variadic arguments are the bind arguments if any, to replace '?' placeholders in the statement string
+  template<typename... Args>
+  boost::optional<double> execAndReturnFirstDouble(const std::string& statement, Args&& ... args) const {
+    boost::optional<double> result;
+    if (m_impl){
+      result = m_impl->execAndReturnFirstDouble(statement, std::forward<Args>(args)...);
```
jmarrec, line 827: Not sure whether I should use `std::forward<Args>(args)...` here actually. Same comment as in SqlFile_Impl.hpp

jmarrec:
A little test I've done manually with the bindings too:

```

z = m.getThermalZones[0]
z.setName("A zone with an apostrophe' in it")

w = OpenStudio::WorkflowJSON.new
m.setWorkflowJSON(w)
w.setOswPath("./in.osw")
w.setSeedFile("model.osm")
w.setWeatherFile('in.epw')

m.save('model.osm')
w.save

#####
# openstudio run -w in.osw
##### After run

m = osload('model.osm')
sqlFile = OpenStudio::SqlFile.new('run/eplusout.sql')
m.setSqlFile(sqlFile)

z = m.getThermalZones[0]
z.isConditioned

# OS 3.0.0: [openstudio.model.ThermalZone] <1> Query for Object of type 'OS:ThermalZone' and named 'A zone with an apostrophe' in it' isConditioned failed.
# This PR: Works fine, and z.isConditioned.get => "Yes"
```

jmarrec:
**TODO**

* Run the OpenStudio Results measure (and others) with this PR to ensure nothing broke.

jmarrec:
@DavidGoldwasser Could you do testing with this PR please?

tijcolem:
Looks like the osx-incremental has a build error.  https://ci.commercialbuildings.dev/blue/organizations/jenkins/openstudio-incremental-osx/detail/PR-4008/3/pipeline

It could be a result of the incremental build repo state. I kicked off a full build now so we can see the results of this. 
https://ci.commercialbuildings.dev/blue/organizations/jenkins/openstudio-develop-nightly/detail/openstudio-develop-nightly/596/pipeline

jmarrec:
It's a real error, clang appears to not like a function signature that gcc is fine with.

```
/Users/jenkins/git/OpenStudioIncremental/develop/src/model/../utilities/sql/PreparedStatement.hpp:74:101: error: missing default argument on parameter 'args'

  PreparedStatement(const std::string &t_stmt, sqlite3 *t_db, bool t_transaction = false, Args&&... args)

```

It's complaining because I have a parameter with a default, which is supposed to come last. But packs (`Args&& ... args`) also have to come last.

@lefticus If you have some remaining time on OS, could you help out please?

src/model/ModelObject.cpp
kbenne: This is much better. I like.

src/utilities/sql/SqlFile.hpp
kbenne: well the Impls are breaking through all over the place as a matter of practice, because of how casting works. I'm wondering more about how these variadic templates look to swig. Have you evaluated this api from the Ruby point of view?

kbenne:
The main thing that is holding this up from my point of view is the following reoccurring build error.
```

In file included from /Users/jenkins/git/OpenStudioIncremental/develop/src/model/../utilities/sql/SqlFile_Impl.hpp:39:
/Users/jenkins/git/OpenStudioIncremental/develop/src/model/../utilities/sql/PreparedStatement.hpp:74:101: error: missing default argument on parameter 'args'
  PreparedStatement(const std::string &t_stmt, sqlite3 *t_db, bool t_transaction = false, Args&&... args)
```

Secondly, I'm wondering how this looks from the Measure point of view in Ruby. Surely these variadic templates aren't "just working" are they?

I think this is impressive work. I like the interface from a C++ point of view and it is way better than building up string streams.

lefticus:
@jmarrec Good looking work.

clang is definitely correct on this one, and I'm guessing a newer GCC would complain as well.

So many different things could bind to that `bool` for `t_transaction` that I think the API is problematic anyhow, consider:

`PreparedStatement("statement", connection, "string1", "string2");`

That `"string1"` will pretty easily get interpreted as the `bool`.

My initial reaction is that you should do something like

```cpp
enum class Transaction {
  true,
  false
};

// force the user to choose transaction or not and make it more strongly typed.
// slightly more typing but much safer
PreparedStatement(const std::string &stmt, sqlite3 *t_db, Transaction transaction, Args && ... args);
```

`PreparedStatement("statement", connection, Transaction::true, "string1", "string2");`

What do you think?

jmarrec:
> @jmarrec Good looking work.
> 
> clang is definitely correct on this one, and I'm guessing a newer GCC would complain as well.
> 
> So many different things could bind to that `bool` for `t_transaction` that I think the API is problematic anyhow, consider:
> 
> `PreparedStatement("statement", connection, "string1", "string2");`
> 
> That `"string1"` will pretty easily get interpreted as the `bool`.
> 
> My initial reaction is that you should do something like
> 
> ```c++
> enum class Transaction {
>   true,
>   false
> };
> 
> // force the user to choose transaction or not and make it more strongly typed.
> // slightly more typing but much safer
> PreparedStatement(const std::string &stmt, sqlite3 *t_db, Transaction transaction, Args && ... args);
> ```
> 
> `PreparedStatement("statement", connection, Transaction::true, "string1", "string2");`
> 
> What do you think?

Yeah that's what I had in mind, minus the strong typing, which is a good idea.

We need to be careful about the impact it will have on the bindings (ruby in particular). There are tons of measures who use `sqlFile.execAndReturnXXX` which I've messed with as well. I haven't had time to test how ruby handles the pack but if it's too awful we could use something like a vector/list/array of strings with a default of empty or something (in Python I'd do that as `def __init__(self, stmt, t_db, transation=False, args=None)`). Do you have an opinion on that?

lefticus:
> 
> We need to be careful about the impact it will have on the bindings (ruby in particular). There are tons of measures who use `sqlFile.execAndReturnXXX` which I've messed with as well. I haven't had time to test how ruby handles the pack but if it's too awful we could use something like a vector/list/array of strings with a default of empty or something (in Python I'd do that as `def __init__(self, stmt, t_db, transation=False, args=None)`). Do you have an opinion on that?

Last time I personally worked on the ruby bindings was before we had variadic templates. *But* knowing how they worked with ChaiScript, you're going to have to come up with some fixed concept of what's allowed. Likely just a list of strings as you stated.

jmarrec:
@lefticus Thanks! did consider just linking sqlite3 to openstudiolib but figured I should try to avoid it since after all it was working without it before. Is this a regression?

lefticus:
> 
> 
> @lefticus Thanks! did consider just linking sqlite3 to openstudiolib but figured I should try to avoid it since after all it was working without it before. Is this a regression?

From what I can tell this is new behavior because sqlite3 was not previously part of the public interface, and it is now. We could probably work within what you have to remove it from the public interface again, by writing some simple shim functions that call back into a new .cpp file.

tijcolem:
@DavidGoldwasser Would you be able to test out these new installer packages to confirm that it works when using Ruby? I think this is pretty much ready to merge from an C++ implementation point of view.  Great work to @jmarrec on this.

DavidGoldwasser:
Looks great, I extended an OpenStudio Results test I have that used periods in construction name to use a wide variety of special characters including ~`@#$%^&{['";:,? at the beginning, middle, and end of zones or construction names.

tijcolem:
@DavidGoldwasser Thanks for testing!   I think this PR is ready to go.

tijcolem (approved):

PR #4023, opened 2020-07-22
Addresses #3936, subSurface Type reset when assigned to surface

### LINKED ISSUE #3936: SubSurface Type reset when assigned to surface
Issue overview
--------------
Running the following script:
```
require 'openstudio'

surf_polygon = OpenStudio::Point3dVector.new
surf_polygon << OpenStudio::Point3d.new(0, 0, 0)
surf_polygon << OpenStudio::Point3d.new(0, 10, 0)
surf_polygon << OpenStudio::Point3d.new(0, 10, 10)
surf_polygon << OpenStudio::Point3d.new(0, 0, 10)
surf = OpenStudio::Model::Surface.new(surf_polygon, model)

door_polygon = OpenStudio::Point3dVector.new
door_polygon << OpenStudio::Point3d.new(0, 1, 1)
door_polygon << OpenStudio::Point3d.new(0, 9, 1)
door_polygon << OpenStudio::Point3d.new(0, 9, 9)
door_polygon << OpenStudio::Point3d.new(0, 1, 9)
subsurf = OpenStudio::Model::SubSurface.new(door_polygon, model)
subsurf.setSubSurfaceType('Door')

puts subsurf.subSurfaceType # should be Door, and is Door

subsurf.setSurface(surf)

puts subsurf.subSurfaceType # should still be Door, but is now FixedWindow
```

I expect it to say Door twice, but instead it produces:
```
Door
FixedWindow
```

I can workaround this by setting the subsurface type _after_ the door is assigned to the surface.

### Environment

 - Platform (Operating system, version): Windows 10
 - Version of OpenStudio (if using an intermediate build, include SHA): OpenStudio 2.9.1

joseph-robertson:
Pull request overview
---------------------

 - Fixes #3936

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


jmarrec:
@joseph-robertson @kbenne 

I wonder if there is really any interest in actually calling `assignDefaultSubSurfaceType()`. We could just do this instead, which would perhaps be less surprising?

```
  std::string SubSurface_Impl::subSurfaceType() const {
    std::string result;
    if (isSubSurfaceTypeDefaulted()) {
        result = this->defaultSubSurfaceType();
    } else {
       result = getString(OS_SubSurfaceFields::SubSurfaceType,true,true));
    }

    return result;
  }
```

(The FixedWindow IDD Default would be pointless in that case).

joseph-robertson:
@jmarrec 

Can you explain this further? What would that block of code replace? What could be the surprise if it's left the way it is now?

jmarrec:
> Can you explain this further? What would that block of code replace? What could be the surprise if it's left the way it is now?

Well, the API is actually **assigning** a value based on logic, which is what I would call a (smart) default. There will be no way of knowing whether the user actually **assigned** a value or if it was automatically assigned by Openstudio.
That would also mean that the logic to default the `subSurfaceType` would happen the first time you call `setSurface` but not the second. I doubt that you would do that often, but you might, if doing it in the bindings and messing it up the first time.

eg:

Make two surfaces: a wall, and a ceiling
Make a subSurface, do not explicitly set the subSurfaceType

# First time
subSurface.setSurface(Wall) => the subSurfaceType was blank, so assignDefault is called and it is now **harcoded** to "FixedWindow"

# Second time
Oops, I messed up, Let's correct it: subSurface.setSurface(Ceiling) => the subSurfaceType being hardcoded already, it is **not** changed so I end up with a "FixedWindow" on a ceiling, instead of a "Skylight".

Perhaps it's not too big of a deal in E+ context though: I did check the FT, and it appears that the result in E+ will end up fine, as the only thing that matters for E+ is 'Door', 'GlassDoor' or 'Window'.


----

Side note: This blocks need review, and definitely corrections to the "changing SubSurfaceType to Dloor/GlassDoor" warnings

https://github.com/NREL/OpenStudio/blob/ff0e898ff96e0653e1b72da4200a3403b927fdaa/src/energyplus/ForwardTranslator/ForwardTranslateSubSurface.cpp#L71-L93

Perhaps we should also double check that the flow control is handling every case... And we should probably catch all other situations (but Door/GlassDoor) where a non-opaque type is used and the construction is opaque (eg FixedWindow, but using an Opaque construction... though perhaps that should be done in model API)

jmarrec:
Side note: This blocks need review, and definitely corrections to the "changing SubSurfaceType to Dloor/GlassDoor" warnings

https://github.com/NREL/OpenStudio/blob/ff0e898ff96e0653e1b72da4200a3403b927fdaa/src/energyplus/ForwardTranslator/ForwardTranslateSubSurface.cpp#L71-L93

https://github.com/NREL/OpenStudio/blob/ff0e898ff96e0653e1b72da4200a3403b927fdaa/src/energyplus/ForwardTranslator/ForwardTranslateSubSurface.cpp#L86-L88

Should be: `LOG(Warn, "SubSurface '" << modelObject.name().get() << "' uses fenestration construction, changing SubSurfaceType to GlassDoor");`

Perhaps we should also double check that the flow control is handling every case... And we should probably catch all other situations (but Door/GlassDoor) where a non-opaque type is used and the construction is opaque (eg FixedWindow, but using an Opaque construction... though perhaps that should be done in model API)

jmarrec:
@kbenne Can you review https://github.com/NREL/OpenStudio/pull/4023#issuecomment-668433729 and decide whether we choose change things or just go with this PR please? I am not extremely opinionated on the subject so it'd be great to have another pair of eyes.

src/energyplus/ForwardTranslator/ForwardTranslateSubSurface.cpp
```diff
@@ -84,10 +84,10 @@ boost::optional<IdfObject> ForwardTranslator::translateSubSurface( model::SubSur
     idfObject.setString(FenestrationSurface_DetailedFields::ConstructionName, construction->name().get());
 
     if (subSurfaceType == "Door" && construction->isFenestration()){
-      LOG(Warn, "SubSurface '" << modelObject.name().get() << "' uses fenestration construction, changing SubSurfaceType to Door");
+      LOG(Warn, "SubSurface '" << modelObject.name().get() << "' uses fenestration construction, changing SubSurfaceType to GlassDoor");
       subSurfaceType = "GlassDoor";
     } else if (subSurfaceType == "GlassDoor" && !construction->isFenestration()){
-      LOG(Warn, "SubSurface '" << modelObject.name().get() << "' uses non-fenestration construction, changing SubSurfaceType to GlassDoor");
+      LOG(Warn, "SubSurface '" << modelObject.name().get() << "' uses non-fenestration construction, changing SubSurfaceType to Door");
```
jmarrec, line 90: :+1:

PR #4026, opened 2020-07-26
Fix OpenStudio Windows crashed during save attempt 

macumber:
Pull request overview
---------------------

 - Currently `replaceDir` will throw an uncaught exception if the user has the save directory (or child subdirectory) open in Windows explorer.  This change no longer deletes the save directory if that exists, it clears all contents inside of it.  Code attempts to sync as many changes as possible and does not generate uncaught exceptions.

### Pull Request Author

 - [ x ] Checked behavior in OpenStudioApplication, adjusted policies as needed (`src/openstudio_lib/library/OpenStudioPolicy.xml`)

### Review Checklist

 - [ ] Labeling is ok


jmarrec:
Windows is offline, and mac is failing on the usual suspect `	2366 - EnergyPlusFixture.ForwardTranslatorTest_MultiThreadedLogMessages (Failed)
`

src/model/FileOperations.cpp
```diff
@@ -123,15 +126,19 @@ namespace model {
     }
 
     // Now we recreate it, and it must work!
-    if (!openstudio::filesystem::create_directory(dest))
-    {
-      LOG_FREE(Error, "replaceDir", "Could not create destination directory: " << toString(dest));
-      return false;
+    if (!openstudio::filesystem::exists(dest)) {
```
jmarrec, line 129: @macumber As far as I can tell you're actually no longer trying to remove the directory itself above (should you be after you tried removing everything inside?), so this block is pointless no?

macumber: No I think you can hit this if you open a bare OSM file with no companion directory, the companion directory would be created in temp and then would be synched back.

jmarrec: Good point!

src/model/FileOperations.cpp
```diff
@@ -100,21 +102,22 @@ namespace model {
...
-    {
+    // If the dest dir exists, remove all entries in it
+    if (openstudio::filesystem::exists(dest)) {
       if (openstudio::filesystem::is_directory(dest)) {
-        try {
-          openstudio::filesystem::remove_all(dest);
-        } catch (const std::exception &e) {
-          LOG_FREE(Error, "replaceDir", "Destination directory could not be removed: " << toString(dest) << "error: " << e.what());
-          return false;
+        for (openstudio::filesystem::directory_iterator end_dir_it, it(dest); it != end_dir_it; ++it) {
```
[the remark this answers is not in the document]
macumber: Yeah I think that would work too, but the main thing is don't throw an unhandled exception if we try to delete something (file or folder) and it fails.  Just log the file and return false.  We should try to clean out as many files as we can.

tijcolem:
@macumber Thanks for the fix for this! I know many app users will be very happy as this has been frustrating when saving your work only to have it crash.  This looks good to drop to me.

Issue #4043, opened 2020-08-12
Fix or disable failing unit tests


jmarrec:
Issue overview
--------------

At least these tests are failing **sometimes** (but often enough to be really annoying). They should be fixed or disabled.

* [ ] EnergyPlusFixture.ForwardTranslatorTest_MultiThreadedLogMessages (Failed): 
    * this is known, cf https://github.com/NREL/OpenStudio/pull/3912 
* BCLFixture: I think this looks like a timeout issue from web requests... they fail after 60s of not completing. May be a firewall thing.
    * [ ] BCLFixture.RemoteBCLTest (SIGTRAP)
    * [ ] BCLFixture.GetComponentByUID (SIGTRAP)

@kbenne @joseph-robertson @tijcolem seen any other?

The list of tests that CI has filtered out is this FYI:

```
  failingTests = [
    'ModelFixture.Space_intersectSurfaces_degenerate1',
    'ModelFixture.Space_intersectSurfaces_degenerate2',
    'ModelFixture.Space_intersectSurfaces_degenerate3',
    'ModelFixture.ThreeJSReverseTranslator_FloorplanJS_SimAUD_Paper',
    'CLITest-test_encodings-encoding',
    'CLITest-test_encodings-encoding2'
  ];
```

## Details

```
616/3011 Testing: BCLFixture.GetComponentByUID
616/3011 Test: BCLFixture.GetComponentByUID
Command: "/Users/jenkins/git/OpenStudioIncremental/develop/build/Products/openstudio_utilities_tests" "--gtest_filter=BCLFixture.GetComponentByUID"
Directory: /Users/jenkins/git/OpenStudioIncremental/develop/build/src/utilities
"BCLFixture.GetComponentByUID" start time: Aug 12 05:22 MDT
Output:
----------------------------------------------------------
Running main() from gmock_main.cc
Note: Google Test filter = BCLFixture.GetComponentByUID
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from BCLFixture
[ RUN      ] BCLFixture.GetComponentByUID
/Users/jenkins/git/OpenStudioIncremental/develop/src/utilities/bcl/test/BCL_GTest.cpp:327: Failure
Value of: component
  Actual: false
Expected: true
<end of output>
Test time =  60.19 sec
----------------------------------------------------------
Test Failed.
"BCLFixture.GetComponentByUID" end time: Aug 12 05:23 MDT
[...]

PR #4047, opened 2020-08-14
Pin the actual recipe revisions for all conan dependencies

jmarrec:
Pro: we have zero chance of pulling something else. Fully reproducible builds

Cons: whenever we want to update a dependency, we'll have to bump the revision in OpenStudio/ConanInstall.cmake

If we do not pin dependencies, we could simply upload a new recipe revision, and delete the old one from the remote, or pass UPDATE to conan cmake run. I don't like passing UPDATE to conan_cmake_run actually  because it has to query bintray for changes everytime you run cmake (which takes time, doesn't work offline, etc)

Note: the pinning of deps really matter a lot of openstuido_ruby but not so much for the rest of the deps. Because we're unlikely going to bump a dep like boost for no reason without also bumping the version (eg boost/1.74...)

### DIFF
--- a/ConanInstall.cmake
+++ b/ConanInstall.cmake
@@ -37,16 +37,30 @@ if(NOT CONAN_OPENSTUDIO_ALREADY_RUN)
   #conan_add_remote(NAME jmarrec
   #  URL https://api.bintray.com/conan/jmarrec/testing)
 
+  # Enable revisions in conan
+  execute_process(
+    COMMAND ${CONAN_CMD} config get general.revisions_enabled
+    OUTPUT_VARIABLE CONAN_REV_STATUS
+    ERROR_VARIABLE CONAN_REV_STATUS
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+  message(STATUS "Conan: config get general.revisions_enabled=${CONAN_REV_STATUS}")
+  if (NOT "${CONAN_REV_STATUS}" STREQUAL "True")
+    message(AUTHOR_WARNING "Conan: Force enabling revisions (conan config set general.revisions_enabled=True)")
+    execute_process(COMMAND ${CONAN_CMD} config set general.revisions_enabled=True)
+  endif()
+
   list(APPEND CONAN_OPTIONS "zlib:minizip=True")
   # TODO:  list(APPEND CONAN_OPTIONS "fmt:header_only=True")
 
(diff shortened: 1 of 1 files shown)


ConanInstall.cmake
```diff
@@ -37,16 +37,30 @@ if(NOT CONAN_OPENSTUDIO_ALREADY_RUN)
   #conan_add_remote(NAME jmarrec
   #  URL https://api.bintray.com/conan/jmarrec/testing)
 
+  # Enable revisions in conan
```
jmarrec, line 40:
```suggestion
  # Enable revisions in conan: check if they are already enabled, if not do it an warn user
```

jmarrec:
@tijcolem we'll have to wipe the entire conan cache on the build agents) to make this work, because there are conflicts with the packages already there BEFORE the `general.revisions_enabled=True`

ConanInstall.cmake
```diff
@@ -57,29 +71,30 @@ if(NOT CONAN_OPENSTUDIO_ALREADY_RUN)
...
-    bzip2/1.0.8
+    #"bzip2/1.0.8#d4a5c7144832d75fc3f349c5346160b0"
+    #"libyaml/0.2.5#9e234874df88c3ba7249f6d1368fceaf"
     BASIC_SETUP CMAKE_TARGETS NO_OUTPUT_DIRS
     OPTIONS ${CONAN_OPTIONS}
     BUILD ${CONAN_BUILD}
     # Passes `-u, --update`    to conan install: Check updates exist from upstream remotes
     # That and build=outdated should ensure we track the right
-    UPDATE
+    # UPDATE
```
jmarrec, line 97:
```suggestion
    UPDATE
```

Actually it seems that UPDATE is needed to go check on the remote to see if we can find the right revisions in case we have some revisions already and they don't match. Too bad.

jmarrec:
Mac is just failing because of the usual suspect (cf #4043)
```
	2368 - EnergyPlusFixture.ForwardTranslatorTest_MultiThreadedLogMessages (Failed)
```

tijcolem:
I cleared the ~/.conan/data on the remote agents and replayed the CI.  It sounds like this may not have been necessary due to your recent change using `UPDATE` but I thought we'd start w/o any conan data to make sure things work as expected.   The nightly build agents do not have any conan cache so we don't need to do anything with those.

PR #4065, opened 2020-09-04
Fix AVMList clone between different models

### LINKED ISSUE #4033: Limited functionality of clone AirLoopHVAC
Issue overview
--------------
My use case is cloning from one model to another model, not within a model. I may find more issues, but I wanted to file with this initial problem.

- [ ] AvailabilityManagerAssignmentList isn't cloned with the air loop. This results in crash of app when go to control tab for the air loop, or failed export to IDF. This may be general clone Loop issue and not specific to plant loops. I have not tested plant loops yet.

## Current Behavior

Screen Shot 2020-08-05 at 10 23 33 AM

## Expected Behavior
Clone should result in valid air loop 

## Steps to Reproduce

1.
2.
3.
4.

## Possible Solution

Here is my ruby work around, not proposed solution, but shows what needs to change to make the loop valid
```
    # create and assign AvailabilityManagerAssignmentList to cloned air loops
    model.getAirLoopHVACs.sort.each do |air_loop|
      if air_loop.getString(4).to_s == ""
        list = OpenStudio::Model::AvailabilityManagerAssignmentList.new(air_loop)
        # line above adds list but doesn't associate it with air_loop. Line below is a work around to force it in
        air_loop.setString(4,list.handle.to_s)
      end
    end
```

### Environment

### Context
Clone air loops from one model to another as part of effort to combine many OSM's to one OSM.

### LINKED ISSUE #4034: Limited functionality of clone PlantLoop
Issue overview
--------------
Similar issue to Issue #4033 with missing availability manger assignment list in cloned plant loop.

## Steps to Reproduce

1.
2.
3.
4.

## Possible Solution
The work around is similar to air loop bud different field to change in the plant loop.
```
    # todo - code below here should be removed when SDK updated to handle this with clone plant loop method
    # create and assign AvailabilityManagerAssignmentList to cloned plant loops
    model.getPlantLoops.sort.each do |plant_loop|
      if plant_loop.getString(22).to_s == ""
        list = OpenStudio::Model::AvailabilityManagerAssignmentList.new(plant_loop)
        # line above adds list but doesn't associate it with air_loop. Line below is a work around to force it in
        plant_loop.setString(22,list.handle.to_s)
      end
    end
```

### Environment

kbenne:
close #4033
close #4034

Pull request overview
---------------------

 - DESCRIBE PURPOSE OF THIS PULL REQUEST

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/model/AvailabilityManagerAssignmentList.hpp
+++ b/src/model/AvailabilityManagerAssignmentList.hpp
@@ -60,6 +60,8 @@ class MODEL_API AvailabilityManagerAssignmentList : public ModelObject {
    */
   explicit AvailabilityManagerAssignmentList(const Loop& loop);
 
+  explicit AvailabilityManagerAssignmentList(const Model& model);
+
   virtual ~AvailabilityManagerAssignmentList() {}
 
   //@}
--- a/src/model/AvailabilityManagerAssignmentList.hpp
+++ b/src/model/AvailabilityManagerAssignmentList.hpp
@@ -39,6 +39,8 @@ namespace model {
 namespace detail {
 
   class AvailabilityManagerAssignmentList_Impl;
+  class AirLoopHVAC_Impl;
+  class PlantLoop_Impl;
 
 } // detail
 
@@ -59,8 +61,6 @@ class MODEL_API AvailabilityManagerAssignmentList : public ModelObject {
    * At least until we implement AVMLists for ZoneHVAC equipment
    */
   explicit AvailabilityManagerAssignmentList(const Loop& loop);
-
-  explicit AvailabilityManagerAssignmentList(const Model& model);
 
   virtual ~AvailabilityManagerAssignmentList() {}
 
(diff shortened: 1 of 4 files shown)


kbenne:
what do you think @jmarrec?

src/model/AvailabilityManagerAssignmentList.hpp
```diff
@@ -60,6 +60,8 @@ class MODEL_API AvailabilityManagerAssignmentList : public ModelObject {
    */
   explicit AvailabilityManagerAssignmentList(const Loop& loop);
 
+  explicit AvailabilityManagerAssignmentList(const Model& model);
```
jmarrec, line 63: Can we make it protected/private? Or at least ignore in the bindings... I don't like the fact that this becomes public, as users shouldn't have any business creating orphaned ones.

kbenne: yeah. I can do that. note that even if you use `AvailabilityManagerAssignmentList(const Loop& loop)` the new AVM is still orphaned. The loop argument is actually only used to set the name of the AVM. This itself should probably be considered a bug. https://github.com/NREL/OpenStudio/blob/fa9a77bf173556a2a044f64789a7f4f01b0df076/src/model/AvailabilityManagerAssignmentList.cpp#L409

jmarrec: I suppose to some extent and to echo #4072 AvailabilityManagerAssignmentList doesn't _really_ have any business having a name in the OSM since the user isn't meant to interact with it... It could be created at FT time.

kbenne: @jmarrec you are certainly welcome to add to the list of objects in #4077 to "un-name" If it is not user facing then I think it is fair game. For those that have non uuid names, and perhaps even those that do, I would anticipate a performance advantage.

kbenne: Certainly users will appreciate having nice names in the finished EnergyPlus idf. AVMlist named after the loop it belongs to etc.

PR #4066, opened 2020-09-04
Multiple shading controls referenced by a single subsurface

joseph-robertson:
Pull request overview
---------------------

 - Related EnergyPlus PR: https://github.com/NREL/EnergyPlus/pull/8196
 - SDK proposal from @shorowit: https://github.com/NREL/OpenStudio/issues/4042
 - [x] TODO: write VT
 - [x] TODO: ensure VT tests look ok 
 - [x] TODO: update model tests
 - [x] TODO: ensure model tests ok
 - [x] TODO: ensure FT and tests look ok
 - [x] TODO: ensure RT and tests look ok
 - [x] Add Release notes in https://github.com/NREL/OpenStudio/blob/develop/developer/doc/ReleaseNotes/OpenStudio_Release_Notes_3_1_0.md
 - [ ] Does Radiance FT need update?
 - [ ] Does the logic in the ForwardTrnaslator.cpp need to be updated: https://github.com/NREL/OpenStudio/pull/4066/files/005f0a48fa0934614d79ae3b6026702e2f7d6c6a#diff-2db10a6b3649fcfe9504db5c88d82d70
 - [x] Test C# bindings on windows

SubSurface
 - `boost::optional<ShadingControl> shadingControl() const` (OS_DEPRECATED; only returns if there is 1 shading control)
 - `std::vector<ShadingControl> shadingControls() const` (NEW)
 - `unsigned int numberofShadingControls() const` (NEW)
 - `bool setShadingControl(const ShadingControl& shadingControl)` (OS_DEPRECATED; will remove any existing shading controls and then add the one shading control)
 - `void resetShadingControl()` (OS_DEPRECATED; will remove any existing shading controls)
 - `bool addShadingControl(const ShadingControl& shadingControl)` (NEW)
 - `bool addShadingControls(const std::vector<ShadingControl> &shadingControls)` (NEW)
 - `void removeShadingControl(const ShadingControl& shadingControl)` (NEW)
 - `void removeAllShadingControls()` (NEW)

ShadingControl

 - `unsigned int numberofSubSurfaces() const` (NEW)
 - `bool addSubSurface(const SubSurface& subSurface)` (NEW)
 - `bool addSubSurfaces(const std::vector<SubSurface> &subSurfaces)` (NEW)
 - `void removeSubSurface(const SubSurface& subSurface)` (NEW)
 - `void removeAllSubSurfaces()` (NEW)
- `bool glareControlIsActivate() const` (NEW)
[...]

PR #4103, opened 2020-09-29
Update ShadingControl

joseph-robertson:
Pull request overview
---------------------

 - **DO NOT MERGE**
 - Prevent adding different-zone sub surfaces on a single shading control

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


src/model/ShadingControl.cpp
```diff
@@ -463,17 +467,29 @@ namespace model {
...
+                if (thermalZone->handle() != thermalZone2->handle()) {
+                  LOG(Warn, briefDescription() << ": cannot add two surfaces that are in different Thermal Zones ('" << thermalZone->nameString()
+                                               << "' versus existing '" << thermalZone2->nameString() << "')");
+                  return false;
+                }
+              }
+            }
+          }
+        }
+      }
```
jmarrec, line 485: I thought @shorowit said that this is actually fine now to have a multiple zones referenced in the same ShadingControl? https://github.com/NREL/OpenStudio/pull/4066#issuecomment-694289272

src/energyplus/ForwardTranslator.cpp
```diff
@@ -358,49 +358,26 @@ namespace energyplus {
...
-              it->second.push_back(clone);
-              clone.additionalProperties().setFeature("Shading Control Sequence Number", (int)it->second.size());
+            auto it = zoneHandleToShadingControlVectorMap.find(zoneHandle);
+            if (it == zoneHandleToShadingControlVectorMap.end()) {
+              zoneHandleToShadingControlVectorMap.insert(std::make_pair(zoneHandle, std::vector<ShadingControl>()));
             }
+            it = zoneHandleToShadingControlVectorMap.find(zoneHandle);
+            OS_ASSERT(it != zoneHandleToShadingControlVectorMap.end());
+            it->second.push_back(shadingControl);
+            shadingControl.additionalProperties().setFeature("Shading Control Sequence Number", (int)it->second.size());
```
jmarrec, line 380: The FT change seems to make sense though. What's weird and not handled (I have no clue how to properly handle that though) is the case where you have something like:

* Shading Control 1 controls Zone1-SurfaceA1
* Shading Control 2 controls Zone2-SurfaceA2 and Zone2-SurfaceA2

Then what happens to the Shading Control Sequence Number? Shading Control 2 would ahve a Shading Control Sequence Number of 2 (assuming it wasn't renamed, since the sort is done on name...), but Zone2-SurfaceA2 is referenced by only one shading control with a sequence number of 2. Does that make sense (I genuinely don't know).


@shorowit thoughts please? It's been a while I've used ShadingControls; and never since the E+ 9.4 changes, and I don't think the E+ API got any clearer in this specific case...

src/model/ShadingControl.cpp
shorowit: I'm not an expert on this object and there are so many ways to use it that it becomes very confusing. But as far as I know, it has always been fine to have subsurfaces of multiple zones referenced by a single ShadingControl, at least under some circumstances. When we did this, we used the "OnIfScheduleAllows" [shading control type](https://bigladdersoftware.com/epx/docs/9-4/input-output-reference/group-thermal-zone-description-geometry.html#field-shading-control-type). If you instead used a control type that depends on the [thermal zone](https://bigladdersoftware.com/epx/docs/9-4/input-output-reference/group-thermal-zone-description-geometry.html#field-zone-name-29-001) specified, I don't know what would happen.

src/energyplus/ForwardTranslator.cpp
shorowit: I have no experience with the Shading Control Sequence Numbers, sorry. (I've never used the ShadingControl object in a way where the sequence numbers mattered.)

joseph-robertson:
@jmarrec @shorowit See https://github.com/NREL/OpenStudio/pull/4066#issuecomment-868760359.

joseph-robertson:
Closing for now. Issue captured in https://github.com/NREL/OpenStudio/issues/4602.

PR #4115, opened 2020-10-07
4109 sql parse crash

### LINKED ISSUE #4109: 3.1.0-rc1 Ruby bindings on Windows sql parse segfaults
This issue originated and discovered by @mdahlhausen. See the issue/problem below. 

--------------
Using ruby bindings on Windows segfaults when parsing eplusout.sql results file. 

## Expected Behavior
It should parse the eplusout.sql or fail gracefully if the eplusout.sql file is corrupt and/or returning incorrect info (e.g. bad query). 

## Steps to Reproduce
Using openstudio 3.1.0-rc1](https://github.com/NREL/OpenStudio/releases/tag/v3.1.0-rc1) run the following script using ruby bindings. See attached eplusout.zip (unzip first). 

```
require 'openstudio'
model = OpenStudio::Model::Model.new()
sql_fle_path = "#{File.dirname(__FILE__)}/eplusout.sql"
sql_file = OpenStudio::SqlFile.new(OpenStudio::Path.new(sql_fle_path))
model.setSqlFile(sql_file)
error_query = "SELECT ErrorMessage FROM Errors WHERE ErrorType in(1,2)"
errs = model.sqlFile.get.execAndReturnVectorOfString(error_query)
```
[eplusout.zip](https://github.com/NREL/OpenStudio/files/5334732/eplusout.zip)

### Environment

 - Windows 10 Pro x64
-  [Ruby 2.5.5 ](https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-2.5.5-1/rubyinstaller-devkit-2.5.5-1-x64.exe)

 - [OpenStudio-3.1.0-rc1+062e7a8260-Windows.exe](https://github.com/NREL/OpenStudio/releases/download/v3.1.0-rc1/OpenStudio-3.1.0-rc1+062e7a8260-Windows.exe)

lefticus:
Pull request overview
---------------------

 - Fixes #4109


src/utilities/UtilitiesSql.i
```diff
@@ -41,4 +41,32 @@
 
 // order that these are loaded matters, e.g TimeSeries in data needs to know about DateTime in time
 %include <utilities/sql/SqlFile.i>
+
+%{
```
jmarrec, line 45: http://www.swig.org/Doc4.0/SWIG.html#SWIG_nn44 seems to place it in a `%init %{ ... }%` block. Is this on purpose that this doesn't have the `%init `  specifier?

lefticus: It was on purpose. The documentation and the code samples do not agree with each other. It says some targets do not support `%init` so you might need to use a `static` variable, then goes on to put the `static` inside of an `%init`! If the target doesn't support `%init` then that should have no impact, right?

I figured I would do the ultimately most generic option because of C# on Windows, just to be safe.

tijcolem:
@lefticus Confirmed via internal testing that this fix works!  This is great. 

Already merged to the v3.1.0-rc3 branch which will go to develop.

Issue #4126, opened 2020-10-21
SqlFile Crash


kbenne:
A bug was recently identified and fixed when using the SqlFile class from Ruby on windows https://github.com/NREL/OpenStudio/issues/4109

More recently it has been uncovered that the same issue exists anytime the SqlFile class is used from code compiled into a DLL and then linked to the OpenStudio.dll. This happens because of a couple of aligning facts.

1. The sqlite3 library depends on some global data
2. Much of the SqlFile implementation is now inlined
3. On windows the global data used by sqlite3 is actually scoped to the dll boundary and not in fact truly global

The proposed fix is to refactor all of the sqlite3 api calls into cpp source files, and remove these api calls from the inline header implementation.

lefticus:
See #4128 for a proposed fix.

PR #4128, opened 2020-10-22
V3.1.0 fixed sqlite3 linkage

lefticus:
Addresses #4126 

 - Move implementation of PreparedStatement into .cpp file
 - Remove hack for previous ruby fix
 - Remove now unnecessary extra linkage to sqlite3

Note: this has been tested to make sure the previous Ruby / Sqlite bug had not regressed.

### DIFF
--- a/src/utilities/sql/PreparedStatement.hpp
+++ b/src/utilities/sql/PreparedStatement.hpp
@@ -44,7 +44,7 @@ struct sqlite3_stmt;
 
 namespace openstudio {
 
-struct PreparedStatement
+class UTILITIES_API PreparedStatement
 {
 private:
   sqlite3 *m_db;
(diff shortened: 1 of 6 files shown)


src/utilities/sql/PreparedStatement.hpp
```diff
@@ -33,114 +33,61 @@
...
   sqlite3 *m_db;
   sqlite3_stmt *m_statement;
   bool m_transaction;
 
-  PreparedStatement & operator=(const PreparedStatement&) = delete;
-  PreparedStatement(const PreparedStatement&) = delete;
+  struct InternalConstructor { };
+
+  // a helper constructor to delegate to. It has the first parameter as
+  // InternalConstructor so it's distinguishable from the on with the defaulted
```
jmarrec, line 57:
```suggestion
  // InternalConstructor so it's distinguishable from the one with the defaulted
```

jmarrec (approved):
That looks really good! 

I added two very minor comments.

PR #4130, opened 2020-10-22
In SqlFile trap case when EnergyPlus year is 0

### LINKED ISSUE #4129: SqlFile::timeSeries broken when year = 0 in EnergyPlus database
There is code to catch this in OpenStudio for EP version < 9.2, but the database Time table's year column will still be populated with year = 0 for design days, even in EnergyPlus 9.4. The OpenStudio catch for this scenario is disabled for EP versions above 9.2, resulting in an exception. Code in question is here https://github.com/NREL/OpenStudio/blob/db4dab8965f82770872928a17c79789ea67c4a1d/src/utilities/sql/SqlFile_Impl.cpp#L448

kbenne:
close #4129

### DIFF
--- a/src/utilities/sql/SqlFile_Impl.cpp
+++ b/src/utilities/sql/SqlFile_Impl.cpp
@@ -447,16 +447,16 @@ namespace openstudio{
       // m_hasYear & m_hasIlluminanceMapYear are both default initialized to true
       if (version < VersionString(9, 2)) {
         m_hasIlluminanceMapYear = false;
+      }
 
-        if (version < VersionString(8, 9)) {
+      if (version < VersionString(8, 9)) {
+        m_hasYear = false;
+      } else {
+        // Check if zero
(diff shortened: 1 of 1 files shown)


src/utilities/sql/SqlFile_Impl.cpp
```diff
@@ -447,16 +447,16 @@ namespace openstudio{
...
+      }
 
-        if (version < VersionString(8, 9)) {
+      if (version < VersionString(8, 9)) {
+        m_hasYear = false;
+      } else {
+        // Check if zero
+        boost::optional<int> maxYear = execAndReturnFirstInt("SELECT MAX(Year) FROM Time");
+        if (!maxYear.is_initialized() || maxYear.get() <= 0) {
+          LOG(Warn, "Using EnergyPlusVersion version " << version.str() << " which should have 'Year' field, but it's always zero");
```
jmarrec, line 458: Ok, that can't really hurt to be more careful here. 
Initially I added that becaus the 9.0 was problematic, others would work fine.

jmarrec (approved):

PR #4138, opened 2020-11-09
Fix EnergyPlus md5 check for Ubuntu 20.4

kbenne:
@jmarrec or @tijcolem Someone did a really nice job adding logic to support Ubuntu 18.04 and 20.04, and I'm guessing it was one of you two although I did not check the history. I just updated my Linux computer to 20.04 so this was very nice. For me though, I needed to make this small change.


CMakeLists.txt
```diff
@@ -573,7 +573,7 @@ if(UNIX)
     set(ENERGYPLUS_EXPECTED_HASH DOES_NOT_EXIST_02ed618aab65bf46c96bf1faf2e1905b)
     set(ENERGYPLUS_PLATFORM "Redhat-x86_64")
   else()
-    if(LSB_RELEASE_ID_SHORT MATCHES "20.04")
+    if(LSB_RELEASE_VERSION_SHORT MATCHES "20.04")
```
jmarrec, line 576: :+1: I forgot to update that

src/measure/OSArgument.cpp
```diff
@@ -863,15 +863,15 @@ bool OSArgument::setStringInternal(OSArgumentVariant& variant, const std::string
       auto const double_val = std::stod(value, nullptr);
       variant = double_val;
       result = true;
-    } catch (std::exception e) {
+    } catch (std::exception & e) {
```
jmarrec, line 866: :+1: https://stackoverflow.com/questions/6755991/catching-stdexception-by-reference

jmarrec (approved):

PR #4139, opened 2020-11-09
chore: Bulk update using clang-format style options 

### LINKED ISSUE #3806: clang-format script feedback and adjustments
#3721 fixed #3718, by adding a .clang-format file, and @tijcolem added as a pipeline on jenkins.

**Fully enforcing coding styles / passing clang-format test will require reformatting the entire codebase in one go** to avoid having diffs popping up in files you touched (since the script added on CI only diffs files you actually changed in your PR) but in areas of the file you didn't modify: this would clutter the diffs and therefore the review process.

**Before we resort to that, we should monitor closely what the clang-format script is proposing as changes on PRs that are open, and adjust the policies to better align with the coding style we do want to enforce. This issue is meant to centralize this feedback/tweaking.**

* If you do run ./ci/clang-format.sh locally on one of your PRs, please commit the result in a specific commit so we can inspect it easily and include it here for others to review.
* If you spot areas of improvements, please highlight them.

tijcolem:
From an earlier discussion, we developed a clang-format style guide. resolves #3806. This is PR applies these format rules to a sample set of the codebase. Right now It's just a few .cpp, .hpp files for us to comment on and approve before updating the entire code base.  

As this commit is huge and will touch just about every part of the code, git diffs from previous commits will be very hard (e.g. git blame) but I don't see a great way to handle this. @jmarrec I think you had some way of dealing with this before? 

From what I am seeing, options for these scenarios are. 

1 ) You could replay all commits one-by-one with new formating applied but all of these commits would have different shas. This seems really dangerous. I don't like it. 

2 ) Update the code as one commit.  

3 ) We could incrementally update the code and apply the rules to new and change files as they are updated.  That would be messy and would take a long time for a codebase this large., 

I think there are some options you can add to git blame when looking at the history to omit white space. Such as this.
`git blame -w -M`

Going forward, the CI will check only new and updated code and apply style rules as needed. Developers should run the format script in [/ci/clang-format.sh](https://github.com/NREL/OpenStudio/blob/develop/ci/clang-format.sh) locally and apply these changes before submitting PRs. 

Here are the [clang-format style rules](https://github.com/NREL/OpenStudio/blob/develop/src/.clang-format) that are currently applied.


tijcolem:
Seems like as of git 2.23+ you can use `--ignore-rev` or pass in a file `--ignore-revs-file <file>` with a list of git commits for git-blame to ignore. I do not think this will work for the GitHub GUI blame feature though. 
See docs:  https://git-scm.com/docs/git-blame

tijcolem:
Not supported on GitHub UI - https://github.community/t/support-ignore-revs-file-in-githubs-blame-view/3256/8

jmarrec:
@tijcolem Yeah I added a `.git-blame-ignore-revs` file to the root of the repo, where you can add commits to be ignored locally when running git blame, after you have run `git config blame.ignoreRevsFile .git-blame-ignore-revs` to register it.

https://github.com/NREL/OpenStudio/pull/3899#issuecomment-593419510

It does not work on Github UI indeed right now.

tijcolem:
I applied the updated clang-format style rules. A whopping 3,268 cpp|hpp|c|h files changed.

tijcolem:
Narrowing things down a bit. All files under /src/model/ are fine with the new clang-format settings. 
Moving along to determine the offending files.

tijcolem:
narrowed it down to files in the  `./src/utilities/idd/`  We could set this up to exclude these files from the clang-format scrub.

tijcolem:
Looks like it's isolated to these files that are causing the problem. 
[./src/utilities/idd/IddFieldProperties.cpp](https://github.com/NREL/OpenStudio/blob/issue_3806/src/utilities/idd/IddFieldProperties.cpp)
[./src/utilities/idd/IddFieldProperties.hpp](https://github.com/NREL/OpenStudio/blob/issue_3806/src/utilities/idd/IddFieldProperties.hpp)

jmarrec:
@tijcolem Great detective work! I found the offending bit and reverted it in https://github.com/NREL/OpenStudio/pull/4139/commits/63a03caf908ab57e412b7873214513d77aa355b1

I proceeded to "protect" the calls to `OPENSTUDIO_ENUM` throughout by wrapping them between

`// clang-format off` / `// clang-format on`

jmarrec:
Ok, with e1f9d2c, the usual suspects are failing, and that's it.
Linux: 

```
The following tests FAILED:
	620 - BCLFixture.RemoteBCLTest (SIGTRAP)
	622 - BCLFixture.GetComponentByUID (SIGTRAP)
	3034 - CLITest-test_bundle-bundle (Failed)
	3035 - CLITest-test_bundle-bundle_git (Failed)
```

tijcolem:
@jmarrec Nice. I saw those enums and it looked suspicious post-clang-format. Glad that is now ignored! 

I included a GitHub actions job to apply the clang-format as this is a light-duty job. 
This PR needs to be merged in first before GitHub actions will run it.  

I need to update the developer workflow and the PR template so that contributors understand that they will need to run the clang-format and apply these rules for the change they change/add. 

I think this is ready though.

PR #4140, opened 2020-11-10
include pyenergyplus directory from the installer

bonnema:
avoid this error:
```
** Severe  ** Failed to import module "<Module Name>"
**   ~~~   ** Python error description follows: 
**   ~~~   ** ModuleNotFoundError("No module named 'pyenergyplus'")
```
when trying to run pyems via openstudio.


tijcolem (approved):
@bonnema Thanks for adding.

PR #4151, opened 2020-12-04
fix issue_4132 - hande lower case window drive letters

### LINKED ISSUE #4132: CLI error using Windows Command Prompt
Issue overview
--------------
I usually test the CLI on Windows using PowerShell, and it works fine:
image

But if I try to use Windows Command Prompt instead, it fails:
image

The error is:
```
Exception: Embedded file not found ':c:/ruby/2.5.0/net/protocol.rb'
eval:172:in `getFileAsString': unexpected return
```

### Environment

 - Platform (Operating system, version): Windows 10
 - Version of OpenStudio (if using an intermediate build, include SHA): OpenStudio 3.1.0

tijcolem:
fix #4132   Add to the existing regex to match for lower and upper case letters

### DIFF
--- a/src/cli/embedded_help.rb
+++ b/src/cli/embedded_help.rb
@@ -16,7 +16,7 @@ module OpenStudio
       absolute_path = File.expand_path p2
 
       # strip Windows drive letters
-      if /[A-Z]\:/.match(absolute_path)
+      if /[A-Za-z]\:/.match(absolute_path)
         absolute_path = absolute_path[2..-1]
       end
       absolute_path = ':' + absolute_path
--- a/src/cli/embedded_help.rb
+++ b/src/cli/embedded_help.rb
@@ -16,7 +16,7 @@ module OpenStudio
       absolute_path = File.expand_path p2
 
       # strip Windows drive letters
-      if /[A-Za-z]\:/.match(absolute_path)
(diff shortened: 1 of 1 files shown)


src/cli/embedded_help.rb
```diff
@@ -16,7 +16,7 @@ def self.get_absolute_path(p)
       absolute_path = File.expand_path p2
 
       # strip Windows drive letters
-      if /[A-Z]\:/.match(absolute_path)
+      if /[A-Za-z]\:/.match(absolute_path)
```
jmarrec, line 19: Fix is simple enough, problem was a needle in a haystack though. Great detective work!

jmarrec: Side note: I doubt this happens in real life, but perhaps we should anchor the pattern to the start of the string, like this `/^[A-Za-z]\:/.match(absolute_path.strip)`

jmarrec, line 19:
```suggestion
      if /^[A-Za-z]\:/.match(absolute_path.strip)
```

jmarrec (approved):

src/cli/embedded_help.rb
tijcolem: make sense to add in a regex anchor. like you said, very unlikely to hit that but can't hurt.

PR #4180, opened 2021-01-08
Addresses #4175, issues with ft for ElectricLoadCenter:Distribution

joseph-robertson:
Pull request overview
---------------------

 - Related to #4175
 - Adds `energyplus/Test/ElectricLoadCenterDistribution_GTest.cpp` test file with:
   - `ForwardTranslator_ElectricLoadCenterDistribution_NoInverter`
   - `ForwardTranslator_ElectricLoadCenterDistribution_NoGenerators` 
- Allows `ElectricLoadCenter:Distribution` objects with no generators/transformers (e.g., `LrgOff_GridStorageDemandLeveling.idf`) by updating `ForwardTranslator.cpp`
- Don't create `ElectricLoadCenter:Distribution` objects in ctors for:
  - `GeneratorFuelCell`
  - `GeneratorMicroTurbine`
  - `GeneratorPhotovoltaic`
  - `GeneratorPVWatts`

Please read [OpenStudio Pull Requests](https://github.com/NREL/OpenStudio/wiki/OpenStudio-Pull-Requests) to better understand the OpenStudio Pull Request protocol.

### Pull Request Author

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.

 - [ ] Model API Changes / Additions
 - [ ] Any new or modified fields have been implemented in the EnergyPlus ForwardTranslator (and ReverseTranslator as appropriate)
 - [ ] Model API methods are tested (in `src/model/test`)
 - [ ] EnergyPlus ForwardTranslator Tests (in `src/energyplus/Test`)
 - [ ] If a new object or method, added a test in NREL/OpenStudio-resources: Add Link
 - [ ] If needed, added VersionTranslation rules for the objects (`src/osversion/VersionTranslator.cpp`)
 - [ ] Checked behavior in OpenStudioApplication, adjusted policies as needed (`src/openstudio_lib/library/OpenStudioPolicy.xml`)
 - [ ] Verified that C# bindings built fine on Windows, partial classes used as needed, etc.
 - [ ] All new and existing tests passes
 - [ ] If methods have been deprecated, update rest of code to use the new methods

**Labels:**

 - [ ] If change to an IDD file, add the label `IDDChange`
 - [ ] If breaking existing API, add the label `APIChange`
 - [ ] If deemed ready, add label `Pull Request - Ready for CI` so that CI builds your PR

### Review Checklist

This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
[...]

PR #4192, opened 2021-01-27
Addresses #4182, provide OS SDK/API support for the EnergyPlus "Generator:WindTurbine" object

### LINKED ISSUE #4182: Provide OS SDK/API support for the EnergyPlus "Generator:WindTurbine" object
Enhancement Request
--------------
To provide parity of "behind the meter" Renewable energy generation analysis using OpenStudio, provide OS SDK / API access to the EnergyPlus "Generator:WindTurbine" object.

## Detailed Description
OpenStudio currently does not support direct access to the EnergyPlus Generator:WindTurbine object via the OS API. 

While  the "Generator:WindTurbine" is available to OS users by authoring an EnergyPlus measure, this requires renewable energy (comparison and integration) scripted workflows to handle Wind Generation objects separately from OS methods used to generate PV or other onsite generation and storage options. Specifically, EnergyPlus measures that model wind generators that need to communicate with different configurations and sizes of inverters, converters and battery storage objects, as well as Electric Load Center charge/discharge management strategies (those objects are all available via the OS API.). 

## Possible Implementation
Suggest a 1:1 mapping to the Generator:WindTurbine object as the simplest and most cost effective effective approach to wrapping.

joseph-robertson:
Pull request overview
---------------------

 - Fixes #4182
 - Also add RT 
 - Questions:
   - Which, if any, of the following 7 should be passed into the ctor (or should they get values from the docs, which is what this PR current does)?
     - `Rated Rotor Speed`
     - `Rotor Diameter`
     - `Overall Height`
     - `Rated Power`
     - `Rated Wind Speed`
     - `Cut In Wind Speed`
     - `Cut Out Wind Speed`  

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


resources/model/OpenStudio.idd
```diff
@@ -30404,6 +30404,151 @@ OS:Generator:FuelSupply,
...
+      \type handle
+      \required-field
+  A2, \field Name
+      \required-field
+      \reference GeneratorNames
+  A3, \field Availability Schedule Name
+      \note Availability schedule name for this system. Schedule value > 0 means the system is available.
+      \note If this field is blank, the system is always available.
+      \type object-list
+      \object-list ScheduleNames
```
jmarrec, line 30420:
```suggestion
  A3, \field Availability Schedule Name
      \note Availability schedule name for this system. Schedule value > 0 means the system is available.
      \note If this field is blank, the system is always available.
      \type object-list
      \object-list ScheduleNames
      \required-field
```

Set to `model.alwaysOnDiscreteSchedule()` in Ctor:

resources/model/OpenStudio.idd
```diff
@@ -30404,6 +30404,151 @@ OS:Generator:FuelSupply,
...
+  N11,\field Maximum Power Coefficient
+      \note This field should be input if the rotor type is Horizontal Axis Wind Turbine
+      \type real
+      \minimum> 0.0
+      \maximum 0.59
+      \required-field
+  N12,\field Annual Local Average Wind Speed
+      \type real
+      \units m/s
+      \minimum> 0.0
```
jmarrec, line 30501: Not required indeed :+1: 

> If this field is not entered, then the model will use the wind speed from the design day or weather file information with only adjustment at the height of the rotor.

resources/model/OpenStudio.idd
```diff
@@ -30404,6 +30404,151 @@ OS:Generator:FuelSupply,
...
+  N13,\field Height for Local Average Wind Speed
+      \type real
+      \units m
+      \minimum> 0.0
+      \required-field
+  N14,\field Blade Chord Area
+      \type real
+      \units m2
```
jmarrec, line 30509:
```suggestion
  N14,\field Blade Chord Area
      \type real
      \units m2
      \required-field
      \minimum 0.0
```

Same as Blad Draft Coef and Lift Coef

jmarrec: E+ code has checks to see if VAWT is used while these fields are zero: https://github.com/NREL/EnergyPlus/blob/d6cf92a2a7dc8ca3489d39792c4315f6f3bea957/src/EnergyPlus/WindTurbine.cc#L515-L562

resources/model/OpenStudio.idd
```diff
@@ -30404,6 +30404,151 @@ OS:Generator:FuelSupply,
...
+      \type real
+      \required-field
+  N21,\field Power Coefficient C5
+      \minimum> 0.0
+      \type real
+      \required-field
+  N22;\field Power Coefficient C6
+      \minimum> 0.0
+      \type real
+      \required-field
```
jmarrec, line 30550: Make extensible and use max fields to limit to 6? @joseph-robertson 

This isn't mandatory in this case, as it's not like there will be new coefficients added.

jmarrec: I went to look for the equation in the source code, and it seems I discovered a bug... During the refactor to C++ it seems that one term of the equation corresponding to C3 has been omitted: https://github.com/NREL/EnergyPlus/blob/d6cf92a2a7dc8ca3489d39792c4315f6f3bea957/src/EnergyPlus/WindTurbine.cc#L905-L907

jmarrec: Well perhaps that's done on purpose,b ut both `C3` and `C4` are completely unused... I opened https://github.com/NREL/EnergyPlus/issues/8490

src/energyplus/ForwardTranslator.cpp
```diff
@@ -402,6 +402,16 @@ namespace energyplus {
...
 
+    // remove orphan Generator:WindTurbine
+    for (auto& wt : model.getConcreteModelObjects<GeneratorWindTurbine>()) {
+      if (!wt.electricLoadCenterDistribution()) {
+        LOG(Warn,
+            "GeneratorWindTurbine " << wt.name().get() << " is not referenced by any ElectricLoadCenterDistribution, it will not be translated.");
+        wt.remove();
+        continue;
+      }
+    }
```
jmarrec, line 413: So I had a similar realization in https://github.com/NREL/OpenStudio/pull/4194/commits/949fe92542a7ec28122c6fecb511a3e226a74335.

This stuff shouldn't exist. The children of ELCD (Generators, etc) should not be added to the list of objects to always translate. The ElectricLoadCenterDistribution only should be added to that list, and it should be responsible to trigger translation of its generators (and others) instead. 

I'm pretty sure this is just an historical side effect to please the sketchup plugin.

src/energyplus/ForwardTranslator.cpp
```diff
@@ -3262,6 +3277,7 @@ namespace energyplus {
 
     result.push_back(IddObjectType::OS_Generator_Photovoltaic);
     result.push_back(IddObjectType::OS_Generator_PVWatts);
+    result.push_back(IddObjectType::OS_Generator_WindTurbine);
```
jmarrec, line 3280: See comment above, that shouldn't be added to the list I think.

src/model/GeneratorWindTurbine.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+    double powerCoefficientC6() const;
+
+    //@}
+    /** @name Setters */
+    //@{
+
+    bool setAvailabilitySchedule(Schedule& schedule);
+
+    void resetAvailabilitySchedule();
```
jmarrec, line 118: Remove

src/model/GeneratorWindTurbine.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+    double cutOutWindSpeed() const;
+
+    double fractionSystemEfficiency() const;
+
+    double maximumTipSpeedRatio() const;
+
+    double maximumPowerCoefficient() const;
+
+    boost::optional<double> annualLocalAverageWindSpeed() const;
```
jmarrec, line 90: Optional OK

src/model/GeneratorWindTurbine.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+    double heightforLocalAverageWindSpeed() const;
+
+    boost::optional<double> bladeChordArea() const;
```
jmarrec, line 94: Non optional

src/energyplus/CMakeLists.txt
```diff
@@ -200,6 +200,7 @@ set(${target_name}_src
   ForwardTranslator/ForwardTranslateGeneratorFuelCellStackCooler.cpp
   ForwardTranslator/ForwardTranslateGeneratorFuelCellWaterSupply.cpp
   ForwardTranslator/ForwardTranslateGeneratorFuelSupply.cpp
+  ForwardTranslator/ForwardTranslateGeneratorWindTurbine.cpp
```
jmarrec, line 203: ReverseTranslator?

src/model/GeneratorWindTurbine.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+    virtual ~GeneratorWindTurbine() {}
+
+    //@}
+
+    static IddObjectType iddObjectType();
+
+    /** @name Getters */
+    //@{
+
```
jmarrec, line 63: You're missing the getter for `availabilitySchedule`, which should return a non optional schedule.

jmarrec: Ah shoot, I didn't realize that Generator.hpp defined it https://github.com/NREL/OpenStudio/blob/ae47a41ab6eb0b8a49f971becce106d5a5dbd448/src/model/Generator.hpp#L61

It shouldn't have been an optional IMHO, but changing would break api so...

src/model/GeneratorWindTurbine_Impl.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+      double maximumTipSpeedRatio() const;
+
+      double maximumPowerCoefficient() const;
+
+      boost::optional<double> annualLocalAverageWindSpeed() const;
+
+      double heightforLocalAverageWindSpeed() const;
+
+      boost::optional<double> bladeChordArea() const;
```
jmarrec, line 109: non optional

src/model/GeneratorWindTurbine.cpp
```diff
@@ -0,0 +1,639 @@
...
+      : Generator_Impl(other, model, keepHandle) {}
+
+    const std::vector<std::string>& GeneratorWindTurbine_Impl::outputVariableNames() const {
+      static const std::vector<std::string> result{"Generator Produced AC Electricity Rate",       "Generator Produced AC Electricity Energy",
+                                                   "Generator Turbine Local Wind Speed",           "Generator Turbine Local Air Density",
+                                                   "Generator Turbine Power Coefficient",          "Generator Turbine Tip Speed Ratio",
+                                                   "Generator Turbine Chordal Component Velocity", "Generator Turbine Normal Component Velocity",
+                                                   "Generator Turbine Relative Flow Velocity",     "Generator Turbine Attack Angle"};
+      return result;
+    }
```
jmarrec, line 75: :+1:

src/model/GeneratorWindTurbine.cpp
```diff
@@ -0,0 +1,639 @@
...
+    setRatedPower(55000.0);    // from docs
+    setRatedWindSpeed(11.0);   // from docs
+    setCutInWindSpeed(3.5);    // from docs
+    setCutOutWindSpeed(25.0);  // from docs
+    setFractionSystemEfficiency(0.835);
+    setMaximumTipSpeedRatio(5.0);
+    setMaximumPowerCoefficient(0.25);
+    setHeightforLocalAverageWindSpeed(50.0);
+    setBladeDragCoefficient(0.9);
+    setBladeLiftCoefficient(0.05);
```
jmarrec, line 409: Since you default to `HorizontalAxisWindTurbine`, perhaps we should make all unrelated coefficients to 0, and add the Chord Area once it's made required?

```suggestion
    setBladeDragCoefficient(0.0);
    setBladeLiftCoefficient(0.0);
    setBladeChordArea(0.0);
```

src/model/GeneratorWindTurbine.cpp
```diff
@@ -0,0 +1,639 @@
...
+    setPowerCoefficientC1(0.5176);
+    setPowerCoefficientC2(116.0);
+    setPowerCoefficientC3(0.4);
+    setPowerCoefficientC4(0.0);
+    setPowerCoefficientC5(5.0);
+    setPowerCoefficientC6(21.0);
+
+    //Add ElectricLoadCenterDistribution to get ElectricLoadCenterGenerators
+    ElectricLoadCenterDistribution elcd(model);
+    elcd.addGenerator(*this);
```
jmarrec, line 419:
```suggestion
```

Remove this to match #4180

src/energyplus/ForwardTranslator/ForwardTranslateGeneratorWindTurbine.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+  /** This method forward translates the OS:Generator:WindTurbine as well the optional attached OS:Generator:WindTurbine:HeatRecovery
+ to the EnergyPlus Generator:WindTurbine object */
+  boost::optional<IdfObject> ForwardTranslator::translateGeneratorWindTurbine(GeneratorWindTurbine& modelObject) {
+    boost::optional<std::string> s;
+    boost::optional<double> d;
+    boost::optional<int> i;
+
+    // Name
+    IdfObject idfObject = createRegisterAndNameIdfObject(openstudio::IddObjectType::Generator_WindTurbine, modelObject);
```
jmarrec, line 58: I would delay creation of the object. Then add checks based on rotorType

```
if rotorType = 'Vertical':
   if (Blade Chord Area == 0) || (Blade Drag Coefficient ==0) || (Blade Lift Coefficient == 0):
       LOG(Error) && return boost::none

```

resources/model/OpenStudio.idd
```diff
@@ -30404,6 +30404,151 @@ OS:Generator:FuelSupply,
...
+      \minimum> 0.0
+      \required-field
+  N14,\field Blade Chord Area
+      \type real
+      \units m2
+  N15,\field Blade Drag Coefficient
+      \note This field is only for Vertical Axis Wind Turbine..
+      \note The user must input this field if the rotor type is Vertical Axis Wind Turbine.
+      \type real
+      \required-field
```
jmarrec, line 30514:
```suggestion
      \required-field
      \minimum 0.0
```

resources/model/OpenStudio.idd
```diff
@@ -30404,6 +30404,151 @@ OS:Generator:FuelSupply,
...
+  N16,\field Blade Lift Coefficient
+      \note This field is only for Vertical Axis Wind Turbine..
+      \note The user must input this field if the rotor type is Vertical Axis Wind Turbine.
+      \type real
+      \required-field
```
jmarrec, line 30519:
```suggestion
      \required-field
      \minimum 0.0
```

src/model/GeneratorWindTurbine.cpp
jmarrec: Actually, perhaps we should leave it as is...

src/model/GeneratorWindTurbine.cpp
```diff
@@ -0,0 +1,639 @@
...
+
+    void GeneratorWindTurbine_Impl::resetAvailabilitySchedule() {
+      bool result = setString(OS_Generator_WindTurbineFields::AvailabilityScheduleName, "");
+      OS_ASSERT(result);
+    }
+
+    bool GeneratorWindTurbine_Impl::setRotorType(std::string rotorType) {
+      bool result = setString(OS_Generator_WindTurbineFields::RotorType, rotorType);
+      return result;
+    }
```
jmarrec, line 262: (Not sold on it, but should we clear/zero out the fields that become moot?)

jmarrec (changes requested):

src/energyplus/CMakeLists.txt
jmarrec: @joseph-robertson As a reminder, GenerateClass.rb will generate one for you that should require minimal changes.

Full command
```
cd developer/ruby/
$os_build/Products/openstudio GenerateClass.rb -c "GeneratorWindTurbine" -b "Generator"  -i "OS:Generator:WindTurbine" -s model -o /home/julien/Software/Others/OpenStudio/src/model/ -p -f -r
```

I think this should only do the RT:

```
cd developer/ruby/
$os_build/Products/openstudio GenerateClass.rb -c "GeneratorWindTurbine" -b "Generator"  -i "OS:Generator:WindTurbine" -r
```

jmarrec:
@joseph-robertson I left some comments, and some - I think - required changes (making one field required for eg, removing the creation of the ECLD in the Ctor too). Address what you think makes sense, just mark comments resolved for what doesn't (I left everything so you see them, even those where I was wrong such as availability schedule).

To answer the question in the OP:

> Which, if any, of the following 7 should be passed into the ctor (or should they get values from the docs, which is what this PR current does)

I think your current implemenation is fine.

src/model/GeneratorWindTurbine.hpp
jmarrec: I would **still** remove this even though the getter will be optional per Generator.hpp.

resources/model/OpenStudio.idd
jmarrec: I would **still** make this required-field and default to model.alwaysOnDiscreteSchedule in Ctor even though getter will be optional per Generator.hpp

src/model/GeneratorWindTurbine.hpp
jmarrec: (Please resolve this comment once read... I was wrong)

resources/model/OpenStudio.idd
joseph-robertson: Should it be `\minimum> 0.0` then?

src/model/GeneratorWindTurbine.cpp
joseph-robertson: What should Blade Chord Area default to?

resources/model/OpenStudio.idd
joseph-robertson: The idf in testfiles has blanks for the three Blade fields...

jmarrec: I know, since the field isn't used if horizontal, but that doesn't mean it hurts any to set it to zero by default.

src/model/GeneratorWindTurbine.cpp
jmarrec: It won't hurt to have it in the FT (and I wrote it before I realized you put sensible defaults here) so I pushed it anyways in ff5f76814e46a4d6b26723dd8638d1994b73033c

jmarrec: @joseph-robertson I'll let you decide on this one, I'm fine with either zero or the current defaults you set!

jmarrec (approved):

jmarrec:
Tested that OS-res test in https://github.com/NREL/OpenStudio-resources/pull/128 is still passing. This can merge, good job @joseph-robertson !

resources/model/OpenStudio.idd
joseph-robertson: Decided not to go the extensible route.

src/model/GeneratorWindTurbine.cpp
joseph-robertson: We'll leave it up to the user to control the blade fields.

PR #4208, opened 2021-02-16
Python measure pointer

jmarrec:
Pull request overview
---------------------

@lefticus @brianlball @kbenne

### DIFF
--- a/src/measure/Measure.i
+++ b/src/measure/Measure.i
@@ -92,7 +92,9 @@
     os << *self;
     return os.str();
   }
+};
 
+%extend openstudio::measure::OSRunner {
 
   // The below is added for type conversions from ruby to python via PyCall (for Python measures)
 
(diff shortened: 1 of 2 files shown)


lefticus:
I know you named it with an ugly `__` name so people don't try to use it, but just an FYI, that using `__` anywhere in an identifier is undefined behavior because it's reserved for the implementation. Also a leading `_` followed by a lower case letter is reserved in the global namespace.

jmarrec:
I'm aware of that yes. Because it's done at swig level it doesn't concern me that much, especially given than you have the `__str__` one right above already anyways (as per the swig docs)

There's no easy way to do otherwise too. For python there is, I could define it as xxxxToIntPtr or whatnot, then use `%pythoncode` to alias it. But it won't work for ruby as there's no rubycode.

If you can propose an alternative I'll take it, otherwise I'll pretend I didn't see your warning :)

jmarrec:
I've opened a companion PR at https://github.com/NREL/OpenStudio-workflow-gem/pull/129. This isn't working yet.

jmarrec:
I'm merging this one, happy to change the "__" stuff if need be. But I'd like to get the python_measure PR to trigger a rebuild of the python bindings so one can install them for testing.

jmarrec:
refer to https://github.com/NREL/OpenStudio/pull/4190 if you need to find the latest of the development. Python bindings will be posted to testpypi, and installers (which include ruby bindings) posted as comments on that PR in a few hours.

Issue #1586, opened 2015-04-08
joinAll method seems to be order dependent


macumber:
Need more testing around this method

macumber:
@asparke2 FYI

jmarrec:
No steps to reproduce. #4221 may have fixed this. Closing until further information is provided.

Issue #1398, opened 2015-02-05
Native support to calculate building perimeter


DavidGoldwasser:
@asparke2 do you need this at all for ref buildings. I needed it for AEDG, but wrote something in ruby. The comment below as info from original Pivotal ticket.

Here is a link to helper method I use for Coffee. It takes the model as an argument and looks for edges that are used on one surface with ground an another with exterior. It returns the perimeter length.
https://github.com/NREL/cofee-measures/blob/master/lib/os_lib_geometry.rb#L374

Nick pointed out this should probably take in a story, because sometimes you may want to do it on specific story.

DavidGoldwasser:
I need to know the perimeter for the exterior lighting measures (11 and 12). I show this as optional because I have ideas for how to do this in ruby if necessary. Below are two options:

Old idea was to loop through exterior wall surfaces for spaces on the first floor and calculate the area/avg surface height. This had a number of issues, in particular wasn't sure how split surfaces would work.

Newer idea is to still loop through those same walls, but loop through have a user input for height above space origin to include in perimeter. I would then loop through pairs of vertices that both lie below that level. Then I would take the distance after equalizing the z value (so it is horizontal distance). For edges that span over the height I could figure out what fraction of it to count.

asparke2:
@DavidGoldwasser I don't need it for the reference buildings.  However, we would need it for creating code baseline models at some point.

mattdoiron:
This feature would be very useful for the work we do at our company - we do a lot of compliance-type modelling. I would love to take a look at the sample linked to in the original comment, but the link appears to be broken.

macumber:
What are the specific methods that people want? 

double BuildingStory::exteriorPerimeter() const
double Space::exteriorPerimeter() const
double Surface::effectiveWidth() const
double Surface::effectiveHeight() const

?

mattdoiron:
I can only speak for myself, but for compliance with things like ASHRAE 90.1 there is a need to create F-factor constructions for slab-on-grade surfaces. These require the "exposed perimeter" of a given surface as an input and I _just_ finished doing it manually for 100+ surfaces (in Sketchup plugin)! So I would suggest Surface::exteriorPerimeter() or something similar. The space-level and building-story level perimiters would also be useful for the other applications mentioned by others.

DavidGoldwasser:
Matt, sorry about delay re the bad link. It is a private repo. The sample code is here on unmet hours as well
[...]

PR #4221, opened 2021-02-23
Geometry improvements

### LINKED ISSUE #1614: joinAll fails on cases with inner loop in result
joinall

ggartside:
Pull request overview
---------------------

Adds Polygon3d class to geometry utility, adds new join and joinAll methods to intersection. Existing join and joinAll methods are not affected.

Adds unit tests in Intersection to test new join and joinAll methods, and unit tests go Geometry to test basic functions of the pOlygon3d class such as area and normal calculations

This will fix #1614  - joinAll fails on cases with inner loop in result

### Review Checklist

 - [ ] Labeling is ok


src/utilities/geometry/Polygon.cpp
```diff
@@ -0,0 +1,143 @@
...
+	Polygon3d::Polygon3d(Point3dVector outerPath) {
+    for (auto p : outerPath)
+      points.push_back(p);
+  }
+
+	/// <summary>
+	/// Adds a point to the polygon perimeter
+	/// </summary>
+	/// <param name="point"></param>
+	void Polygon3d::AddPoint(Point3d& point) {
```
macumber, line 47: Suggest `addPoint`

ggartside: Have renamed all methods to follow naming conventions

jmarrec:
I just dropped by and saw clang-format/cppcheck being unhappy so I fixed that. I see you tried to clang-format @ggartside already, but you probably have an older version of clang-format or something.

I commented some unused variables in tests to please cppcheck in https://github.com/NREL/OpenStudio/pull/4221/commits/ea25e0ec71f7528115ffbb5edd2d1c5780f8d312, but it would be better to actually use these variables in the test probably, like adding an `EXPECT_EQ(xx, perimeter)`, I'll let you decide.

jmarrec:
Unix has -Wall -Werror and is throwing at least here:

```
[  4%] Generating ruby_OpenStudioGBXML_wrap.cxx

/srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Polygon.cpp: In member function ‘double openstudio::Polygon3d::getPerimeter()’:

/srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Polygon.cpp:113:22: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]

   for (long i = 0; i < points.size(); i++) {

                    ~~^~~~~~~~~~~~~~~

```

src/utilities/geometry/Test/Intersection_GTest.cpp
```diff
@@ -1835,3 +1836,239 @@ TEST_F(GeometryFixture, simplify7) {
...
+/// |        +---------+         |
+/// |        |         |         |
+/// |        |         |         |
+/// |        |         |         |
+/// |--------+---------+---------|
+/// |                            |
+/// |       Polygon B            |
+/// |                            |
+/// |                            |
+/// +----------------------------+
```
jmarrec, line 1862: :+1:

jmarrec:
```
/srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Test/Geometry_GTest.cpp: In member function ‘virtual void GeometryFixture_Polygon_Basic_Test::TestBody()’:

/srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Test/Geometry_GTest.cpp:1130:24: error: cannot bind non-const lvalue reference of type ‘openstudio::Point3d&’ to an rvalue of type ‘openstudio::Point3d’

   testPolygon.addPoint(Point3d(0, 0, 0));

                        ^~~~~~~~~~~~~~~~

In file included from /srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Test/Geometry_GTest.cpp:36:0:

/srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Test/../Polygon.hpp:52:8: note:   initializing argument 1 of ‘void openstudio::Polygon3d::addPoint(openstudio::Point3d&)’

   void addPoint(Point3d& point);

        ^~~~~~~~

/srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Test/Geometry_GTest.cpp:1131:24: error: cannot bind non-const lvalue reference of type ‘openstudio::Point3d&’ to an rvalue of type ‘openstudio::Point3d’

   testPolygon.addPoint(Point3d(100, 0, 0));

                        ^~~~~~~~~~~~~~~~~~

In file included from /srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Test/Geometry_GTest.cpp:36:0:

/srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Test/../Polygon.hpp:52:8: note:   initializing argument 1 of ‘void openstudio::Polygon3d::addPoint(openstudio::Point3d&)’

   void addPoint(Point3d& point);

        ^~~~~~~~

/srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Test/Geometry_GTest.cpp:1132:24: error: cannot bind non-const lvalue reference of type ‘openstudio::Point3d&’ to an rvalue of type ‘openstudio::Point3d’

   testPolygon.addPoint(Point3d(100, 100, 0));

                        ^~~~~~~~~~~~~~~~~~~~

In file included from /srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Test/Geometry_GTest.cpp:36:0:

/srv/jenkins/openstudio/git/incremental/OpenStudio/src/utilities/geometry/Test/../Polygon.hpp:52:8: note:   initializing argument 1 of ‘void openstudio::Polygon3d::addPoint(openstudio::Point3d&)’

   void addPoint(Point3d& point);

        ^~~~~~~~

```

ggartside:
> 
> 
> I just dropped by and saw clang-format/cppcheck being unhappy so I fixed that. I see you tried to clang-format @ggartside already, but you probably have an older version of clang-format or something.
> 
> I commented some unused variables in tests to please cppcheck in [ea25e0e](https://github.com/NREL/OpenStudio/commit/ea25e0ec71f7528115ffbb5edd2d1c5780f8d312), but it would be better to actually use these variables in the test probably, like adding an `EXPECT_EQ(xx, perimeter)`, I'll let you decide.
> 
> I also tagged as `Pull Request - Ready for CI`, I'm looking forward to test results (to see if these fixes the `Space_GTest` degenerate tests etc (which I locally disabled today as it's been annoying me for a while to see them fail))

Can you tell me which tests these are and I'll look out for them - we are mainly addressing a limited number of outstanding issues and those may include the space tests or not, I don't know at this stage

src/utilities/geometry/Polygon.hpp
```diff
@@ -0,0 +1,93 @@
...
+  Point3dVectorVector getInnerPaths() const;
+
+  void addPoint(Point3d& point);
+  void addHole(Point3dVector hole);
+  Vector3d newellVector();
+  Vector3d outwardNormal();
+  /// <summary>
+  /// Calculates the gross area of the polygon (area excluding holes)
+  /// </summary>
+  /// <returns></returns>
```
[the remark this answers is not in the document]
ggartside: So no '///' commenting?

I guess that's a visual studio thing anyway probably

jmarrec:
@ggartside I opened a PR to this branch with proposed changes, that should fix your build issues. https://github.com/NREL/OpenStudio/pull/4238. I'll let you review and determine whether they make sense or not.

----

The failing geometry tests I was talking about can be run via `ctest` or `Products/openstudio_model_tests` at the root of the build folder:

```
ModelFixture.Space_intersectSurfaces_degenerate1
ModelFixture.Space_intersectSurfaces_degenerate2
ModelFixture.Space_intersectSurfaces_degenerate3
ModelFixture.ThreeJSReverseTranslator_FloorplanJS_SimAUD_Paper
```

(Your PR doesn't affect their status yet, they still fail)

src/utilities/geometry/Polygon.hpp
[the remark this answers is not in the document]
jmarrec: The format used, (with tags like summary and returns (blank)) looks overkill and heavy is all.

src/utilities/geometry/Polygon.cpp
```diff
@@ -0,0 +1,147 @@
...
+  }
+}
+
+void Polygon3d::addPoint(const Point3d& point) {
+  outerPath.push_back(point);
+}
+
+
+void Polygon3d::setOuterPath(Point3dVector outerPath) {
+  outerPath = outerPath;
```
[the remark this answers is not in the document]
ggartside: Ok that was bad! 

The visual studio compiler didn't complain though no doubt there are warnings somewhere, clang though, I'm not familiar with clang, is this something I can run locally as part of the build?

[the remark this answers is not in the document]
jmarrec: clang is LLVM's compiler. Typically found on mac, as Xcode is apple-clang (a fork of clang). You could run that on Windows but I would't bother, that's why we have CI in place to build on MSVC (Windows), Clang (Mac) and GCC (Ubuntu)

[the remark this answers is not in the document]
ggartside: Two of those degenerate space/surface tests are fixed with the latest check in, the one not fixed, 2, has a couple of very small surfaces that are not eliminated but make the combined roof areas outside tolerance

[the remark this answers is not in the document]
ggartside: Julien, do you have any idea what this syntax error might be?

[src/utilities/geometry/Intersection.cpp:82]:(warning),[constStatement],Found suspicious operator '>'
[src/utilities/geometry/Intersection.cpp:83]:(warning),[constStatement],Found suspicious operator '>'
[src/utilities/geometry/Intersection.cpp:84]:(warning),[constStatement],Found suspicious operator '>'
[src/utilities/geometry/Intersection.cpp:1295]:(warning),[constStatement],Found suspicious operator '>'

The lines in the file are
  BoostMultiPolygon resultExpand;
  BoostMultiPolygon resultShrink;
  BoostMultiPolygon result;

And BoostMultiPolygon is defined as

typedef boost::geometry::model::multi_polygon<BoostPolygon> BoostMultiPolygon;

BoostMultiPolygon is also used in Space.cpp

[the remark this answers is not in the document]
jmarrec: @ggartside Thanks for the excel sheet and the explanation. I've got to admit I get slightly lost with this geometry stuff as it don't think about it often and it requires me to concentrate really hard and take everything step by step. From your description and the sheet, I think what you wrote makes total sense and it does seem to be a combination of 1.) Comparing two slightly different things and 2.) very low tolerance (the units are indeed in meters, so 4mm2 isn't a huge deal).

---- 

These warnings `Found suspicious operator '>'` are issued by cppcheck, and cppcheck especially given that we pass the `--inconclusive` flag can throw false positive (and has some parse errors sometimes).

 I was going to revert your commit https://github.com/NREL/OpenStudio/pull/4221/commits/8b5e98a62467e9e5583c855f3c7c098b55e2a58c which I think is unnecessary but I see you reverted the changes in https://github.com/NREL/OpenStudio/pull/4221/commits/212aea62b0440526a0cccd3797bd4068004bed69 already. I will fix the cppcheck warnings with an inline suppression. 

----

But as it stands two things: 
* The build fails due to warnings treated as error in the Space_GTest.cpp you added (and a few other places) 
    * This is easy to fix (I'll replace `int i = 0` with `size_t i = 0`)
    * But you seem to be using OSM test file(s) that you didn't include in source control. **I included an example of how to properly register test OSMs in CMakeLists.txt and how to use them in the last commit**. Please adjust that accordingly.
* Your changes created some new test failures too, in https://github.com/NREL/OpenStudio/pull/4221/commits/8b5e98a62467e9e5583c855f3c7c098b55e2a58c these are:

```
The following tests FAILED:
	417 - GeometryFixture.Polygon3d_Join (Failed)
	418 - GeometryFixture.Polygon3d_JoinAll_1614 (Failed)
	419 - GeometryFixture.Polygon3d_JoinAllPolygons_1614 (Failed)
	420 - GeometryFixture.JoinAll_2527 (Failed)
	2210 - ModelFixture.Surface_Intersect_DifferentHeight_PartialOverlap (Failed)
	2211 - ModelFixture.Surface_Intersect_DifferentHeight_ShareOneEdge_PartialOverlap (Failed)
```

resources/CMakeLists.txt
```diff
@@ -111,6 +111,8 @@ set(ENERGYPLUS_OUTPUTS "")
 set(model_resources_src
   model/OpenStudio.idd
   model/schedulefile.csv
+
+  model/two_stories_pre_intersect.osm
```
jmarrec, line 115: I just "touched" this file in `<root>/resources/model/two_stories_pre_intersect.osm` so obviously you need to replace with the actual file.

src/model/test/Space_GTest.cpp
```diff
@@ -2038,10 +2038,13 @@ TEST_F(ModelFixture, Space_intersectSurfaces_degenerate3) {
...
 /// Second of all we use a different removeSpikes method that shrinks and expands the polygon
 TEST_F(ModelFixture, RemoveSpikesAndOverlaps) {
-  osversion::VersionTranslator translator;
-  //model::OptionalModel model = translator.loadModel(toPath("./whole_bldg_partially_matched.osm"));
-  //boost::optional<Model> model = Model::load(toPath("./7-7 Windows Complete.osm"));
-  boost::optional<Model> model = Model::load(toPath("./two_stories_pre_intersect.osm"));
+
+  openstudio::path path = resourcesPath() / toPath("model/two_stories_pre_intersect.osm");
+  // openstudio::path path = resourcesPath() / toPath("model/whole_bldg_partially_matched.osm");
+  // openstudio::path path = resourcesPath() / toPath("model/7-7 Windows Complete.osm");
```
jmarrec, line 2044: If you need to test all three, you can make a free function in the test file, with their EXPECT/ASSERT statements, and pass the OSM as argument.

src/model/test/Space_GTest.cpp
```diff
@@ -2013,6 +2027,90 @@ TEST_F(ModelFixture, Space_intersectSurfaces_degenerate3) {
...
+
+  for (const auto& space : spaces) {
+    std::string name = space.name().value();
+    Transformation spaceTransformation = space.transformation();
+
+    std::vector<Surface> surfaces = space.surfaces();
+    // What is this?
+    //if (name == "Space TZ46-3") {
+    //  int hello = 1;
+    //}
```
jmarrec, line 2065: This was throwing an unused-variable warning treated as error. I didn't know what to do with it

src/model/test/Space_GTest.cpp
```diff
@@ -2013,6 +2027,90 @@ TEST_F(ModelFixture, Space_intersectSurfaces_degenerate3) {
...
+            std::reverse(faceVertices.begin(), faceVertices.end());
+          }
+          boost::optional<Vector3d> otherOutwardNormal = getOutwardNormal(otherFaceVertices);
+          if (otherOutwardNormal->z() > -1 + tol) {
+            std::reverse(otherFaceVertices.begin(), otherFaceVertices.end());
+          }
+          boost::optional<IntersectionResult> intersection = openstudio::intersect(faceVertices, otherFaceVertices, tol);
+          if (intersection) {
+            // Should this be an EXPECT_TRUE(false) << "Space...."?
+            LOG(Error, "Space " << name << " has surfaces that intersect with each other, namely " << thisName << " and " << otherName);
```
jmarrec, line 2095: Generally speaking, I would expect some ASSERT/EXPECT in the test

src/model/test/Space_GTest.cpp
```diff
@@ -2013,6 +2027,90 @@ TEST_F(ModelFixture, Space_intersectSurfaces_degenerate3) {
...
+
+  for (const auto& space : blacklist) {
+    auto it = std::find(spaces.begin(), spaces.end(), space);
+    auto index = distance(spaces.begin(), it);
+    spaces.erase(spaces.begin() + index);
+  }
+
+  // Don't sort by area, use the shrink and expand to remove spikes
+  intersectSurfaces(spaces);//false, true);
+  matchSurfaces(spaces);
```
jmarrec, line 2111: Generally speaking, I would expect some ASSERT/EXPECT in the test

jmarrec (commented):
Left comments to explain my changes

ggartside:
> 
> 
> (This will need to become a path that's not harcoded to one machine)

Yeah I know, that wasn't supposed to happen!

ggartside:
> 
> 
> @ggartside have you considered porting [Topolys](https://github.com/automaticmagic/topolys) to C++ to calculate exposed perimeter?

I haven't heard of it, but I followed the link to topologic.app and I know Robert Aish from way back when!

tijcolem:
@ggartside looks like this is PR close to completion? Nice work on all of this. It'll be great to have these resolved in 3.2! 

Looks like windows is complaining about 1 bad swig type (below). Once that is fixed, CI on all platforms should pass. 

```
1/1 Test #3292: CSharpTest-FindMissingSWIGTypes ...***Failed    0.58 sec
In total, 1 Bad swigs found

{

  "OpenStudioUtilitiesGeometry": [

    "SWIGTYPE_p_boost__optionalT_std__vectorT_std__vectorT_openstudio__Point3d_t_t_t.cs"

  ]

}

Note: 2 errors have been ignored
0% tests passed, 1 tests failed out of 1
Total Test time (real) =   2.13 sec

The following tests FAILED:
	3292 - CSharpTest-FindMissingSWIGTypes (Failed)
powershell.exe : Errors while running CTest
```

ggartside:
I didn't realize that one was something I caused, I thing I fixed it in he .i file (fingers crossed).

Hopefully we got all the bugs too and happy to see those existing failing tests now passing.

tijcolem:
@ggartside  Looks like your last commit fixed it! Nice work.  Is this PR all ready to go to be merged?

resources/CMakeLists.txt
kbenne: I believe adding these osm files to this list will include them in the embedded file system. I'm not sure that is desired or necessary.

ruby/test/Issue_2560_Test.rb
```diff
@@ -0,0 +1,168 @@
...
+        #OptionalSpace
+        sp4 = OpenStudio::Model::Space::fromFloorPrint(perimeter,3,model)
+        assert(sp4)
+        sp4 = sp4.get
+        sp4.setBuildingStory(story1)
+        tz4 = OpenStudio::Model::ThermalZone.new(model)
+        sp4.setThermalZone(tz4)
+
+        # Model before intersection
+        outpath = File.join(File.expand_path('../../.') + "/build/resources/model/","2560_before.osm")
```
kbenne, line 96: I think the "build" directory needs to be configured. I think this is based on an assumption about where the build directory is located, which cannot be guaranteed. @jmarrec do you think you could help @ggartside configure this?

ruby/test/Issue_2560_Test.rb
```diff
@@ -0,0 +1,168 @@
...
+        #spaces = spaces.get
+        spacesX = OpenStudio::Model::SpaceVector.new
+        spaces.each do |s|
+            spacesX << s
+        end
+
+        OpenStudio::Model::intersectSurfaces(spacesX)
+        OpenStudio::Model::matchSurfaces(spacesX)
+
+        outpath = File.join(File.expand_path('../../.') + "/build/resources/model/","2560_after.osm")
```
kbenne, line 109: I suspect this assumption will repeat. I won't comment on any remaining cases.

src/utilities/CMakeLists.txt
```diff
@@ -343,6 +345,13 @@ set(${target_name}_test_src
   bcl/test/BCLMeasure_GTest.cpp
 )
 
+## # Example: Selectively add test files
```
kbenne, line 348: @ggartside this was an example if you needed it, but if you want I think it can be removed now.

src/utilities/geometry/Polygon3d.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+  // gets an edge on thge outer path
+  Point3dVector getEdge(size_t i);
+
+  // Returns the ovelapping segments between this polygon and the line
+  std::vector<Point3dVector> overlap(const Point3dVector& line) const;
+
+ private:
+  // The polygon's outer path
+  Point3dVector m_outerPath;
```
kbenne, line 110: @jmarrec I would not hold this PR up for this comment, but do you think this class should follow PIMPL to hide these members? 

@macumber might have an opinion too. 

Regardless I think that could be done retroactively if we choose.

kbenne (changes requested):
This is impressive to me. A lot of the implementation is beyond my qualifications to critique, but it appears well covered by tests and nicely executed. 

I had a couple of comments. One relating to including the osm files in the model resources. I'm pretty sure that is not needed and will add heft to the openstudio CLI. The other main comment was about what I think is an assumption about where the build directory is located. If @jmarrec is willing to take a look at that I would appreciate it, otherwise I can help out in this area too.

Again awesome, and thank you!

resources/CMakeLists.txt
jmarrec: This will just copy them to the `build/resources` directory, not embed them.

ruby/test/Issue_2560_Test.rb
jmarrec: The main point here is that these are **ruby** tests instead of being **Google Test**s (C++). Google Tests would have access to resourcesPath https://github.com/NREL/OpenStudio/blob/a007b7801e3bb05731f012f585facdd814ef31fb/resources/resources.hxx.in#L6-L9

I'm not sure why these are ruby tests, whether that's written in the scope of work or not. I also realize it would take a bit of work to make them C++ tests, so I don't really have an opinion. I can propose a safer workaround for now.

resources/CMakeLists.txt
```diff
@@ -111,6 +111,11 @@ set(ENERGYPLUS_OUTPUTS "")
...
+  model/floorplan_school.osm
+  model/7-7_Windows_Complete.osm
```
jmarrec, line 117: `7-7_Windows_Complete.osm`  appears unused anywhere in the code.

jmarrec: There's something that needs cleanup here: https://github.com/NREL/OpenStudio/blob/c0e8704114bbc582d1e9180b89bb853045ca6930/src/model/test/Space_GTest.cpp#L2576-L2578

src/utilities/geometry/Polygon3d.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+  for (size_t i = 0; i < m_outerPath.size(); i++) {
+    Point3d pp1 = m_outerPath[i];
+    Point3d pp2 = m_outerPath[(i + 1) % m_outerPath.size()];
+
+    Point3d p1 = ti * m_outerPath[i];
+    Point3d p2 = ti * m_outerPath[(i + 1) % m_outerPath.size()];
+
+    // The two line segments must be parallel and coincident within tolerance
+    if (abs(p1.y()) > 0.01 || abs(p2.y()) > 0.01) continue;
```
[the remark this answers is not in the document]
jmarrec: cf https://github.com/NREL/OpenStudio/pull/4275/commits/8c07f5a61862b1b0f0dbd57a8fe65049c6d8ae1a

src/model/Surface.cpp
```diff
@@ -1034,10 +1034,17 @@ namespace model {
     }
 
     boost::optional<SurfaceIntersection> Surface_Impl::computeIntersection(Surface& otherSurface) {
-      double tol = 0.01;  // 1 cm tolerance
+      double tol = 0.01;       // 1 cm tolerance
+      double areaTol = 0.001;  // 10mm2 tolerance
```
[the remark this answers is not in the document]
jmarrec: cf 0be4881386579520fa0cb37bde5d2c49567790e4

src/model/Surface.cpp
```diff
@@ -1034,10 +1034,17 @@ namespace model {
...
 
       boost::optional<Space> space = this->space();
       boost::optional<Space> otherSpace = otherSurface.space();
+
+      std::string spaceName = space->nameString();
+      std::string otherSpaceName = otherSpace->nameString();
+      std::string surfaceName = nameString();
+      std::string otherSurfaceName = otherSurface.nameString();
```
jmarrec, line 1046: All unused. cf ad9901650a4fd017c35c1b98bae77350e914c598

src/model/Surface.cpp
```diff
@@ -2011,6 +2021,30 @@ namespace model {
...
+      if (surfaceType() == "Floor" && outsideBoundaryCondition() == "Ground") {
+        auto vertices = this->vertices();
+        if (vertices.size() > 0 && vertices[0].z() == 0) {
+          vertices = tr * vertices;
+          for (size_t i = 0; i < vertices.size(); i++) {
+            Point3dVector line;
+            line.push_back(vertices[i]);
+            line.push_back(vertices[(i + 1) % vertices.size()]);
+            Point3dVectorVector overlaps = buildingPerimeter.overlap(line);
+            for (auto overlap : overlaps) {
```
jmarrec, line 2038: const auto&

ruby/test/Issue_2560_Test.rb
jmarrec: I removed all save code from the ruby tests in fb25ab9186fbd6fb71ed041858e18380bd436ca2

jmarrec: The questiona bout whether we should change these ruby tests to be C++ tests still stands. (Also, I think there are a couple of tests that are both in C++ and ruby tests)

ruby/test/Issue_2560_Test.rb
jmarrec: https://github.com/NREL/OpenStudio/commit/fb25ab9186fbd6fb71ed041858e18380bd436ca2

src/utilities/CMakeLists.txt
jmarrec: removed in bff04eb459d0bd3c3e4f2588239ee3f348159cec

src/utilities/geometry/Polygon3d.hpp
jmarrec: I don't think we need Pimpl here. This is short-lived, for computing intersections and co, and isn't something like a ModelObject that we need to serialize and also query all objects like model.objects()

ruby/test/RemoveSpikes&Overlaps_TZ46_TZ47_Test.rb
```diff
@@ -0,0 +1,412 @@
...
+        OpenStudio::Model::intersectSurfaces(spacesF)#,false,true)
+        OpenStudio::Model::matchSurfaces(spacesF)
+
+        outpath = OpenStudio::Path.new(File.dirname(__FILE__) + "/output/"+@osm_name)
+        model.save(outpath, true);
+        #end
+
+
+
+        #include AMod
```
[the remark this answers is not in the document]
jmarrec: Removed in ef3201bd71f80c12bb1f4139606e23d20ff72d8e

ruby/test/RemoveSpikes&Overlaps_TZ46_TZ47_Test.rb
```diff
@@ -0,0 +1,412 @@
...
+    end
+end
+#model = OpenStudio::Model::Model.new
+#space = Openstudio::Model::Space::Space(model)
```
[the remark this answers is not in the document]
jmarrec: Removed in ef3201bd71f80c12bb1f4139606e23d20ff72d8e

ruby/test/RemoveSpikesAndOverlaps_TZ46_TZ47_Test.rb
```diff
@@ -0,0 +1,404 @@
...
+        model.save(File.expand_path('../../.') + "/build/resources/model/removeSpikesAndOverlaps_TZ46_TZ47.osm", true);
+        #end
+
+
+
+        #include AMod
+    end
+end
+#model = OpenStudio::Model::Model.new
+#space = Openstudio::Model::Space::Space(model)
```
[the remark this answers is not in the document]
jmarrec: Removed in ef3201bd71f80c12bb1f4139606e23d20ff72d8e

ruby/test/RemoveSpikesAndOverlaps_TZ46_TZ47_Test.rb
```diff
@@ -0,0 +1,404 @@
...
+                end
+            end
+        end
+
+        # Don't sort by area, use the shrink and expand to remove spikes
+        OpenStudio::Model::intersectSurfaces(spacesF)#,false,true)
+        OpenStudio::Model::matchSurfaces(spacesF)
+
+        # outpath = OpenStudio::Path.new(File.dirname(__FILE__) + "/output/"+@osm_name)
+        model.save(File.expand_path('../../.') + "/build/resources/model/removeSpikesAndOverlaps_TZ46_TZ47.osm", true);
```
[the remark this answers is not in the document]
jmarrec: Done in fb25ab9186fbd6fb71ed041858e18380bd436ca2

src/model/Building.cpp
```diff
@@ -932,6 +933,27 @@ namespace model {
       return Transformation::rotation(Vector3d(0, 0, 1), -degToRad(this->northAxis()));
     }
 
+    double Building_Impl::exteriorPerimeter() {
+
+      Point3dVectorVector polygons;
+
+      for (auto space : model().getConcreteModelObjects<Space>()) {
```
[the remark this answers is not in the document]
jmarrec: ad9901650a4fd017c35c1b98bae77350e914c598

src/model/Building.cpp
```diff
@@ -932,6 +933,27 @@ namespace model {
...
+        Transformation spaceTransformation = space.transformation();
+        for (auto surface : space.surfaces()) {
+          if (surface.outsideBoundaryCondition() == "Ground" || surface.surfaceType() == "Floor") {
+            Point3dVector points = spaceTransformation * surface.vertices();
+            if (points.size() > 0 && points[0].z() == 0) polygons.push_back(points);
```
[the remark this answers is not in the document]
jmarrec: ad9901650a4fd017c35c1b98bae77350e914c598

src/model/Space.cpp
```diff
@@ -2224,6 +2225,8 @@ namespace model {
         return;
       }
 
+      std::string spaceName = name().value();
```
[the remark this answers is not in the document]
jmarrec: I removed it in ad9901650a4fd017c35c1b98bae77350e914c598

src/model/Surface.cpp
jmarrec: done in ad9901650a4fd017c35c1b98bae77350e914c598 along with many others

ruby/test/Issue_2560_Test.rb
ggartside: We are contracted to provide both c++ and ruby tests so we started with a set of c++ tests then duplicated them in ruby at the end of the project, There are some tests we haven't pushed yet because we were trying to find out how to specify the location of model files but now that might not be an issue any more if some of these tests are not require or we create the model through the API

jmarrec:
Aside from the changes I've suggested in #4275, this looks very good to me! Nice work @ggartside

shorowit:
Is the "exposed perimeter" defined as the edges of the floor surface along the footprint of the building? Note how [exposed perimeter is defined for the EnergyPlus Kiva foundation model](https://bigladdersoftware.com/epx/docs/9-4/input-output-reference/group-advanced-surface-concepts.html#surfaceproperty-exposedfoundationperimeter).

tijcolem:
@jmarrec thanks for review/changes in #4275. That has been merged in and all passing on CI. 

@ggartside I think the remaining parts now are to tidy up the unit tests and use create the model via the API. 
I know this is not possible for all tests but for the ones you can this would be great.  Let us know if there are any other loose ends, otherwise, once those are tests are updated and passing CI we can merge!

src/utilities/geometry/Polygon3d.hpp
macumber: I agree with @jmarrec here.  The other geometry classes are Plain Old Data types

src/model/Building.cpp
```diff
@@ -932,6 +933,30 @@ namespace model {
...
+    double Building_Impl::exteriorPerimeter() {
+
+      Point3dVectorVector polygons;
+
+      for (const auto& space : model().getConcreteModelObjects<Space>()) {
+        Transformation spaceTransformation = space.transformation();
+        for (const auto& surface : space.surfaces()) {
+          if (surface.outsideBoundaryCondition() == "Ground" || surface.surfaceType() == "Floor") {
+            Point3dVector points = spaceTransformation * surface.vertices();
+            if (!points.empty() && points[0].z() == 0.0) {
```
macumber, line 945: Might want to check that all the points have z == 0?

macumber: Another approach would be to project every floor onto z=0 by just setting z = 0 on all the polygons

src/model/Building.cpp
```diff
@@ -932,6 +933,30 @@ namespace model {
       return Transformation::rotation(Vector3d(0, 0, 1), -degToRad(this->northAxis()));
     }
 
+    double Building_Impl::exteriorPerimeter() {
```
macumber, line 936: You might see if ASHRAE, BEDES, or some other standard has a definition of exterior perimeter you can reference

src/model/Space.cpp
```diff
@@ -2727,6 +2729,30 @@ namespace model {
       return result;
     }
 
+    double Space_Impl::exposedPerimeter(const Polygon3d& buildingPerimeter) const {
```
macumber, line 2732: Where do you get buildingPerimeter from?  Maybe it should get perimeter from the current story?

macumber: Depending on your definitions buildingPerimeter could be multiple disjoint polygons.

src/utilities/geometry/Intersection.cpp
```diff
@@ -64,10 +65,56 @@ namespace openstudio {
...
+    }
+    return result[0];
+  }
+}
+
 BoostPolygon removeSpikes(const BoostPolygon& polygon) {
   BoostPolygon temp(polygon);
   boost::geometry::remove_spikes(temp);
-  return temp;
+  return removeSpikesEx(temp);
```
macumber, line 117: Do you need both remove spikes calls?

src/utilities/geometry/Polygon3d.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include "Geometry.hpp"
+#include "Intersection.hpp"
+#include <utilities/geometry/Transformation.hpp>
+
+namespace openstudio {
+
+Polygon3d::Polygon3d() {}
+
+Polygon3d::Polygon3d(const Point3dVector& outerPath) {
+  for (const auto& p : outerPath) {
```
macumber, line 41: Should you throw if outerPath is not planar?

src/utilities/geometry/Polygon3d.cpp
```diff
@@ -0,0 +1,236 @@
...
+Point3dVector Polygon3d::getOuterPath() const {
+  return m_outerPath;
+}
+
+Point3dVectorVector Polygon3d::getInnerPaths() const {
+  return m_innerPaths;
+}
+
+void Polygon3d::addHole(const Point3dVector& hole) {
+  m_innerPaths.emplace_back(hole);
```
macumber, line 75: Should you check that hole is on the same plane as the outer path? We have a Plane class you could use

macumber: Should you check the hole is contained in the outer path?

src/utilities/geometry/Polygon3d.cpp
```diff
@@ -0,0 +1,236 @@
...
+    Point3d p1 = m_outerPath[i];
+    Point3d p2 = m_outerPath[(i + 1) % m_outerPath.size()];
+    perimeter += openstudio::getDistance(p1, p2);
+  }
+
+  return perimeter;
+}
+
+bool Polygon3d::isClockwise() {
+  if (OptionalVector3d normal = getOutwardNormal(m_outerPath)) {
```
macumber, line 124: This doesn't really make sense, it's only clockwise if viewed from above or something, do you need this method? I think it might be backwards too.  At least as I think about it a clockwise polygon has an outward normal that goes down.

src/utilities/geometry/Polygon3d.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+Point3d Polygon3d::centroid() {
+  if (auto p = openstudio::getCentroid(m_outerPath)) {
+    return p.get();
+  }
+
+  return Point3d();
+}
+
+bool Polygon3d::pointInPolygon(const Point3d& point, double tol) {
```
macumber, line 139: I don't get the differences between pointInPolygon, within, and in.  pointInPolygon doesn't exclude holes, within does, and in is the union of both?

src/utilities/geometry/Polygon3d.cpp
```diff
@@ -0,0 +1,236 @@
...
+bool Polygon3d::inside(const Point3d& point, double tol) {
+  return pointInPolygon(point, tol) || within(point, tol);
+}
+
+size_t Polygon3d::getSize() {
+  return m_outerPath.size();
+}
+
+Point3dVector Polygon3d::getEdge(size_t i) {
+  Point3d p1 = m_outerPath[i];
```
macumber, line 177: check that i < size?

src/utilities/geometry/Polygon3d.cpp
```diff
@@ -0,0 +1,236 @@
...
+  Point3d p2 = m_outerPath[(i + 1) % m_outerPath.size()];
+  return Point3dVector{p1, p2};
+}
+
+std::vector<Point3dVector> Polygon3d::overlap(const Point3dVector& line) const {
```
macumber, line 182: This handles the case of a line intersecting a U shaped polygon twice?  Does it find intersections with holes too?  What is this used for?

src/utilities/geometry/Polygon3d.hpp
```diff
@@ -0,0 +1,117 @@
...
+  // Gets the perimeter of the outer path of the polygon
+  double perimeter();
+
+  // Determines whether the polygon os clockwise (normal down) or anti-clockwise (normal up)
+  bool isClockwise();
+
+  // Calculates the centroid of the polygon
+  Point3d centroid();
+
+  // Point is on the outer path or one of the inner paths
```
macumber, line 90: Shouldn't the inner paths be inside the outer path?  Do you mean "on the outer path" or "inside the outer path"?

src/utilities/geometry/Polygon3d.hpp
```diff
@@ -0,0 +1,117 @@
...
+  bool pointInPolygon(const Point3d& point, double tol = 0.01);
+
+  // Point inside outer path and not inside a hole
+  bool within(const Point3d& point, double tol = 0.01);
+
+  // Point is inside the outer path and not inside a hole or on the outer path or one of the inner paths
+  bool inside(const Point3d& point, double tol = 0.01);
```
macumber, line 97: Seems redundant to pointInPolygon

macumber:
Nice job, I like the Polygon class and especially all the tests!  I left some comments.  My hope is that one day we could port [Topolys](https://github.com/automaticmagic/topolys) to C++ and add to OpenStudio.  Maybe consider that for future work?  I would be happy for DA to do the port if it is useful

jmarrec:
@ggartside I modified your Original post above to actually write Fix #1614  (this has the effect of automatically closing the associated issue when this PR is merged). I noticed two issues you may be fixing as well:

Fix #1398 - Native Support to calculate building perimeter

Fix #1586 - joinAllMethod seems be to order dependent

Please add them to your original post ahead if it does fix them.

src/utilities/geometry/Polygon3d.cpp
ggartside: Yep twice or as many times as necessary

It finds all the segments on a polygon that overlap a line, so the polygon would be the building story footprint and the line would be an edge of a floor surface and the overlapping segments would give you the total length of the surface that is coincident with the building story shape, the exposed part of the floor.

ggartside:
> 
> 
> Nice job, I like the Polygon class and especially all the tests! I left some comments. My hope is that one day we could port [Topolys](https://github.com/automaticmagic/topolys) to C++ and add to OpenStudio. Maybe consider that for future work? I would be happy for DA to do the port if it is useful

Thanks! I see the polygon class as away to do the surface intersection and matching without having to immediately triangulate to remove holes. If you defer triangulation to the end then in a lot of cases you will end up with a simpler set of surfaces. Some times the holes will be removed by subsequent intersections so it isn't event necessary to remove them. If you look at page 4 of the attached document you'll see what I mean. And also on page 5 you'll see a bunch of off angle edges when most surfaces are oriented vertically or horizontally, those extra edges are the result of triangulation.

And finally I think you can get a big performance gain by matching when you intersect I I think you are aware of this because there is a comment int he code about this). When you intersect two surfaces and get the common are between them you have by definition already found the matching pair of surfaces so there is no need to have to 're-find' them later

I will have to take a look at Topolys to get a better understanding of it. 

[Phase 3 - Prototyping.docx](https://github.com/NREL/OpenStudio/files/6286612/Phase.3.-.Prototyping.docx)

kbenne (approved):

tijcolem:
@ggartside CI failures are due to scheduled brown outs at conan - https://status.bintray.com/.  
Should be back up later today and I will re-run for the latest commits. 

The bintray is something we are aware of and are actively moving away.

src/utilities/geometry/Polygon3d.hpp
ggartside: Depends on if you need define inside being within the polygon bounds but not on one of the paths or whether inside is defined as being within the polygon bounds or on one of the paths. The calling context would dictate that so these methods just provide the flexibility

And also you may want to know only whether a point is on one of the paths

src/utilities/geometry/Intersection.cpp
ggartside: I didn't see any harm in leaving the original call in so I left it. That way there's two attempts to remove spikes.

Maybe in the future the boost call can be removed

src/utilities/geometry/Polygon3d.cpp
ggartside: You are right it is reversed (well with RHS coordinates it is). On reflection I think it would be best to remove this method

jmarrec:
> @ggartside I modified your Original post above to actually write Fix #1614 (this has the effect of automatically closing the associated issue when this PR is merged). I noticed two issues you may be fixing as well:
> 
> Fix #1398 - Native Support to calculate building perimeter
> 
> Fix #1586 - joinAllMethod seems be to order dependent
> 
> Please add them to your original post ahead if it does fix them.

@ggartside Ping on this one please. I think #1398 at least got fixed by this PR. I would appreciate if you would scan the list of open issues assigned to you and inform whether you think this PR fixed them (https://github.com/NREL/OpenStudio/issues?q=assignee%3Aggartside+is%3Aopen) Thanks!

(cc @tijcolem )

PR #4225, opened 2021-02-24
Addresses #4223, wrap the new ElectricLoadCenter:Storage:LiIonNMCBattery object in the OS SDK

### LINKED ISSUE #4223: E+ 9.5.0: Wrap the new ElectricLoadCenter:Storage:LiIonNMCBattery object in the OS SDK
Enhancement Request
--------------
E+ 9.5.0: Wrap the new ElectricLoadCenter:Storage:LiIonNMCBattery object in the OS SDK New battery model from SAM NREL/EnergyPlus#8499

Ref https://github.com/NREL/OpenStudio/issues/4210

joseph-robertson:
Pull request overview
---------------------

 - Fixes #4223

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


kbenne:
@joseph-robertson are you ready for me to merge this?

resources/model/OpenStudio.idd
```diff
@@ -31338,6 +31338,141 @@ OS:ElectricLoadCenter:Storage:Simple,
...
+        \required-field
+   A2,  \field Name
+        \required-field
+        \type alpha
+        \reference ElecStorageList
+   A3,  \field Availability Schedule Name
+        \note Availability schedule name for this system. Schedule value > 0 means the system is available.
+        \note If this field is blank, the system is always available.
+        \type object-list
+        \object-list ScheduleNames
```
jmarrec, line 31361:
```suggestion
        \required-field
        \type object-list
        \object-list ScheduleNames
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery.hpp
```diff
@@ -0,0 +1,193 @@
...
+    /** @name Constructors and Destructors */
+    //@{
+
+    explicit ElectricLoadCenterStorageLiIonNMCBattery(const Model& model);
+
+    virtual ~ElectricLoadCenterStorageLiIonNMCBattery() {}
+
+    //@}
+
+    static IddObjectType iddObjectType();
```
jmarrec, line 62: Missing `static std::vector<std::string> lifetimeModelValues();`

src/model/ElectricLoadCenterStorageLiIonNMCBattery.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+    /** @name Getters */
+    //@{
+
+    // Convenience method to return the electricalLoadCenter on which it's assigned (optional)
+    // In parent
+    // boost::optional<ElectricLoadCenterDistribution> electricLoadCenterDistribution() const;
+
+    Schedule availabilitySchedule() const;
+    bool isAvailabilityScheduleDefaulted() const;
```
jmarrec, line 72:
```suggestion
    Schedule availabilitySchedule() const;
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery.hpp
```diff
@@ -0,0 +1,193 @@
...
+    double chargeRateatWhichVoltagevsCapacityCurveWasGenerated() const;
+
+    double batteryCellInternalElectricalResistance() const;
+
+    //@}
+    /** @name Setters */
+    //@{
+
+    bool setAvailabilitySchedule(Schedule& schedule);
+    void resetAvailabilitySchedule();
```
jmarrec, line 120:
```suggestion
    bool setAvailabilitySchedule(Schedule& schedule);
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+    // TODO: Included in parent class, shouldn't need to define it...
+    //virtual bool setThermalZone(ThermalZone& zone) override;
+    //virtual void resetThermalZone() override;
+
+    bool setRadiativeFraction(double radiativeFraction);
+
+    bool setLifetimeModel(std::string lifetimeModel);
```
jmarrec, line 128:
```suggestion
    bool setLifetimeModel(const std::string& lifetimeModel);
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery_Impl.hpp
```diff
@@ -0,0 +1,196 @@
...
+      //@}
+      /** @name Getters */
+      //@{
+
+      // Convenience method to return the electricalLoadCenter on which it's assigned (optional)
+      // Included in Base Class
+      // boost::optional<ElectricLoadCenterDistribution> electricLoadCenterDistribution() const;
+
+      Schedule availabilitySchedule() const;
+      bool isAvailabilityScheduleDefaulted() const;
```
jmarrec, line 78:
```suggestion
      Schedule availabilitySchedule() const;
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery_Impl.hpp
```diff
@@ -0,0 +1,196 @@
...
+      virtual std::vector<EMSActuatorNames> emsActuatorNames() const override;
+
+      virtual std::vector<std::string> emsInternalVariableNames() const override;
+
+      //@}
+      /** @name Setters */
+      //@{
+
+      bool setAvailabilitySchedule(Schedule& schedule);
+      void resetAvailabilitySchedule();
```
jmarrec, line 129:
```suggestion
      bool setAvailabilitySchedule(Schedule& schedule);
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery_Impl.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+      // Override ElectricalStorage (Base class) methods
+      virtual bool setThermalZone(const ThermalZone& thermalZone) override;
+      virtual void resetThermalZone() override;
+
+      bool setRadiativeFraction(double radiativeFraction);
+
+      bool setLifetimeModel(std::string lifetimeModel);
```
jmarrec, line 137:
```suggestion
      bool setLifetimeModel(const std::string& lifetimeModel);
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery_Impl.hpp
```diff
@@ -0,0 +1,196 @@
...
+      //@}
+      /** @name Other */
+      //@{
+
+      // TODO: Remove if not used. Don't think I need to override any of these
+      //ModelObject clone(Model model) const override;
+
+      //std::vector<openstudio::IdfObject> remove() override;
+
+      //std::vector<ModelObject> children() const override;
```
jmarrec, line 182:
```suggestion
```

src/model/test/ElectricLoadCenterStorageLiIonNMCBattery_GTest.cpp
```diff
@@ -0,0 +1,172 @@
...
+    ::testing::ExitedWithCode(0), "");
+
+  Model m;
+  ElectricLoadCenterStorageLiIonNMCBattery battery(m);
+
+  Schedule schedule = battery.availabilitySchedule();
+  boost::optional<ScheduleConstant> scheduleConstant = schedule.optionalCast<ScheduleConstant>();
+  ASSERT_TRUE(scheduleConstant);
+  EXPECT_EQ((*scheduleConstant).value(), 1.0);
+  EXPECT_TRUE(battery.isAvailabilityScheduleDefaulted());
```
jmarrec, line 65:
```suggestion
```

src/model/test/ElectricLoadCenterStorageLiIonNMCBattery_GTest.cpp
```diff
@@ -0,0 +1,172 @@
...
+  EXPECT_TRUE(battery.setFractionofCellCapacityRemovedattheEndofExponentialZone(0.73));
+  EXPECT_TRUE(battery.setFractionofCellCapacityRemovedattheEndofNominalZone(0.74));
+  EXPECT_TRUE(battery.setChargeRateatWhichVoltagevsCapacityCurveWasGenerated(4000));
+  EXPECT_TRUE(battery.setBatteryCellInternalElectricalResistance(6000));
+
+  Schedule schedule = battery.availabilitySchedule();
+  boost::optional<ScheduleConstant> scheduleConstant = schedule.optionalCast<ScheduleConstant>();
+  ASSERT_TRUE(scheduleConstant);
+  EXPECT_EQ((*scheduleConstant).value(), 0.5);
+  EXPECT_FALSE(battery.isAvailabilityScheduleDefaulted());
```
jmarrec, line 122:
```suggestion
```

src/model/test/ElectricLoadCenterStorageLiIonNMCBattery_GTest.cpp
```diff
@@ -0,0 +1,172 @@
...
+  EXPECT_EQ(3.25, battery.cellVoltageatEndofExponentialZone());
+  EXPECT_EQ(3.35, battery.cellVoltageatEndofNominalZone());
+  EXPECT_EQ(3.15, battery.defaultNominalCellVoltage());
+  EXPECT_EQ(3.12, battery.fullyChargedCellCapacity());
+  EXPECT_EQ(0.73, battery.fractionofCellCapacityRemovedattheEndofExponentialZone());
+  EXPECT_EQ(0.74, battery.fractionofCellCapacityRemovedattheEndofNominalZone());
+  EXPECT_EQ(4000, battery.chargeRateatWhichVoltagevsCapacityCurveWasGenerated());
+  EXPECT_EQ(6000, battery.batteryCellInternalElectricalResistance());
+
+  battery.resetAvailabilitySchedule();
```
jmarrec, line 144:
```suggestion
```

src/model/test/ElectricLoadCenterStorageLiIonNMCBattery_GTest.cpp
```diff
@@ -0,0 +1,172 @@
...
+  battery.resetThermalZone();
+
+  EXPECT_TRUE(battery.isAvailabilityScheduleDefaulted());
```
jmarrec, line 147:
```suggestion
```

src/model/test/ElectricLoadCenterStorageLiIonNMCBattery_GTest.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+  battery.setRadiativeFraction(0.6);
+
+  ElectricLoadCenterStorageLiIonNMCBattery batteryClone = battery.clone(m).cast<ElectricLoadCenterStorageLiIonNMCBattery>();
+  EXPECT_EQ(0.6, batteryClone.radiativeFraction());
+
+  Model m2;
+  ElectricLoadCenterStorageLiIonNMCBattery batteryClone2 = battery.clone(m2).cast<ElectricLoadCenterStorageLiIonNMCBattery>();
+  EXPECT_EQ(0.6, batteryClone2.radiativeFraction());
+}
```
jmarrec, line 163: Should cloning reset the thermalZone? I would test either way

src/model/test/ElectricLoadCenterStorageLiIonNMCBattery_GTest.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+TEST_F(ModelFixture, ElectricLoadCenterStorageLiIonNMCBattery_Remove) {
+  Model m;
+  auto size = m.modelObjects().size();
+  ElectricLoadCenterStorageLiIonNMCBattery battery(m);
+  EXPECT_EQ(size + 1, m.modelObjects().size());
+  EXPECT_FALSE(battery.remove().empty());
+  EXPECT_EQ(size, m.modelObjects().size());
+}
```
jmarrec, line 172: Check whether thermalZone isn't deleted perhaps?

src/model/ElectricLoadCenterStorageLiIonNMCBattery.cpp
```diff
@@ -0,0 +1,657 @@
...
+    //    {
+    //      if (elecStor->handle() == this->handle())
+    //      {
+    //        value = elcd;
+    //      }
+    //    }
+    //  }
+    //  return value;
+
+    //}
```
jmarrec, line 130:
```suggestion
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery.cpp
```diff
@@ -0,0 +1,657 @@
...
+        //LOG_AND_THROW(briefDescription() << " does not have an Availability Schedule attached.");
+        // I'm choosing to just return alwaysOnDiscreteSchedule
+        return this->model().alwaysOnDiscreteSchedule();
+      }
+      return value.get();
+    }
+
+    bool ElectricLoadCenterStorageLiIonNMCBattery_Impl::isAvailabilityScheduleDefaulted() const {
+      return isEmpty(OS_ElectricLoadCenter_Storage_LiIonNMCBatteryFields::AvailabilityScheduleName);
+    }
```
jmarrec, line 148:
```suggestion
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery.cpp
```diff
@@ -0,0 +1,657 @@
...
+    bool ElectricLoadCenterStorageLiIonNMCBattery_Impl::setAvailabilitySchedule(Schedule& schedule) {
+      bool result = setSchedule(OS_ElectricLoadCenter_Storage_LiIonNMCBatteryFields::AvailabilityScheduleName,
+                                "ElectricLoadCenterStorageLiIonNMCBattery", "Availability", schedule);
+      return result;
+    }
+
+    void ElectricLoadCenterStorageLiIonNMCBattery_Impl::resetAvailabilitySchedule() {
+      bool result = setString(OS_ElectricLoadCenter_Storage_LiIonNMCBatteryFields::AvailabilityScheduleName, "");
+      OS_ASSERT(result);
+    }
```
jmarrec, line 286:
```suggestion
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery.cpp
```diff
@@ -0,0 +1,657 @@
...
+      bool result = setString(OS_ElectricLoadCenter_Storage_LiIonNMCBatteryFields::ZoneName, "");
+      OS_ASSERT(result);
+    }
+
+    bool ElectricLoadCenterStorageLiIonNMCBattery_Impl::setRadiativeFraction(double radiativeFraction) {
+      bool result = setDouble(OS_ElectricLoadCenter_Storage_LiIonNMCBatteryFields::RadiativeFraction, radiativeFraction);
+      return result;
+    }
+
+    bool ElectricLoadCenterStorageLiIonNMCBattery_Impl::setLifetimeModel(std::string lifetimeModel) {
```
jmarrec, line 305:
```suggestion
    bool ElectricLoadCenterStorageLiIonNMCBattery_Impl::setLifetimeModel(const std::string& lifetimeModel) {
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery.cpp
```diff
@@ -0,0 +1,657 @@
...
+  }  // namespace detail
+
+  /* Constructor. Defaults the availabilitySchedule to alwaysOnDiscreteSchedule */
+  ElectricLoadCenterStorageLiIonNMCBattery::ElectricLoadCenterStorageLiIonNMCBattery(const Model& model)
+    : ElectricalStorage(ElectricLoadCenterStorageLiIonNMCBattery::iddObjectType(), model) {
+    OS_ASSERT(getImpl<detail::ElectricLoadCenterStorageLiIonNMCBattery_Impl>());
+
+    // already defaults to alwaysOn
+    //Schedule schedule = model.alwaysOnDiscreteSchedule();
+    //setAvailabilitySchedule(schedule);
```
jmarrec, line 425:
```suggestion
    Schedule schedule = model.alwaysOnDiscreteSchedule();
    bool ok = setAvailabilitySchedule(schedule);
    OS_ASSERT(ok);
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery.cpp
```diff
@@ -0,0 +1,657 @@
...
+    setFractionofCellCapacityRemovedattheEndofExponentialZone(0.8075);
+    setFractionofCellCapacityRemovedattheEndofNominalZone(0.976875);
+    setChargeRateatWhichVoltagevsCapacityCurveWasGenerated(1);
+    setBatteryCellInternalElectricalResistance(0.09);
+  }
+
+  IddObjectType ElectricLoadCenterStorageLiIonNMCBattery::iddObjectType() {
+    return IddObjectType(IddObjectType::OS_ElectricLoadCenter_Storage_LiIonNMCBattery);
+  }
+
```
jmarrec, line 451:
```suggestion
  std::vector<std::string> ElectricLoadCenterStorageLiIonNMCBattery
::lifetimeModelValues() {
    return getIddKeyNames(IddFactory::instance().getObject(iddObjectType()).get(), OS_ElectricLoadCenter_Storage_LiIonNMCBattery
::LifetimeModel);
  }

```

src/model/ElectricLoadCenterStorageLiIonNMCBattery.cpp
```diff
@@ -0,0 +1,657 @@
...
+  // Setters
+  bool ElectricLoadCenterStorageLiIonNMCBattery::setAvailabilitySchedule(Schedule& schedule) {
+    return getImpl<detail::ElectricLoadCenterStorageLiIonNMCBattery_Impl>()->setAvailabilitySchedule(schedule);
+  }
+
+  void ElectricLoadCenterStorageLiIonNMCBattery::resetAvailabilitySchedule() {
+    getImpl<detail::ElectricLoadCenterStorageLiIonNMCBattery_Impl>()->resetAvailabilitySchedule();
+  }
+
```
jmarrec, line 556:
```suggestion
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery.cpp
```diff
@@ -0,0 +1,657 @@
...
+  // }
+
+  Schedule ElectricLoadCenterStorageLiIonNMCBattery::availabilitySchedule() const {
+    return getImpl<detail::ElectricLoadCenterStorageLiIonNMCBattery_Impl>()->availabilitySchedule();
+  }
+
+  bool ElectricLoadCenterStorageLiIonNMCBattery::isAvailabilityScheduleDefaulted() const {
+    return getImpl<detail::ElectricLoadCenterStorageLiIonNMCBattery_Impl>()->isAvailabilityScheduleDefaulted();
+  }
+
```
jmarrec, line 466:
```suggestion
```

src/model/ElectricLoadCenterStorageLiIonNMCBattery.cpp
```diff
@@ -0,0 +1,657 @@
...
+  //}
+  //void ElectricLoadCenterStorageLiIonNMCBattery::resetThermalZone() {
+  //  getImpl<detail::ElectricLoadCenterStorageLiIonNMCBattery_Impl>()->resetThermalZone();
+  //}
+
+  bool ElectricLoadCenterStorageLiIonNMCBattery::setRadiativeFraction(double radiativeFraction) {
+    return getImpl<detail::ElectricLoadCenterStorageLiIonNMCBattery_Impl>()->setRadiativeFraction(radiativeFraction);
+  }
+
+  bool ElectricLoadCenterStorageLiIonNMCBattery::setLifetimeModel(std::string lifetimeModel) {
```
jmarrec, line 569:
```suggestion
  bool ElectricLoadCenterStorageLiIonNMCBattery::setLifetimeModel(const std::string& lifetimeModel) {
```

jmarrec (changes requested):
Requesting simple changes, and suggested them all (should work, typos aside)

src/model/ElectricLoadCenterStorageLiIonNMCBattery.hpp
jmarrec, line 62:
```suggestion
    static IddObjectType iddObjectType();
    
    static std::vector<std::string> lifetimeModelValues();
```

jmarrec:
@joseph-robertson, addressing my own review comments in https://github.com/NREL/OpenStudio/pull/4225/commits/5ec67802d855e2a4d30f8a43a19828a11b892791,  you can revert part or all if you feel strongly, but I prefer to keep API consistent

src/model/test/ElectricLoadCenterStorageLiIonNMCBattery_GTest.cpp
```diff
@@ -0,0 +1,176 @@
...
+  ElectricLoadCenterStorageLiIonNMCBattery battery(m);
+  EXPECT_TRUE(battery.setThermalZone(z));
+
+  battery.setRadiativeFraction(0.6);
+
+  ElectricLoadCenterStorageLiIonNMCBattery batteryClone = battery.clone(m).cast<ElectricLoadCenterStorageLiIonNMCBattery>();
+  EXPECT_EQ(0.6, batteryClone.radiativeFraction());
+  // TODO: should the clone retain the thermalZone or clear it?
+  EXPECT_TRUE(batteryClone.thermalZone());
+  EXPECT_EQ(z, batteryClone.thermalZone());
```
jmarrec, line 159: @kbenne @joseph-robertson This is the last TODO IMHO

jmarrec (changes requested):

src/model/test/ElectricLoadCenterStorageLiIonNMCBattery_GTest.cpp
jmarrec: I noticed that ElectricLoadCenterStorageConverter for eg keeps it, so I'm fine leave it as is

jmarrec (approved):

jmarrec:
LGTM, I suggest merging @kbenne

joseph-robertson:
@jmarrec @kbenne Not quite ready to merge yet; still waiting for feedback on constructor(s) from @nmerket.

src/energyplus/ReverseTranslator/ReverseTranslateElectricLoadCenterStorageLiIonNMCBattery.cpp
```diff
@@ -0,0 +1,200 @@
...
+    d = workspaceObject.getDouble(ElectricLoadCenter_Storage_LiIonNMCBatteryFields::BatteryMass);
+    OS_ASSERT(d);
+    const double mass = *d;
+
+    // Battery Surface Area
+    d = workspaceObject.getDouble(ElectricLoadCenter_Storage_LiIonNMCBatteryFields::BatterySurfaceArea);
+    OS_ASSERT(d);
+    const double surfaceArea = *d;
+
+    openstudio::model::ElectricLoadCenterStorageLiIonNMCBattery elcStorLiIonNMCBattery(m_model, nSeries, nParallel, mass, surfaceArea);
```
nmerket, line 79: @joseph-robertson I made these changes to get the fields we need to construct the object before doing so. Because they're required, I'm no longer doing the `if(i) ...` before setting it because it needs to be there. Is using `OS_ASSERT` a good way to do that here, or do I need to have it return a more gentile error message?

src/model/ElectricLoadCenterStorageLiIonNMCBattery.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+  /** ElectricLoadCenterStorageLiIonNMCBattery is a ElectricalStorage that wraps the OpenStudio IDD object 'OS:ElectricLoadCenter:Storage:LiIonNMCBattery'. */
+  class MODEL_API ElectricLoadCenterStorageLiIonNMCBattery : public ElectricalStorage
+  {
+   public:
+    /** @name Constructors and Destructors */
+    //@{
+
+    explicit ElectricLoadCenterStorageLiIonNMCBattery(const Model& model, const int numberofCellsinSeries, const int numberofStringsinParallel,
+                                                      const double batteryMass, const double batterySurfaceArea);
```
nmerket, line 57: I changed the constructor to require these additional fields.

nmerket (commented):
I still want to try to come up with an additional alternative constructor and possibly a setter or two that takes more user-friendly inputs like "target battery capacity [kwh]" and generates some of the other inputs like series and parallel from that. However, if we need to merge this sooner, I'm fine with this the way it is and I'll do that in a later version or in our measure code.

src/energyplus/ReverseTranslator/ReverseTranslateElectricLoadCenterStorageLiIonNMCBattery.cpp
joseph-robertson: I'd say follow this as an example: https://github.com/NREL/OpenStudio/blob/develop/src/energyplus/ReverseTranslator/ReverseTranslateSurfaceControlMovableInsulation.cpp. Both `surface` and `material` are passed into the ctor.

nmerket:
@joseph-robertson Do you know what's going wrong with the windows incremental build?

joseph-robertson:
@nmerket I don't think it's related to anything you're doing. Question for @tijcolem.

src/model/ElectricLoadCenterStorageLiIonNMCBattery.hpp
jmarrec: I'm not a big fan of this change. If we can provide reasonable defaults (I think we can, just pick whatever is your the E+ example file) I'd prefer to have a `ElectricLoadCenterStorageLiIonNMCBattery(const Model& model)` ctor.

You can have 2 constructors if you want this explicit one. And you could still throw in the ReverseTranslator if the IDF doesn't have one of these fields if you like (or issue a Warn/Error, and let the default pass in)

(This isn't "out the blue", we've had lengthy discussions about this in the past).

jmarrec:
> @joseph-robertson Do you know what's going wrong with the windows incremental build?

MSVC is throwing because of your RT:

```
    OptionalInt nSeries;
    nSeries = workspaceObject.getInt(ElectricLoadCenter_Storage_LiIonNMCBatteryFields::NumberofCellsinSeries);
    if (not nSeries) {     //// <============= THIS
```

jmarrec:
Ok, now Windows reports a failure that you can ignore, it's a know ruby test that's failing

```
The following tests FAILED:
	3000 - RubyTest-Date_Test-ymd_constructor (Failed)
```

joseph-robertson:
@nmerket @jmarrec So what's the decision here? To merge as it is now? Or to provide alternative ctor(s)?

jmarrec:
I would personally like an alternative, model only, ctor. I'm fine with the rest. @kbenne thoughts?

kbenne:
> I would personally like an alternative, model only, ctor. I'm fine with the rest. @kbenne thoughts?

I also prefer that ModelObject constructors offer one overload that requires only a model argument. Nothing wrong with other more explicit overloads, but I feel like the simple construction with one argument makes the API more approachable.

nmerket:
@jmarrec @kbenne I intentionally removed the "model only" constructor because I felt like some fields should be provided by the user. The additional constructor I am thinking of adding is one that will calculate the number of series and parallel based on a desired capacity. I'm worried that a user will just use the default constructor and assume that it's a good model, when it's not. Maybe I'm not understanding the paradigm of how the OpenStudio API should work.

nmerket:
Upon further consideration, I think I'm going to wait until the next release to add another constructor with the capacity that calculates the series and parallel. There's still some logic we need to figure out on what should be defaulted and how that will work in practice. 

As far as I'm concerned, I'm happy with this the way it is to merge, so don't wait on me. I'm headed out of town next week and won't be able to work on this, so if you feel very strongly that it needs a model only constructor, please go ahead and put it back in.

Thanks for being patient with me on this.

jmarrec:
> @jmarrec @kbenne I intentionally removed the "model only" constructor because I felt like some fields should be provided by the user. The additional constructor I am thinking of adding is one that will calculate the number of series and parallel based on a desired capacity. I'm worried that a user will just use the default constructor and assume that it's a good model, when it's not. Maybe I'm not understanding the paradigm of how the OpenStudio API should work.

@nmerket I totally understand your initial point of view on this, but I'll try to provide some insight as to why I think differently. I think providing a Ctor that takes the required argument(s) is fine and helps drive the point that they should be provided. But I still want a Ctor that takes `Model&` only **if we can** (the cases where we can't are very few). The only arguments that would be valid to pass in - IMHO - are required **objects** (Curves, HVACComponents, etc). And in most cases I'd still want to have one that takes these, and one that takes `Model&` only and **instantiates** the rest so we have:

1. An easy way to use the object and **inspect** it by providing `Model&` only
2. An advanced way of passing everything required to avoid duplicating objects (especially for ResourceObjects, such as Curves) if you are concerned about speed/performance, file size, or avoid repeating yourself

The reason is because it's much harder to use the API if you don't. When I need to write a measure or some code, I jump into an interactive ruby prompt (via `pry`, which is `irb` but with better tab completion). Most of the time, I would try to construct the object I'm interested it by just providing `Model` to the class Ctor, and most of the time it works. If not, I can generally figure it out. Once I have the object, I can print it, inspect its methods, etc, so I know how to use the object without any external resources. Having to go look up SDK documentation is annoying (mind that in most cases you also are already looking at the E+ I/O ref...)

In your specific case, you're trying to require numeric fields. Most objects in the API have double/integer/choice fields that must be tweaked in order to produce realistic results (or just because they are marked `\required-field` in the IDD). If we suddenly made Ctors that take in all of these, most Ctors would be 5 to 10 arguments long and the API would be a nightmare to use.
It's the user's responsibility to tweak these, or embrace the Garbage-In Garbage-Out (GIGO) principle.

---


**Some Demonstration**

Most of the time, I can construct an object

```ruby
[1] julien(main)> include OpenStudio::Model
=> Object
[2] julien(main)> m = Model.new
=> #<OpenStudio::Model::Model:0x000055b4845e08d8
 @__swigtype__="_p_openstudio__model__Model">
[3] julien(main)> b = BoilerHotWater.new(m)
=> #<OpenStudio::Model::BoilerHotWater:0x000055b4840b3450
```

Sometimes it doesn't work, but I can trick ruby into giving me more info by passing bogus args, and I get enough info to go on (eg: below, from domain knowledge and argument type, I know it's the `AvailabilitySchedule` that's required).

```ruby
[6] julien(main)> atu = AirTerminalSingleDuctConstantVolumeNoReheat.new(m)
ArgumentError: wrong # of arguments(1 for 2)
from (pry):6:in `initialize'

# Pass bogus arg
[7] julien(main)> atu = AirTerminalSingleDuctConstantVolumeNoReheat.new(m, "")
TypeError: Expected argument 1 of type openstudio::model::Schedule &, but got String ""
	in SWIG method 'AirTerminalSingleDuctConstantVolumeNoReheat'
from (pry):7:in `initialize'
```

Now, what happens when you provide a Ctor that needs a lot of arguments, and with the same type for different args? Well, there is **no way** other than to go to the SDK documentation to look it up, **which is clumsy and makes you lose time**.

```
[8] julien(main)> ptac = ZoneHVACPackagedTerminalAirConditioner.new(m)
ArgumentError: wrong # of arguments(1 for 5)
from (pry):8:in `initialize'
[9] julien(main)> ptac = ZoneHVACPackagedTerminalAirConditioner.new(m, "", "", "", "")
TypeError: Expected argument 1 of type openstudio::model::Schedule &, but got String ""
	in SWIG method 'ZoneHVACPackagedTerminalAirConditioner'
from (pry):9:in `initialize'
[10] julien(main)> ptac = ZoneHVACPackagedTerminalAirConditioner.new(m, m.alwaysOnDiscreteSchedule, "", "", "")
TypeError: Expected argument 2 of type openstudio::model::HVACComponent &, but got String ""
	in SWIG method 'ZoneHVACPackagedTerminalAirConditioner'

# At this point I know it's expecting either the Fan, the Heating COil or the Cooling coil, which one is it?!
```

I need several clicks to get here: https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.1.0-doc/model/html/classopenstudio_1_1model_1_1_zone_h_v_a_c_packaged_terminal_air_conditioner.html and even the signature isn't enough, I need an extra click until I see the **arguments names**

image

**One I have the object, I can inspect it, look at its methods, etc**

```ruby
[13] julien(main)> puts atu
OS:AirTerminal:SingleDuct:ConstantVolume:NoReheat,
  {d12f35b0-b126-4a67-a928-d5e5e5eb8da8}, !- Handle
  Air Terminal Single Duct Constant Volume No Reheat 1, !- Name
  {b22711ff-fbcd-4967-8fe1-70b8c804d423}, !- Availability Schedule Name
  ,                                       !- Air Inlet Node Name
  ,                                       !- Air Outlet Node Name
  AutoSize;                               !- Maximum Air Flow Rate {m3/s}

=> nil
```

This is OpenStudio, so most of the time by looking at a field I can infer its corresponding getter/setter. `Maximum Air Flow Rate` => I expect the getter to be `maximumAirFlowRate()` and the setter to be `setMaximumAirFlowRate`. I can also use the `methods` function, chain it with grep and a regex, to find all Setters that apply.

```ruby
[14] julien(main)> atu.methods.grep(/^set/)
=> [:setMaximumAirFlowRate,
 :setAvailabilitySchedule,
 :setParent,
 :setPointer,
 :setFieldComment,
 :setString,
 :setDouble,
 :setQuantity,
 :setComment,
 :setUnsigned,
 :setInt,
 :setName]
```
```

jmarrec:
I put back a Model only ctor in https://github.com/NREL/OpenStudio/pull/4225/commits/4096a0be4fdc12554b7850484511de6e5c3526f4. I left the (unusual in my opinion) ctor that takes numeric field in place though.

PR #4238, opened 2021-03-09
Lint geometry_improvements and use const refs

jmarrec:
Pull request overview
---------------------

Proposed changes to PR https://github.com/NREL/OpenStudio/pull/4221

Lint geometry_improvements and use const refs + rename to NewallVecto…r (to match existing `openstudio::getNewallVector`)

@ggartside  I'll let you review and determine whether they make sense or not, and whether you want to pull them into your branch.


jmarrec:
I temporarily pointed to `develop` branch so that CI would build this branch, tests are fine, so switching back to branch for #4221

PR #4251, opened 2021-03-19
Cbecc 9 4

kbenne:
Pull request overview
---------------------

The purpose of this PR is to introduce the latest features of the CBECC-Com - OpenStudio translator.


kbenne:
@jmarrec your review is especially welcome on a couple of items that extend outside of the CBECC translator

1. A change to the built in always on / off schedules that the model generates to avoid duplicates.
2. An addition of radiant surface type "wall"

@tijcolem CBECC folks use their own builds of OpenStudio from their own branch, but I do like to keep OS develop up to speed with these changes before release.

kbenne:
Hey @tijcolem and @jmarrec I need to stage this a bit differently by creating a CBECC-9-5 branch. So I'm going to pull this back and create a new PR.

PR #4275, opened 2021-04-07
Geometry improvements review

jmarrec:
Proposed changes to #4221. @ggartside I didn't want to hijack your branch so I created this one, to facilitate your review of my proposed changes. Most of them are cleanup/cosmetic, but I also re-enabled the tests to run on unix. I switched our Jenkins CI library so it would also run the disabled intersect tests (`Space_intersectSurfaces_degenerate3` for eg)

Note: for now I'm pointing to develop so that CI builds it and tests it. Then I'll change the base to @ggartside 's branch.

### DIFF
--- a/Jenkinsfile_develop_osx
+++ b/Jenkinsfile_develop_osx
@@ -1,6 +1,6 @@
 //Jenkins pipelines are stored in shared libaries. Please see: https://github.com/NREL/cbci_jenkins_libs
 
-@Library('cbci_shared_libs') _
+@Library('cbci_shared_libs@geometry_improvements') _
 
 // Build for PR to develop branch only.
 if ((env.CHANGE_ID) && (env.CHANGE_TARGET) ) {
--- a/ruby/test/Issue_2560_Test.rb
+++ b/ruby/test/Issue_2560_Test.rb
@@ -35,6 +35,7 @@ require 'csv'
 
 class Issue_2560_Test < Minitest::Test
   def test_Issue_2560
+
     model = OpenStudio::Model::Model.new
     story2 = OpenStudio::Model::BuildingStory.new(model)
 
@@ -93,8 +94,9 @@ class Issue_2560_Test < Minitest::Test
     sp4.setThermalZone(tz4)
 
     # Model before intersection
-    outpath = File.join(File.expand_path('../../.') + "/build/resources/model/","2560_before.osm")
-    model.save(outpath, true)
+    # outdir = File.absolute_path(File.join(OpenStudio::getEnergyPlusDirectory.to_s, "../resources/model"))
+    # outpath = File.join(outdir, "2560_before.osm")
+    # model.save(outpath, true)
 
     spaces = model.getSpaces
     #spaces = spaces.get
@@ -106,9 +108,10 @@ class Issue_2560_Test < Minitest::Test
     OpenStudio::Model::intersectSurfaces(spacesX)
     OpenStudio::Model::matchSurfaces(spacesX)
 
-    outpath = File.join(File.expand_path('../../.') + "/build/resources/model/","2560_after.osm")
-    model.save(outpath, true)
+    # outpath = File.join(outdir, "2560_after.osm")
+    # model.save(outpath, true)
 
+    # TODO: this block achieves nothing, is it for debugging purposes?
     spacesX.each do |space|
       spaceName = space.name.get
       puts spaceName
@@ -124,9 +127,9 @@ class Issue_2560_Test < Minitest::Test
 
     # Verify that the floor surfaces on space 1 are matched
     space1Surfaces = sp1.surfaces;
-    index_surface1 = nil
-    index_surface27 = nil
-    index_surface28 = nil
+    # index_surface1 = nil
+    # index_surface27 = nil
+    # index_surface28 = nil
     _space1Surface_1 = nil
     _space1Surface_27 = nil
     _space1Surface_28 = nil
--- a/src/model/Building.cpp
+++ b/src/model/Building.cpp
@@ -937,21 +937,24 @@ namespace model {
 
       Point3dVectorVector polygons;
 
-      for (auto space : model().getConcreteModelObjects<Space>()) {
+      for (const auto& space : model().getConcreteModelObjects<Space>()) {
         Transformation spaceTransformation = space.transformation();
-        for (auto surface : space.surfaces()) {
+        for (const auto& surface : space.surfaces()) {
           if (surface.outsideBoundaryCondition() == "Ground" || surface.surfaceType() == "Floor") {
             Point3dVector points = spaceTransformation * surface.vertices();
-            if (points.size() > 0 && points[0].z() == 0) polygons.push_back(points);
+            if (!points.empty() && points[0].z() == 0.0) {
+              polygons.push_back(points);
+            }
           }
         }
       }
 
       auto result2 = openstudio::joinAllPolygons(polygons, 0.01);
-      if (result2.size() == 1)
+      if (result2.size() == 1) {
         return result2[0].perimeter();
-      else
+      } else {
         return 0.0;
+      }
     }
 
     std::vector<std::vector<Point3d>> Building_Impl::generateSkylightPattern(double skylightToProjectedFloorRatio, double desiredWidth,
--- a/src/model/Surface.cpp
+++ b/src/model/Surface.cpp
@@ -1034,8 +1034,8 @@ namespace model {
     }
 
     boost::optional<SurfaceIntersection> Surface_Impl::computeIntersection(Surface& otherSurface) {
-      double tol = 0.01;       // 1 cm tolerance
-      double areaTol = 0.001;  // 10mm2 tolerance
+      double tol = 0.01;       //  1 cm tolerance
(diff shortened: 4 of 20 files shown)


jmarrec:
OSX reports clean: 100% tests passed, 0 tests failed out of 3309

https://ci.commercialbuildings.dev/blue/organizations/jenkins/openstudio-incremental-osx/detail/PR-4275/1/pipeline/48

src/model/test/Space_GTest.cpp
```diff
@@ -2029,25 +2030,21 @@ TEST_F(ModelFixture, Space_intersectSurfaces_degenerate3) {
...
-/// <summary>
-/// Illustrates a fix for surface intersection getting stuck in a loop
-/// First of all we need to remove surfaces that overlap within the same space
-/// Second of all we use a different removeSpikes method that shrinks and expands the polygon
-/// </summary>
-/// <param name=""></param>
-/// <param name=""></param>
+// Illustrates a fix for surface intersection getting stuck in a loop
+// First of all we need to remove surfaces that overlap within the same space
+// Second of all we use a different removeSpikes method that shrinks and expands the polygon
```
jmarrec, line 2039: Now all tests are also enabled on Unix

src/model/test/Space_GTest.cpp
```diff
@@ -2029,25 +2030,21 @@ TEST_F(ModelFixture, Space_intersectSurfaces_degenerate3) {
...
 TEST_F(ModelFixture, RemoveSpikesAndOverlaps_TZ46_TZ47) {
   Model model;
   openstudio::path path = resourcesPath() / toPath("/model/RemoveSpikesAndOverlaps_TZ46_TZ47");
 
+#ifdef WIN32
 #  pragma region SPACE 1(TZ46 - 81)
+#endif
+
```
jmarrec, line 2047: This is necessary because GCC doesn't udnerstand `#pragma region` which is a MSVC thing (clang does support it though)

jmarrec: Honestly I'd probably just remove that altogether, but whatever

src/model/test/Space_GTest.cpp
```diff
@@ -2235,7 +2235,7 @@ TEST_F(ModelFixture, RemoveSpikesAndOverlaps_TZ46_TZ47) {
     nSurfaces += space.surfaces().size();
   }
 
-  ASSERT_EQ(nSurfaces, 16);
+  EXPECT_EQ(nSurfaces, 16);
```
jmarrec, line 2238: From now on, lots of changing ASSERT_XXX to EXPECT_XXX. ASSERT is helpful when it would throw after it. eg `ASSERT_EQ(1, vector.size(); vector[0]....`. Otherwise EXPECT_ is better because it lets your test continue to report additional fail/success.

src/model/test/Space_GTest.cpp
```diff
@@ -2720,52 +2543,224 @@ TEST_F(ModelFixture, ExposedPerimeter) {
...
+
+  // openstudio::path outPath = resourcesPath() / toPath("model/15023_Model12_after.osm");
+  // model->save(outPath, true);
+}
+
+// Skipping this one because this is outside of the current scope.
+// To correct this intersection and matching needs to allow holes and then
+// decompose polygons with holes as the last step of the process. Often as
+// is shown here the holes are removed by the intersection process anyway
+TEST_F(ModelFixture, DISABLED_ShatteredModel_Existing_3424) {
```
jmarrec, line 2632: Instead of wrapping this one in a macro, I'm marking it DISABLED_ which will make it not run by default. You could call `Products/openstudio_model_tests --gtest_also_run_disabled_tests` to check the status. This one fails.

src/model/test/Space_GTest.cpp
```diff
@@ -2720,52 +2543,224 @@ TEST_F(ModelFixture, ExposedPerimeter) {
...
+  model::OptionalModel model = translator.loadModel(modelPath);
+  EXPECT_TRUE(model);
+
+  SpaceVector spaces = model->getConcreteModelObjects<Space>();
+  intersectSurfaces(spaces);
+  matchSurfaces(spaces);
+
+  // openstudio::path outPath = resourcesPath() / toPath("model/15023_Model12_after.osm");
+  // model->save(outPath, true);
+}
```
jmarrec, line 2626: I'm not sure what to do with these two. They pass, but they are slow. I know you reported one took 16minutes originally, but now it still takes 20secs which is too long for my taste.

Is there anything in these two files that **isn't** already covered in a model GTest? If so, can we do that? Then we can just remove the two OSM and tests here

@ggartside @tijcolem @kbenne thoughts?

Jenkinsfile_develop_osx
```diff
@@ -1,6 +1,6 @@
 //Jenkins pipelines are stored in shared libaries. Please see: https://github.com/NREL/cbci_jenkins_libs
 
-@Library('cbci_shared_libs') _
+@Library('cbci_shared_libs@geometry_improvements') _
```
jmarrec, line 3: This is just me saying to jenkins that it should use the branch I created where I re-enabled the degenerate intersection tests (which now pass)

ruby/test/Issue_2560_Test.rb
```diff
@@ -106,9 +108,10 @@ def test_Issue_2560
     OpenStudio::Model::intersectSurfaces(spacesX)
     OpenStudio::Model::matchSurfaces(spacesX)
 
-    outpath = File.join(File.expand_path('../../.') + "/build/resources/model/","2560_after.osm")
-    model.save(outpath, true)
+    # outpath = File.join(outdir, "2560_after.osm")
+    # model.save(outpath, true)
 
+    # TODO: this block achieves nothing, is it for debugging purposes?
```
jmarrec, line 114: https://github.com/NREL/OpenStudio/pull/4221#discussion_r608450589

src/model/Building.cpp
```diff
@@ -937,21 +937,24 @@ namespace model {
...
+      for (const auto& space : model().getConcreteModelObjects<Space>()) {
         Transformation spaceTransformation = space.transformation();
-        for (auto surface : space.surfaces()) {
+        for (const auto& surface : space.surfaces()) {
           if (surface.outsideBoundaryCondition() == "Ground" || surface.surfaceType() == "Floor") {
             Point3dVector points = spaceTransformation * surface.vertices();
-            if (points.size() > 0 && points[0].z() == 0) polygons.push_back(points);
+            if (!points.empty() && points[0].z() == 0.0) {
+              polygons.push_back(points);
+            }
```
jmarrec, line 947: Some changes like const-refs, double == double comparison, and breaking one-line ifs

src/model/Surface.cpp
```diff
@@ -1034,17 +1034,12 @@ namespace model {
     }
 
     boost::optional<SurfaceIntersection> Surface_Impl::computeIntersection(Surface& otherSurface) {
-      double tol = 0.01;       // 1 cm tolerance
-      double areaTol = 0.001;  // 10mm2 tolerance
+      double tol = 0.01;       //  1 cm tolerance
+      double areaTol = 0.001;  // 10 cm2 tolerance
```
jmarrec, line 1038: Units

src/model/Surface.cpp
```diff
@@ -1110,9 +1105,9 @@ namespace model {
         return boost::none;
       }
 
-      // DA - Change tolerance. Current tolerance is 0.0001 which is 1mm2 which is unrealistic
-      // tolerance could be fixed, say 10,,2 or as a proportion of the area of the polygon. 4mm2
-      // on  apolygon of area 570m2 is a tiny fraction
+      // DA - Change tolerance. Current tolerance is 0.0001 which is 1cm2 which is unrealistic
+      // tolerance could be fixed, say 10cm2 or as a proportion of the area of the polygon. 4cm2
+      // on a polygon of area 570m2 is a tiny fraction
```
jmarrec, line 1110: Units

src/utilities/geometry/Polygon3d.cpp
```diff
@@ -204,7 +203,9 @@ std::vector<Point3dVector> Polygon3d::overlap(const Point3dVector& line) const {
     Point3d p2 = ti * m_outerPath[(i + 1) % m_outerPath.size()];
 
     // The two line segments must be parallel and coincident within tolerance
-    if (abs(p1.y()) > 0.01 || abs(p2.y()) > 0.01) continue;
+    if (std::abs(p1.y()) > 0.01 || std::abs(p2.y()) > 0.01) {
```
jmarrec, line 206: std::abs(float) instead of abs(int)

ruby/test/Issue_2560_Test.rb
```diff
@@ -93,8 +94,9 @@ def test_Issue_2560
     sp4.setThermalZone(tz4)
 
     # Model before intersection
-    outpath = File.join(File.expand_path('../../.') + "/build/resources/model/","2560_before.osm")
-    model.save(outpath, true)
+    # outdir = File.absolute_path(File.join(OpenStudio::getEnergyPlusDirectory.to_s, "../resources/model"))
```
kbenne, line 97: Thank you @jmarrec

kbenne (approved):
Looks great to me.

PR #4292, opened 2021-04-16
Coordinate default values for outdoor air method

### LINKED ISSUE #4131: Misaligned sizing fields for VAV system sizing
Issue overview
--------------
Currently, the default sizing behavior for systems in OpenStudio (also with the App) is:
'ZoneSum' in the 'System Outdoor Air Method' field of the 'Sizing:System` object
'VentilationRateProcedure' in the 'System Outdoor Air Method' field of the 'Controller:MechanicalVentilation' object, which is not exposed in the app.

These two values should be aligned.

## Current Behavior
For system sizing in cold climate zones (e.g. 6A), this issue results in undersizing of the main heating coil and supply air temperature during the design day run being ~1
[heating_sizing_debugging.xlsx](https://github.com/NREL/OpenStudio/files/5431841/heating_sizing_debugging.xlsx)
5F instead of the intended 55F.  The coil sizing details assume some amount of mixed air, even if the '100% outdoor air in heating' field is set to Yes in the Sizing:System object.  I'm not exactly clear on the 

## Expected Behavior
There should be warning or way to default the System Outdoor Air Method to use either the field in the Controller:MechanicalVentilation object, or the one in the Sizing System object.  I don't see a reason why they should be different during design sizing.

## Steps to Reproduce
Here is an example base model (120 zone office, served by VAV, in climate zone 6A).
[heating_sizing_debugging.zip](https://github.com/NREL/OpenStudio/files/5431847/heating_sizing_debugging.zip)
I've included an output spreadsheet of coil sizing results.

## Possible Solution
We need to have some discussion here on the best way to resolve this - either having one field default to the value in the other field, adding a warning message, or at least making sure this is available in the app.

### Environment

 - Platform (Operating system, version):  Windows
 - Version of OpenStudio (if using an intermediate build, include SHA): OS > 3.0

### Context
This is causing an issue for any models using default HVAC constructors in openstudio-standards.  I can set both fields in the API, so there is no pressing need for a new release, but several other users may run into this sizing issue.

Thanks to @asparke2 and @kbenne for helping narrow down the issue.

kbenne:
This aligns the default value of the oa method between
ControllerMechanicalVentilation and SizingSystem.

close #4131

### DIFF
--- a/src/model/ControllerMechanicalVentilation.cpp
+++ b/src/model/ControllerMechanicalVentilation.cpp
@@ -130,7 +130,7 @@ namespace model {
         result = "VentilationRateProcedure";
       }
 
-      return value.get();
+      return result.get();
     }
 
     bool ControllerMechanicalVentilation_Impl::isSystemOutdoorAirMethodDefaulted() const {
--- a/src/model/ControllerMechanicalVentilation.cpp
+++ b/src/model/ControllerMechanicalVentilation.cpp
@@ -29,10 +29,16 @@
 
 #include "ControllerMechanicalVentilation.hpp"
 #include "ControllerMechanicalVentilation_Impl.hpp"
+#include "AirLoopHVACOutdoorAirSystem.hpp"
+#include "AirLoopHVACOutdoorAirSystem_Impl.hpp"
+#include "AirLoopHVAC.hpp"
+#include "AirLoopHVAC_Impl.hpp"
 #include "ControllerOutdoorAir.hpp"
 #include "ControllerOutdoorAir_Impl.hpp"
 #include "Schedule.hpp"
 #include "Schedule_Impl.hpp"
+#include "SizingSystem.hpp"
+#include "SizingSystem_Impl.hpp"
 #include "Model.hpp"
 #include "Model_Impl.hpp"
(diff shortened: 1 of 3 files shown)


src/model/ControllerMechanicalVentilation.cpp
```diff
@@ -103,9 +109,28 @@ namespace model {
     }
 
     std::string ControllerMechanicalVentilation_Impl::systemOutdoorAirMethod() const {
-      boost::optional<std::string> value = getString(OS_Controller_MechanicalVentilationFields::SystemOutdoorAirMethod, true);
-      OS_ASSERT(value);
-      return value.get();
+      boost::optional<std::string> result;
+      const auto value = getString(OS_Controller_MechanicalVentilationFields::SystemOutdoorAirMethod, true);
+      if (value && !value->empty()) {
+        result = value;
```
jmarrec, line 115: (I would have used `getString(XXX, true, true)` (signature is `boost::optional<std::string> getString(unsigned index, bool returnDefault = false, bool returnUninitializedEmpty = false) const;`) but this will do as well)

jmarrec:
This should be added to the (pre) release notes probably.

PR #4307, opened 2021-04-27
Fix #4289 - Standardize Curve IDD references like E+ does (Uni/Bi...variateFunctions) and remove hard check in model code

### LINKED ISSUE #4289: OpenStudio is enforcing some curve type limits that are too restrictive
This bug report originated from @TiejunWu and I am pasting in full because I think it very clearly explains the issue and there is no use reformulating an already well defined issue. This is a high priority. 

> I met some problem when I tried to use look up table for DX Coil performance in OpenStudio. Several of the functions like “setTotalCoolingCapacityFunctionOfTemperatureCurve”, returned a false when I tried to use a 2-D look up table as the parameter.
>  
> Then I looked into the OpenStudio source code here:
>  
> https://github.com/NREL/OpenStudio/blob/f2930bbeb1031b731747c3393343f741b0fc11fa/src/model/CoilCoolingDXSingleSpeed.cpp#L234-L247
>  
> Looks like in this function there is a check on the parameter “curve”, if will just return false if the parameter cannot be casted to be a Biquadratic curve.
>  
> However, such a check does not exist for the function setEnergyInputRatioFunctionOfTemperatureCurve. And I found that when I set a 2D lookup table to this function, it returned a true.
>  
> https://github.com/NREL/OpenStudio/blob/f2930bbeb1031b731747c3393343f741b0fc11fa/src/model/CoilCoolingDXSingleSpeed.cpp#L285-L298
>  
> When I model a DX coil directly in EnergyPlus, there is no such restrictions, I can use  2D lookup tables for Capacity as a function of temperature curve, Energy Input Ratio  Function of Temperature Curve and use 1D lookup table for Capacity Function of Flow Fraction Curve and Energy Input Ratio Function of Flow Fraction Curve, etc.
>  
> Can these checks be modified or removed for future versions of OpenStudio SDK so that lookup tables can be used to specify DX coil performance? Potentially we have to do so for all the DX coils.

jmarrec:
Pull request overview
---------------------

- Fix #4289 
- Standardize Curve IDD references like E+ does (Uni/Bi...variateFunctions) and remove hard check in model code

Technically I suppose we could go above and create a ModelObject::setCurve that would be a wrapper around setPointer that would also check that if a TableMultiLookupVariable is supplied it has the correct number of independent variables based on IDD choices, but E+ will have an informative message so the performance penalty for this extra check (and effort to modify all setters that take a Curve as a param) isn't worth it IMHO.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/developer/python/Check_Correct_IDD_curve_refs_between_ep_and_os.ipynb
+++ b/developer/python/Check_Correct_IDD_curve_refs_between_ep_and_os.ipynb
@@ -1,0 +1,212 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "ccdd4142",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import openstudio\n",
+    "# pip install fuzzywuzzy[speedup]\n",
+    "from fuzzywuzzy import process"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d804b2de",
+   "metadata": {},
+   "source": [
+    "I'm comparing the IDD references for objects referencing curves between E+ and OS IDDs.\n",
+    "The fuzzywuzzy python module is used to match the OS field name to the EP IDD one (because we may be outdated or deviated on purpose, like removing numbers when making extensible or removing 'Name' at end)\n",
+    "\n",
+    "See https://github.com/NREL/OpenStudio/issues/4289"
+   ]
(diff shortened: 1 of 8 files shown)


jmarrec:
@kbenne No problem running the regression tests

kbenne (approved):

PR #4333, opened 2021-05-27
test_ymd_constructor windows ruby test

tijcolem:
A particular ruby unit test `test_ymd_constructor` fails on windows due to select constructor args causing segfaults vs the expected RuntimeErrror. 

e.g.
```
assert_raises(RuntimeError){OpenStudio::Date.new("Jan".to_MonthOfYear,32,2008)} => segfaults
assert_raises(RuntimeError){OpenStudio::Date.new("Dec".to_MonthOfYear,0,2008)} =>  throws RuntimeErrror as expected
```

Interesting, `OpenStudio::Date.new("Jan".to_MonthOfYear,32,2008)` when ran outside the context of Minitest, throws RuntimeErrror which is expected so it seems this issue is related to Minitest and Ruby 2.7.2 on windows.  For the sake of excluding this test on windows, I changed some of the tests a bit which IMO has good enough coverage.  

Along the unit test failures. With this fix we are down to two unit test failures that were excluded from runs. 

```
CLITest-test_encodings-encoding
CLITest-test_encodings-encoding
```

I looked into these.  Since OpenStudio >= 3.0.0 the returned string encoded value is ASCII-8BIT vs the expected UTF-8. 
The byte sequence is valid UTF-8 though and can be converted by doing *string*.encode("UTF-8") 

For example
```
test_string = "Some_test_string"
test_string = test_string.encode("UTF-8")  # => UTF-8

space = OpenStudio::Model::Space.new(model)

space.setName(test_string)

name_string = space.nameString # => ASCII-8BIT 
name_string = name_string.encode("UTF-8") # => UTF-8
```

It's not clear to me if this is a ruby binding issue that changed with ruby version upgrades or something in OpenStudio encoding. I know ASCII-8BIT is essentially a raw bytes with Ruby.  I suppose the test can be changed to covert ASCII-8BIT to UTF-8 but maybe there is more going on to this that needs to be looked at.

### DIFF
--- a/Jenkinsfile_develop_osx
+++ b/Jenkinsfile_develop_osx
@@ -1,6 +1,6 @@
 //Jenkins pipelines are stored in shared libaries. Please see: https://github.com/NREL/cbci_jenkins_libs
 
-@Library('cbci_shared_libs') _
+@Library('cbci_shared_libs@remove_excluded_tests') _
 
 // Build for PR to develop branch only.
 if ((env.CHANGE_ID) && (env.CHANGE_TARGET) ) {
--- a/Jenkinsfile_develop_osx
+++ b/Jenkinsfile_develop_osx
@@ -1,6 +1,6 @@
 //Jenkins pipelines are stored in shared libaries. Please see: https://github.com/NREL/cbci_jenkins_libs
 
-@Library('cbci_shared_libs@remove_excluded_tests') _
+@Library('cbci_shared_libs') _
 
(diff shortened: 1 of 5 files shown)


tijcolem:
Perhaps the reason for the ASCII-8BIT is due to the default assignment here to UTF-8 (https://github.com/NREL/OpenStudio/blob/develop/src/cli/main.cpp#L556) but not the init for utf_8 is commented out.

PR #4342, opened 2021-06-02
Fix #4327 - FanSystemModel electricPowerFraction should be optional

### LINKED ISSUE #4327: FanSystemModel electricPowerFraction should be optional
Issue overview
--------------
No matter which `addSpeed` method is used for the `FanSystemModel` object, the SDK requires the `electricPowerFraction` to be entered. However, EnergyPlus allows two different behaviors:
- Don't specify the `electricPowerFraction` for each speed, in which case the `electricPowerFunctionofFlowFractionCurve` is used
- Specify the `electricPowerFraction` for each speed, in which case the `electricPowerFunctionofFlowFractionCurve` is not used

This behavior is described [here](https://bigladdersoftware.com/epx/docs/9-5/input-output-reference/group-fans.html#field-speed-power-fraction-fansysmodel) in the EnergyPlus documentation.

I'd like to do the former, but OpenStudio forces the latter. Not sure if this can be accommodated without breaking the API...

jmarrec:
Pull request overview
---------------------

Fix #4327 - FanSystemModel electricPowerFraction should be optional

The API changes are only in the helper class FanSystemModelSpeed that I create just for the extensible groups, so I think it's fine. This prevents a feature users want (at least @shorowit) so this small api change is warranted and worth it IMHO. @kbenne please advise.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/model/FanSystemModel.hpp
+++ b/src/model/FanSystemModel.hpp
@@ -52,18 +52,18 @@ namespace model {
   class MODEL_API FanSystemModelSpeed
   {
    public:
-    /* Only accepts ModelObjects that are of type Surface, Subsurface or InternalMass, will throw otherwise */
+    FanSystemModelSpeed(double flowFraction);
+
     FanSystemModelSpeed(double flowFraction, double electricPowerFraction);
 
     double flowFraction() const;
-    double electricPowerFraction() const;
+    boost::optional<double> electricPowerFraction() const;
 
-    // this operator is to support sorting of TableMultiVariableLookupPoint in the order required by EnergyPlus Table:Lookup object
     bool operator<(const FanSystemModelSpeed& other) const;
 
    private:
     double m_flowFraction;
-    double m_electricPowerFraction;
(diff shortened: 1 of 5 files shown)


src/model/FanSystemModel.hpp
```diff
@@ -52,18 +52,18 @@ namespace model {
...
   {
    public:
-    /* Only accepts ModelObjects that are of type Surface, Subsurface or InternalMass, will throw otherwise */
+    FanSystemModelSpeed(double flowFraction);
+
     FanSystemModelSpeed(double flowFraction, double electricPowerFraction);
 
     double flowFraction() const;
-    double electricPowerFraction() const;
+    boost::optional<double> electricPowerFraction() const;
```
jmarrec, line 60: API change

jmarrec: @kbenne

PR #4371, opened 2021-07-02
Github Actions to publish python bindings to pypi when it's a release and testing worked

jmarrec:
Pull request overview
---------------------

Github Actions to publish python bindings to pypi when it's a release and testing worked

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


.github/workflows/python_bindings.yml
```diff
@@ -102,6 +102,13 @@ jobs:
           echo "$MSVC_DIR\VC\Auxiliary\Build" >> $GITHUB_PATH
         fi;
 
+        if [[ "$GITHUB_REF" == *"refs/tags"* ]]; then
+          PYTHON_PIP_REPOSITORY=pypi
+        else
+          PYTHON_PIP_REPOSITORY=testpypi
+        fi;
+        echo "PYTHON_PIP_REPOSITORY=$PYTHON_PIP_REPOSITORY" >> $GITHUB_ENV
```
jmarrec, line 110: So I'm not 100% sure this is going to work in all cases... There is one path where it would fail:
* Our project is at 3.3.0.
* We compute the next available version by looking at pypi, where it doesn't exist yet. so 3.3.0 is selected.
* We upload to testpyi with this version, and it works

* In the event the push to testpypi works BUT the bindings don't during testing, it is NOT uplaod to pypi
* We rerun the workflow: we still compute 3.3.0 because we looked at pypi which still doesn't have it
* Now the push to testpypi won't work since it already exists

jmarrec: Perhaps the easiest way would be to build the python package twice with the testpypi calculated version and the pypi calculated one...

jmarrec: or, in the event we do a release, don't even bother trying to upload to testpypi, just upload as an artifact and download that directly in the testing stage

.github/workflows/python_bindings.yml
```diff
@@ -178,7 +185,7 @@ jobs:
         set -x
         echo "Launching a build with BUILD_PYTHON_BINDINGS=ON and BUILD_PYTHON_PIP_PACKAGE=ON, turning off as many other things as possible to speed it up"
         cmake -DCMAKE_BUILD_TYPE=Release \
-              -DBUILD_PYTHON_BINDINGS=ON -DBUILD_PYTHON_PIP_PACKAGE=ON -DPYTHON_VERSION=${{ matrix.python-version }} \
+              -DBUILD_PYTHON_BINDINGS=ON -DBUILD_PYTHON_PIP_PACKAGE=ON -DPYTHON_PIP_REPOSITORY=${{ env.PYTHON_PIP_REPOSITORY }} -DPYTHON_VERSION=${{ matrix.python-version }} \
```
jmarrec, line 188: Passing PYTHON_PIP_REPOSITORY conditions how we look for the next available version

.github/workflows/python_bindings.yml
```diff
@@ -190,9 +197,8 @@ jobs:
         set -x
         cmake --build . --target package -j 2
 
+    # We always upload to testpypi
```
jmarrec, line 200: Should we?

.github/workflows/python_bindings.yml
```diff
@@ -206,9 +212,21 @@ jobs:
...
+      working-directory: ./build/Products/python_package/
+      shell: bash
+      run: |
+        tar -cvzf openstudio-${{ matrix.name }}.tar.gz dist/
+
+    - name: Upload .whl to artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: openstudio-${{ matrix.name }}  # something like openstudio-Windows_py37
+        path: ./build/Products/python_package/openstudio-${{ matrix.name }}.tar.gz
```
jmarrec, line 225: Use artifact upload to pass information between jobs. The build uploads, the test job downloads

.github/workflows/python_bindings.yml
```diff
@@ -251,13 +269,84 @@ jobs:
...
+      env:
+        TWINE_USERNAME: __token__
+        TWINE_REPOSITORY: pypi
+        # Go to repo on github > Settings > Secrets and add it
+        TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
+      shell: bash
+      run: |
+        tar -xzvf "openstudio-${{ matrix.name }}.tar.gz";
+        ls -R
+        python -m twine upload --skip-existing dist/*
```
jmarrec, line 304: That would allow **partial** upload to pypi.

.github/workflows/python_bindings.yml
```diff
@@ -251,13 +269,84 @@ jobs:
...
+
+  upload_python_bindings_to_pypi:
+    # if: contains(github.ref, 'refs/tags')
+    needs: test_python_bindings
+    name: Upload to PyPi
+    runs-on: ubuntu-20.04
+    steps:
```
jmarrec, line 311: This job is run if the entire test job works. And we try to upload everything in one go. (Obviously in this case since we already did above, the upload would be skipped...)

.github/workflows/python_bindings.yml
```diff
@@ -251,13 +269,84 @@ jobs:
...
+        for f in **/*.tar.gz; do
+          mv $f .
+        done;
+
+        for f in *.tar.gz; do
+          tar -xzvf "$f";
+        done;
+
+        ls -R
+        python -m twine upload --skip-existing --repository pypi dist/*
```
jmarrec, line 352: Download all artifacts, they all end up in the same dist/ subfolder that we upload.

python/test/test_find_pypi_tag.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        new_v = compute_appropriate_version(
+            current_v=version.parse(v),
+            releases=[version.parse(v) for v in releases],
+            current=False
+        )
+        self.assertEqual(version.parse(expected_v), new_v)
+
+    def test_when_not_exist_with_prereleasetag(self):
+        self.compare(releases=["3.2.0"], v="3.2.1-alpha", expected_v="3.2.1a0")
```
jmarrec, line 23: I had to add an actual unit test to do TDD for the find_pypi_tag.py file, I wasn't managing to do it without it. This shows pretty well how it functions.

.github/workflows/python_bindings.yml
tijcolem: Hmm, good point. What about upload the bindings to testpypi using the normal sequence by appending the test version(e.g. 3.2.2a4, 3.2.2a5), test those packages, and then add another step to change the name of the bindings version to the official release  (3.2.2) and upload those?  Then upload those to pypi. 

Another option would be if the testpypi version doesn't pass the tests we yank the version?  I don't love that idea but it's also test server so maybe this okay to do this. Not sure if we can do this with cli calls.

python/test/test_find_pypi_tag.py
```diff
@@ -0,0 +1,66 @@
...
+        # Here I actually expect a post release to be appended
+        self.compare(releases=["3.2.0", "3.2.1a1", "3.2.1rc1"],
+                     v='3.2.1-rc1', expected_v="3.2.1rc1.post0")
+
+        self.compare(releases=["3.2.0", "3.2.1a1", "3.2.1rc1"],
+                     v='3.2.1-rc2', expected_v="3.2.1rc2")
+
+        # Now we release official
+        self.compare(releases=["3.2.0", "3.2.1a1", "3.2.1rc1"],
+                     v='3.2.1', expected_v="3.2.1")
```
jmarrec, line 61: Especially this

.github/workflows/python_bindings.yml
tijcolem: I think so, for pushes to develop branch.

.github/workflows/python_bindings.yml
tijcolem: Pretty clever use of artifacts here. 👍

tijcolem:
This is all great. I'm going to add another if condition where we only only upload to testpypi if all builds succeed vs the per build we have configured now.  I've noticed that the Ubuntu bindings can fail occasionally, and while it's easy to replay it, if this is a tagged release (e.g. 3.2.2), the previous successful uploads to testpypi would fail and consequently uploads to pypi would never run.

tijcolem (approved):
I believe all is now ready to be merged.

PR #4424, opened 2021-08-30
Issue 4361

### LINKED ISSUE #4361: OpenStudio API-reported surface "netArea" (when dealing with Frame & Divider objects)
Issue overview
--------------
It seems the OpenStudio API is reporting erroneous [netArea](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.2.0-doc/model/html/classopenstudio_1_1model_1_1_planar_surface.html#ae757533696d34e5bc50bd396c9e1eab0) values for opaque (planar) surfaces with subsurfaces holding _Frame & Divider_ objects (which in turn affects OpenStudio API-reported WWR). For models with many windows (especially with narrow aspect ratios), OpenStudio API-reported WWR may be off by 5% to 10%.

## Current Behavior
As described in the _EnergyPlus Input Output Reference Manual_, subsurface vertices "are those of the glazed part of the window, not the frame" when dealing with [Frame & Divider](https://bigladdersoftware.com/epx/docs/9-5/input-output-reference/group-thermal-zone-description-geometry.html#windowpropertyframeanddivider) objects. "EnergyPlus automatically subtracts the area of the frame - determined from the glazing dimensions and the frame width - from the area of the wall containing the window". Although OpenStudio's forward translation to EnergyPlus remains valid (EnergyPlus correctly reports surface/subsurface net vs gross areas, valid WWR, etc.) when dealing with _Frame & Divider_ objects, the OpenStudio API opaque surface _netArea_ method reports values as if the subsurface vertices described the _rough opening_ (and not the glazed portion).

## Expected Behavior
When dealing with _Frame & Divider_ objects, OpenStudio API should report _pre-simulation_ exactly what EnergyPlus reports _post-simulation_ e.g., opaque surface _net area_, WWR, etc.

## Steps to Reproduce
The easiest is to go through the tests [here](https://github.com/rd2/tbd/blob/d758edcc051067e845a413ec3d71ab5b6b830c04/spec/tbd_tests_spec.rb#L4414), where both issue and proposed [workaround](https://github.com/rd2/tbd/blob/d758edcc051067e845a413ec3d71ab5b6b830c04/lib/framedivider.rb#L124) are tested.

## Possible Solution
See aforementioned [workaround](https://github.com/rd2/tbd/blob/d758edcc051067e845a413ec3d71ab5b6b830c04/lib/framedivider.rb#L124) (this would require some cleaning up and porting it to C++). This would in turn likely require adjusting API WWR [calculations](https://github.com/NREL/OpenStudio/blob/70a5549c439eda69d6c514a7275254f71f7e3d2b/src/model/Surface.cpp#L1446).

### Environment

Tests are initially written for OpenStudio v2.9.1 (on a Mac), although PRs to TBD systematically go through automated 3.x Docker'ered [testing](https://github.com/rd2/tbd/actions) as well. Same behaviour.

### Context
This was recently observed when trying to accommodate subsurfaces (with Frame & Divider objects) for the _Thermal Bridging & Derating_ ([TBD](https://github.com/rd2/tbd)) OpenStudio Measure. Window (door, skylight) _rough opening_ perimeters generally do represent notable thermal bridges (mainly due to a greater concentration of structural supports e.g., lintels). So we knew we had to generate _substitute_ subsurface vertices for our _rough openings_ in our [Topolys](https://github.com/automaticmagic/topolys) back engine. The workaround not only generates the right _rough opening_ vertices (and hence the right edge lengths), it also validates whether the _rough opening_ fits within the parent surface and avoids any overlaps with its siblings (i.e., other surface's subsurfaces). The function also reports the correct _rough opening_ area (m2), which is crucial in TBD's thermal _derating_ step (based on parent surface _net area_). In addition, TBD will _soon_ (a month from this posting?) offer an _enhanced_ UA method (which will factor-in _major_ thermal bridging, as well as WWR), the calculation of which would be enabled without running an EnergyPlus simulation (i.e., _Apply Measures Now_). For these reasons, TBD no longer relies (for now) on OpenStudio API-reported _netArea_ (_derating_ workaround can be found [here](https://github.com/rd2/tbd/blob/d758edcc051067e845a413ec3d71ab5b6b830c04/lib/psi.rb#L1560)).

ggartside:
Pull request overview
---------------------

 - Fixes #4361

### Pull Request Author

 - [x] If a new object or method, added a test in NREL/OpenStudio-resources: Add Link
 - [x] If needed, added VersionTranslation rules for the objects (`src/osversion/VersionTranslator.cpp`)
 - [x] Verified that C# bindings built fine on Windows, partial classes used as needed, etc.
 - [x] If methods have been deprecated, update rest of code to use the new methods

**Labels:**

### Review Checklist

 - [x] Perform a Code Review on GitHub
 - [x] Code Style, strip trailing whitespace, etc.
 - [x] All related changes have been implemented: model changes, model tests, FT changes, FT tests, VersionTranslation, OS App
 - [x] Labeling is ok
 - [x] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [x] If feature, test running new feature, try creative ways to break it
 - [x] CI status: all green or justified


src/model/SubSurface.hpp
```diff
@@ -263,6 +263,12 @@ namespace model {
 
     boost::optional<AirflowNetworkSurface> airflowNetworkSurface() const;
 
+    /* Get the total area of the sub surface including frame and divider area */
+    double roughOpeningArea() const;
+
+    /* Get the rough opening vertices for the subsurface */
+    std::vector<Point3d> roughOpening() const;
```
macumber, line 270:
```suggestion
    std::vector<Point3d> roughOpeningVertices() const;
```

src/model/SubSurface.hpp
```diff
@@ -263,6 +263,12 @@ namespace model {
 
     boost::optional<AirflowNetworkSurface> airflowNetworkSurface() const;
 
+    /* Get the total area of the sub surface including frame and divider area */
```
macumber, line 266:
```suggestion
    /* Get the total area of the sub surface rough area which includes the frame */
```

The divider doesn't extend outside of the glazed area.

src/model/SubSurface.hpp
```diff
@@ -263,6 +263,12 @@ namespace model {
...
+    double roughOpeningArea() const;
+
+    /* Get the rough opening vertices for the subsurface */
```
macumber, line 269:
```suggestion
    /* Get the rough opening vertices for the sub surface including the frame */
```

src/energyplus/Test/SubSurface_GTest.cpp
```diff
@@ -127,3 +127,224 @@ TEST_F(EnergyPlusFixture, ForwardTranslator_SubSurface) {
...
+
+  WindowPropertyFrameAndDivider frame1(model);
+  frame1.setFrameWidth(0.030);
+  subSurface1.setWindowPropertyFrameAndDivider(frame1);
+
+  WindowPropertyFrameAndDivider frame2(model);
+  frame2.setFrameWidth(0.030);
+  subSurface2.setWindowPropertyFrameAndDivider(frame2);
+
+  windowWallRatio = surface.windowToWallRatio();
```
macumber, line 348: You'll want to check the standards to see if you should include frame area and/or divider area in the WWR calculations.

src/model/SubSurface.hpp
```diff
@@ -263,6 +263,12 @@ namespace model {
 
     boost::optional<AirflowNetworkSurface> airflowNetworkSurface() const;
```
macumber, line 264: Suggest adding `frameArea` and `dividerArea` methods

src/model/SubSurface.hpp
macumber: This could be implemented to point at a `grossAreaVertices` method on `PlanarSurface`, `grossAreaVertices` could be used to implement `PlanarSurface::overlaps(const PlanarSurface& other) const`

0001-Adds-exposedPerimeter-for-Surface.patch
```diff
@@ -0,0 +1,117 @@
+From d17df783191c744c6edb4dfdab45bfb074e2f7cc Mon Sep 17 00:00:00 2001
```
macumber, line 1: You don't mean to commit these patches right?

tijcolem:
@ggartside Can you remove those .patch files? I don't think you want to include these @macumber mentioned.

tijcolem:
@ggartside Yep, I just pushed up a commit to remove those.

src/energyplus/Test/SubSurface_GTest.cpp
```diff
@@ -127,3 +127,223 @@ TEST_F(EnergyPlusFixture, ForwardTranslator_SubSurface) {
...
+  // Create a surface and subsurface
+  // get the surface net and gross area & subsurface gross area
+  // surface gross area == surface net area + subsurface gross area
+  // That is before the subsurface is assigned a frame and divider
+
+  std::vector<Point3d> vertices;
+  vertices.push_back(Point3d(0, 0, 2));
+  vertices.push_back(Point3d(0, 0, 0));
+  vertices.push_back(Point3d(2, 0, 0));
+  vertices.push_back(Point3d(2, 0, 2));
```
jmarrec, line 149:
```suggestion
  std::vector<Point3d> vertices{
    {0, 0, 2},
    {0, 0, 0},
    {2, 0, 0},
    {2, 0, 2},
  };
```

src/energyplus/Test/SubSurface_GTest.cpp
```diff
@@ -127,3 +127,223 @@ TEST_F(EnergyPlusFixture, ForwardTranslator_SubSurface) {
...
+  Surface surface(vertices, model);
+  surface.setSpace(space);
+  surface.setSurfaceType("Wall");
+
+  vertices.clear();
+  vertices.push_back(Point3d(0.5, 0, 1.5));
+  vertices.push_back(Point3d(0.5, 0, 0.5));
+  vertices.push_back(Point3d(1.5, 0, 0.5));
+  vertices.push_back(Point3d(1.5, 0, 1.5));
```
jmarrec, line 158:
```suggestion
vertices.clear();
vertices = {
    {0.5, 0, 1.5},
    {0.5, 0, 0.5},
    {1.5, 0, 0.5},
    {1.5, 0, 1.5},
  };
```

src/energyplus/Test/SubSurface_GTest.cpp
```diff
@@ -127,3 +127,223 @@ TEST_F(EnergyPlusFixture, ForwardTranslator_SubSurface) {
...
+  // Create a surface and subsurface
+  // get the surface net and gross area & subsurface gross area
+  // surface gross area == surface net area + subsurface gross area
+  // That is before the subsurface is assigned a frame and divider
+
+  std::vector<Point3d> vertices;
+  vertices.push_back(Point3d(0, 0, 2));
+  vertices.push_back(Point3d(0, 0, 0));
+  vertices.push_back(Point3d(2, 0, 0));
+  vertices.push_back(Point3d(2, 0, 2));
```
jmarrec, line 205:
```suggestion
  std::vector<Point3d> vertices{
    {0, 0, 2},
    {0, 0, 0},
    {2, 0, 0},
    {2, 0, 2},
  };
```

src/energyplus/Test/SubSurface_GTest.cpp
```diff
@@ -127,3 +127,223 @@ TEST_F(EnergyPlusFixture, ForwardTranslator_SubSurface) {
...
+  Surface surface(vertices, model);
+  surface.setSpace(space);
+  surface.setSurfaceType("Wall");
+
+  vertices.clear();
+  vertices.push_back(Point3d(0.5, 0, 1.99));
+  vertices.push_back(Point3d(0.5, 0, 0.99));
+  vertices.push_back(Point3d(1.5, 0, 0.99));
+  vertices.push_back(Point3d(1.5, 0, 1.99));
```
jmarrec, line 214:
```suggestion
  vertices = {
    {0.5, 0, 1.99},
    {0.5, 0, 0.99},
    {1.5, 0, 0.99},
    {1.5, 0, 1.99},
  };
```

src/energyplus/Test/SubSurface_GTest.cpp
```diff
@@ -127,3 +127,223 @@ TEST_F(EnergyPlusFixture, ForwardTranslator_SubSurface) {
...
+  double subSurfaceTotalArea = subSurface.roughOpeningArea();
+
+  EXPECT_EQ(surfaceGrossArea, surfaceNetArea + subSurfaceGrossArea);
+  double windowWallRatio = surface.windowToWallRatio();
+  EXPECT_EQ(windowWallRatio, 0.25);
+
+  // Then assign a frame and divider to the subsurface
+  // then do the same thing with the areas
+  // The area should be unchanged because the subsurface is not enclosed
+  // by the parent serface after including the frame
```
jmarrec, line 232:
```suggestion
  // by the parent surface after including the frame
```

src/model/SubSurface.cpp
```diff
@@ -1595,6 +1596,37 @@ namespace model {
   void SubSurface::resetShadingControl() {
     removeAllShadingControls();
   }
+
+  std::vector<Point3d> SubSurface::roughOpeningVertices() const {
+    auto frameAndDivider = windowPropertyFrameAndDivider();
+    if (frameAndDivider) {
```
jmarrec, line 1602:
```suggestion
    if (auto frameAndDivider = windowPropertyFrameAndDivider()) {
```

src/model/SubSurface.cpp
```diff
@@ -1595,6 +1596,37 @@ namespace model {
...
+      double fw = frameAndDivider->frameWidth();
+      // Get a transform to change the points to x/y
+      Transformation faceTransform = Transformation::alignFace(this->vertices());
+      std::vector<Point3d> faceVertices = faceTransform.inverse() * this->vertices();
+      // Offset the points by the framewidth
+      boost::optional<std::vector<Point3d>> offset = openstudio::buffer(faceVertices, fw, 0.01);
```
jmarrec, line 1608: This offset here is unused unless it failed the first time. Is that normal?

src/model/SubSurface.cpp
```diff
@@ -1595,6 +1596,37 @@ namespace model {
...
+      if (!offset) {
+        // If offset failed it is because the points are in the wrong order
+        // If it fails again then something went awry with boost::buffer
+        faceVertices = openstudio::reverse(faceVertices);
+        offset = openstudio::buffer(faceVertices, fw, 0.01);
+        std::vector<Point3d> roughOpeningVertices = faceTransform * offset.get();
+        return roughOpeningVertices;
+      }
+    }
```
jmarrec, line 1617: Should it be this instead?

```suggestion
      boost::optional<std::vector<Point3d>> offset = openstudio::buffer(faceVertices, fw, 0.01);
      if (!offset) {
        // If offset failed it is because the points are in the wrong order
        // If it fails again then something went awry with boost::buffer
        faceVertices = openstudio::reverse(faceVertices);
        offset = openstudio::buffer(faceVertices, fw, 0.01);
      }
      std::vector<Point3d> roughOpeningVertices = faceTransform * offset.get();
      return roughOpeningVertices;
    }
```

src/model/SubSurface.cpp
```diff
@@ -1595,6 +1596,37 @@ namespace model {
...
+    return this->vertices();
+  }
+
+  double SubSurface::roughOpeningArea() const {
+
+    boost::optional<double> area = openstudio::getArea(roughOpeningVertices());
+    if (area)
+      return *area;
+    else
+      return grossArea();
```
jmarrec, line 1628:
```suggestion
    if (boost::optional<double> area = openstudio::getArea(roughOpeningVertices())) {
      return area.get();
    }
    
    return grossArea();
```

src/model/SubSurface.hpp
jmarrec: * [ ] TODO:  I think that makes sense indeed.

src/model/Surface.cpp
```diff
@@ -1440,16 +1440,101 @@ namespace model {
         return result;
       }
 
-      double windowArea = 0.0;
+      //double windowArea = 0.0;
+      //for (const SubSurface& subSurface : this->subSurfaces()) {
+      //  if (istringEqual(subSurface.subSurfaceType(), "FixedWindow") || istringEqual(subSurface.subSurfaceType(), "OperableWindow")) {
+      //    windowArea += subSurface.multiplier() * subSurface.roughOpeningArea();
+      //  }
+      //}
```
jmarrec, line 1448: * [ ] TODO: Delete?

src/model/Surface.cpp
```diff
@@ -1440,16 +1440,101 @@ namespace model {
...
+
+      return wwr;
+    }
+
+    // Calcuklates and returns the toital area of the subsurfaces
+    // If any subsrface extends outside the parent surface or overlaps an adjacent subsurface
+    // then that subsurface's vertices rather than rough opening vertioces are used to calculate the area
+    // NOTE: I'm starting to think a better way to do this would be to use polygon booleans
+    // 1 - Intersect the subsurface rough opening with the parent surface to get the area inside the parent surface
+    // 1 - Add all the subsurface rough openings together (so overlap areas dont count twice)
```
jmarrec, line 1461:
```suggestion
    // Calcuklates and returns the toital area of the subsurfaces
    // If any subusrface extends outside the parent surface or overlaps an adjacent subsurface
    // then that subsurface's vertices rather than rough opening vertices are used to calculate the area
    // NOTE: I'm starting to think a better way to do this would be to use polygon booleans
    // 1 - Intersect the subsurface rough opening with the parent surface to get the area inside the parent surface
    // 2 - Add all the subsurface rough openings together (so overlap areas don't count twice)
```

src/model/Surface.cpp
```diff
@@ -1440,16 +1440,101 @@ namespace model {
...
+          if (norm && norm->z() > 0) {
+            std::reverse(opening.begin(), opening.end());
+          }
+          roughOpenings[subSurface] = opening;
         }
       }
 
-      double wwr = windowArea / grossArea;
+      // Check each rough opening against the parent surface, if any vertices are outside
+      // the parent surfave then revert back to the original vertices
```
jmarrec, line 1495:
```suggestion
      // the parent surface then revert back to the original vertices
```

src/model/Surface.cpp
```diff
@@ -1440,16 +1440,101 @@ namespace model {
...
-      return wwr;
+        if (!openstudio::polygonInPolygon(vertices, parentVertices, tol)) {
+          // One or more vertices not within the parent surface so replace with the original vertices
+          auto vertices = parentToXY * subSurface.vertices();
+          auto norm = openstudio::getOutwardNormal(vertices);
+          if (norm && norm->z() > 0) {
+            std::reverse(vertices.begin(), vertices.end());
+          }
+          roughOpenings[subSurface] = vertices;
+          break;
```
jmarrec, line 1508: * [ ] TODO: Is the break intended? I don't understand everything that's going on, but don't you want to fix all of them?
Also, why not make this check directly in the loop above when populating roughOpenins?

src/model/Surface.cpp
```diff
@@ -1440,16 +1440,101 @@ namespace model {
...
+        }
+      }
+
+      // Check subsurfaces for overlaps
+      // (if I knew how I'd order themn left to right in x then we could test adjacent ones only)
+      for (const auto& opening1 : roughOpenings) {
+        for (const auto& opening2 : roughOpenings) {
+          if (opening1 == opening2) continue;
```
jmarrec, line 1516: I'd like a place to put a breakpoint. If clang-format isn't throwing on the lack of braces, something is wrong with our rules.

```suggestion
          if (opening1 == opening2) {
            continue;
          }
```

src/model/Surface.cpp
```diff
@@ -1440,16 +1440,101 @@ namespace model {
...
+
+          auto result = openstudio::intersect(opening1.second, opening2.second, tol);
+          if (result) {
```
jmarrec, line 1519:
```suggestion
          if (auto result = openstudio::intersect(opening1.second, opening2.second, tol)) {
```

src/model/Surface.cpp
```diff
@@ -1440,16 +1440,101 @@ namespace model {
...
+            roughOpenings[opening1.first] = vertices;
+          }
+        }
+      }
+
+      // Accumulate the areas
+      double area = 0;
+      for (const auto& roughOpening : roughOpenings) {
+        area += openstudio::getArea(roughOpening.second).value() * roughOpening.first.multiplier();
+      }
```
jmarrec, line 1535: Maybe it doesn't really improve the lisibility, but std::accumulate (`#include <numeric>`) is meant for that.

```suggestion
      // Accumulate the areas
      double area = std::accumulate(roughOpenings.cbegin(), roughOpenings.cend(), 0.0, [](double sum, auto& roughOpening) {
        return sum + openstudio::getArea(roughOpening.second).value() * roughOpening.first.multiplier();
      });
```

src/utilities/geometry/Intersection.cpp
```diff
@@ -505,6 +505,32 @@ std::vector<Point3d> removeSpikes(const std::vector<Point3d>& polygon, double to
...
+    if (abs(point.z()) > tol) {
+      return false;
+    }
+  }
+
+  for (const Point3d& point : points) {
+    boost::tuple<double, double> p = boostPointFromPoint3d(point, allPoints, tol);
+    BoostPoint boostPoint(p.get<0>(), p.get<1>());
+    double distance = boost::geometry::distance(boostPoint, *boostPolygon);
+    if (distance >= 0.0001) return false;
```
jmarrec, line 529:
```suggestion
    if (distance >= 0.0001) {
      return false;
    }
```

jmarrec (commented):
I applied my nitpicking changes and moved the tests to ModelFixture. I resolved all outdated review comments as a result. You have 4 reviews TODO left in the review above.

src/model/SubSurface.hpp
ggartside: I added these methods as suggested. Frame area is pretty obvious but for divider area I used the number of horizontal dividers, the divider width and the width of the sub surface and similarly for vertical with the height of the sub surface. Any better ideas?

macumber: I don't see those implementations in this PR?  Did they get moved?

jmarrec: Probably just a git push missing @ggartside

tijcolem:
@ggartside  Just following up on this PR.  Have you addressed the TODO @jmarrec requested?  Let's do another review and then we can hopefully get this merged in.

src/model/test/SubSurface_GTest.cpp
```diff
@@ -1280,3 +1281,244 @@ TEST_F(ModelFixture, SubSurface_isXXXAutocalculated_4399) {
...
+
+  WindowPropertyFrameAndDivider frame(model);
+  frame.setFrameWidth(0.030);
+
+  subSurface.setWindowPropertyFrameAndDivider(frame);
+
+  double frameArea = subSurface.frameArea();
+  EXPECT_NEAR(frameArea, 0.1236, 0.01);
+  double divArea = subSurface.dividerArea();
+  EXPECT_DOUBLE_EQ(divArea, 0);
```
jmarrec, line 1341: I'd like to see a test for this (that isn't 0)

ggartside: Ok

ggartside: Done

tijcolem:
Looks like the review comments have been addressed on this so I'm going to merge.

PR #4431, opened 2021-09-01
Fixed an issue in Surface splitSurfaceForSubSurfaces which caused the…

ggartside:
…w routine to return way too may overlapping surfaces

Added a unit test Issue_4374 which uses one surface from the secondary school ref building as test dats

Pull request overview
---------------------

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


tijcolem:
@ggartside This is great as this is an important fix. 

@DavidGoldwasser Would you be able to use one of the installers to verify this fixes the problem?

DavidGoldwasser:
Not working yet, see my updates on this issue in measure repo. 
https://github.com/NREL/openstudio-model-articulation-gem/issues/76#issuecomment-927178641

DavidGoldwasser:
@ggartside ok, so I think we are good with what was in this installer without the new proposed change to change where the split happens. This was misunderstanding on my part. I didn't realize it splits between all sub-surfaces instead of just doors. Here are my proposed fixes to the measure. One removes windows before splitting so it only splits at doors. The other is change in tolerance on rectangle check. https://github.com/NREL/openstudio-model-articulation-gem/commit/3769173f610d2de25374813be9f06be5a156b9a9

DavidGoldwasser:
@ggartside and @tijcolem minor correction to my last comment. I'm using ins taller from a few days ago, I'm wondering if the last commit from today should be removed? or maybe it won't create any problems. I'm concerned that when I do clear out windows before split that the new split won't be tight against the door as is expected.
https://github.com/NREL/OpenStudio/pull/4431/commits/7035a3edf9aede060b50039ffc68244c6cfdc473

tijcolem:
@ggartside Sounds like @DavidGoldwasser was able to successfully use your split surface fix ending on commit 
https://github.com/NREL/OpenStudio/pull/4431/commits/42398b6f5a84b0efc9103bf1a48b291bacb537d0 in this branch. Since this is working, it's probably best to revert to that commit.  You don't need to do force push as we can have a record of the other changes (maybe it'll still come into play).

ggartside:
@DavidGoldwasser If the intention is to only split the surface around doors then I could modify the original routine to skip window openings. 

@tijcolem I will revert back to the original method, 

One thing I don't understand yet is why this routine is necessary, maybe you can enlighten me?

DavidGoldwasser:
Task linked: [CU-1h1w527 Review Surface matching PR (03)](https://app.clickup.com/t/1h1w527)

tijcolem:
@ggartside Thanks for reverting!  This looks good and I'll drop in.

PR #4434, opened 2021-09-07
Addresses #4409, forward translate Space

### LINKED ISSUE #4409: E+ 9.6.0: Wrap Space and SpaceList
Enhancement Request
--------------

joseph-robertson:
Pull request overview
---------------------

 - Fixes #4409

Start stubbing this out so we can identify all that needs to be done.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


kbenne:
One observation is that this (naturally) has the 9.6 update within it. Of course the update is required for this feature, but it makes the review more challenging. Perhaps the PR could retarget the EnergyPlus 9.6 branch, or perhaps it just hangs out until the 9.6 update is merged, and then we review this. Thoughts?

jmarrec:
I think we need to revert https://github.com/NREL/OpenStudio/pull/4434/commits/8c89b742cd0cb9a6173df7ea41728d550a58c8fb while this is in development no?

We need to check both, but until we actually get somewhere where our unit tests pass with the Space Translation ON, I think testing that may be more informative than checking we didn't break the case where Space Translation is Off.

tijcolem:
@jmarrec Yeah, good point. I think I asked Joe to set this to Off as this will be the default but we need to get all the tests passing first before doing so. Let's revert back. @jmarrec or @joseph-robertson can either you push up a commit to do so?

joseph-robertson:
@tijcolem @jmarrec Setting `m_excludeSpaceTranslation = true` would be the default going forward. This translates to the old behavior (no space translation).

joseph-robertson:
Maybe I'm not following...

tijcolem:
@joseph-robertson Yes, right before this PR gets merged in we want to set `m_excludeSpaceTranslation = true;`, but right now as @jmarrec pointed out, the unit tests are passing (which is good) but we want to correct any test failures when `m_excludeSpaceTranslation = false;`  which I think there are still some failing.  

Going back a few commits when `m_excludeSpaceTranslation = false;` shows:  https://ci.commercialbuildings.dev/blue/organizations/jenkins/openstudio-incremental-osx/detail/PR-4434/16/pipeline

I tried to run regression tests using the installer from this branch and was getting E+ errors. I think first step is get everything passing here with `m_excludeSpaceTranslation = false;` and then move on to reg tests.

tijcolem:
Thanks @joseph-robertson.  Okay now we can resume and try to work sort out these test failures.

src/energyplus/ForwardTranslator.cpp
```diff
@@ -108,9 +108,13 @@ namespace energyplus {
     m_excludeSQliteOutputReport = false;
     m_excludeHTMLOutputReport = false;
     m_excludeVariableDictionary = false;
+    m_excludeSpaceTranslation = false;
```
kbenne, line 111: Don't we want the default to be true for the moment? Or is the thinking that we handle this defaulting at the osw level?

src/energyplus/ForwardTranslator.cpp
```diff
@@ -108,9 +108,13 @@ namespace energyplus {
...
   }
 
   Workspace ForwardTranslator::translateModel(const Model& model, ProgressBar* progressBar) {
+
+    // When m_excludeSpaceTranslation is false, could we skip the (expensive) clone since we aren't combining spaces?
+    // No, we are still doing stuff like removing orphan loads, spaces not part of a thermal zone, etc
```
kbenne, line 117: Agree. I can't readily provide an example, but im confident that the translated model (a copy) is mutated in other ways besides combine spaces in various parts of the translation process. The unwritten rules of the translator are that it is ok to mutate the model.

src/energyplus/ForwardTranslator.cpp
```diff
@@ -3073,6 +3083,7 @@ namespace energyplus {
   }
 
   std::vector<IddObjectType> ForwardTranslator::iddObjectsToTranslate() {
+    // TODO: avoid the call to iddObjectsToTranslateInitializer with the ton of push_backs
```
kbenne, line 3086: This is a great performance catch. Something needed to be done.

I might be off base here, but this has a bit of a code smell to my nose. it does the job sure. but maybe we can use a member variable and make use of member initialization to do this job. local statics are questionable to me.

src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipment.cpp
```diff
@@ -57,23 +57,25 @@ namespace openstudio {
 namespace energyplus {
 
   boost::optional<IdfObject> ForwardTranslator::translateElectricEquipment(ElectricEquipment& modelObject) {
-    IdfObject idfObject(openstudio::IddObjectType::ElectricEquipment);
-    m_idfObjects.push_back(idfObject);
+
+    IdfObject idfObject = createRegisterAndNameIdfObject(openstudio::IddObjectType::ElectricEquipment, modelObject);
```
kbenne, line 61: yes, I think createRegisterAndNameIdfObject is the correct way. I'm probably the most guilty party of not using this method, although in this case I don't think it was me. I think it is more common in the oldest type translations, which were written before the createRegisterAndName method existed.

src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipment.cpp
```diff
@@ -57,23 +57,25 @@ namespace openstudio {
...
 
-    idfObject.setString(ElectricEquipmentFields::Name, modelObject.name().get());
-
     boost::optional<Space> space = modelObject.space();
     boost::optional<SpaceType> spaceType = modelObject.spaceType();
     if (space) {
-      boost::optional<ThermalZone> thermalZone = space->thermalZone();
-      if (thermalZone) {
-        idfObject.setString(ElectricEquipmentFields::ZoneorZoneListorSpaceorSpaceListName, thermalZone->name().get());
+      if (m_excludeSpaceTranslation) {
```
kbenne, line 72: I expect that in going through this, we'll see this block of code often. I wonder if it can / should be factored out?

src/energyplus/ForwardTranslator/ForwardTranslateSpace.cpp
```diff
@@ -51,22 +74,53 @@ namespace energyplus {
...
+      }
+
+      idfObject.setDouble(SpaceFields::FloorArea, modelObject.floorArea());
+
+      if (boost::optional<SpaceType> spaceType = modelObject.spaceType()) {
+        idfObject.setString(SpaceFields::SpaceType, spaceType->name().get());
+      }
+
+      // Translate all Space-specific loads (and geometry children)
+      auto translateSpaceLoads = [this](auto loads) {
```
kbenne, line 94: oh my auto.

kbenne: maybe there should be a method `SpaceLoadInstance Space::loads` that returns all of the loads and then this is less needed.

src/energyplus/ForwardTranslator/ForwardTranslateThermalZone.cpp
```diff
@@ -169,159 +169,775 @@ namespace openstudio {
...
+      // Translate all Space-specific loads
+      auto translateSpaceLoads = [this](auto loads) {
+        std::sort(loads.begin(), loads.end(), WorkspaceObjectNameLess());
+        for (auto& load : loads) {
+          translateAndMapModelObject(load);
+        }
+      };
+
+      // TODO: THIS SHOULD HAPPEN. The translateSpace should do it!
+      translateSpaceLoads(space.shadingSurfaceGroups());
```
kbenne, line 210: Why can't we just let translateSpace to this work for both cases m_excludeSpaceTranslation == true and m_excludeSpaceTranslation == false

src/energyplus/ForwardTranslator/ForwardTranslateThermalZone.cpp
```diff
@@ -169,159 +169,775 @@ namespace openstudio {
...
 
-      // translate other equipment
-      OtherEquipmentVector otherEquipment = spaces[0].otherEquipment();
-      std::sort(otherEquipment.begin(), otherEquipment.end(), WorkspaceObjectNameLess());
-      for (OtherEquipment& equipment : otherEquipment) {
-        translateAndMapModelObject(equipment);
-      }
+      }  // end spaces
 
       // translate daylighting controls
```
kbenne, line 942: I think this is common to the pathway that does not translate to EnergyPlus Space. Can we factor it out?

kbenne: Actually everything past this point might be common. or very very similar.

src/energyplus/ForwardTranslator/ForwardTranslateThermalZone.cpp
```diff
@@ -169,159 +169,775 @@ namespace openstudio {
...
+        // if all spaces have same directionOfRelativeNorth use that, otherwise clear it
+        if (!space.isDirectionofRelativeNorthDefaulted()) {
+          if (directionofRelativeNorth && (*directionofRelativeNorth == space.directionofRelativeNorth())) {
+            // no-op
+          } else {
+            directionofRelativeNorth.reset();
+          }
+        }
 
-      idfObject.setDouble(openstudio::ZoneFields::YOrigin, spaces[0].yOrigin());
```
kbenne, line 222: Is this method of computing the origin taken from the comnbine spaces algorithm?

src/energyplus/ForwardTranslator/ForwardTranslateSpace.cpp
jmarrec: I think you'll still need the lambda somewhere for the space infiltration stuff, which has to reference a zone and not a space/spacelist.

I have a doubt about whether that actually already exists. I'll check

src/energyplus/ForwardTranslator.cpp
jmarrec: See conversation above. Basically we set it to false now while we work on the PR so we get the CI results (and so we don't have to tweak the OS-resources repo either) . We'll change when we feel PR is ready

src/energyplus/ForwardTranslator.cpp
jmarrec: A couple lines below in translateModelPrivate  it starts by removing orphan loads etc.
While we could do things differently (catch that in the FT function specific to each class), this is only going to increase complexity of this already complex PR.
I might revisit this for performance reasons in the future.

src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipment.cpp
jmarrec: The GenerateClass.rb portion that adds automatic FT code is quite new too, I added that a year or two ago, so there were no boilerplate code in the past.
The automatic FT does createNameAndRegister by default but has a TODO with commented code to explain that you could do it differently in the case where you might actually test for a condition where the object shouldn't be translated (missing required children or invalid combinations)

src/energyplus/ForwardTranslator/ForwardTranslateElectricEquipment.cpp
jmarrec: Yeah I think something should be done to factor it out (used by 10+ functions) but I didn't feel I had the right construct to do it yet

PR #4485, opened 2021-10-27
4438 gbxml export

### LINKED ISSUE #4438: gbXML export is not reproducible
Issue overview
--------------
Objects export order is not reproducible in the gbXML ForwardTranslator.  The ForwardTranslator should sort objects to ensure the export is reproducible.

## Current Behavior
Objects are exported in non-reproducible order.

## Expected Behavior
Objects are exported in reproducible order.

## Steps to Reproduce
Create an OSM model via a script, export to gbXML.  Repeat, exported gbXML is not in same order.  This is especially problematic for interior floor/ceilings as sometimes the floor is exported and sometimes the ceiling is.  Sorting by surface type would ensure that interior floors are always exported (the interior ceiling would be listed as the adjacent surface).

## Possible Solution
Sort all objects when doing gbXML export.  Need to sort surfaces by surface type (Wall, Roof, Floor, etc) first.  Need to figure out sorting criteria for each objects (probably want to sort spaces by story or min z coord, then maybe min x and min y coords, then name)

### Environment

### Context
Topologic Energy uses OpenStudio's gbXML export, this issue causes confusion and possible wrong results when the results imported into other tools such as Spider Aragog gbXML Viewer.  

Reported by @wassimj at https://unmethours.slack.com/archives/C0ADY1ZRB/p1631800632015300

ggartside:
Pull request overview
---------------------

 - Fixes #4438 

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


src/model/Model.cpp
```diff
@@ -1518,7 +1518,13 @@ namespace model {
     buildingStory.setNominalFloortoFloorHeight(3);
 
     // create spaces from floor print
-
+    // GG 10/26/21 - Change the spac egeneration to use fromFloorPrint instead of using Clone
```
macumber, line 1521: Clone would give each object a new unique UUID handle but that should not matter if you are properly sorting the surfaces by name.  Another good test would be to write the model to file and re-read it back in, that is a typical use case.

src/model/Model.cpp
```diff
@@ -1548,18 +1554,24 @@ namespace model {
     space1->setThermalZone(thermalZone);
     space1->setBuildingStory(buildingStory);
 
-    ModelObject clone = space1->clone(model);
```
macumber, line 1551: Oh I see, this probably wasn't deterministic because clone isn't deterministic.  You could make clone sort the children to fix that:

https://github.com/NREL/OpenStudio/blob/develop/src/model/ParentObject.cpp#L106

src/model/Model.cpp
ggartside: I think what I was trying to do here was create what I know to be identical models in OpenStudio and export them to gbxml and then using an external file comparison verified that the files are identical. When I used Clone the order of the surfaces came back different each time so the files would never be the same (i.e. for example Surface 5 in one model was a floor and in the other model a ceiling) so I used fromfloorPrint to generate the spaces instead.

src/model/Model.cpp
ggartside: See my above comment - would sorting fix the issue I was addressing? Namlelt that creating the export model twice did not produce identical models?

macumber: Yes sorting the result of children in ParentObject_Impl::clone should go a long way towards making the models identical.  The objects will have different UUIDs so the models won't be completely identical.

tijcolem:
@macumber @ggartside Looks like this is good to go?

macumber:
I don't know if @ggartside plans to update `ParentObject_Impl::clone` to sort children or not.  If so, some of the changes to `Model::exampleModel` could be reverted

src/model/Model.cpp
ggartside: I don't know much about the clone itself but that seems to imply that children() doesn't return a specific order of children each time it is called? Its not just that they are coming back in a different order each time clone is called the surfaces are different, for example the first time clone is called I get this

image

and the second time I get this

image

(screenshots are comparing the two xml files)

ggartside: I think my best course of action is to create my own simpler variant of exampleModel specifically for my test

ggartside:
> 

No - I created a method that creates a simple model that can be used for my test, I reverted exampleModel back to its original incarnation

PR #4505, opened 2021-11-18
Addresses #4477, improve handling of invalid values passed to SDK

### LINKED ISSUE #4477: Improve handling of invalid values passed to SDK
Enhancement Request
--------------
Improve the handling of invalid argument values passed to SDK methods. The current behavior is arguably dangerous and can result in software bugs and incorrect simulation results that developers/users are likely not even aware of.

## Detailed Description
Currently when an OpenStudio measure passes an invalid argument value to an SDK method, the method will not apply the change and simply return false. This means that a default value may get used in the EnergyPlus simulation unbeknownst to the user, which, depending on the EnergyPlus input, could case a severe impact to simulation results. 

A couple real-world examples where this occurred:
- OpenStudio-HPXML: It was possible to calculate a negative water heater tank loss coefficient and try to [assign that to the water heater](https://github.com/NREL/OpenStudio-HPXML/blob/184ca0855d29ac97d5c042caebfd368200319c72/HPXMLtoOpenStudio/resources/waterheater.rb#L1609). (We added [error-checking](https://github.com/NREL/OpenStudio-HPXML/pull/198) to fix this.) The bug resulted in very high water heating energy use in the simulation results.
- OpenStudio-standards: EnergyPlus changed its fuel type enumerations but an old value is [still being used](https://github.com/NREL/openstudio-standards/issues/1100) by openstudio-standards. The bug results in the wrong heating fuel type in the simulation results.

I'm sure there are other bugs I'm unaware of. 😱 

It is very easy for a developer/user to be blissfully unaware that an invalid value was passed and a default value is being used instead. The only way to currently identify these issues is to check the return value of _every_ SDK method call, but this would make measures' code significantly more complex.

## Possible Implementation
**Option 1.** I would strongly advocate for the SDK raising an exception that would cause the OS measure to fail, so that the measure can be fixed. An on/off switch could be allowed if some developers don't want their measures to break and want to continue living in ignorant bliss.

**Option 2.** Another option would be to add `model.warnings` and `model.errors` methods, similar to the [forward translator](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.2.1-doc/energyplus/html/classopenstudio_1_1energyplus_1_1_forward_translator.html).

**Option 3.** One last option would be to allow the invalid arguments to persist in the OS model. Then they could A) get validated _once_ at the end of model creation or B) get passed to EnergyPlus and result in a simulation error, allowing the developer to fix the issue. This might have unintended consequences? But it would also probably improve performance pretty significantly.

joseph-robertson:
Pull request overview
---------------------

 - Fixes #4477

I think we are leaning toward an implementation of 3B: optional switch for turning off idd validity checking.

**PR summary**:
- [x] Rename "None" strictness level to "Minimal"
- [x] Introduce new "None" strictness level
- [x] Modify the following methods of `src/utilities/idf/WorkspaceObject.cpp`:
  - [x] `setName`: enforce name unicity for None and Minimal; **this is a "breaking change bug fix" for Minimal**
  - [x] `setString`: avoid checking validity when None
  - [x] `setPointer`: avoid checking validity when None
- [x] New test in `src/energyplus/Test/WaterHeaterMixed_GTest.cpp` for checking the setting of valid/invalid values across None, Minimal, Draft strictness levels
- [ ] **HELP NEEDED** New test in `src/utilities/idf/Test/Workspace_GTest.cpp` for checking None vs Minimal

Questions after setting `checkValidity=false` for `IdfObject_Impl::setDouble` and `IdfObject_Impl::setString`:
- [x] This is starting to uncover some issues with setters. For example, this conflicts with the idd: https://github.com/NREL/OpenStudio/blob/develop/src/model/CoolingTowerSingleSpeed.cpp#L993.
- [x] Appears we do not need to set `IdfObject_Impl::setString` in addition to `WorkspaceObject_Impl::setString`
- [x] Do we want to set `checkValidity=false` for `WorkspaceObject_Impl::setPointer`? If we don't, this method checks for validity over everything. So after setting an "invalid" value, this method would return false and clear out all references.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


jasondegraw:
How is this an improvement over the current handling? Isn't there already a way to make sure this is handled properly without introducing a flag to turn off something that seems pretty central to the way that OpenStudio has behaved?

shorowit:
We now have three real-world examples of bugs found by this PR:
1. As described in the PR description, the CoolingTowerSingleSpeed constructor was setting an invalid value.
2. I ran this build through our OpenStudio-ERI tests and found a bug where we would sometimes calculate a negative conductivity for a material. Before, we were calculating a wrong ERI score but were unaware. Now we get an EnergyPlus error (`   ** Severe  ** <root>[Material][my material][conductivity] - "-5.737216" - Expected number greater than 0.000000`) so we can fix the problem.
3. When we were transitioning Home Energy Score to OpenStudio/EnergyPlus, there were situations where window U-factors could be greater than the limit allowed by EnergyPlus (which we are [_finally_ removing](https://github.com/NREL/EnergyPlus/pull/9117)), but we were unaware and produced simulation results using the OpenStudio default value. This default value is nowhere near the intended U-factor, so the results were completely bogus and would have resulted in wrong scores for real homes. We luckily caught the issue before we rolled out the new Home Energy Score version, but it was only because we were comparing results to the previous version of Home Energy Score using DOE2.

On the bright side, I ran this build through a couple other repos spanning several hundred test files and they ran successfully. So thankfully I don't have lots of bugs everywhere.

This PR brings me a great deal of relief. There is nothing worse than producing wrong results and not knowing about it. Or banging your head against the wall and wasting hours because you can't figure out why OpenStudio doesn't accept the value you're setting (see recent Unmet Hours slack conversation).

shorowit:
Found another bug -- a negative WaterUse:Equipment peak flow rate was being calculated.

shorowit:
And found another one -- ZoneInfiltration:EffectiveLeakageArea with stack and wind coefficients that are zero, which is not allowed.

jmarrec:
I still tend to personally think the code calling openstudio SDK should be the one checking the return value... And we should do better with `OS_ASSERT` in our internal calls (such as Ctors) to find problems: we still have a bunch of old ctors that do not use `OS_ASSERT` or not enough, but we could and should fix those by temporarilly throwing inside `setXXX`.
To me the validity checking that OS does is one of the key value proposition that the SDK has.

But anyways, given you have a strong use case for it, that there is consensus on adopting this change, **and given it'll be an opt-in feature**, we should just implement and stop discussing the why.

---

Thinking about it some more, I think we shouldn't add yet another global boolean flag on `Workspace`. What I would propose to do is:

* **Rely on the Workspace's `StrictnessLevel`**
* Revamp setString/setName/setPointer in `Workspace_Impl` to not do *any* kind of checking if the strictness level says so 
    *  `StrictnessLevel::None` is probably appropriate, the others are `Draft` and `Final` (Final I don't think we ever really use...)
    * Currently there are still some amount of checking done if the strictness level is `None`
    * Worse case we add a 4th StrictnessLevel (eg: change the current `None` to `Minimal`, and make `None` do nothing)
* User wanting to not have any checking would do `model.setStrictnessLevel("None".to_StrictnessLevel)`
* We should ensure we pass that to the FT, either by just checking that if model has a strictnessLevel of `None`, then the IDF Workspace should be `None` to, or provide a specific FT config option

What do you think of this @joseph-robertson @shorowit ?

mbadams5:
I've been silently following this PR thus far. I agree with @jmarrec on this subject.

I believe all the examples you gave @shorowit are due to not checking the return value from the setters. If a user is not checking the return value from the setter then that isn't a bug in OpenStudio but rather a bug in their code.

Now the issue with the CoolingTowerSingleSpeed constructor is definitely a bug in OpenStudio and should be addressed. And as @jmarrec mentioned, internally OpenStudio should be checking the return values from setters as well, especially in the CTORs with `OS_ASSERT`.

The validation is performed on every setter due to the OpenStudio App GUI. It uses the `bool` return value to inform the user if they've entered an invalid entry (and it wouldn't update the model until the input is valid). Without this, you would not have the real time feedback to fix errors until you went to simulation the model and then you'd have to map the EnergyPlus errors back to the correct OpenStudio object/field.

Now as people have started to use only the SDK more heavily, especially using it to programmatically generate buildings (up to millions of buildings), that per setter validation becomes less important. When you are programmatically generating buildings, there is greater control over input parameters and it all runs sequentially (and "behind the scenes") so that instantaneous feedback needed for the GUI is not needed for the SDK. However, you can (and should) still check the `bool` return to make sure you are entering a valid parameter if there is any doubt about the input parameter (value, range, type, etc.). 

I highly disagree that this new behavior should be the default behavior. That is a huge change from how OpenStudio has always functioned. I agree with @jmarrec that this should be *opt-in* and either you either use a specific setter or use the `StrictnessLevel` like he suggested. And I actually like his "worse case" option to "change the current `None` to `Minimal`, and make `None` do nothing".

All that said, I am definitely onboard with having an opt-in setter or `StrictnessLevel` to eliminate validation *if you know what you are doing*. Validation has always been a large hotspot in OpenStudio code and if you know you are programmatically creating valid OpenStudio/EnergyPlus models then you can avoid that performance hit, especially when generating thousands to millions of models.

shorowit:
> However, you can (and should) still check the `bool` return to make sure you are entering a valid parameter if there is any doubt about the input parameter (value, range, type, etc.).

It's worth pointing out that the examples you give here (value, range, type, etc.) _are not treated the same_. If you pass a wrong data type or wrong OSM object type to a setter, it raises an error, whereas an invalid value (enumeration, range) returns false. As a developer, I find it confusing that I get different results depending on the type of error in my code.

For the record, my preference here was to have invalid values raise an immediate error (see Option 1 in the [original GitHub issue](https://github.com/NREL/OpenStudio/issues/4477)). The OpenStudio Application would still be able to notify users of an invalid value that way. But short of that solution, I'll take anything I can get. Just having the build from this PR has been super helpful to me to find problems.

> I believe all the examples you gave @shorowit are due to not checking the return value from the setters. If a user is not checking the return value from the setter then that isn't a bug in OpenStudio but rather a bug in their code.

I somewhat agree with you in theory, but not in practice. In ruby land, I'm not aware of a single example where developers are checking return values from setters. It's not happening in openstudio-standards. It's not happening in BCL measures. It's barely mentioned in the measure writing guide or the documentation. No one is doing it in practice. And so there's a lot of dangerous analysis being done today because **end users** aren't aware that their results may be compromised.

shorowit:
Using an opt-in approach is obviously an easier decision to make because no one will complain -- but in practice, it means that most users/workflows will not be opted in and so these kinds of bugs, producing erroneous simulation results, will continue to go largely unnoticed.

jasondegraw:
I pretty much agree with everything @mbadams5 has written. The current behavior is intentional, so the current behavior isn't a bug. We may disagree with the decision that was made, but that doesn't make it a bug.

If the "problem" is that return values are being discarded, I'm not especially impressed. Bad programming practice is bad programming practice. Using bad programming to justify dropping validation as the default behavior when that's always been the way that OpenStudio has behaved is just wrong IMHO. My personal preference would be to leave things as they are.

There is an argument to be had about performance, and that could be enough to do this. But not enough to make it the default, and this PR doesn't look like it's on the right track. If we really must do this, then please leave the current behavior as the default.

joseph-robertson:
Give the PR another look (specifically [a3bc942](https://github.com/NREL/OpenStudio/pull/4505/commits/a3bc9429ccfc7f6a7fb64b96ab669eb0b355703e)). We are keeping the current behavior as the default behavior.

src/utilities/idf/ValidityEnums.hpp
```diff
@@ -39,20 +39,23 @@ namespace openstudio {
...
 OPENSTUDIO_ENUM(StrictnessLevel,
   ((None))
+  ((Minimal))
   ((Draft))
   ((Final))
 );
  *  \endcode */
 OPENSTUDIO_ENUM(StrictnessLevel,
   ((None))
+  ((Minimal))
```
jmarrec, line 58: I know I'm the one that proposed changing the current `None` to `Minimal`, but that's actually a breaking change... Since it's in the public api: a user is free to instantiate a workspace themselves eg: `w = OpenStudio::Workspace.new("None".to_StrictnessLevel, "EnergyPlus".to_IddFileType)`

I don't know how much we care though.

An alternative would be to add another name completely, like `Unsafe`, `FallThrough`, or something. Otherwise if we're good with this, we should make it part of the release notes (and label  this PR as `API Change`)

src/utilities/idf/WorkspaceObject.cpp
```diff
@@ -422,10 +422,16 @@ namespace detail {
 
   // Pre-condition:  Object valid at Workspace's strictness level.
   bool WorkspaceObject_Impl::setString(unsigned index, const std::string& value, bool checkValidity) {
+    StrictnessLevel level = m_workspace->strictnessLevel();
+
+    // StrictnessLevel::None
+    if (level < StrictnessLevel::Minimal) {
+      checkValidity = false;
+    }
```
jmarrec, line 430: Not a big fan of simply overriding the arg, but that's mostly cosmetic at this point, but maybe I'll feel even more strongly about this once I've looked at all the changes.

src/energyplus/ForwardTranslator.cpp
```diff
@@ -678,7 +678,7 @@ namespace energyplus {
       this->createStandardOutputRequests();
     }
 
-    Workspace workspace(StrictnessLevel::None, IddFileType::EnergyPlus);
+    Workspace workspace(StrictnessLevel::Minimal, IddFileType::EnergyPlus);
```
jmarrec, line 681: Do you need a way to go to `None` here? either via checking if model's strictness is None => None, or via a ft.setStrictnesslevel method or something

Maybe you don't, I'm not 100% sure.

src/utilities/idf/WorkspaceObject.cpp
jmarrec: The alternative is to do something like that on line 461 below for eg:

```diff
-      if (checkValidity) {
+      if (checkValidity && level > StrictnessLevel::None) {
         if (!fieldDataIsValid(index, level).empty()) {
```

jmarrec: Function calls like `convertToTargetHandle(value, iddObject().objectLists(index), checkValidity);` would (or should after modification) work because they do stuff like this already:

```
    if ((!checkValidity) || (m_workspace->strictnessLevel() < StrictnessLevel::Draft)) {
```

joseph-robertson:
@jmarrec Can we create a test that shows the difference between None and Minimal? Should that test go in `src/utilities/idf/Test/Workspace_GTest.cpp`?

PR #4513, opened 2022-01-04
Addresses #4457, support gbXML translation where user-input <Name> is different from the id

### LINKED ISSUE #4457: Support gbXML translation where user-input <Name> is different from the id
Enhancement Request
--------------
For the majority of objects in gbXML, there is both an `id` that is used to reference the object throughout the XML but there is also a human-readable `<Name>` tag for the object. Here is an example with Space:

gbXML_Name_and_id

Currently, there is no way to produce gbXML files with the OpenStudio SDK where objects have a `<Name>` that is different than their `id` so both fields suffer from the same restrictions. This also means that the `id` of the object is lost when importing from gbXML to OpenStudio and the resulting gbXML gets imported with the user-input `<Name>` set as the EnergyPlus Name, meaning the imported object name often contains characters that are not appropriate for EnergyPlus.

## Detailed Description

In many workflows, it makes sense for the [Name](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-2.0.3-doc/utilities/html/classopenstudio_1_1_workspace.html#a67f42851af364e8c6e8faa8835bfafa8) of the OpenStudio workspace object to become gbXML `id` (and vice versa on import) since OpenStudio Names are ultimately what identify the objects within IDF and in the EnergyPlus results. Both gbXML ids and OpenStudio/EnergyPlus Names need to follow typical character restrictions and they need to be unique since they are used to identify objects across a model. So it would be helpful if the gbXML importer (aka. [ReverseTranslator](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.2.1-doc/gbxml/html/classopenstudio_1_1gbxml_1_1_reverse_translator.html)) could have an option to import the gbXML ids as OpenStudio/EnergyPlus Names instead of always trying to make the gbXML `<Name>` act like an identifier.

Conversely, given that the `<Name>` tags in gbXML are often specified by end users, such `<Name>` tags can often be duplicated in the objects across a gbXML model or they can contain characters that aren't legal for EnergyPlus. It would be nice to have some way to support this usage of the gbXML `<Name>` tag in the OpenStudio SDK since, right now, such usage can result in imported gbXML models that are not simulate-able in EnergyPlus. Or, if they are simulate-able, the EnergyPlus results cannot be easily matched back to the original gbXML because OpenStudio escapes special characters in the gbXML `<Names>` when it translates them to IDF.

## Possible Implementation

Working with the features currently supported in the OpenStudio SDK, we can see that [all ModelObjects in OpenStudio have an addiditonalProperties](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.2.1-doc/model/html/classopenstudio_1_1model_1_1_model_object.html#a86a8b998aa97760411f84e745a23c6a9) attribute that can be used to store custom types of user inputs (like user-input names).

Within the [Ladybug Tools software](https://www.ladybug.tools/) that I support, we are currently planning to store the user-input names of objects under these [`AdditionalProperties`](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.2.1-doc/model/html/classopenstudio_1_1model_1_1_additional_properties.html). We plan to call this user-input name under `AdditionalProperties` a `DisplayName` and we'll use it to ensure that these user-input names can survive the translation from OpenStudio and back to our software, while still ensuring that the IDF and E+ Results use ids that are appropriate for E+.

Perhaps, if we could agree that that `DisplayName` (or some other property) could be a standard `additionalProperty` that was available on OpenStudio Model Objects, we can use it to support the the current intended behavior of the `<Name>` tag in gbXML. This would free up the gbXML id to act as an identifier across the gbXML > OSM > IDF > E+ Results translation process.

## Note

I realize that this request is probably not that technically difficult to implement but I'm sure that many of us will have strong opinions. So let's all be open-minded that there may be a few good possible implementations to making this type of interoperability with gbXML work.

CC: @tijcolem , @tanushree04 , @macumber

joseph-robertson:
Pull request overview
---------------------

 - Fixes #4457

Checklist:
- [x] Add helper methods on `ModelObject` for getting/setting `gbXMLId`, `CADObjectId`, `displayName`
- [x] New FT test
- [x] Update existing RT tests
- [x] New RT test

Questions:
- [ ] `gbXMLId` is not FT'd for `Building`
- [ ] `displayName` is not FT'd for `Building`
- [ ] `CADObjectId` is not FT'd for `Building`, `BuildingStory`, `Construction`, `Material`

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


macumber:
This looks good @joseph-robertson !

joseph-robertson:
> 

@jmarrec Can you summarize what this means?

chriswmackey:
I just tested it and it works beautifully! I know this is going to make a lot of our users happy.

Very excited to have this in the next release!

src/gbxml/ForwardTranslator.cpp
```diff
@@ -1350,6 +1327,40 @@ namespace gbxml {
...
+      model::AdditionalProperties additionalProperties = modelObject.additionalProperties();
+      if (additionalProperties.hasFeature("gbXMLId")) {
+        id = additionalProperties.getFeatureAsString("gbXMLId").get();
+      }
+    }
+    auto idElement = parentElement;
+    idElement.append_attribute("id") = escapeName(id).c_str();
+    result = idElement;
+    return result;
+  }
```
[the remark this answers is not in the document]
jmarrec: I realize now you probably just did the boost::optional<pugi::xml_node> return type to match the existing `translateCADObjectId`.  But I'd say this should be changed too, there isn't a single use where the return type isn't discarded.

Ideally all of our getters would be marked `[[nodiscard]]` and this type of code wouldn't compile anymore.

src/gbxml/ForwardTranslator.hpp
```diff
@@ -113,6 +113,8 @@ namespace gbxml {
     boost::optional<pugi::xml_node> translateLayer(const openstudio::model::Material& material, pugi::xml_node& parent);
     boost::optional<pugi::xml_node> translateMaterial(const openstudio::model::Material& material, pugi::xml_node& parent);
     boost::optional<pugi::xml_node> translateConstructionBase(const openstudio::model::ConstructionBase& constructionBase, pugi::xml_node& parent);
+    boost::optional<pugi::xml_node> translateId(const openstudio::model::ModelObject& modelObject, pugi::xml_node& parentElement);
+    boost::optional<pugi::xml_node> translateName(const openstudio::model::ModelObject& modelObject, pugi::xml_node& parentElement);
```
jmarrec, line 117: Some docstring explaining how the mapping is done would be nice IMHO.

src/gbxml/ReverseTranslator.cpp
```diff
@@ -1046,6 +1020,30 @@ namespace gbxml {
...
+                                                                                 openstudio::model::ModelObject& modelObject) {
+    std::string id = element.attribute("id").value();
+    m_idToObjectMap.insert(std::make_pair(id, modelObject));
+    modelObject.setName(id);
+
+    model::AdditionalProperties result = modelObject.additionalProperties();
+    result.setFeature("gbXMLId", id);
+
+    return result;
+  }
```
jmarrec, line 1033: I know you're matching existing code, but the return type bothers me a bit here as well.
It's always initialized to an `AdditionalProperties` object. And it will never be used.

src/gbxml/ReverseTranslator.hpp
```diff
@@ -98,6 +98,8 @@ namespace gbxml {
     boost::optional<openstudio::model::ModelObject> translateSpace(const pugi::xml_node& element, openstudio::model::Model& model);
     boost::optional<openstudio::model::ModelObject> translateSurface(const pugi::xml_node& element, openstudio::model::Model& model);
     boost::optional<openstudio::model::ModelObject> translateSubSurface(const pugi::xml_node& element, openstudio::model::Surface& surface);
+    boost::optional<openstudio::model::ModelObject> translateId(const pugi::xml_node& element, openstudio::model::ModelObject& modelObject);
+    boost::optional<openstudio::model::ModelObject> translateName(const pugi::xml_node& element, openstudio::model::ModelObject& modelObject);
```
jmarrec, line 102: docstring maybe?

src/gbxml/ForwardTranslator.cpp
```diff
@@ -1350,6 +1327,40 @@ namespace gbxml {
...
+    boost::optional<pugi::xml_node> result;
+
+    std::string id;
+    id = modelObject.name().get();
+    if (modelObject.hasAdditionalProperties()) {
+      model::AdditionalProperties additionalProperties = modelObject.additionalProperties();
+      if (additionalProperties.hasFeature("gbXMLId")) {
+        id = additionalProperties.getFeatureAsString("gbXMLId").get();
+      }
+    }
```
jmarrec, line 1340: You added a helper  `boost::optional<std::string> ModelObject_Impl::gbXMLId() const`, so use that instead.

src/model/ModelObject.cpp
```diff
@@ -752,6 +752,51 @@ namespace model {
...
+      if (hasAdditionalProperties()) {
+        AdditionalProperties additionalProperties = this->additionalProperties();
+        if (additionalProperties.hasFeature("CADObjectId")) {
+          result = additionalProperties.getFeatureAsString("CADObjectId");
+        }
+      }
+      return result;
+    }
+
+    bool ModelObject_Impl::setCADObjectId(std::string cadObjectId) {
```
jmarrec, line 766: Do not make a copy of std::string here. Use a const ref
```suggestion
    bool ModelObject_Impl::setCADObjectId(const std::string& cadObjectId) {
```

src/model/ModelObject.cpp
```diff
@@ -851,6 +896,30 @@ namespace model {
     return getImpl<detail::ModelObject_Impl>()->removeAdditionalProperties();
   }
 
+  boost::optional<std::string> ModelObject::cadObjectId() const {
+    return getImpl<detail::ModelObject_Impl>()->cadObjectId();
+  }
+
+  bool ModelObject::setCADObjectId(std::string cadObjectId) {
+    return getImpl<detail::ModelObject_Impl>()->setCADObjectId(cadObjectId);
```
jmarrec, line 904: Same for public side, use a const ref, (or take by val but `std::move` it)
```suggestion
  bool ModelObject::setCADObjectId(const std::string& cadObjectId) {
    return getImpl<detail::ModelObject_Impl>()->setCADObjectId(cadObjectId);
```

src/gbxml/ReverseTranslator.cpp
```diff
@@ -1046,6 +1020,30 @@ namespace gbxml {
...
+  boost::optional<openstudio::model::ModelObject> ReverseTranslator::translateName(const pugi::xml_node& element,
+                                                                                   openstudio::model::ModelObject& modelObject) {
+    model::AdditionalProperties result = modelObject.additionalProperties();
+
+    auto name = element.child("Name").text();
+    if (!name.empty()) {
+      result.setFeature("displayName", name.as_string());
+    }
+
+    return result;
```
jmarrec, line 1044: use the helper you added no? `bool ModelObject::setDisplayName(const std::string& displayName)`

jmarrec (changes requested):
Some minor changes requested, mostly docstring, use the helpers, and avoid copies by taking std::string as const references

Otherwise looks good to me!

src/gbxml/ForwardTranslator.cpp
```diff
@@ -1350,22 +1327,39 @@ namespace gbxml {
...
-        if (cadObjectId) {
-          if (additionalProperties.hasFeature("programIdRef")) {
-            boost::optional<std::string> programIdRef = additionalProperties.getFeatureAsString("programIdRef");
-            if (programIdRef) {
-              auto cadObjectIdElement = parentElement.append_child("CADObjectId");
-              cadObjectIdElement.append_attribute("programIdRef") = (*programIdRef).c_str();
-              result = cadObjectIdElement;
-            }
+      boost::optional<std::string> cadObjectId = modelObject.cadObjectId();
+      if (cadObjectId) {
```
jmarrec, line 1355: This is kinda a side note, but I would prefer always using the initialize in if construct. It limits scope, so you cannot try to access cadObjectId by mistake after the if block.

```suggestion
      if (boost::optional<std::string> cadObjectId = modelObject.cadObjectId()) {
```

PR #4522, opened 2022-01-26
Refactor GLTF

jmarrec:
Pull request overview
---------------------

This is a refactor of the code on https://github.com/NREL/OpenStudio/pull/4520

### DIFF
--- a/src/model/GltfForwardTranslator.cpp
+++ b/src/model/GltfForwardTranslator.cpp
@@ -76,26 +76,28 @@
 #define STB_IMAGE_IMPLEMENTATION
 #define STB_IMAGE_WRITE_IMPLEMENTATION
 
-#include "tiny_gltf.h"
+#include <tiny_gltf.h>
 
-#include <thread>
+#include <algorithm>
 #include <cmath>
+#include <iterator>
+#include <numeric>
 #include <stack>
+#include <limits.h>
 
 namespace openstudio {
 namespace model {
 
   namespace GLTF = tinygltf;
-  using namespace GLTF;
 
   // For raw values for GLTF Materials
-  struct mat_Data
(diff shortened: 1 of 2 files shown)


src/model/GltfForwardTranslator.cpp
```diff
@@ -108,40 +110,49 @@ namespace model {
...
   // param : _materials
   void initMaterials(std::vector<GLTF::Material>& _materials) {
     // default material is always set at index 0;
-    const float grey = 0.8f;
-    const float alpha = 1.0f;
-    GLTF::Material m = createMaterial("Default material", grey, grey, grey, alpha, true);
-    _materials.push_back(m);
+    const int grey = 204;
+    const double alpha = 1.0;
+    createMaterial(_materials, MaterialData{.matName = "Default material", .r = grey, .g = grey, .b = grey, .a = alpha, .isDoubleSided = true});
```
Brijendra21, line 155: We're getting an error here "**use of designated initializers requires at least '/std:c++20**'"
if we're bumping up the language standard to **"ISO C++20 Standard (/std:c++20)"** from "ISO C++17 Standard (/std:c++17)" then more errors are introduced. Please Suggests...

jmarrec: This worked on gcc with c++17, my bad.
Remove the designated initializers then.

tijcolem:
@jmarrec  Thanks for all the great suggestions with this PR.  It looks like @Brijendra21 implemented these one by one as separate commits.  Going forward, if we have a PR such as this, we'll want to merge it as the commit history can get a bit messy to understand what was implemented.

tijcolem:
Closing since these were added as separate commits to #4520

PR #4523, opened 2022-01-28
Performance benchmarks

jmarrec:
Pull request overview
---------------------

This extracts some benchmarks on several branches I have. 

I'm not sure what's the most practical way to do this. Some of these benchmarks are testing variations of something, like Current , Proposed1, Proposed2. If we're comparing two builds, then we should only have 'Current' (and this will change as we do changes).

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


jmarrec:
@kbenne @tijcolem I created an external repo at https://github.com/jmarrec/OpenStudio-benchmarks

My idea is that it would be more convenient to have it externally and just link to libopenstudiolib. It'll allow adding benchmarks at any point and test against a previous installer. Building the benchmarks is quite quick, but we could even override which libopenstudiolib.so it loads by using LD_LIBRARY_PATH.

What do you think?

tijcolem:
@jmarrec  I built your branch and enabled the optional BUILD_BENCHMARK. All the statistics that are generated by the python driver script are really impressive!  I do like the way it's laid out now but perhaps we can expand the current script to include post-process results from the current build and a comparison build that could show differences as percentage?  We could also keep track of these stats on CI for every build and upload results to a bucket and pull from there vs having to re-run constantly.  

Also, I'd be fine with using the env LD_LIBRARY_PATH to toggle between different libopenstudiolibs.  I don't think it works on windows though?  Maybe it has some other equivalent way to load dynamic libs.  For now though, I think we are running all the perf stuff in linux so prob not an immediate concern.  

This is awesome stuff!!

jmarrec:
@tijcolem I'm thinking it may be better to have these in an external repo (https://github.com/jmarrec/OpenStudio-benchmarks), so it's easier to add a benchmark **after**, versus having to put the benchmark in develop, then rebase your feature branch onto it so you can compare. Thoughts?

tijcolem:
@jmarrec Following up on this. I like you idea of having this in a separate repo where we can link to different versions of libopenstudiolibs.  I created a new repo. https://github.com/NREL/OpenStudio-benchmarks.  Could you move your repo to this?

tijcolem:
@jmarrec Also, since this will be separate from the main OpenStudio and it doesn't much to build we can take use GitHub actions https://github.com/NREL/OpenStudio-benchmarks as the driver to track performance metrics.

jmarrec:
@tijcolem  pushed to https://github.com/NREL/OpenStudio-benchmarks

tijcolem:
@jmarrec Thanks for moving to the new repo! I think this a good home for this so I'm going to close this.

PR #4527, opened 2022-02-03
Gerard/4372 intersection issue found using create bar

### LINKED ISSUE #4372: Intersection issue found when using create bar
Issue overview
--------------
There is small identified error in pre/post intersection area, but that doesn't account for larger error in expected exterior perimeter reported. 

            "stdout" : "[openstudio.model.Surface] <1> Initial area of other surface 'Surface 316' 57.2551 does not equal post intersection area 57.1732\n[openstudio.model.Surface] <1> Initial area of surface 'Surface 326' 13.9355 does not equal post intersection area 13.8945\n[openstudio.model.Surface] <1> Initial area of surface 'Surface 338' 13.9355 does not equal post intersection area 13.8945\n",
            "step_errors" : 
            [
               "East/West walls don't have the expected area (actual 5,492.4418 ft^2, target 3,661.6279 ft^2)"
            ],

## Current Behavior
That error is called here
https://github.com/NREL/openstudio-extension-gem/blob/develop/lib/openstudio/extension/core/os_lib_model_generation.rb#L2543

## Expected Behavior
Expected values should match and meaasure and method shoudl run without an error

## Steps to Reproduce
I'm going to add a test on branch of extension gem with these inputs and should be able to replicate it

### Environment

### Context
Is blocking URBANopt project until I figure out work arond

ggartside:
Pull request overview
---------------------

 - Fixes #4372

Fixes an issue revealed by surface matching when small surfaces created in intersection were reduced to 0 area surfaces and discarded. The tolerance used in removeSpikesEx should have been the same value as the amount used to shrink and expand surfaces to remove spikes 

PR includes the fix (one line change) a test model and a unit test in Model_Gtest to verify the surfaces are paired correctly

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


ggartside:
The issue was a fault in removeSpikesEx. In this method a polygon is expanded by `amount` and then shrunk by `amount` to remove small spikes, the resulting vertices were compared to the original vertices and adjusted accordingly so that the original coordinates could be returned wherever possible. The tolerance used to compare vertices was too large and resulted in the polygon being reduced to a 0 area polygon and rejected. By changing the tolerance to the amount expanded and shrunk the issue was fixed.

Polygons are rejected based on area tolerance in intersection - they have to be greater than tol * tol which is 100mm2, surface matching uses a linear tolerance of 0.01, 10mm. If a polygon is rejected because it is less than 100mm2 it is possible that vertices in that polygon could be > 10mm apart and hence surface matching could fail under certain circumstances. 

Currently, if two identical surfaces are intersected their vertices are not changed by the intersection method, which makes perfect sense. However if two almost identical surfaces are intersected and the resulting difference between them is too small to be considered then the original surface vertices are not changed either even though they are slightly different. 

These two surfaces are therefore considered to be identical after intersection is performed but may not be considered identical as far as surface matching is concerned (because they are not identical)

To eliminate this error intersection should always return **two identical surfaces that will be determined to be equal when passed to circulartEqual**

ggartside:
I resolved the two test fails. One of them was caused by adding the circularEqual test in Surface intersect. The circulaEqual is removing inline points and so the count of vertices dropped from 6 to 4. This may be an unwanted artifact so to be prudent it was removed

src/model/test/Model_GTest.cpp
```diff
@@ -790,3 +792,57 @@ TEST_F(ModelFixture, Ensure_Name_Unicity_SpaceAndSpaceGroupNames) {
...
+TEST_F(ModelFixture, Issue_4372) {
+
+  // Open OffsetTests
+  // INtersect surfaces
+  // Match Surfaces
+  // Check the internal walls are paired
+  // Surface 4 Space 101 with SPace 102 surface 12
+  // Surface 2 Space 102 with
+
+  openstudio::path modelPath = resourcesPath() / "model" / toPath("offset_tests.osm");
```
jmarrec, line 805:
```suggestion
  openstudio::path modelPath = resourcesPath() / toPath("model/offset_tests.osm");
```

src/model/test/Model_GTest.cpp
```diff
@@ -790,3 +792,57 @@ TEST_F(ModelFixture, Ensure_Name_Unicity_SpaceAndSpaceGroupNames) {
...
+  ASSERT_TRUE(openstudio::filesystem::exists(modelPath));
+
+  openstudio::osversion::VersionTranslator vt;
+  boost::optional<openstudio::model::Model> model = vt.loadModel(modelPath);
+  ASSERT_TRUE(model);
+
+  std::vector<Space> spaces = model->getModelObjects<Space>();
```
jmarrec, line 812:
```suggestion
  std::vector<Space> spaces = model->getConcreteModelObjects<Space>();
```

src/model/test/Model_GTest.cpp
```diff
@@ -790,3 +792,57 @@ TEST_F(ModelFixture, Ensure_Name_Unicity_SpaceAndSpaceGroupNames) {
...
+  intersectSurfaces(spaces);
+
+  matchSurfaces(spaces);
+  std::vector<Surface> surfacesAfter = model->getModelObjects<Surface>();
```
jmarrec, line 816:
```suggestion
  std::vector<Surface> surfacesAfter = model->getConcreteModelObjects<Surface>();
```

src/model/test/Model_GTest.cpp
```diff
@@ -790,3 +792,57 @@ TEST_F(ModelFixture, Ensure_Name_Unicity_SpaceAndSpaceGroupNames) {
...
+  for (auto surface : surfacesAfter) {
```
jmarrec, line 817:
```suggestion
  for (const auto& surface : surfacesAfter) {
```

src/model/test/Model_GTest.cpp
```diff
@@ -790,3 +792,57 @@ TEST_F(ModelFixture, Ensure_Name_Unicity_SpaceAndSpaceGroupNames) {
...
+
+    std::string name = surface.name().value();
+    OptionalSurface otherSurface;
+    if (name == "Surface 4") {
+      otherSurface = surface.adjacentSurface();
+      ASSERT_TRUE(otherSurface);
+      ASSERT_EQ(otherSurface->name().value(), "Surface 8");
```
jmarrec, line 824:
```suggestion
      EXPECT_EQ(otherSurface->nameString(), "Surface 8");
```

Same for all similar below.

* EXPECT_EQ is enough (only use ASSERT_EQ when not passing the condition would mean a segfault, such as prior to dereferencing an optional or somethng)
* nameString() is more readable than name().get() or name().value(), though both are correct.

src/model/test/Model_GTest.cpp
```diff
@@ -790,3 +792,57 @@ TEST_F(ModelFixture, Ensure_Name_Unicity_SpaceAndSpaceGroupNames) {
...
+      otherSurface = surface.adjacentSurface();
+      ASSERT_TRUE(otherSurface);
+      ASSERT_EQ(otherSurface->name().value(), "Surface 26");
+    } else if (name == "Surface 28") {
+      otherSurface = surface.adjacentSurface();
+      ASSERT_TRUE(otherSurface);
+      ASSERT_EQ(otherSurface->name().value(), "Surface 32");
+    }
+
+    if (otherSurface) LOG(Info, "Surface " << surface.name().value() << " is paired with " << otherSurface->name().value());
```
jmarrec, line 843:
```suggestion
    // if (otherSurface) {
    //   LOG(Info, "Surface " << surface.name().value() << " is paired with " << otherSurface->name().value());
    // }
```

src/model/test/Model_GTest.cpp
```diff
@@ -790,3 +792,57 @@ TEST_F(ModelFixture, Ensure_Name_Unicity_SpaceAndSpaceGroupNames) {
...
+  }
+
+  modelPath = resourcesPath() / "model" / toPath("offset_tests_matched.osm");
+  model->save(modelPath, true);
```
jmarrec, line 847:
```suggestion
  // modelPath = resourcesPath() / "model" / toPath("offset_tests_matched.osm");
  // model->save(modelPath, true);
```

src/model/test/Space_GTest.cpp
```diff
@@ -1726,6 +1727,7 @@ TEST_F(ModelFixture, Space_intersectSurfaces_degenerate2) {
   boost::optional<Space> bottomCore = Space::fromFloorPrint(vertices, 3, m);
   ASSERT_TRUE(bottomCore);
   bottomCore->setZOrigin(0);
+  bottomCore->setBuildingStory(bottomStory);
```
jmarrec, line 1730: Could you explain why you are modifying all these tests without doing extra checks please?

src/utilities/geometry/Intersection.cpp
```diff
@@ -105,7 +110,7 @@ std::vector<BoostPolygon> removeSpikesEx(const BoostPolygon& polygon) {
         for (unsigned j = 0; j < polygon.outer().size(); j++) {
           Point3d p1(polygon.outer()[j].x(), polygon.outer()[j].y(), 0);
           // Two points are within tolerance set the result to the original input point
-          if (getDistance(point3d, p1) <= 0.05) {
+          if (getDistance(point3d, p1) <= amount) {
```
jmarrec, line 113: This is existing code, but I wonder if we shouldn't use a better variable name than `amount`

jmarrec (changes requested):
Some minor changes, otherwise everything looks good!

src/model/test/Model_GTest.cpp
ggartside: What is the difference between the two calls? Should I always use getConcreteModelObjects or are the occasions where getModelObjects is better?

tijcolem (approved):
@jmarrec Thanks for reviewing. Just running the windows CI again and if that passed I'll merge.

PR #4576, opened 2022-04-26
#Fix 4544 - CoilCoolingDX::clone crashes when called with another model

jmarrec:
Pull request overview
---------------------

- #Fix 4544
- CoilCoolingDX::clone crashes when called with another model, because we have an issue where a component has a resourceobject which has a resource object, AND everyone lists them in the children() method.
- I went around in circles trying to make it work and retain it in the children method, but I decided not to, because really they shouldn't be listed

I don't think this qualifies as an "APIChange" because the API didn't change. It'll require downstream users to update their code (OSapp at least)

### Pull Request Author

 - [x) Model API methods are tested (in `src/model/test`)

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/model/CoilCoolingDX.cpp
+++ b/src/model/CoilCoolingDX.cpp
@@ -121,7 +121,7 @@ namespace model {
       std::vector<ModelObject> result;
 
       // This is a ResourceObject, so it shouldn't really be a child except for OS App / IG
-      result.push_back(performanceObject());
+      // result.push_back(performanceObject());
 
       std::vector<AirflowNetworkEquivalentDuct> myAFNItems =
         getObject<ModelObject>().getModelObjectSources<AirflowNetworkEquivalentDuct>(AirflowNetworkEquivalentDuct::iddObjectType());
--- a/src/model/CoilCoolingDXCurveFitOperatingMode.cpp
+++ b/src/model/CoilCoolingDXCurveFitOperatingMode.cpp
@@ -85,8 +85,8 @@ namespace model {
 
     std::vector<ModelObject> CoilCoolingDXCurveFitOperatingMode_Impl::children() const {
       // These are ResourceObjects, so they shouldn't really be children except for OS App / IG
-      std::vector<ModelObject> result = subsetCastVector<ModelObject>(speeds());
-
+      // std::vector<ModelObject> result = subsetCastVector<ModelObject>(speeds());
+      std::vector<ModelObject> result;
       return result;
     }
 
--- a/src/model/CoilCoolingDXCurveFitPerformance.cpp
+++ b/src/model/CoilCoolingDXCurveFitPerformance.cpp
@@ -98,13 +98,13 @@ namespace model {
       std::vector<ModelObject> result;
 
       // These are ResourceObjects, so they shouldn't really be children except for OS App / IG
-      result.push_back(baseOperatingMode());
-      if (auto _mode = alternativeOperatingMode1()) {
-        result.push_back(_mode.get());
-      }
-      if (auto _mode = alternativeOperatingMode2()) {
-        result.push_back(_mode.get());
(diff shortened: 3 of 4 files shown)


src/model/CoilCoolingDXCurveFitOperatingMode.cpp
```diff
@@ -85,8 +85,8 @@ namespace model {
 
     std::vector<ModelObject> CoilCoolingDXCurveFitOperatingMode_Impl::children() const {
       // These are ResourceObjects, so they shouldn't really be children except for OS App / IG
-      std::vector<ModelObject> result = subsetCastVector<ModelObject>(speeds());
-
+      // std::vector<ModelObject> result = subsetCastVector<ModelObject>(speeds());
```
kbenne, line 88: Why not just remove this line instead of a comment?

src/model/CoilCoolingDX.cpp
```diff
@@ -134,7 +134,7 @@ namespace model {
       std::vector<ModelObject> result;
 
       // This is a ResourceObject, so it shouldn't really be a child except for OS App / IG
```
kbenne, line 136: What is the strategy over in the App if performance objects are no longer children? Is there a corresponding change to App to specialize the IG for this type? Or maybe instead it is just ok to not show the performance data? I'm just asking out of curiousity. I agree that the main impact of this PR pertains to the App's inspector.

kbenne (approved):

src/model/CoilCoolingDXCurveFitPerformance.cpp
```diff
@@ -99,13 +99,13 @@ namespace model {
       std::vector<ModelObject> result;
 
       // These are ResourceObjects, so they shouldn't really be children except for OS App / IG
-      result.push_back(baseOperatingMode());
```
wenyikuang, line 102: I rebased the change(s) onto my codebase and rebuild and cause the failure of serveral tests.

```
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] ModelFixture.CoilCoolingDX_clone
[  FAILED  ] ModelFixture.CoilCoolingDXCurveFitPerformance_clone
[  FAILED  ] ModelFixture.CoilCoolingDXCurveFitOperatingMode_clone
```

Do you mind review the tests?

jmarrec: Yeah, I'm still not sure how to fix it properly. I need to revisit this at some point.

wenyikuang: I deep dived into the unittests and find the issue is happening in 


```
  EXPECT_EQ(0u, model.getConcreteModelObjects<CoilCoolingDXCurveFitOperatingMode>().size());
```

Which the `model.getConcreteModelObjects<CoilCoolingDXCurveFitOperatingMode>().size()`'s value is 1.

jmarrec: I have access to Jenkins FYI, so I have access to the failed runs to see the failures.

The solution to having both clone to same model and clone to another is not trivial though. I'll eventually get back to this PR, it's a minor bug with a rather *expensive* solution: one that requires reimplementing or overriding the internals of ResourceObject and ParentObject (getRecursiveResources / getRecursiveChildren) and how things are added at Workspace level (addObjects). That portion of the sdk is complicated and finicky.
I've converted the PR to draft in the meantime for clarity.

jmarrec: If someone is burning to fix this though please go ahead. I think I have a jupyter notebook somewhere where I've rewritten the getRecursiveXXX stuff to be able to test changes, I could try to locate it.

Issue #127, opened 2013-07-19
Fix volume calculation to match EnergyPlus (Bugzilla #730)


axelstudios:
On 2012-06-08 14:19:09, @DavidGoldwasser wrote:

> Just wanted to file this before I forgot about it. Current method will have problem with spaces that have sloped walls or ceilings.

On 2012-09-09 23:24:43, @DavidGoldwasser wrote:

> I added BadSimResults but that would be an edge case where walls or roofs are not vertical.

On 2012-12-17 13:42:44, @DavidGoldwasser wrote:

> This was inadvertently under C#

Keywords: BadSimResults

DavidGoldwasser:
@macumber I'll keep this open, but I know it is more of a feature than a bug, but it can be important for proper calculation of some internal load values on forward translation of complex geometry.

agonzalezesteve:
I propose using the divergence theorem to compute the space volume, instead of multiplying the space area by a "mean" space height.
![image](https://user-images.githubusercontent.com/38532229/39042198-89e2f86e-448a-11e8-9afb-04f47bf9f674.png)
Then, using that every surface belongs to a plane, whose normal is parallel to its coefficients a, b and c;
![image](https://user-images.githubusercontent.com/38532229/39042288-b9944770-448a-11e8-9fa3-f9782df3947a.png)
only a constant integral is needed in every surface of the space.
![image](https://user-images.githubusercontent.com/38532229/39042362-e27e2034-448a-11e8-8453-52c764eab314.png)
Then, the volume integral reduces to
![image](https://user-images.githubusercontent.com/38532229/39042456-2891eb00-448b-11e8-8aaf-37661a4adaf0.png)
where d_i is the d coefficent of the plane where the i-th surface belongs and A_i is the area of the i-th surface.

I've tested it using this ruby code.
![image](https://user-images.githubusercontent.com/38532229/39042540-5e6ecc34-448b-11e8-84aa-fb9d3adf4804.png)

eringold:
@agonzalezesteve thank you, this is awesome. 

I just tested the following single space geometry: 
![image](https://user-images.githubusercontent.com/10799524/39071790-9a1af36a-44a5-11e8-9c28-0fb74ff1fad2.png)

By hand, I calculate the volume at 17.6642 m^3 (measuring from sketchup in ft/in and converting ft^3 to m^3 in excel)
The default `space.volume` method for this space returns 19.0972 m^3
Your method returns 17.5825 m^3.
[...]

Issue #4561, opened 2022-04-04
gbxml export does not transfer thermalZone Volume


wassimj:
Issue overview
--------------

Space Volume calculations in openstudio are not accurate for complex shapes (https://github.com/NREL/OpenStudio/issues/127). As a workaround one can set osThermalZone.setVolume. While this fixes the issue in the .OSM file, the gbxml forward translator does not carry over that value and uses space.volume which can have wrong value. It should use space.thermalzone.volume instead. Or offer the use a method to set the volume for a space (osSpace.setVolume does NOT exist)

## Current Behavior

For complex shapes, the volume of the space is wrong

## Expected Behavior

One or all things should happen:
1. osSpace should calculate the volume in a more accurate fashion. Address https://github.com/NREL/OpenStudio/issues/127
2. osSpace should offer a osSpace.setVolume method
3. If the above is not fixed, The gbxml Forward translator should use the saved value in the thermalZone rather than in space.volume

## Steps to Reproduce
1. Open the attached SpeckleBuilding.zip file and extract the OSM and GBXML files
2. Using a robust geometric method, compute the volumes of Space "8" and Space "9"
3. Compare the Volumes of Space "8" and Space "9" in the gbxml file to your calculations
4. Export the OSM file to gbxml and observe the volume of Space "8" and Space "9" compare to the volume set for their thermal zones.

## Possible Solution

One or allthings should happen:
1. osSpace should calculate the volume in a more accurate fashion. Address https://github.com/NREL/OpenStudio/issues/127
2. osSpace should offer a osSpace.setVolume method
3. If the above is not fixed, The gbxml Forward translator should use the saved value in the thermalZone rather than in space.volume
### Environment
[SpeckleBuilding.zip](https://github.com/NREL/OpenStudio/files/8407059/SpeckleBuilding.zip)

Some additional details about your environment for this issue (if relevant):
 - Platform (Operating system, version):
 - Version of OpenStudio (if using an intermediate build, include SHA): Application: 1.3.0  SDK: 3.3.0

### Context

We are trying to output a robust and dependable gbxml from openstudio for energy analysis and a wrong volume can cause problems down the line.

macumber:
I think that points 1 and 2 should be addressed.  The OS volume calculations should match the E+ calculations and now that E+ has a space object I think it will make sense to be able to set the space volume directly if you know it more accurately than OS/E+.

macumber:
[...]

PR #4579, opened 2022-04-28
Add Volume field to Space

joseph-robertson:
Pull request overview
---------------------

 - Addresses #4561.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


resources/model/OpenStudio.idd
```diff
@@ -5298,9 +5298,12 @@ OS:Space,
   A9, \field Design Specification Outdoor Air Object Name
        \type object-list
        \object-list DesignSpecificationOutdoorAirNames
-  A10; \field Building Unit Name
+  A10, \field Building Unit Name
        \type object-list
        \object-list BuildingUnitNames
+  N5; \field Volume
+       \type real
+       \units m3
```
jmarrec, line 5306:
```suggestion
  N5; \field Volume
       \type real
       \units m3
       \autocalculatable
       \default autocalculate
```

src/model/Space.hpp
```diff
@@ -280,6 +284,8 @@ namespace model {
     /// Does not include space multiplier in calculation.
     double volume() const;
 
+    bool isVolumeAutocalculated() const;
```
jmarrec, line 287:
```suggestion
    bool isVolumeDefaulted() const;

    bool isVolumeAutocalculated() const;
```

src/model/Space.hpp
```diff
@@ -112,6 +112,10 @@ namespace model {
 
     void resetPartofTotalFloorArea();
 
+    bool setVolume(double volume);
+
+    void resetVolume();
+
```
jmarrec, line 118:
```suggestion
    bool setVolume(double volume);

    void resetVolume();

    void autocalculateVolume();

```

jmarrec (changes requested):
Make the field autocalculatable (cf ThermalZone).

There might be other things to consider:
* ThermalZone::volume() should maybe check for its Space(s)'s volume if harcoded?
* How do we deal with the case where the user tries to harcode both the Space(s) and ThermalZone volumes and they don't match? eg if you have a single space inside a zone and put the space volume to be 200 m3 and the zone to be 100 m3?
* Probably want to Forward translate it

jmarrec:
ThermalZone::airVolume is the one that calls into Space::volume(). The ThermalZone::volume() itself just returns the hardcoded one if any. This is existing behavior, so I'm ok with it.

(FYI, Note that https://github.com/NREL/OpenStudio/pull/4592 will fix the issue #127 that prompted the issue #4561 that this fixes (as a mix of workaround/flexibility))

PR #4592, opened 2022-05-23
Fix#127 - Fix volume calculation to match EnergyPlus

### LINKED ISSUE #127: Fix volume calculation to match EnergyPlus (Bugzilla #730)

jmarrec:
Pull request overview
---------------------

- Fixes #127 
- Add a Polyhedron class in utilities/geometry, which reimplements the same logic that E+ does, with two main methods:
    - `isEnclosedVolume()` which check that each edge/segment is used twice exactly, potentially splitting up existing edges (this is similar to what intersectSurfaces would do, but does not reuse that portion)
    - calcPolyhedronVolume, which assuming your volume is indeed enclosed, will return the appropriate volume.

Space::volume is enhanced to do:
* If isEnclosedVolume() => calcPolyhedronVolume
* Otherwise, keep existing behavior (which is take an average of the floor and ceiling heights, multiply by floor area)
    * E+ in fact has a few more potential alternatives, by checking if a floor and ceiling are the same surface or not, if walls are vertical etc. I did not seem it terribly important to also implements those, as the goal here is more to correctly calculate **enclosed** volumes (volumes are meant to be enclosed really)

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


jmarrec:
the windows build error makes no sense.

`Unable to open file D:\jenkins\openstudio\develop\build\src\utilities\csharp_OpenStudioUtilitiesData_wrap.cxx: Permission denied`

tijcolem (approved):
This is great to have this issue resolved! many years.

PR #4571, opened 2022-04-19
Issue 4387 gl tf extras

Brijendra21:
Addition of Metadata and Userdata in the translated glTF 
---------------------

 - Fixes #4387 

### Pull Request Author

 - [ ] Model API Changes / Additions
 - [ ] Any new or modified fields have been implemented in the EnergyPlus ForwardTranslator (and ReverseTranslator as appropriate)
 - [ ] Model API methods are tested (in `src/model/test`)
 - [ ] EnergyPlus ForwardTranslator Tests (in `src/energyplus/Test`)
 - [ ] If a new object or method, added a test in NREL/OpenStudio-resources: Add Link
 - [ ] If needed, added VersionTranslation rules for the objects (`src/osversion/VersionTranslator.cpp`)
 - [ ] Verified that C# bindings built fine on Windows, partial classes used as needed, etc.
 - [ ] All new and existing tests passes
 - [ ] If methods have been deprecated, update rest of code to use the new methods

**Labels:**

 - [ ] If change to an IDD file, add the label `IDDChange`
 - [ ] If breaking existing API, add the label `APIChange`
 - [ ] If deemed ready, add label `Pull Request - Ready for CI` so that CI builds your PR

### Review Checklist

This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] Code Style, strip trailing whitespace, etc.
 - [ ] All related changes have been implemented: model changes, model tests, FT changes, FT tests, VersionTranslation, OS App
 - [ ] Labeling is ok
 - [ ] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified

PR #4597, opened 2022-05-24
Hotfix GLTF extras broke running tests in Debug

jmarrec:
Pull request overview
---------------------

GLTF extras broke (#4571) running ctest in Debug (400 tests segfaulting)

This is an absolute minimal fix to have the tests passing, but I think we should refactor: 
* avoid all static storage objects
* change the data layout and API in the classes that are in GltfForwardTranslator
    * just use `T` (eg double) instead of std::tuple<std::string, tinygltf::Value(T)> 
    * provide a protected toGltfValue that will create the std::map<std::string, tinygltf::Value>
* Move those to separate files (like the wrappers are)
* Remove the wrappers (which are superflous now)

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


tijcolem:
@jmarrec Thanks for the quick fix on this! I looked at the CI runner for code coverage and it's been set ignore the ctest failures as it was only looking at the code coverage. I changed this so it'll fail now if any of the debug tests fail. It looks like a couple of debug tests routinely fail so I added these to filter list.  Going forward we should be able to catch these.

PR #4621, opened 2022-06-29
#4551 - Add minimum/maximum values to numeric OSArguments and use it in validateUserArgument

### LINKED ISSUE #4551: Add minimum/maximum values to numeric OSArguments
Enhancement Request
--------------
Extend OSArgument (integer and double types) to allow setting (and getting) minimum/maximum values, and have these displayed in the XML/JSON when arguments are computed.
e.g. I would like to do something like this:
```
    demand_target = OpenStudio::Measure::OSArgument.makeDoubleArgument('demand_target', true)
    demand_target.setDisplayName("Target Demand Limit")
    demand_target.setDescription("Demand response controls will adjust building loads to attempt to maintain this demand level (kW).")
    demand_target.setDefaultValue(100)
    demand_target.setMinValue(0)
    args << demand_target
```
and when running `openstudio measure --compute_arguments` get:
```
"arguments": [
	{
		"name": "demand_target",
		"display_name": "Target Demand Limit",
		"description": "Demand response controls will adjust building loads to attempt to maintain this demand level (kW).",
		"type": "Double",
		"required": true,
		"model_dependent": false,
		"default_value": 100.0,
		"min_value": 0.0
	}
]
```
Note: right now `setMinValue` and `setMaxValue` both exist, but don't seem to do anything useful. This would allow external interfaces to use the JSON to validate measure input before running the measure, and for measure validation within the run method. Could also extend [`runner.validateUserArguments`](https://github.com/NREL/OpenStudio/blob/49f77e7cb61bde928b25f74f8acdbc07a4110d5e/src/measure/OSRunner.cpp#L395) to do the same.

jmarrec:
Pull request overview
---------------------

 - Fixes #4551

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/cli/measure_manager.rb
+++ b/src/cli/measure_manager.rb
@@ -423,6 +423,8 @@ class MeasureManager
         when 'Integer'
           arg[:units] = argument.units.get if argument.units.is_initialized
           arg[:default_value] = argument.defaultValue.get.to_i if argument.defaultValue.is_initialized
+          arg[:min_value] = argument.minValue.get.to_i if argument.minValue.is_initialized
+          arg[:max_value] = argument.maxValue.get.to_i if argument.maxValue.is_initialized
 
         when 'String'
           arg[:default_value] = argument.defaultValue.get if argument.defaultValue.is_initialized
@@ -468,7 +470,17 @@ class MeasureManager
       elsif type == "Double".to_OSArgumentType
         arg[:units] = argument.units.get if argument.units.is_initialized
         arg[:default_value] = argument.defaultValueAsDouble if argument.hasDefaultValue
-
+        if argument.hasDomain
+          min, max = argument.domainAsDouble
+          # I'm a bit wary of rounding issues... I think 1e308 instead of
+          # Float::MAX (1.7976931348623157e+308) is fine for our applications...
+          if min > -1e308 # Float::MIN (technically Float::MIN is 2.2250738585072014e-308)
+            arg[:min_value] = min
+          end
+          if max < 1e308 # Float::MAX
+            arg[:max_value] = max
+          end
+        end
       elsif type == "Quantity".to_OSArgumentType
         arg[:units] = argument.units.get if argument.units.is_initialized
         arg[:default_value] = argument.defaultValueAsQuantity.value if argument.hasDefaultValue
@@ -476,16 +488,24 @@ class MeasureManager
       elsif type == "Integer".to_OSArgumentType
         arg[:units] = argument.units.get if argument.units.is_initialized
         arg[:default_value] = argument.defaultValueAsInteger if argument.hasDefaultValue
-
+        if argument.hasDomain
+          min, max = argument.domainAsInteger
+          if min != -2147483648
+            arg[:min_value] = min
+          end
+          if max != 2147483648
+            arg[:max_value] = max
+          end
+        end
       elsif type == "String".to_OSArgumentType
         arg[:default_value] = argument.defaultValueAsString if argument.hasDefaultValue
 
       elsif type == "Choice".to_OSArgumentType
         arg[:default_value] = argument.defaultValueAsString if argument.hasDefaultValue
-          arg[:choice_values] = []
-          argument.choiceValues.each {|value| arg[:choice_values] << value}
-          arg[:choice_display_names] = []
-          argument.choiceValueDisplayNames.each {|value| arg[:choice_display_names] << value}
+        arg[:choice_values] = []
+        argument.choiceValues.each {|value| arg[:choice_values] << value}
+        arg[:choice_display_names] = []
+        argument.choiceValueDisplayNames.each {|value| arg[:choice_display_names] << value}
 
       elsif type == "Path".to_OSArgumentType
         arg[:default_value] = argument.defaultValueAsPath.to_s if argument.hasDefaultValue
--- a/src/cli/measure_manager.rb
+++ b/src/cli/measure_manager.rb
@@ -474,7 +474,7 @@ class MeasureManager
           min, max = argument.domainAsDouble
           # I'm a bit wary of rounding issues... I think 1e308 instead of
           # Float::MAX (1.7976931348623157e+308) is fine for our applications...
-          if min > -1e308 # Float::MIN (technically Float::MIN is 2.2250738585072014e-308)
+          if min > -1e308 # -Float::MAX
             arg[:min_value] = min
           end
           if max < 1e308 # Float::MAX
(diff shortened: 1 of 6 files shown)


src/cli/measure_manager.rb
```diff
@@ -86,7 +86,7 @@ def initialize(logger=nil)
     @measures = {} # measure_dir => BCLMeasure
     @measure_info = {} # measure_dir => {osm_path => RubyUserScriptInfo}
 
-    eval(OpenStudio::Ruleset::infoExtractorRubyFunction)
+    eval(OpenStudio::Measure::infoExtractorRubyFunction)
```
jmarrec, line 89: Modernize our script to not use the deprecated namespace

src/cli/measure_manager.rb
```diff
@@ -468,24 +470,42 @@ def get_arguments_from_measure_info(measure_info)
...
+          min, max = argument.domainAsDouble
+          # I'm a bit wary of rounding issues... I think 1e308 instead of
+          # Float::MAX (1.7976931348623157e+308) is fine for our applications...
+          if min > -1e308 # -Float::MAX
+            arg[:min_value] = min
+          end
+          if max < 1e308 # Float::MAX
+            arg[:max_value] = max
+          end
+        end
```
jmarrec, line 483: `openstudio measure --compute-arguments` will now output the min/max.

src/measure/OSArgument.cpp
```diff
@@ -763,7 +763,7 @@ namespace measure {
       return false;
     }
 
-    double minValue = std::numeric_limits<double>::min();
+    double minValue = std::numeric_limits<double>::lowest();
```
jmarrec, line 766: Using `min()` here means using the smallest positive value. NOT what was expected.

Use `lowest()` which equals `-max()`

cf https://godbolt.org/z/K1aq87Yjh

```
std::numeric_limits<double>::max()   =  1.7976931348623157e+308
std::numeric_limits<double>::min()   =  2.2250738585072014e-308  # POSITIVE minimum finite value
std::numeric_limits<double>::lowest()= -1.7976931348623157e+308
```

src/measure/OSMeasureInfoGetter.cpp
```diff
@@ -161,7 +161,27 @@ namespace measure {
...
+          auto domain = argument.domainAsDouble();
+          auto& low = domain.front();
+          auto& high = domain.back();
+          if (low > std::numeric_limits<double>::lowest()) {
+            minValue = std::to_string(low);
+          }
+          if (high < std::numeric_limits<double>::max()) {
+            maxValue = std::to_string(high);
+          }
+        }
```
jmarrec, line 184: I noticed this TODO so I also addressed it. It will get passes to BCLMeasureArgument => the `measure.xml` now will also include the min/max values


image

jmarrec:
@DavidGoldwasser Could you take a look at this one please? And confirm adding the min_value / max_value to the measure.xml is ok with you?

---

One thing I also wanted to do was to replace `JSON.generate` by `JSON.pretty_generate` for --compute-arguments and --update (not update_all), so the resulting JSON is actually readable by a human instead of being a condensed one-liner.

https://github.com/NREL/OpenStudio/blob/0e7dda7436acd575285c7894b43d1951a5309682/src/cli/openstudio_cli.rb#L1483

https://github.com/NREL/OpenStudio/blob/0e7dda7436acd575285c7894b43d1951a5309682/src/cli/openstudio_cli.rb#L1543

Thoughts anyone?

src/cli/measure_manager.rb
```diff
@@ -468,24 +470,42 @@ def get_arguments_from_measure_info(measure_info)
       elsif type == "Double".to_OSArgumentType
         arg[:units] = argument.units.get if argument.units.is_initialized
         arg[:default_value] = argument.defaultValueAsDouble if argument.hasDefaultValue
-
+        if argument.hasDomain
```
joseph-robertson, line 473: Does it still have domain if either min or max are set but not both?

src/cli/measure_manager.rb
```diff
@@ -468,24 +470,42 @@ def get_arguments_from_measure_info(measure_info)
...
         arg[:units] = argument.units.get if argument.units.is_initialized
         arg[:default_value] = argument.defaultValueAsQuantity.value if argument.hasDefaultValue
 
       elsif type == "Integer".to_OSArgumentType
         arg[:units] = argument.units.get if argument.units.is_initialized
         arg[:default_value] = argument.defaultValueAsInteger if argument.hasDefaultValue
-
+        if argument.hasDomain
+          min, max = argument.domainAsInteger
+          if min != -2147483648
```
joseph-robertson, line 493: I'm confused what the purpose of this is? Same for above with 1e308.

src/measure/test/OSMeasure_GTest.cpp
```diff
@@ -389,3 +392,92 @@ TEST_F(MeasureFixture, RegisterValueNames) {
...
+
+  virtual std::vector<OSArgument> arguments(const Model& model) const override {
+    std::vector<OSArgument> result;
+
+    OSArgument arg = OSArgument::makeDoubleArgument("double_arg", true);
+    arg.setMaxValue(10.0);
+    result.push_back(arg);
+
+    arg = OSArgument::makeIntegerArgument("int_arg", true);
+    arg.setMinValue(0);
```
joseph-robertson, line 411: Test with a negative?

src/cli/measure_manager.rb
jmarrec: Domain is set as `[std::numeric_limits<double>::lowest(), max]` if max only is provided, and if min only then `[min, std::numeric_limits<double>::max()]`

src/cli/measure_manager.rb
jmarrec: Same answer as above... When only mix or max is provided, one of the domain boundaries is going to be `std::numeric_limits<T>::lowest()` or `std::numeric_limits<T>::max()`.
In the case of an integer that's +- 2147483648, and for a double that's 1.7976931348623157e+308

joseph-robertson: So if you did indeed set a min less than `lowest()` or a max greater than `max()`, wouldn't it not set/write those values?

jmarrec: I don't see how you could do that though. It'd be an overflow error, but mostly you just can't do it,  bindings included.

cf:

```
$ irb
2.7.2 :002 > 1e308
 => 1.0e+308 
2.7.2 :003 > 1e309
 => Infinity 

```

joseph-robertson (approved):

Issue #4653, opened 2022-08-23
Python Measures: Introduce "PythonEngine" library


kbenne:
Similar to `RubyEngine`, add a library with an API that derives from `ScriptEngine`. Look at the OpenStudioLabs project for a reference. It would be nice to have consistency between the ruby cmake target names, and the python targets.

jmarrec:
Fixed via #4688

PR #4674, opened 2022-09-09
4641 ruby shared lib

kbenne:
Pull request overview
---------------------

- Fix #4641

This PR moves the Ruby related parts of OpenStudio into a shared library that is dynamically loaded by the CLI.

There is significant project cleanup and reorganization. 

## Pull Request Author

 - [ ] Everything builds
 - [ ] Ruby bindings can be imported in ruby
     - [ ] Windows
     - [ ] Linux
     - [ ] Mac
- [ ] CLI works on all platforms
     - [ ] Windows
     - [ ] Linux
     - [ ] Mac
 - [ ] NREL/OpenStudio-resources testing
 - [ ] Verified that C# bindings built fine on Windows, partial classes used as needed, etc.
 - [ ] All new and existing tests passes
 - [ ] If methods have been deprecated, update rest of code to use the new methods

**Labels:**

 - [x] If change to an IDD file, add the label `IDDChange`
 - [x] If breaking existing API, add the label `APIChange`
 - [x] If deemed ready, add label `Pull Request - Ready for CI` so that CI builds your PR

### Review Checklist

This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] Code Style, strip trailing whitespace, etc.
 - [ ] All related changes have been implemented: model changes, model tests, FT changes, FT tests, VersionTranslation, OS App
 - [x] Labeling is ok
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified

PR #4688, opened 2022-09-20
4653 python engine

kbenne:
Pull request overview
---------------------

Address #4653

This PR introduces a PythonEngine class that can be used to access Python from the OpenStudio cli.

### Pull Request Author

### Review Checklist

 - [ ] Labeling is ok


kbenne:
This should happen after the [Ruby PR](https://github.com/NREL/OpenStudio/pull/4674)

python/engine/PythonEngine.cpp
```diff
@@ -0,0 +1,170 @@
...
+namespace openstudio {
+
+static constexpr auto importScript = R"(
+import sys
+sys.path.insert(0, '{}')
+import {}
+)";
+
+PythonEngine::PythonEngine(int argc, char *argv[])
+    : ScriptEngine(argc, argv), program(Py_DecodeLocale("python3.8", nullptr)) {
```
jmarrec, line 29: Why use python3.8 and not argv[0]? If we start making python 3.8 **mandatory** then I expect to see a change in the find_package(Python in CMakeLists.txt.

In fact, I do expect to see the BUILD_PYTHON_BINDINGS option be removed or something if this becomes part of the regular CLI

jmarrec: (Or if we allow a minimal build, some kind of cmake-foo + ifdef SCRIPT_ENGINE_PYTHON or something)

python/engine/PythonEngine.cpp
```diff
@@ -0,0 +1,170 @@
...
+#endif
+
+
+namespace openstudio {
+
+static constexpr auto importScript = R"(
+import sys
+sys.path.insert(0, '{}')
+import {}
+)";
```
jmarrec, line 26: This looks weird to me to combine this + exec. It probably works fine.

But I was thinking it would be using the C API

```
PyObject* sys = PyImport_ImportModule( "sys" );
PyObject* sys_path = PyObject_GetAttrString( sys, "path" );
PyObject* folder_path = PyUnicode_FromString( "<insert folder path>" );
PyList_Append( sys_path, folder_path );
```

python/engine/PythonEngine.cpp
```diff
@@ -0,0 +1,170 @@
...
+    : ScriptEngine(argc, argv), program(Py_DecodeLocale("python3.8", nullptr)) {
+  Py_SetProgramName(program); // optional but recommended
+
+  Py_Initialize();
+  importOpenStudio();
+}
+
+PythonEngine::~PythonEngine() {
+  if (Py_FinalizeEx() < 0) {
+    exit(120);
```
jmarrec, line 38: Why 120? Because that's what python uses for an interpreter not terminating correctly.

python/engine/PythonEngine.cpp
```diff
@@ -0,0 +1,170 @@
...
+  }
+  PyMem_RawFree(program);
+}
+
+void PythonEngine::importOpenStudio() {
+  if (moduleIsRunningFromBuildDirectory()) {
+    const auto bindingsDir = getOpenStudioModuleDirectory();
+    exec(fmt::format(importScript, bindingsDir.string(), "openstudiodev"));
```
jmarrec, line 46: IS exec better than PyRun_SimpleString ?

python/engine/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  MODULE
+  PythonEngine.hpp
+  PythonEngine.cpp
+  ${CMAKE_CURRENT_BINARY_DIR}/SWIGPythonRuntime.hxx
+)
+
+file(
+  GENERATE OUTPUT $<TARGET_FILE_DIR:pythonengine>/openstudiodev.py
+  INPUT openstudiodev.py.in
+)
```
jmarrec, line 21: fancy. configure_file would have done though no? I'm just curious

CMakeLists.txt
```diff
@@ -916,40 +916,7 @@ if(BUILD_NODE_MODULES)
...
-      message(FATAL_ERROR "PYTHON_PIP_REPOSITORY must be one of ${PipRepositoryValues}")
-    endif()
-
-    set(PYTHON_PACKAGE_FOLDER "${PROJECT_BINARY_DIR}/Products/python_package")
-    file(MAKE_DIRECTORY "${PYTHON_PACKAGE_FOLDER}")
-    set(ALL_PYTHON_PACKAGE_TARGETS "") # global list of python bindings
-    set(ALL_PYTHON_PACKAGE_GENERATED_SRCS "") # global list of generated .py files for pyhton_package
-
-  endif()
+  include(python/SetupPython.cmake)
```
jmarrec, line 919: Yeah well if BUILD_PYTHON_BINDINGS is off, then what happens to python engine, module and the CLI using it? It'll break

jmarrec (commented):
Naive initial pass

python/engine/PythonEngine.cpp
kbenne: I think argv[0] might be brittle due to symbolic linking and what have you, but I'm not sure. 

But your larger point seems important to me, which is that this should be configured with as you say cmake-foo.

python/engine/PythonEngine.cpp
kbenne: Yeah I do feel a little guilty writing this in python. Would you feel better if I moved it over to C? 

I actually think PyConfig https://docs.python.org/3/c-api/init_config.html might be the most formal way to do this, but I looked at it and it seemed more complicated than I wanted.

python/engine/PythonEngine.cpp
kbenne: I think ScriptEngine::exec is good because it is a nice abstraction that is the same among scripting languages. Internally PyRun_SipleString would be ok.

python/engine/CMakeLists.txt
kbenne: configure_file does not allow generator expressions as far as I'm aware. The whole point of these shenanigans is to handle the configuration subdirectory in multi config builds. file(GENERATE ..) will create one file per config of the target build system. I think this approach is robust.

Full disclaimer I haven't tested on a multi config system yet, but my Windows build is running right now.

CMakeLists.txt
kbenne: I agree. The BUILD_PYTHON_BINDINGS option should be removed. I'll add that change to the PR.

python/engine/PythonEngine.cpp
jmarrec: I was saying Why 120? Then I looked it up and it made sense; No action needed

python/module/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+)
+
+
+if (BUILD_PYTHON_PIP_PACKAGE)
+
+  # configure_file(${CMAKE_CURRENT_SOURCE_DIR}/openstudio_dynamic.py ${PROJECT_BINARY_DIR}/Products/python/openstudio_dynamic.py COPYONLY)
+  configure_file(${CMAKE_CURRENT_SOURCE_DIR}/openstudio.py ${PYTHON_PACKAGE_FOLDER}/openstudio/__init__.py COPYONLY)
+
+  if(PYTHON_PIP_REPOSITORY STREQUAL "testpypi")
+    execute_process( COMMAND ${Python_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/find_pypi_tag.py" OUTPUT_VARIABLE PYPI_VERSION)
```
jmarrec, line 27: Path is wrong, find_pypi_tag.py is one level above currently.

kbenne: Haven't tested the pip package. The could be plenty of issues introduced by my rearrangement of the deck chairs.

jmarrec: Fixed this specific issue in c6fa66eed9bbe1f0b44cde8cce96fb4dbbfc76ce

kbenne: Awesome. Thanks!

resources/Examples/compact_osw/measures/IncreaseRoofRValuePython/measure.py
```diff
@@ -0,0 +1,282 @@
...
+        return args
+
+    def run(self,
+            model: openstudio.model.Model,
+            runner: openstudio.measure.OSRunner,
+            user_arguments: openstudio.measure.OSArgumentMap):
+        """
+        define what happens when the measure is run
+        """
+
```
jmarrec, line 46: Not sure if I need `super().run(model, runner, user_arguments)` here

CMakeLists.txt
kbenne: I just realized BUILD_RUBY_BINDINGS is still an option too. Should we remove that as well?

jmarrec: Well, the github CI workflows for python bindings and csharp do `-DBUILD_RUBY_BINDINGS=OFF -DBUILD_CLI=OFF` to speed up the build (and remove the need to configure the runner for the other bindings too)

kbenne:
@tijcolem This is the python measures PR to be aware of. I just changed the base branch to develop.

Issue #4776, opened 2022-12-21
Missing wall sub-surfaces and wall surface not subdivided correctly


ggartside:
Issue overview
--------------
This is an issue in both the ThreeJS reverse translator and the Floorspace reverse translator so that suggests the issue lies in code that is common to both import paths. 

## Current Behavior
One external wall is missing a window subsurface and is not subdivided as it should be as part of the wall surface is internal and part is external

ThreeJS....
![image](https://user-images.githubusercontent.com/21084412/208980456-c83814b5-6d40-4aea-b17e-a6ea88b2a420.png)

Floorspace
![image](https://user-images.githubusercontent.com/21084412/208980652-a29bed9b-fca3-4c12-a530-a785ef18de91.png)

## Expected Behavior
The wall should be subdivided and contain a window sub surface like this...
![image](https://user-images.githubusercontent.com/21084412/208980768-e992d4cb-5d1e-4747-bb74-77b5cbb9b612.png)

As can be seen in the original floorspace model
![image](https://user-images.githubusercontent.com/21084412/208981213-d4fb95c2-3a74-4bb7-a08f-8113646646eb.png)

## Steps to Reproduce

1.
2.
3.
4.

### Environment

Some additional details about your environment for this issue (if relevant):
 - Platform (Operating system, version):
 - Version of OpenStudio (if using an intermediate build, include SHA):

PR #4777, opened 2022-12-21
4766 different surface counts

### LINKED ISSUE #4766: Different number of wall surfaces between ThreeJS reverse translator and Floorspace reverse translatot
Issue overview
--------------
The Floorspace and ThreeJS reverse translators should produce identical models as far as surface count and vertex matching of surfaces goes. 

## Current Behavior
Using the sample file provided in #4222 floorplan_mcve.json Space 1-6 has a different count of internal wall surfaces
This difference appears to be an issue with surface alignment that differ between the two imports

## Expected Behavior
Surface count should be the same and surface matching vertices should produce no errors

## Steps to Reproduce

The source json data looks like this

image

There is an inline vertex indicated by the red dot
In the ThreeJS import the surfaces along this edge are aligned as follows

image

In the Floorspace import he surfaces along this edge do not align

image

It is possible that somewhere in the Floorspace import the inline vertex has been removed somehow resulting in combining two adjacent edges into one edge that is not parallel with the edge on space 1-12. The result is the surfaces on Space 1-6 and Space 1-12 are not considered to have equal planes and therefore are not intersected

For an idea of scale of what we are dealing with here, this is an overview of the section in question with the inline vertex identified by the red dot

image

As can be seen the inline vertex is very close to the corner vertex of Space 1-12 which may be the reason it has been removed, the distance between the two vertices is approximately 2" so that should be outside the tolerance of merging vertices 

image

## Possible Solution
Find out where the vertex is removed and do not remove it.

### Environment

ggartside:
Pull request overview
---------------------

 - Fixes #4766

To resolve an issue caused by removing inline vertices that altered the direction vector of the combined edges enough to make surface matching fail

This call was originally added to resolve the issue with Space 2 - 1 in the floorplan_school model where the vertices are not ordered correctly and the resulting surfaces are overlapping. Removing this call meant that the way wall sub-surfaces are made and associated with the walls had to be modified to allow for the situation that we are no longer able to match edge ids

This fix has also resulted in creating issue #4776 and #4783
 

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


src/utilities/geometry/Point3d.cpp
```diff
@@ -113,4 +113,25 @@ std::ostream& operator<<(std::ostream& os, const std::vector<Point3d>& pointVect
...
+      os << "\"Y\":" << points[j].y() << ",";
+      os << "\"Z\":" << points[j].z();
+      os << "},";
+    }
+    os << "],";
+  }
+  os << "]";
+
+  return os;
+}
```
jmarrec, line 135: * Range-based for loop
* Point3d has an ostream operator already, can't we just use that already? At least I find uppercase X, Y, Z quite weird

Proposed implementation:

```
std::ostream& operator<<(std::ostream& os, const std::vector<std::vector<Point3d>>& pointVectorVector) {
  os << "[\n";
  for (const auto& points: pointVectorVector) {
    os << "  [";
    for (const auto& point: points) {
      os << "\n    " << point << ", ";
    }
    os << "\n  ],\n";
  }
  os << "]";

  return os;
}
```

And alternative relying on `<fmt/ranges.h>` and `<fmt/ostream.h>`

```c++
std::ostream& operator<<(std::ostream& os, const std::vector<std::vector<Point3d>>& pointVectorVector) {
  os << fmt::format("{}", pointVectorVector);
  return os;
}
```

src/utilities/geometry/Test/Intersection_GTest.cpp
```diff
@@ -2349,33 +2349,39 @@ TEST_F(GeometryFixture, Offset) {
   poly1.push_back(Point3d(8, 10, 0));
   poly1.push_back(Point3d(11, 10, 0));
   poly1.push_back(Point3d(11, 7, 0));
+  Point3dVector poly2;
+  poly2.push_back(Point3d(7.5, 10.5, 0));
+  poly2.push_back(Point3d(11.5, 10.5, 0));
+  poly2.push_back(Point3d(11.5, 6.5, 0));
+  poly2.push_back(Point3d(7.5, 6.5, 0));
```
jmarrec, line 2356: An FYI but going forward you can do:

```
Point3dVector poly2 {
        {7.5, 10.5, 0.0},
        {11.5, 10.5, 0.0},
        {11.5, 6.5, 0.0},
        {7.5, 6.5, 0.0},
    };
```

This will avoid costly vector resize operations. (granted, this is a test here...).
The copy constructor will be called 7 times in the case of 4 push_backs, versus 4 times with initializer list.

The most efficient is this but it's annoying to type in a test

```
Point3dVector poly2;
poly2.reserve(4);
poly2.emplace_back(7.5, 10.5, 0);
poly2.emplace_back(11.5, 10.5, 0);
poly2.emplace_back(11.5, 6.5, 0);
poly2.emplace_back(7.5, 6.5, 0);
```

here's a testbed if you want to see for yourself: https://compiler-explorer.com/z/zf9EEr31j

src/utilities/geometry/Test/Intersection_GTest.cpp
```diff
@@ -2349,33 +2349,39 @@ TEST_F(GeometryFixture, Offset) {
...
 
   boost::optional<std::vector<Point3d>> result1 = openstudio::buffer(poly1, 0.5, 0.01);
   ASSERT_TRUE(result1);
+  Point3dVectorVector tmp;
+  tmp.push_back(poly1);
+  tmp.push_back(*result1);
```
jmarrec, line 2362: Same here: `Point3dVectorVector tmp{poly1, *result1}`;

jmarrec (changes requested):

src/model/Space.cpp
```diff
@@ -2314,18 +2314,16 @@ namespace model {
       Transformation transformation = this->transformation().inverse() * other.transformation();
 
       for (Surface surface : this->surfaces()) {
-
-        std::vector<Point3d> vertices = removeCollinear(surface.vertices());
-
+        if (surface.adjacentSurface()) continue;
```
jmarrec, line 2317: braces

src/model/FloorspaceReverseTranslator.cpp
```diff
@@ -563,34 +559,18 @@ namespace model {
...
-        wallVertices.push_back(Point3d(v1.x(), v1.y(), minZ));
-      }
+      Point3dVector wallVertices;
+      wallVertices.push_back(Point3d(p2.x(), p2.y(), maxZ));
+      wallVertices.push_back(Point3d(p1.x(), p1.y(), maxZ));
+      wallVertices.push_back(Point3d(p1.x(), p1.y(), minZ));
+      wallVertices.push_back(Point3d(p2.x(), p2.y(), minZ));
+      Point3dVector wallSegment;
+      wallSegment.push_back(p1);
+      wallSegment.push_back(p2);
```
jmarrec, line 573: this is production code here, so I would go with reserve + emplace_back

src/model/FloorspaceReverseTranslator.cpp
```diff
@@ -602,34 +582,43 @@ namespace model {
 
           // Create a window subsurface for every window that is on this edge
           for (const auto& window : m_currentFSStory->windows()) {
-            if (window.edge()->id() == edgeRef.edge().id()) {
-              createWindowSubsurface(window, surface, edgeRef, minZ, maxZ);
+            // Get window's x/y position which is the alpha based on its edge
+            // windows can have multiple alphas but we only need one
+            auto edge = *window.edge();
```
jmarrec, line 587: We're sure it's inintialized? (I realize this was existing code, the -> operator was called already before)

src/model/Space.cpp
```diff
@@ -2404,12 +2408,20 @@ namespace model {
...
               hasSubSurfaceMap[otherSurfaceHandle] = !otherSurface.subSurfaces().empty();
               hasAdjacentSurfaceMap[otherSurfaceHandle] = otherSurface.adjacentSurface().has_value();
             }
 
             if (hasSubSurfaceMap[otherSurfaceHandle] || hasAdjacentSurfaceMap[otherSurfaceHandle]) {
+
+              auto subSurfaces = otherSurface.subSurfaces();
+              for (auto subSurface : subSurfaces) {
+                std::string subSurfaceName = subSurface.nameString();
+              }
```
jmarrec, line 2424: 1. This does nothing so it's useless
2. If it did do something, this creates a copy for each iteration, use const auto& or auto&

src/model/Space.cpp
```diff
@@ -2404,12 +2408,20 @@ namespace model {
 
           for (Surface otherSurface : otherSurfaces) {
             std::string otherSurfaceHandle = toString(otherSurface.handle());
+            std::string otherSurfaceName = otherSurface.nameString();
+            std::string otherSurfaceType = otherSurface.surfaceType();
```
jmarrec, line 2412: this doesn't appear to be used anywhere

src/model/Space.cpp
```diff
@@ -2392,6 +2394,8 @@ namespace model {
 
         for (Surface surface : surfaces) {
           std::string surfaceHandle = toString(surface.handle());
+          std::string surfaceName = surface.nameString();
+          std::string surfaceType = surface.surfaceType();
```
jmarrec, line 2398: This doesn't appear to be used anywhere

src/model/Surface.cpp
```diff
@@ -1027,6 +1027,8 @@ namespace model {
       double tol = 0.01;       //  1 cm tolerance
       double areaTol = 0.001;  // 10 cm2 tolerance
 
+      bool extraLogging = false;
```
jmarrec, line 1030: Same idea. Should be a constexpr

src/model/Surface.cpp
```diff
@@ -1088,13 +1090,26 @@ namespace model {
       //std::reverse(otherFaceVertices.begin(), otherFaceVertices.end());
 
       //LOG(Info, "Trying intersection of '" << this->name().get() << "' with '" << otherSurface.name().get());
-
+      if (extraLogging) {
```
jmarrec, line 1093: if-constexpr

src/utilities/floorspace/FSModel.hpp
```diff
@@ -332,7 +332,7 @@ class UTILITIES_API FSFiller : public FSBase
   boost::optional<FSSpace> space() const;
   std::vector<double> alphas() const;
   // Calculates and returns the center vertex position of the window or door
-  Point3d centerVertex(double alpha);
+  Point3d centerVertex(double alpha) const;
```
jmarrec, line 335: :+1:

jmarrec (changes requested):

jmarrec:
at 5e06a4 you have failing tests. There are also conflicts.

The following tests FAILED:

	2524 - ModelFixture.FloorspaceReverseTranslator_Issue_4312 (Failed)
	2526 - ModelFixture.FloorspaceReverseTranslator_Issue_4488 (Failed)
	2530 - ModelFixture.FloorspaceReverseTranslator_Issue_4237 (Failed)

Errors while running CTest

resources/CMakeLists.txt
```diff
@@ -68,6 +68,13 @@ set(utilities_resources_src
   utilities/Geometry/surface_match_floorplan.json
   utilities/Geometry/threejs.json
   utilities/Geometry/window_floorplan.json
+  utilities/Geometry/issue-4222.json
+  utilities/Geometry/issue-4312.json
+  utilities/Geometry/issue-4203.json
+  utilities/Geometry/issue-4488.json
+  utilities/Geometry/issue-4670.json
+  utilities/Geometry/issue-4313.json
+  utilities/Geometry/floorplan_mcve.json
```
jmarrec, line 77: 4237 missing here

src/model/test/FloorspaceReverseTranslator_GTest.cpp
```diff
@@ -1389,3 +1406,480 @@ TEST_F(ModelFixture, FloorspaceReverseTranslator_FloorplanJS_Site_ClimateZones_4
...
+  }
+
+  CompareTwoModels(*model1, *model);
+}
+
+TEST_F(ModelFixture, FloorspaceReverseTranslator_Issue_4237) {
+
+  ThreeJSReverseTranslator rt;
+
+  openstudio::path p = resourcesPath() / toPath("utilities/Geometry/issue_4237.json");
```
jmarrec, line 1735: wrong name here

src/model/test/FloorspaceReverseTranslator_GTest.cpp
```diff
@@ -1389,3 +1406,480 @@ TEST_F(ModelFixture, FloorspaceReverseTranslator_FloorplanJS_Site_ClimateZones_4
...
+  model1->save(resourcesPath() / toPath("utilities/Geometry/issue_4237-direct.osm"), true);
+
+  CompareTwoModels(*model1, *model);
+}
+
+TEST_F(ModelFixture, FloorspaceReverseTranslator_Issue_4776) {
+
+  ThreeJSReverseTranslator rt;
+
+  openstudio::path p = resourcesPath() / toPath("utilities/Geometry/issue-4203.json");
```
jmarrec, line 1761: Is this normal the json and issue aren't matching?

jmarrec (changes requested):

resources/CMakeLists.txt
jmarrec: Done in 853811433efa6480d6b14f252d22ed3642b2e6b2

src/model/test/FloorspaceReverseTranslator_GTest.cpp
jmarrec: Done in 853811433efa6480d6b14f252d22ed3642b2e6b2

src/utilities/geometry/Intersection.cpp
```diff
@@ -746,6 +827,7 @@ boost::optional<IntersectionResult> intersect(const std::vector<Point3d>& polygo
   std::vector<std::vector<Point3d>> newPolygons1;
   std::vector<std::vector<Point3d>> newPolygons2;
 
+  bool extraLogging = false;
```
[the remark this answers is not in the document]
jmarrec: I went with constexpr bool in 2cca961e2235296f840dc89c1664a97571ea77eb. You can revisit if you think it's good to pass an optional parameter (but penalize everyone)

src/model/Space.cpp
jmarrec: Done in db79456fe87054ed8957888e22f329598bdcec6f

src/model/FloorspaceReverseTranslator.cpp
jmarrec: Done in db79456fe87054ed8957888e22f329598bdcec6f

src/model/Space.cpp
jmarrec: I removed it in db79456fe87054ed8957888e22f329598bdcec6f

src/model/Space.cpp
jmarrec: Removed in db79456fe87054ed8957888e22f329598bdcec6f

src/model/Space.cpp
jmarrec: Removed in db79456fe87054ed8957888e22f329598bdcec6f

src/model/Surface.cpp
jmarrec: constexpr done in db79456fe87054ed8957888e22f329598bdcec6f

src/utilities/geometry/Intersection.cpp
```diff
@@ -771,13 +859,24 @@ boost::optional<IntersectionResult> intersect(const std::vector<Point3d>& polygo
...
+  }
+
   // check if intersection is empty
   if (intersectionResult.empty()) {
     //LOG_FREE(Info, "utilities.geometry.intersect", "Intersection is empty");
     return boost::none;
   }
 
   intersectionResult = removeSpikes(intersectionResult);
+  if (intersectionResult.size() == 0) {
```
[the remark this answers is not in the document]
jmarrec: (done)

src/utilities/geometry/Intersection.cpp
```diff
@@ -762,6 +844,12 @@ boost::optional<IntersectionResult> intersect(const std::vector<Point3d>& polygo
     return boost::none;
   }
 
+  BoostMultiPolygon polys;
+  boost::optional<BoostPolygon> poly = boostPolygonFromVertices(polygon2, allPoints, tol);
```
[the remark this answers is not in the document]
jmarrec: done.

src/utilities/geometry/Intersection.cpp
```diff
@@ -401,7 +482,7 @@ std::vector<Point3d> verticesFromBoostPolygon(const BoostPolygon& polygon, std::
              "Converting polygon with " << polygon.inners().size() << " inner loops to OpenStudio vertices, inner loops ignored");
   }
 
-  result = removeCollinearLegacy(result);
+  if (removeCollinear) result = removeCollinearLegacy(result);
```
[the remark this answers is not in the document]
jmarrec: (done)

src/utilities/geometry/Intersection.cpp
```diff
@@ -1243,12 +1326,13 @@ Polygon3d PolygonFromBoostPolygon(const BoostPolygon& boostPolygon, Point3dVecto
...
       Point3d point3d(inner[i].x(), inner[i].y(), 0.0);
-      Point3d resultPoint = getCombinedPoint(point3d, allPoints, tol);
+      const Point3d resultPoint = getCombinedPoint(point3d, allPoints, tol);
       // don't keep repeated vertices
       if ((i > 0) && (hole.back() == resultPoint)) {
         continue;
       }
-      hole.push_back(point3d);
+      // TODO: shouldn't this be resultPoint instead?! Or maybe the repeated check should use point3d in which case resultPoint is pointless
+      hole.emplace_back(std::move(point3d));
```
jmarrec, line 1335: @ggartside here's someone I noticed while doing the refactor. I think there's something wrong here. Please advise

jmarrec (changes requested):

jmarrec:
@ggartside I've merged develop and resolved conflicts, and cleaned up a bunch of things. There are still some remaining comments that are left open.

src/utilities/geometry/Intersection.cpp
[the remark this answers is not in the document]
ggartside: Braces it is - but I see you already made this change?

src/utilities/geometry/Intersection.cpp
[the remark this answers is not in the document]
ggartside: The intention was to provide for extra logging without having to change the code - just change the value in the debugger. Logging gets clogged up with lots of unnecessary content otherwise.

But - whatever you prefer, definitely don't want to add an extra argument to the function

src/utilities/geometry/Intersection.cpp
```diff
@@ -762,6 +844,12 @@ boost::optional<IntersectionResult> intersect(const std::vector<Point3d>& polygo
     return boost::none;
   }
 
+  BoostMultiPolygon polys;
+  boost::optional<BoostPolygon> poly = boostPolygonFromVertices(polygon2, allPoints, tol);
+  if (extraLogging) {
+    polys.push_back(*poly);
```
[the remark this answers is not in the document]
ggartside: Technically true but this is code is only supposed to be activated when debugging so I didn't think it would be an issue

src/utilities/geometry/Point3d.cpp
ggartside: I'll make that change - the initial implementation was formatted so I could load the json file in to Simergy for debugging purposes so I could view the polygons output from doing intersections but on reflection it would be better to modify the back end code to accept an list of three doubles.

Done

src/utilities/geometry/Test/Intersection_GTest.cpp
ggartside: I'll use the initializer list - it's also closer to how it would be constructed in C# so I like that.

src/utilities/geometry/Test/Intersection_GTest.cpp
ggartside: Done

src/model/Space.cpp
ggartside: I think this was some debug code that I didn't remove

src/model/Surface.cpp
ggartside: Thanks

src/utilities/geometry/Intersection.cpp
ggartside: @jmarrec Me too - I think resultPoint should be used as getCombinedPoint might have modified the original Point3d

src/utilities/geometry/Intersection.cpp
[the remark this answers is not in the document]
ggartside: Done

src/utilities/geometry/Intersection.cpp
ggartside: @jmarrec Yes I think this should use resultPoint as the original point has potentially been modified by getCombinedPoint so I've made the change

src/model/test/FloorspaceReverseTranslator_GTest.cpp
ggartside: @jmarrec The test is re-using the same json file.

src/utilities/geometry/Point3d.cpp
jmarrec: Thanks for the change. Note that i don't think there's anything wrong with producing a valid json for a purpose, I didn't have the context

ggartside: This was creating a json file I could load in to Simergy to see the output from intersections but I agree a more generic list of doubles is better so I used your implementation and changed our code to suit

src/utilities/geometry/Intersection.cpp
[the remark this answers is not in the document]
jmarrec: Just for context, the reason I am wary of doing extra work all the time only for debugging purposes is that I know (from profiling and benchmarking) that the geometry stuff can take a long time and lots of resources for specific models.

jmarrec:
Thanks for the changes @ggartside . I have fixed up the cppcheck and merged latest develop in. There are only 2 open comments that needs addressing at this point.

src/model/FloorspaceReverseTranslator.cpp
ggartside: Yes confident about that otherwise the floorspace model would be invalid but lets make 100% sure

ggartside:
Added a new commit that applies a scale factor when converting between openstudio and boost data formats to improve the numerical accuracy of the boolean operations.

Idea came from this comment in the boostorg/geometry repo https://github.com/boostorg/geometry/issues/1034#issuecomment-1284180101 where the author indicates that scaling the values by 10 improved the result and talks about rounding to an integer grid, so by increasing the range of values we pass to boost we should be improving the resolution of the integer rounding

src/utilities/geometry/Intersection.cpp
```diff
@@ -102,36 +102,113 @@ inline bool has_self_intersections(Geometry const& geometry, bool throw_on_self_
...
+  os << "[";
+  for (const auto& poly : boostPolygons) {
+    os << poly;
+  }
+  os << "]";
+  return os;
+}
+
+// Scale factor for parameters passed to boost
+double scaleBy = 1000.0;
```
jmarrec, line 155: Never use a global variable like this.

```suggestion
// Scale factor for parameters when converting between openstudio and boost data formats to improve the numerical accuracy of the boolean operations.
// Idea came from this comment in the boostorg/geometry repo: https://github.com/boostorg/geometry/issues/1034#issuecomment-1284180101
// where the author indicates that scaling the values by 10 improved the result and talks about rounding to an integer grid, 
// so by increasing the range of values we pass to boost we should be improving the resolution of the integer rounding
static constexpr double scaleBy = 1000.0;
```

jmarrec:
Ubuntu test failures are unrelated:

```
The following tests FAILED:
	639 - BCLFixture.RemoteBCLTest (Failed)
	640 - BCLFixture.RemoteBCLTest2 (Failed)
	642 - BCLFixture.RemoteBCLMetaSearchTest (Failed)
	643 - BCLFixture.RemoteBCL_EncodingURI (Failed)
```

PR #4857, opened 2023-04-07
Schedule:File ctor with path

### LINKED ISSUE #4763: Allow `ScheduleFile` constructor with path instead of ExternalFile argument
Issue overview
--------------
Running the following command on Windows:
`external_file = OpenStudio::Model::ExternalFile::getExternalFile(@model, schedules_path)`

It's throwing an error:
`OS Message: Failed to copy file from "C:/Users/shorowit/AppData/Local/Temp/1/schedules20221212-6228-2tv9lb.csv" to "C:/Program Files (x86)/NREL/BEopt_3.0.0/./files/schedules20221212-6228-2tv9lb.csv"`

https://github.com/NREL/OpenStudio/blob/70df42e000cfe751e28a1e8c33b3c0c2e13786be/src/model/ExternalFile.cpp#L277

I can't tell if the problem is that it doesn't have permission or that there's a space in the path. Either way, I don't think there's a way to tell OpenStudio where to copy this file, so it'd be nice if it could more gracefully handle this.

joseph-robertson:
Pull request overview
---------------------

 - Fixes #4763

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


src/model/ScheduleFile.hpp
```diff
@@ -58,6 +58,8 @@ namespace model {
 
     ScheduleFile(const ExternalFile& externalfile, int column = 1, int rowsToSkip = 0);
 
+    ScheduleFile(const Model& model, const std::string& filename, int column = 1, int rowsToSkip = 0);
```
jmarrec, line 61:
```suggestion
    ScheduleFile(const Model& model, const openstudio::path& filePath, int column = 1, int rowsToSkip = 0);
```

jmarrec: Like I said on https://github.com/NREL/OpenStudio/issues/4763#issuecomment-1502822755, I think ExternalFile::getExternalFile should have taken a path to begin with, but here this is new so I'd say let's go with path.

src/model/ScheduleFile.cpp
```diff
@@ -412,6 +412,22 @@ namespace model {
...
 
+  ScheduleFile::ScheduleFile(const Model& model, const std::string& filename, int column, int rowsToSkip)
+    : ScheduleInterval(ScheduleFile::iddObjectType(), model) {
+    OS_ASSERT(getImpl<detail::ScheduleFile_Impl>());
+
+    boost::optional<ExternalFile> externalfile = ExternalFile::getExternalFile(model, filename, false);
+    OS_ASSERT(externalfile);
+
+    bool ok;
+    ok = setPointer(OS_Schedule_FileFields::ExternalFileName, externalfile->handle());
```
jmarrec, line 423: getExternalFile can fail, at which point you want to remove() the object being created and throw. don't commit this as is, I left a comment in there 

```suggestion
  ScheduleFile::ScheduleFile(const Model& model, const openstudio::path& filePath, int column, int rowsToSkip)
    : ScheduleInterval(ScheduleFile::iddObjectType(), model) {
    OS_ASSERT(getImpl<detail::ScheduleFile_Impl>());

    // NOTE: if you don't end up modifying the type argument in getExternalFile, a toString(filePath) is needed
     boost::optional<ExternalFile> externalFile = ExternalFile::getExternalFile(model, filePath, false);
     if (!externalFile) {
       remove();
       LOG_AND_THROW("Couldn't find the ExternalFile at " << filePath);
    }

    bool ok = setPointer(OS_Schedule_FileFields::ExternalFileName, externalFile->handle());
```

src/model/ExternalFile.cpp
```diff
@@ -136,6 +136,9 @@ namespace model {
     path ExternalFile_Impl::filePath() const {
       path result;
       path fname = toPath(fileName());
+      if (openstudio::filesystem::exists(fname) || openstudio::filesystem::is_regular_file(fname)) {
```
jmarrec, line 139: Did you mean `&&` instead of `||`. The path must exist AND it must be a regular file no?

src/model/ExternalFile.cpp
```diff
@@ -206,12 +209,17 @@ namespace model {
...
     if (!p.has_filename()) {
       return boost::none;
     }
-    std::string s = toString(p.filename());
+    std::string s;
+    if (copy) {
+      s = toString(p.filename());
+    } else {
+      s = filename;
+    }
```
jmarrec, line 222: I'm not certain about this. Do we allow several files named the same in different directories?
I guess that's fine, since it's filePath() that's used in the FT.

https://github.com/NREL/OpenStudio/blob/42683d615dd470d90390175c1fdc884039dbd767/src/energyplus/ForwardTranslator/ForwardTranslateScheduleFile.cpp#L68

src/model/ExternalFile.hpp
```diff
@@ -69,7 +69,7 @@ namespace model {
 
     static std::vector<std::string> columnSeparatorValues();
 
-    static boost::optional<ExternalFile> getExternalFile(const Model& model, const std::string& filename);
+    static boost::optional<ExternalFile> getExternalFile(const Model& model, const std::string& filename, bool copy = true);
```
jmarrec, line 72: I think we want a public facing docstring to describe the method and this parameter. And maybe a more explicit name, `copyFile` or even `copyFileToWorkflowDirectory`?

First draft:

```c++
    /** This factory method will look for an existing ExternaFile with the same file name. If not it will try to create one.
      * - If copyFile is true, the file will be copied to the first entry of the WorkflowJSON's absoluteFilePaths() (or absoluteRootDir() if empty);
      * - If copyFile is false, it assumes this is an absolute path.
      * In both cases, the filename should resolve to a valid path or this will return an empty optional. */
```

jmarrec (commented):
Aside from minor adjustments, @joseph-robertson 's proposed solution is pretty much exactly what I had in mind and described in https://github.com/NREL/OpenStudio/issues/4763#issuecomment-1502822755 before reviewing.

We probably want to have a short OpenStudio-resources reg test (maybe just a throw-away one, not sure...) that exercises this new absolute / no copy path, and the overloaded ScheduleFile ctor to ensure we do get an E+ model that runs.

joseph-robertson:
OpenStudio-resources regression test here: https://github.com/NREL/OpenStudio-resources/pull/197

src/model/ScheduleFile.hpp
joseph-robertson: Oh I see. You're advocating for a potential API break on `getExternalFile`. Hmm...

joseph-robertson: Could we deprecate `getExternalFile` with `filename` argument in favor of a new one with `filepath` argument?

src/model/ExternalFile.cpp
joseph-robertson: I did mean `II`. I copied that line straight from below in the `absoluteFilePaths` loop.

src/model/ExternalFile.hpp
joseph-robertson: Went with `copyFile`.

shorowit (approved):
I did some testing:
1. Tried the build without any changes to our code -- it behaved identical as before.
2. Tried the build and updated our code to use the new constructor -- it behaved as desired (no copy of schedule file).

I haven't look at the code but it worked perfectly in my testing. 👍

src/model/ExternalFile.cpp
jmarrec: that doesn't make sense to me. try calling is_regular_file on an non existing path and it'll throw.
if it exists, then the is_regular_file is never called.

jmarrec: bool is_regular_file(const path& p);

    Returns: is_regular_file(status(p)).

    Throws: filesystem_error if status(p) would throw filesystem_error.

https://www.boost.org/doc/libs/1_79_0/libs/filesystem/doc/reference.html#is_regular_file

joseph-robertson: Is it wrong down below then, too?

jmarrec: I guess it doesn't throw, but anyways...

https://www.boost.org/doc/libs/1_79_0/libs/filesystem/doc/reference.html#status

> Throws: filesystem_error. [Note: result values of file_status(file_not_found)and file_status(type_unknown) are not considered failures and do not cause an exception to be thrown. —end note]

jmarrec: @joseph-robertson yes it's wrong below as well.

jmarrec (changes requested):

jmarrec:
Change the exists || is_regular_file or exists && is_regular_file and then I'm all fine with this PR!

jmarrec (approved):

Issue #4634, opened 2022-07-29
Python plugin next steps


joseph-robertson:
Extend the python plugin feature that was implemented in #4600. Specifically:
- [ ] optionally provide python script with OS handles/names and then FT does replacement in the python script
- [ ] OS measure generates the python script (this is in the realm of user documentation / demonstration)

See https://github.com/NREL/OpenStudio/issues/3839 for more information.

joseph-robertson:
Extend the python plugin feature that was implemented in #4600. Specifically:
- [ ] optionally provide python script with OS handles/names and then FT does replacement in the python script
- [ ] OS measure generates the python script
  - [ ] Option 1: this is in the realm of user documentation / demonstration (i.e., [OS-resources test](https://github.com/NREL/OpenStudio-resources/blob/develop/model/simulationtests/python_plugin.rb))
  - [ ] Option 2: create OS API for populating a plugin template file. see [proposed design doc](https://docs.google.com/document/d/1plh0t1_9cKNT5m6l3H4apo9iAKBHvJ-QHHF-eteWUqM/edit#heading=h.q6w40yz5atu1)

joseph-robertson:
Another idea mentioned by @DavidGoldwasser, add capability at OSW level for using the plugin. David, can you expand on this?

joseph-robertson:
Notes from 12/7 meeting:

One issue with the current implementation of `PythonPlugin` is **portability**. A single python plugin .py file is not guaranteed to work across multiple models (e.g., zone names are different).

**Approach 1**
(A) To address this, we could use **erb templating**. The gist of this approach would be to pass a erb .py template file into `PythonPluginInstance` **pre-substitution**, and then make the substitutions based on contents in `model`.

(B) The optional extension of this approach would be to wait until all measures are applied before making any substitutions. We could potentially create a new "EMS" class of measures that are called _after_ Model measures.

We plan to prototype (A) from above.

**Approach 2**
Update E+ to load `model` on E+ warmup, save as a global, etc. Then you could access `model` by something like `state.model()` or `state.osmodel()` from the python plugin file.

Issue #4846, opened 2023-03-30
[Enhancement Request] Allow importing Python package with C-extension e.g. numpy


wta12:
Enhancement Request
--------------
Allow importing package with C-extension e.g. numpy

## Detailed Description
 OS SDK 3.5.1 returns error if execute python script with numpy as dependencies 
`"OpenStudio/OpenStudio_3_5_1/bin/openstudio labs --verbose --python_home /usr/local/anaconda3/envs/pythonMeasure --python_path /usr/local/anaconda3/envs/pythonMeasure/lib/python3.8/site-packages  execute_python_script python_script_test.py"`
Return error:
```
Error message: IMPORTANT: PLEASE READ THIS FOR ADVICE ON HOW TO SOLVE THIS ISSUE!
Importing the numpy C-extensions failed. This error can happen for
many reasons, often due to issues with your setup or how NumPy was
installed.
```
## Possible Implementation:
EP instance in same OS SDK installation allows Plugin:Instance (pythons script) with numpy imported

tijcolem:
@Myoldmopar  I think this one is related to how E+ manages python and python packages?

tijcolem:
@wta12  Also, have you tried setting the PYTHONPATH env to point to the pip package repo and then run it? 
 
e.g. `export PYTHONPATH=/path/to/my/modules/`

jmarrec:
I have it fixed locally. But I'd like to discuss with @kbenne about it, because I had to change the Unix (well, Ubuntu to be exact, I think it works on mac...) dlopen flags from `RTLD_LAZY | RTLD_LOCAL` to `RTLD_GLOBAL | RTLD_NOW` when using python.

On unix, the C native ext are NOT linked to libpython.so. Our libpythonengine.so IS linked to it.
If we use RTLD_LOCAL, we do not export these libpython.so symbols and as such we get unresolved symbols at runtime, such as `undefined symbol: PyFloat_Type`. The alternative is to find a clever way also dlopen the libpython.so, this one with the RTLD_GLOBAL visibility.

https://stackoverflow.com/questions/67891197/ctypes-cpython-39-x86-64-linux-gnu-so-undefined-symbol-pyfloat-type-in-embedd



## Currently

### Mac

```
[...]

PR #4858, opened 2023-04-07
Python plugin jinja template

joseph-robertson:
Pull request overview
---------------------

Addresses **Approach 1(A)**, but using a python measure: https://github.com/NREL/OpenStudio/issues/4634#issuecomment-1346858424

I've created two measures:
* `PythonPluginRuby` (uses `python_plugin_program_erb.py` template)
* `PythonPluginPython` (uses `python_plugin_program_jinja.py` template)

Both measures are called by `compact_python_plugin.osw`.

```
$ /c/openstudio-3.6.0-alpha/bin/openstudio.exe labs run -w resources/Examples/compact_osw/compact_python_plugin.osw
```

Appears to be working successfully.

Update: I've removed `sys.path.append` and tried running with OS v3.7.0 (w/out labs), and it appears to be working successfully.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


resources/Examples/compact_osw/files/python_plugin_program_erb.py
```diff
@@ -0,0 +1,44 @@
+from pyenergyplus.plugin import EnergyPlusPlugin
+
+import os, sys
```
jmarrec, line 3:
```suggestion
import os
import sys
```

resources/Examples/compact_osw/files/python_plugin_program_erb.py
```diff
@@ -0,0 +1,44 @@
...
+sys.path.append('C:/Python38/Lib/site-packages') # this should (needs to?) be same version as E+'s version
```
jmarrec, line 4: We're going to have to make this path to system python's site_packages some kind of cmake replacement or a command line arg if we want to test this.

But more broadly, this is going to pose the question of whether we should/need to provide some common python packages out of the box like we do for ruby gems (and take into account the pain that it is to maintain that stuff). I'm specifically thinking about native packages (related to this issue https://github.com/NREL/OpenStudio/issues/4846), (pandas does have a C extension as well, I think something about numpy may be special)

@kbenne @tijcolem

jmarrec: Following https://github.com/NREL/OpenStudio/pull/4868 you can make your test pass the `--python_home` or `--python_path` and it'll work. 

I think we should discuss whether we should always add some common python packages like Jinja2 (https://github.com/NREL/OpenStudio/issues/4906#issuecomment-1621966528)

joseph-robertson: @jmarrec @kbenne 

Has there been any more discussion/movement on #4906?

jmarrec: See here: https://github.com/NREL/OpenStudio/blob/0ab02805c62a5f459e669bf3f4c65c95a1d8eed7/src/cli/CMakeLists.txt#L343-L393

joseph-robertson:
@jmarrec This appears to be working now (i.e., can import pandas and jinja2). How can we formalize running `compact_python_plugin.osw` as a test? Add something to src/cli/CMakeLists?

resources/Examples/compact_osw/measures/PythonPluginPython/measure.py
```diff
@@ -0,0 +1,132 @@
+import os, sys
```
jmarrec, line 1:
```suggestion
import os
import sys
```

Generally speaking you should `pip install black isort` on your system and reformat your python files via

```shell
isort .
black -l 120 .
```

resources/Examples/compact_osw/compact_python_plugin.osw
jmarrec, line 1: Move this to another folder. Or it'll get installed. Suggestion for the layout of `resources/Example`

```diff
 .
 ├── compact_osw
 ├── with_analysis
+└── python_plugin
```

resources/Examples/compact_osw/files/python_plugin_program_erb.py
jmarrec, line 1: This should be inside the measures directory inside resources/ , so it is a proper measure and can be moved around.

jmarrec:
> @jmarrec This appears to be working now (i.e., can import pandas and jinja2). How can we formalize running `compact_python_plugin.osw` as a test? Add something to src/cli/CMakeLists?

Yes. like you've done. But it bothers me we don't check at all if E+ actually runs the python plugins

jmarrec:
@joseph-robertson I've made some adjustments (moved things around, linted) + actually tested the plugins are running.

Assuming CI comes clean and you're ok with the changes, this can drop.

PR #4868, opened 2023-04-25
#4846 - Allow importing Python package with C-extension e.g. numpy in the labs CLI

### LINKED ISSUE #4846: [Enhancement Request] Allow importing Python package with C-extension e.g. numpy

jmarrec:
Pull request overview
---------------------

 - Fixes #4846

- Add some ctest for using numpy with the CLI. 3 flavors
    * Explicitly insert to sys.path the site-packages and lib-dynload folders from the system python
    * Pass --python_path twice (site-packages and lib-dynload)
    * Pass --python_home

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 message(STATUS "Using CMake ${CMAKE_VERSION}")
-cmake_minimum_required(VERSION 3.19.0)
+cmake_minimum_required(VERSION 3.20.0)
 cmake_policy(SET CMP0048 NEW)
 
 set(CMAKE_CXX_STANDARD 20)
--- a/python/SetupPython.cmake
+++ b/python/SetupPython.cmake
@@ -7,9 +7,9 @@ set(Python_USE_STATIC_LIBS OFF)
 # find_package(Python) has the problem that on github actions in particular it'll pick up the most recent python (eg 3.9) from the tool cache
 # even if you have used the setup-python action and set it to 3.8
 if (PYTHON_VERSION)
-  find_package(Python ${PYTHON_VERSION} EXACT COMPONENTS Interpreter Development REQUIRED)
+  find_package(Python ${PYTHON_VERSION} EXACT REQUIRED COMPONENTS Interpreter Development OPTIONAL_COMPONENTS NumPy)
 else()
-  find_package(Python COMPONENTS Interpreter Development REQUIRED)
+  find_package(Python REQUIRED COMPONENTS Interpreter Development OPTIONAL_COMPONENTS NumPy)
 endif()
 
 execute_process(COMMAND ${Python_EXECUTABLE} -m pytest --version
--- a/python/engine/PythonEngine.cpp
+++ b/python/engine/PythonEngine.cpp
@@ -90,6 +90,9 @@ PythonEngine::PythonEngine(int argc, char* argv[]) : ScriptEngine(argc, argv), p
   if (pythonHomePassed) {
     addToPythonPath(pathToPythonPackages);
   }
+#if defined(__APPLE__) || defined(__linux___) || defined(__unix__)
+  addToPythonPath(pathToPythonPackages / "lib-dynload");
+#endif
 
   PyObject* m = PyImport_AddModule("__main__");
   if (m == nullptr) {
--- a/python/engine/PythonEngine.cpp
+++ b/python/engine/PythonEngine.cpp
@@ -52,19 +52,42 @@ void PythonEngine::setupPythonPath(const std::vector<openstudio::path>& includeD
 }
 
 PythonEngine::PythonEngine(int argc, char* argv[]) : ScriptEngine(argc, argv), program(Py_DecodeLocale(pythonProgramName, nullptr)) {
+  // TODO: modernize and use PyConfig (new in 3.8): https://docs.python.org/3/c-api/init_config.html
+
   // this frozen flag tells Python that the package and library have been frozen for embedding, so it shouldn't warn about missing prefixes
   Py_FrozenFlag = 1;
 
-  // Set the PYTHONPATH / PYTHONHOME to the E+ shipped standard library
-  // I think we need to set the python path before initializing the library
+  // Path to the E+ shipped standard library
   auto pathToPythonPackages = getEnergyPlusDirectory() / "python_standard_lib";
-  wchar_t* a = Py_DecodeLocale(pathToPythonPackages.make_preferred().string().c_str(), nullptr);
-  Py_SetPath(a);
-  Py_SetPythonHome(a);
+
+  // The PYTHONPATH / PYTHONHOME should be set before initializing Python
+  // If this Py_SetPath is called before Py_Initialize, then Py_GetPath won't attempt to compute a default search path
+  // The default search path is affected by the Py_SetPythonHome
+  // * if the user passed --python_home, we use that as the Python Home, and do not use Py_SetPath. But later we add the E+ standard_lib anyways
+  //   so it takes precedence (to limit incompatibility issues...)
+  // * If the user didn't pass it, we use Py_SetPath set to the E+ standard_lib
+
+  std::vector<std::string> args(argv, std::next(argv, static_cast<std::ptrdiff_t>(argc)));
+  bool pythonHomePassed = false;
+  auto it = std::find(args.cbegin(), args.cend(), "--python_home");
+  if (it != args.cend()) {
+    openstudio::path pythonHomeDir(*std::next(it));
+    wchar_t* h = Py_DecodeLocale(pythonHomeDir.make_preferred().string().c_str(), nullptr);
+    Py_SetPythonHome(h);
+    pythonHomePassed = true;
+  } else {
+    wchar_t* a = Py_DecodeLocale(pathToPythonPackages.make_preferred().string().c_str(), nullptr);
+    Py_SetPath(a);
+  }
 
   Py_SetProgramName(program);  // optional but recommended
 
   Py_Initialize();
+
+  if (pythonHomePassed) {
+    addToPythonPath(pathToPythonPackages);
+  }
+
   PyObject* m = PyImport_AddModule("__main__");
   if (m == nullptr) {
     throw std::runtime_error("Unable to add module __main__ for python script execution");
--- a/python/engine/PythonEngine.cpp
+++ b/python/engine/PythonEngine.cpp
@@ -30,9 +30,12 @@ void addToPythonPath(const openstudio::path& includePath) {
   if (!includePath.empty()) {
     PyObject* sys = PyImport_ImportModule("sys");
     PyObject* sysPath = PyObject_GetAttrString(sys, "path");
+    Py_DECREF(sys);  // PyImport_ImportModule returns a new reference, decrement it
+
     // fmt::print("Prepending '{}' to sys.path\n", includePath);
     PyObject* unicodeIncludePath = PyUnicode_FromString(includePath.string().c_str());
     PyList_Insert(sysPath, 0, unicodeIncludePath);
+    Py_DECREF(sysPath);  // PyObject_GetAttrString returns a new reference, decrement it
   }
 }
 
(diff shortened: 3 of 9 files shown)


CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 message(STATUS "Using CMake ${CMAKE_VERSION}")
-cmake_minimum_required(VERSION 3.19.0)
+cmake_minimum_required(VERSION 3.20.0)
```
jmarrec, line 2: using `cmake_path` which is new in 3.20

https://cmake.org/cmake/help/latest/command/cmake_path.html

python/SetupPython.cmake
```diff
@@ -7,9 +7,9 @@ set(Python_USE_STATIC_LIBS OFF)
 # find_package(Python) has the problem that on github actions in particular it'll pick up the most recent python (eg 3.9) from the tool cache
 # even if you have used the setup-python action and set it to 3.8
 if (PYTHON_VERSION)
-  find_package(Python ${PYTHON_VERSION} EXACT COMPONENTS Interpreter Development REQUIRED)
+  find_package(Python ${PYTHON_VERSION} EXACT REQUIRED COMPONENTS Interpreter Development OPTIONAL_COMPONENTS NumPy)
```
jmarrec, line 10: I'm using NumPy to test loading a native gem because the FindPython (cmake) has a NumPy component so it's easy to test whether the user actually has it installed.

src/cli/CMakeLists.txt
```diff
@@ -330,6 +330,50 @@ if(BUILD_TESTING)
 
   # ============ EndForward a Path properly no matter the slashes employed ============
 
+  # ============ Native Ruby Gems / Python Modules - C extensions =============
+  if (Python_NumPy_FOUND)
```
jmarrec, line 334: Only enable the tests if numpy found

src/cli/CMakeLists.txt
```diff
@@ -330,6 +330,50 @@ if(BUILD_TESTING)
...
+
+    if(NOT Python_ROOT_DIR)
+      # Python_STDLIB: we expect it to be
+      # Unix: ~/.pyenv/versions/3.8.12/lib/python3.8
+      # Windows C:\Python38\Lib
+      cmake_path(GET Python_STDLIB PARENT_PATH Python_ROOT_DIR)
+      if(UNIX)
+        cmake_path(GET Python_ROOT_DIR PARENT_PATH Python_ROOT_DIR)
+      endif()
+    endif()
```
jmarrec, line 344: I don't like these shenanigans too much, but not sure how to do otherwise.

src/cli/CMakeLists.txt
```diff
@@ -330,6 +330,50 @@ if(BUILD_TESTING)
...
+        WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/test/"
+      )
+    endif()
+
+    add_test(NAME OpenStudioCLI.Labs.execute_python_script.numpy.python_home
+      COMMAND $<TARGET_FILE:openstudio> labs
+                --python_home "$<SHELL_PATH:${Python_ROOT_DIR}>"
+                execute_python_script execute_python_script_with_numpy.py
+      WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/test/"
+    )
```
jmarrec, line 373: Three different tests to cover the scenarios listed in OP

python/engine/PythonEngine.cpp
```diff
@@ -25,9 +25,12 @@ void addToPythonPath(const openstudio::path& includePath) {
   if (!includePath.empty()) {
     PyObject* sys = PyImport_ImportModule("sys");
     PyObject* sysPath = PyObject_GetAttrString(sys, "path");
+    Py_DECREF(sys);  // PyImport_ImportModule returns a new reference, decrement it
```
jmarrec, line 28: Py_DECREF some stuff

python/engine/PythonEngine.cpp
```diff
@@ -36,30 +39,56 @@ void PythonEngine::pyimport(const std::string& importName, const std::string& in
...
-  if (!pythonHomeDir.empty()) {
-    wchar_t* a = Py_DecodeLocale(pythonHomeDir.generic_string().c_str(), nullptr);
-    Py_SetPythonHome(a);
+void PythonEngine::setupPythonPath(const std::vector<openstudio::path>& includeDirs) {
+
+  // Iterate in reverse order since addToPythonPath always inserts at pos 0
+  // --python_path path1 --python_path path2  =>  includeDirs = ["path1", "path2"]
+  // std::ranges::reverse_view needs modern compilers
+  for (auto it = includeDirs.rbegin(); it != includeDirs.rend(); it++) {
+    addToPythonPath(*it);
```
jmarrec, line 48: fixup the order here, unrelated

python/engine/PythonEngine.cpp
```diff
@@ -36,30 +39,56 @@ void PythonEngine::pyimport(const std::string& importName, const std::string& in
...
   }
 }
 
 PythonEngine::PythonEngine(int argc, char* argv[]) : ScriptEngine(argc, argv), program(Py_DecodeLocale(pythonProgramName, nullptr)) {
+  // TODO: modernize and use PyConfig (new in 3.8): https://docs.python.org/3/c-api/init_config.html
```
jmarrec, line 53: TODO for later

python/engine/PythonEngine.cpp
```diff
@@ -36,30 +39,56 @@ void PythonEngine::pyimport(const std::string& importName, const std::string& in
...
+  auto it = std::find(args.cbegin(), args.cend(), "--python_home");
+  if (it != args.cend()) {
+    openstudio::path pythonHomeDir(*std::next(it));
+    wchar_t* h = Py_DecodeLocale(pythonHomeDir.make_preferred().string().c_str(), nullptr);
+    Py_SetPythonHome(h);
+    pythonHomePassed = true;
+  } else {
+    wchar_t* a = Py_DecodeLocale(pathToPythonPackages.make_preferred().string().c_str(), nullptr);
+    Py_SetPath(a);
+  }
```
jmarrec, line 79: Deal with --python_home . Can't do it later, it has to be done before Py_Initialize... so grab it from the argc/argv...

python/engine/PythonEngine.cpp
```diff
@@ -36,30 +39,56 @@ void PythonEngine::pyimport(const std::string& importName, const std::string& in
...
   Py_SetProgramName(program);  // optional but recommended
 
   Py_Initialize();
+
+  if (pythonHomePassed) {
+    addToPythonPath(pathToPythonPackages);
+  }
+#if defined(__APPLE__) || defined(__linux___) || defined(__unix__)
+  addToPythonPath(pathToPythonPackages / "lib-dynload");
+#endif
```
jmarrec, line 90: lib-dynload is present in the E+ python_standard_lib, we should add it

src/cli/main.cpp
```diff
@@ -126,28 +126,33 @@ int main(int argc, char* argv[]) {
     experimentalApp
       ->add_option("-I,--include", includeDirs, "Add additional directory to add to front of Ruby $LOAD_PATH (may be used more than once)")
       ->option_text("DIR")
+      ->check(CLI::ExistingDirectory)
```
jmarrec, line 129: Add some CLI11 validators to ensure the various args are valid directories to begin with and catch mistakes

src/utilities/core/DynamicLibraryPOSIX.hpp
```diff
@@ -26,8 +26,17 @@ struct DynamicLibrary
...
+    int flags = RTLD_LAZY | RTLD_LOCAL;  // NOLINT(misc-const-correctness, hicpp-signed-bitwise)
+
+    // This seems to work on Mac without RTLD_GLOBAL...
+#ifdef __linux__
+    if (m_location.filename().generic_string().find("python") != std::string::npos) {
+      // https://stackoverflow.com/questions/67891197/ctypes-cpython-39-x86-64-linux-gnu-so-undefined-symbol-pyfloat-type-in-embedd
+      flags = RTLD_LAZY | RTLD_GLOBAL;
+    }
+#endif
+    m_handle = {dlopen(m_location.c_str(), flags), m_handle_deleter};
```
jmarrec, line 39: This is the bit that bothers me a bit @kbenne

jmarrec:
38eed7243dc8c01db45613af4ca37cdb4ec078b9 can be reverted, I'm just trying it out

python/engine/PythonEngine.cpp
kbenne: I wonder how many instances there are where we miss these decrements. Or we have one that should not be. I bet there are a few remaining.

python/engine/PythonEngine.cpp
kbenne: If feels like a code smell that we are parsing argc / argv at this point instead of having everything parsed out by the CLI11 library. I think I am guilty of establishing this pattern in ScriptEngine and I think I did it because the ruby init functions (at least one of them) accept argc and argv.

src/utilities/core/DynamicLibraryPOSIX.hpp
kbenne: Are we sure we actually need these flags ever? I'm not sure about that. I think there is a decent chance they got added unnessarily.

src/cli/CMakeLists.txt
```diff
@@ -337,6 +337,61 @@ if(BUILD_TESTING)
...
+          list(APPEND PYTHON_PATH "${EXTRA_LOCAL_DIST}")
+        endif()
+      endif()
+    else()
+      set(PYTHON_PATH "$<SHELL_PATH:${Python_SITELIB}>")
+    endif()
+
+    message(DEBUG "PYTHON_PATH=${PYTHON_PATH}")
+
+    add_test(NAME OpenStudioCLI.Labs.execute_python_script.numpy.explicit_sys_path_insert
```
kbenne, line 373: At first this series of tests kind of seemed off to me, because they are testing the environment of the developer/build environment, but when I think more carefully, I guess that is the point. The dev environment is a proxy for whatever environment a user may have, and since you already have the scaffolding in place to find a random numpy installation it turns out to be a convenient way to formumulate a test. Seems nice to me. 

The only thing this leaves on the table are custom extensions that we choose to ship by default, but I believe that will be addressed in a different issue.

kbenne (approved):

python/engine/PythonEngine.cpp
jmarrec: ruby_sysinit does take argc, argv yes. 

Plus, here this is an order thing, and specific to Python while the CLI only sees a ScriptEngineInstance, so adding an explicit PythonHome argument isn't great either.

If you see another cleaner way to do it, this can be addressed later. There are other things that probably need cleanup too (including simplifying the folder / Cmake structure probably).

Another thing: we need to process the argv data manually at the start of the CLI anyways since we do support (for legacy reasons) some usage like `openstudio labs test.rb` that is translated as being  `openstudio labs execute_ruby_script test.rb` . cf https://github.com/NREL/OpenStudio/blob/2af6123b91304e12060af360d0afb24083336a4c/src/cli/main.cpp#L45C1-L89

src/utilities/core/DynamicLibraryPOSIX.hpp
jmarrec: Which flag(s)? RTLD_LAZY?

PR #4873, opened 2023-05-02
AirLoopHVACUnitarySystem set Method During XXX Operation

### LINKED ISSUE #4695: AirLoopHVACUnitarySystem: `Supply Air Flow Rate Method During <XXX> Operation` should be set via related setters/autosize
Issue overview
--------------

`Supply Air Flow Rate Method During <XXX> Operation` should be automatically infered at model time.

## Current Behavior

`Supply Air Flow Rate Method During <XXX> Operation` isn't set by default. E+ shows a warning in the eplusout.err

```
   ** Warning ** AirLoopHVAC:UnitarySystem = AIR LOOP HVAC UNITARY SYSTEM 1
   **   ~~~   ** Method used to determine the cooling supply air flow rate is not specified when cooling coil is present.
   ** Warning ** AirLoopHVAC:UnitarySystem = AIR LOOP HVAC UNITARY SYSTEM 1
   **   ~~~   ** Method used to determine the heating supply air flow rate is not specified when heating coil is present.
```

## Expected Behavior

eg `setSupplyAirFlowRatePerFloorAreaDuringCoolingOperation(1.0)`should set the `Supply Air Flow Rate Method During Cooling Operation` to `FlowPerFloorArea`

autosizeSupplyAirFlowRateDuringCoolingOperation should set it to `SupplyAirFlowRate`

## Steps to Reproduce

```ruby
m = Model.new
u = AirLoopHVACUnitarySystem.new(m)
u.setSupplyAirFlowRatePerFloorAreaDuringCoolingOperation(0.1)
# => true
u.supplyAirFlowRateMethodDuringCoolingOperation.empty?
# => false
u.supplyAirFlowRateMethodDuringCoolingOperation.get
# => ""
```

## Possible Solution

Straightforward.

### Environment

 - Platform (Operating system, version): All
 - Version of OpenStudio (if using an intermediate build, include SHA): develop circa 3.5.0-alpha, 92da27b27de50e076ce3bdc5032e4f348e2231e8

### Context

Found while adding a new regression test. cf https://github.com/NREL/OpenStudio-resources/pull/176/commits/2057e0db8113160da9d21cac1c69efcfbfa3ab6c

joseph-robertson:
Pull request overview
---------------------

 - Fixes #4695

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


jmarrec (commented):
Shouldn't the `setSupplyAirFlowRateMethodDuringCoolingOperation` / `resetSupplyAirFlowRateMethodDuringCoolingOperation` etc should be marked deprecated and become no-op.

ideally the `boost::optional<std::string> supplyAirFlowRateMethodDuringCoolingOperation()` getter should have an API-breaking change and return a `std::string` instead too in that case.

joseph-robertson:
@jmarrec You're advocating for introducing `\required-field` to OpenStudio.idd fields:
* Supply Air Flow Rate Method During Cooling Operation
* Supply Air Flow Rate Method During Heating Operation
* Fan Placement (optional, not technically related to this PR)

And then API-breaking the getters as well as deprecating the setters (and resetters)?

jmarrec:
@joseph-robertson If they end up being always defined, then yes I guess that's what I'm suggesting. But I'm fine if we don't do it. 

Still, the setters/resetters should probably be deprecated if we mean for others setters to set it **and** it's deterministic (I haven't checked if that's the case), meaning that there isn't a use case for calling the setter manually to do something else. A counter example would be a ChillerElectricEIR's condenser Type: "AirCooled", "WaterCooled", "EvaporativelyCooled". If you connect to a PlantLoop, you can set it to WaterCooled. If you disconnect, you can set it to AirCooled. But the user still has a use case for calling `setCondenserType("EvaporativelyCooled")`

jmarrec:
It seems that the setters should probably stay, because of the `FractionOfAutosizedCoolingValue` and `None` options.

```
[1] u.methods.grep(/^set.*DuringCoolingOperation$/i)
=> [:setSupplyAirFlowRateMethodDuringCoolingOperation,
 :setSupplyAirFlowRateDuringCoolingOperation,
 :setSupplyAirFlowRatePerFloorAreaDuringCoolingOperation,
 :setDesignSupplyAirFlowRatePerUnitofCapacityDuringCoolingOperation]
[2] simulationtests(main)> AirLoopHVACUnitarySystem::supplyAirFlowRateMethodDuringCoolingOperationValues
=> ["None", "SupplyAirFlowRate", "FlowPerFloorArea", "FractionOfAutosizedCoolingValue", "FlowPerCoolingCapacity"]
```

jmarrec:
@joseph-robertson I had a lot of trouble seeing what could or couldn't be done, so I made a proof of concept PR at https://github.com/NREL/OpenStudio/pull/4905. This only changes the DuringCooling one, but can serve as a basis for discussion (and potentially extend the same treatment for the others)

src/model/AirLoopHVACUnitarySystem.hpp
```diff
@@ -106,7 +108,7 @@ namespace model {
     boost::optional<HVACComponent> supplementalHeatingCoil() const;
 
     /** In EnergyPlus 8.3.0 and above this property maps to the EnergyPlus field "Cooling Supply Air Flow Rate Method" **/
-    boost::optional<std::string> supplyAirFlowRateMethodDuringCoolingOperation() const;
+    std::string supplyAirFlowRateMethodDuringCoolingOperation() const;
```
kbenne, line 111: Can we capture this API change for the release notes? @DavidGoldwasser or @tijcolem do we have a place to keep track of these so we don't miss them at release time?

src/model/AirLoopHVACUnitarySystem_Impl.hpp
```diff
@@ -84,12 +84,14 @@ namespace model {
 
       boost::optional<Schedule> supplyAirFanOperatingModeSchedule() const;
 
+      bool hasHeatingCoil() const;  // For speed
```
kbenne, line 87: These are in the impl so I probably shouldn't care. (not that I do) Curious though, what is the purpose of this over just asking for the ::heatingCoil which is optional and therefore behaves a lot like a bool?

kbenne (approved):

src/model/AirLoopHVACUnitarySystem.hpp
joseph-robertson: Capturing API changes in release notes here: https://github.com/NREL/OpenStudio/pull/4873/commits/52b5d438a7753304037628bdef6a3b69f7147ba2

PR #4917, opened 2023-06-14
Add a swig out typemap for Json::Value + openstudio::path typemap in/out for Python + add toJSON() for Workfow files and BCLMeasure

jmarrec:
Pull request overview
---------------------

- Converts the Json::Value
    - Ruby: to a hash, with symbols as keys (not sure that's what we really want or not)
    - Python: a dict.
    - CSharp: ignored (for now at least) 

Context: As part of the rewrite of Measure Manager to C++, I'm going to be using Json::Value outputs a lot more than currently. I'm just trying to isolate changes for easier review.

Test code:

ruby:

```shell
Products/openstudio labs -e "m = OpenStudio::Model::Model.new; ft = OpenStudio::EnergyPlus::ForwardTranslator.new; ft.setExcludeLCCObjects(true); w = ft.translateModel(m); d = OpenStudio::EPJSON::toJSON(w); puts d.class; require 'json'; puts JSON.pretty_generate(d)"
```

python:

```shell
Products/openstudio labs -c "m = openstudio.model.Model(); z = openstudio.model.ThermalZone(m); ft = openstudio.energyplus.ForwardTranslator(); ft.setExcludeLCCObjects(True); w = ft.translateModel(m); d = openstudio.epjson.toJSON(w); print(type(d)); import json; print(json.dumps(d, indent=2))"
```

### Pull Request Author

 - [ ] Model API Changes / Additions
 - [ ] Any new or modified fields have been implemented in the EnergyPlus ForwardTranslator (and ReverseTranslator as appropriate)
 - [ ] Model API methods are tested (in `src/model/test`)
 - [ ] EnergyPlus ForwardTranslator Tests (in `src/energyplus/Test`)
 - [ ] If a new object or method, added a test in NREL/OpenStudio-resources: Add Link
 - [ ] If needed, added VersionTranslation rules for the objects (`src/osversion/VersionTranslator.cpp`)
 - [ ] Verified that C# bindings built fine on Windows, partial classes used as needed, etc.
 - [ ] All new and existing tests passes
 - [ ] If methods have been deprecated, update rest of code to use the new methods

**Labels:**

 - [ ] If change to an IDD file, add the label `IDDChange`
 - [ ] If breaking existing API, add the label `APIChange`
 - [ ] If deemed ready, add label `Pull Request - Ready for CI` so that CI builds your PR

### Review Checklist

This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
[...]

PR #4920, opened 2023-06-23
Fix #4885 - Rewrite MeasureManager in C++

### LINKED ISSUE #4885: Rewrite MeasureManager in C++
Enhancement Request
--------------

Even in the `labs` subcommand, the MeasureManager is purely Ruby, so you can't use it to compute arguments for a Python Measure. 

## Detailed Description

https://github.com/NREL/OpenStudio/blob/f4f51c3e1888721212fb6c2002f75023abfc0011/src/cli/MeasureUpdateCommand.cpp#L369-L389

## Possible Implementation

We have the buildings blocks in place to compute arguments already in the MeasureUpdateCommand::getAndUpdateMeasure. But we'd need to write an HTTP server in C++. We have cpprestsdk that we already use for RemoteBCL. boost asio is another option, so is Poco.

jmarrec:
Pull request overview
---------------------

 - Fix #4885

* **Note: this includes #4917**

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


DavidGoldwasser:
Related to #4929. Maybe some additional work if time

kbenne:
@jmarrec. Check it out https://github.com/NREL/OpenStudio/pull/4920/commits/3c4a1c32fd096ca183c5668e2aafe99ac6564fb4 I think this does the trick. I thnk now we just need to do a bunch of testing to make sure the API and behavior is consistent with the original Ruby implementation.

@DavidGoldwasser

src/cli/MeasureManager.cpp
```diff
@@ -619,11 +598,12 @@ void MeasureManagerServer::handle_get(web::http::http_request message) {
     result["status"] = "running";
     result["my_measures_dir"] = my_measures_dir.generic_string();
     message.reply(web::http::status_codes::OK, toWebJSON(result));
+    return;
   }
 
   // Cpprestsdk has it's own json implementation.....
   if (uri == "/internal_state") {
-    handle_request(message, web::json::value(), std::bind(&MeasureManagerServer::internal_state, this, std::placeholders::_1));  // NOLINT
```
kbenne: Much nicer. Not sure why I made someting simple, more complicated.

src/cli/MeasureManager.cpp
```diff
@@ -634,63 +614,54 @@ void MeasureManagerServer::handle_post(web::http::http_request message) {
   const std::string uri = toString(web::http::uri::decode(message.relative_uri().path()));
```
kbenne, line 614: Can message be passed by reference?

kbenne (approved):

src/cli/MeasureManager.cpp
jmarrec: Should it be passed by reference is the question. All of the example code passes it by value.
Digging down the source code, this is a PIMPL class. Single member

```
    std::shared_ptr<http::details::_http_request> _m_impl;
```

So that'd be the size of two pointers (data + control block) right?

jmarrec: https://github.com/microsoft/cpprestsdk/blob/9c654889efb6f5bda69fe8f52b3f7d3d3fb56cd8/Release/samples/CasaLens/casalens.cpp#L85

jmarrec: Something trivially_copyable should be passed by val if it's < 16 bytes. The reference count needs to be +1 then -1 here since it's not a POD type, but I would reckon the effect are minimal. 

TL;DR: I don't think it matters. But if you prefer passing by ref / const ref, we can do that.

kbenne: Case closed.

jmarrec: I guess it is slightly faster though. Here's a benchmark: https://quick-bench.com/q/gek5jMY8CDeiaU41SZIU30beUjM

image

Same locally on my ubuntu gcc

jmarrec:
@kbenne mac CI has the same crash as I do locally when trying to run a Python measure first THEN a Ruby measure: https://ci.openstudio.net/blue/organizations/jenkins/openstudio-incremental-osx/detail/PR-4920/21/pipeline#step-76-log-7

kbenne:
> @kbenne mac CI has the same crash as I do locally when trying to run a Python measure first THEN a Ruby measure: https://ci.openstudio.net/blue/organizations/jenkins/openstudio-incremental-osx/detail/PR-4920/21/pipeline#step-76-log-7

I opened a new issue about this, because I think it is a separate issue from the Measure Manager itself. https://github.com/NREL/OpenStudio/issues/4945

jmarrec:
Definitely unrelated to this PR.
Should we merge this one then?

kbenne:
@kflemin is testing right now. After she is done, then I think we can merge @jmarrec

src/cli/MeasureManager.cpp
```diff
@@ -1001,6 +998,9 @@ void MeasureManagerServer::handle_request(const web::http::http_request& message
 }
 
 void MeasureManagerServer::do_tasks_forever() {
+  fmt::print("MeasureManager Ready");
+  fmt::print("Accepting requests on: {}\n", m_url);
+  std::cout << std::flush;
```
jmarrec, line 1003: Nitpick but having `fmt::print` followed by an old std::cout feels weird. 

```suggestion
  std::fflush(std::cout);
```

Or even `std::fflush(nullptr);` if you want to flush everything.

kbenne: I agree. It kind of smells to me also, but I don't think the fmt library has an API to flush. @kflemin was observing that the logs were coming in late, so I think we do actually need a flush for PAT to get the feedback it is looking for.

std::fflush(...) seems like an improvement to me over std::cout << std::flush. thanks!

PR #4923, opened 2023-07-04
Fix #4692 - Modify Model::load to use the VersionTranslator

### LINKED ISSUE #4692: Modify Model::load(path&) to use VersionTranslator by default
Enhancement Request
--------------

It's been like this for ever, but it's weird that Model::load just assumes this is at the right version.  It's very confusing to new users too, and they struggle with knowing how to load a model.

https://github.com/NREL/OpenStudio/blob/92da27b27de50e076ce3bdc5032e4f348e2231e8/src/model/Model.cpp#L1836-L1858

 It shouldn't be that verbose to load a model

https://github.com/NREL/OpenStudio-workflow-gem/blob/cf99e6096f33b717df79a71c245a4d7aa4ccb7a6/lib/openstudio/workflow/util/model.rb#L51-L70

## Detailed Description

At the very least we should provide an override or extra bool parameter.

## Possible Implementation

Straightforward

jmarrec:
Pull request overview
---------------------

 - Fixes #4692

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/model/Model.cpp
+++ b/src/model/Model.cpp
@@ -21,8 +21,11 @@
 #include <utilities/idd/IddEnums.hxx>
 #include <utilities/idd/OS_Version_FieldEnums.hxx>
 
+#include "../osversion/VersionTranslator.hpp"
+
 #include "../utilities/core/Assert.hpp"
 #include "../utilities/core/ContainersMove.hpp"
+#include "../utilities/core/Filesystem.hpp"
 #include "../utilities/core/PathHelpers.hpp"
 
 #include "../utilities/idd/IddEnums.hpp"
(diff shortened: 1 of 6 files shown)


src/model/Model.cpp
```diff
@@ -21,8 +21,11 @@
 #include <utilities/idd/IddEnums.hxx>
 #include <utilities/idd/OS_Version_FieldEnums.hxx>
 
+#include "../osversion/VersionTranslator.hpp"
```
jmarrec, line 24: So, there's one thing I'm not sure about here. This ends up referencing the osversion::VersionTranslator, when the osversion::VersionTranslator itself references the model/Schedule.hpp (for some 0.8.x translation). osversion depends on model at the model. But all of them are OBJECT libraries anyways, so that looks to link just fine in the end... @kbenne thoughts please?

src/model/Model.cpp
```diff
@@ -1840,27 +1843,26 @@ namespace model {
...
+    // Load the workflow.osw in the model's companion folder
+    const openstudio::path workflowJSONPath = getCompanionFolder(osmPath) / toPath("workflow.osw");
+    if (exists(workflowJSONPath)) {
+      if (boost::optional<WorkflowJSON> workflowJSON_ = WorkflowJSON::load(workflowJSONPath)) {
+        model_->setWorkflowJSON(*workflowJSON_);
       }
     }
 
-    return result;
+    return model_;
```
jmarrec, line 1865: New load method using the VersionTranslator.

PR #4950, opened 2023-08-29
Addresses #4946, allow TableLookup curves for Coil:*:WaterToAirHeatPump:EquationFit objects

### LINKED ISSUE #4946: WaterToAirHeatPumpEquationFit curves do not support Table:Lookup
Issue overview
--------------
CoilCoolingWaterToAirHeatPumpEquationFit and CoilHeatingWaterToAirHeatPumpEquationFit do not support specifying curve data using Table:Lookup.

They are supposed to accept \reference QuadvariateFunctions but the documentation shows only that Curve:QuadLinear and Curve:QuintLinear are allowed for their respective curve fields

### Context
This is needed for ComStock GSHP measure development
[coil_bug.rb.txt](https://github.com/NREL/OpenStudio/files/12443495/coil_bug.rb.txt)

joseph-robertson:
Pull request overview
---------------------

 - Fixes #4946

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/model/CoilHeatingWaterToAirHeatPumpEquationFit.cpp
+++ b/src/model/CoilHeatingWaterToAirHeatPumpEquationFit.cpp
@@ -623,67 +623,78 @@ namespace model {
   double CoilHeatingWaterToAirHeatPumpEquationFit::heatingCapacityCoefficient1() const {
     DEPRECATED_AT_MSG(3, 2, 0, "Use CoilHeatingWaterToAirHeatPumpEquationFit::heatingCapacityCurve().coefficient1Constant() instead.");
     Curve curve = getImpl<detail::CoilHeatingWaterToAirHeatPumpEquationFit_Impl>()->heatingCapacityCurve();
-    return curve.coefficient1Constant();
+    boost::optional<CurveQuadLinear> curveQuadLinear = curve.optionalCast<CurveQuadLinear>();
+    return curveQuadLinear->coefficient1Constant();
   }
 
   bool CoilHeatingWaterToAirHeatPumpEquationFit::setHeatingCapacityCoefficient1(double heatingCapacityCoefficient1) {
     DEPRECATED_AT_MSG(3, 2, 0, "Use CoilHeatingWaterToAirHeatPumpEquationFit::heatingCapacityCurve().setCoefficient1Constant(double) instead.");
     Curve curve = getImpl<detail::CoilHeatingWaterToAirHeatPumpEquationFit_Impl>()->heatingCapacityCurve();
-    return curve.setCoefficient1Constant(heatingCapacityCoefficient1);
+    boost::optional<CurveQuadLinear> curveQuadLinear = curve.optionalCast<CurveQuadLinear>();
+    return curveQuadLinear->setCoefficient1Constant(heatingCapacityCoefficient1);
   }
 
   double CoilHeatingWaterToAirHeatPumpEquationFit::heatingCapacityCoefficient2() const {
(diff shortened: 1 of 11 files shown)


src/model/CoilHeatingWaterToAirHeatPumpEquationFit.cpp
```diff
@@ -304,22 +306,22 @@ namespace model {
...
 
-    CurveQuadLinear CoilHeatingWaterToAirHeatPumpEquationFit_Impl::heatingCapacityCurve() const {
+    Curve CoilHeatingWaterToAirHeatPumpEquationFit_Impl::heatingCapacityCurve() const {
       WorkspaceObject wo = getTarget(OS_Coil_Heating_WaterToAirHeatPump_EquationFitFields::HeatingCapacityCurveName).get();
-      return wo.cast<CurveQuadLinear>();
+      return wo.cast<Curve>();
     }
 
-    bool CoilHeatingWaterToAirHeatPumpEquationFit_Impl::setHeatingCapacityCurve(const CurveQuadLinear& heatingCapacityCurve) {
+    bool CoilHeatingWaterToAirHeatPumpEquationFit_Impl::setHeatingCapacityCurve(const Curve& heatingCapacityCurve) {
```
jmarrec, line 314: This specifically expects a `QuadvariateFunctions` one. Should we check that `heatingCapacityCurve.numVariables() == 4` here?

Genuine question. Not saying we should necessarily, but since this is the general move of E+ to allow generic curve and table objects everywhere, this is bound to come up again, I think now's a good time to decide what to do/standardize. @kbenne @joseph-robertson 

**Pros:**

* Checking is good because it catches user errors early instead of letting E+ complain (via `CheckCurveDim`)

**Cons:**

* More work for us
* Options:
    * Option1: we are hardcoding a `curve.numVariables()` check in each setter that takes a generic `Curve`, which potentially could fall out of date (though in this case I don't see why the number of expected independent variables would change suddenly in E+)
    * Option 2: The alternative is to add a special case in `WorkspaceObject_Impl::setPointer` or a new method `WorkspaceObject_Impl::setCurvePointer`
        * A new check in `setPointer` is going to impose a performance penalty for all objects
        * A new method specifically for Curves means we gotta be careful to use this one specifically for curves

jmarrec: actually for option 2, the best option is probably to a new `bool ModelObject_Impl::setCurve(unsigned index, const Curve& curve)` method that would check the IDD references and do the numVariables() check for us before setting the handle via `WorkspaceObject_Impl::setPointer` (or `WorkspaceObject_Impl::setString` for speed).

That'd avoid having to do lots of shenanigans in Workspace land to figure out the number of variables a Table object has.

kbenne: Increasingly, I'm starting to appreciate some value in not making OpenStudio too smart and therefore letting EnergyPlus report certain errors. 

That said if we do go with validation in OpenStudio then option 2, but with a new ::setCurve method seems right. Paying the performance penality for every call to the existing ::setPointer seems unacceptable to me.

jmarrec: I'm opening an issue here, that we may just end up closing. I realize the Curve objects are already fine (via the `\reference` check), only the TableLookup has this issue:

* https://github.com/NREL/OpenStudio/issues/4982

jmarrec (approved):

PR #4964, opened 2023-09-11
ZoneHVAC:*: new Design Specification Multispeed Object fields

joseph-robertson:
Pull request overview
---------------------

Support new fields "Design Specification Multispeed Object Type" and "Design Specification Multispeed Object Name" for:
- [ ] `ZoneHVAC:WaterToAirHeatPump`
- [ ] `ZoneHVAC:TerminalUnit:VariableRefrigerantFlow`

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/resources/model/OpenStudio.idd
+++ b/resources/model/OpenStudio.idd
@@ -29063,13 +29063,17 @@ OS:ZoneHVAC:TerminalUnit:VariableRefrigerantFlow,
         \units C
         \required-field
         \note Supplemental heater will not operate when outdoor temperature exceeds this value.
-  A15; \field Controlling Zone or Thermostat Location
+  A15, \field Controlling Zone or Thermostat Location
        \type object-list
        \object-list ThermalZoneNames
        \note Used only for AirloopHVAC equipment on a main branch and defines zone name where thermostat is located.
        \note Not required for zone equipment. Leave blank if terminal unit is used in AirLoopHVAC:OutdoorAirSystem:EquipmentList.
        \note Required when terminal unit is used on main AirloopHVAC branch and coils are not set point controlled.
        \note When terminal unit is used in air loop and is load controlled, this zone's thermostat will control operation.
+  A16; \field Design Specification Multispeed Object Name
+       \type object-list
+       \object-list UnitarySystemPerformaceNames
+       \note The name of the performance specification object used to describe the multispeed coil.
 
 OS:ZoneHVAC:WaterToAirHeatPump,
         \min-fields 25
@@ -29236,10 +29240,14 @@ OS:ZoneHVAC:WaterToAirHeatPump,
         \note Constant results in 100% water flow regardless of compressor PLR
         \note Cycling results in water flow that matches compressor PLR
         \note ConstantOnDemand results in 100% water flow whenever the coil is on, but is 0% whenever the coil has no load
-   A26; \field Design Specification ZoneHVAC Sizing Object Name
+   A26, \field Design Specification ZoneHVAC Sizing Object Name
         \note Enter the name of a DesignSpecificationZoneHVACSizing object.
         \type object-list
         \object-list DesignSpecificationZoneHVACSizingName
+   A27; \field Design Specification Multispeed Object Name
+        \type object-list
+        \object-list UnitarySystemPerformaceNames
+        \note The name of the performance specification object used to describe the multispeed coil.
 
 OS:ZoneHVAC:UnitHeater,
        \min-fields 10
--- a/src/energyplus/ForwardTranslator/ForwardTranslateZoneHVACTerminalUnitVariableRefrigerantFlow.cpp
+++ b/src/energyplus/ForwardTranslator/ForwardTranslateZoneHVACTerminalUnitVariableRefrigerantFlow.cpp
@@ -440,8 +440,9 @@ namespace energyplus {
 
       if (_unitarySystemPerformance && _unitarySystemPerformance->name()) {
         idfObject.setString(ZoneHVAC_TerminalUnit_VariableRefrigerantFlowFields::DesignSpecificationMultispeedObjectType,
-                                _unitarySystemPerformance->iddObject().name());
-        idfObject.setString(ZoneHVAC_TerminalUnit_VariableRefrigerantFlowFields::DesignSpecificationMultispeedObjectName, _unitarySystemPerformance->name().get());
+                            _unitarySystemPerformance->iddObject().name());
+        idfObject.setString(ZoneHVAC_TerminalUnit_VariableRefrigerantFlowFields::DesignSpecificationMultispeedObjectName,
+                            _unitarySystemPerformance->name().get());
       }
     }
 
--- a/src/energyplus/ForwardTranslator/ForwardTranslateZoneHVACTerminalUnitVariableRefrigerantFlow.cpp
+++ b/src/energyplus/ForwardTranslator/ForwardTranslateZoneHVACTerminalUnitVariableRefrigerantFlow.cpp
@@ -29,6 +29,8 @@
 #include "../../model/CoilCoolingDXVariableRefrigerantFlow_Impl.hpp"
 #include "../../model/CoilHeatingDXVariableRefrigerantFlow.hpp"
 #include "../../model/CoilHeatingDXVariableRefrigerantFlow_Impl.hpp"
+#include "../../model/UnitarySystemPerformanceMultispeed.hpp"
+#include "../../model/UnitarySystemPerformanceMultispeed_Impl.hpp"
 #include "../../utilities/core/Logger.hpp"
 #include "../../utilities/core/Assert.hpp"
 #include "../../utilities/math/FloatCompare.hpp"
@@ -429,6 +431,17 @@ namespace energyplus {
         if (auto zone = translateAndMapModelObject(zones.front())) {
           idfObject.setString(ZoneHVAC_TerminalUnit_VariableRefrigerantFlowFields::ControllingZoneorThermostatLocation, zone->name().get());
         }
+      }
+    }
+
+    // Design Specification Multispeed Object Name
+    if (boost::optional<UnitarySystemPerformanceMultispeed> designSpecificationMultispeedObject = modelObject.designSpecificationMultispeedObject()) {
+      boost::optional<IdfObject> _unitarySystemPerformance = translateAndMapModelObject(designSpecificationMultispeedObject.get());
+
+      if (_unitarySystemPerformance && _unitarySystemPerformance->name()) {
+        idfObject.setString(ZoneHVAC_TerminalUnit_VariableRefrigerantFlowFields::DesignSpecificationMultispeedObjectType,
+                                _unitarySystemPerformance->iddObject().name());
+        idfObject.setString(ZoneHVAC_TerminalUnit_VariableRefrigerantFlowFields::DesignSpecificationMultispeedObjectName, _unitarySystemPerformance->name().get());
       }
     }
 
--- a/src/energyplus/ForwardTranslator/ForwardTranslateZoneHVACWaterToAirHeatPump.cpp
+++ b/src/energyplus/ForwardTranslator/ForwardTranslateZoneHVACWaterToAirHeatPump.cpp
@@ -27,6 +27,8 @@
 #include "../../model/CoilHeatingGas_Impl.hpp"
 #include "../../model/CoilHeatingWater.hpp"
 #include "../../model/CoilHeatingWater_Impl.hpp"
+#include "../../model/UnitarySystemPerformanceMultispeed.hpp"
+#include "../../model/UnitarySystemPerformanceMultispeed_Impl.hpp"
 #include <utilities/idd/ZoneHVAC_WaterToAirHeatPump_FieldEnums.hxx>
 #include <utilities/idd/OutdoorAir_Mixer_FieldEnums.hxx>
 #include <utilities/idd/Fan_OnOff_FieldEnums.hxx>
@@ -374,6 +376,17 @@ namespace energyplus {
     idfObject.setString(ZoneHVAC_WaterToAirHeatPumpFields::HeatPumpCoilWaterFlowMode, modelObject.heatPumpCoilWaterFlowMode());
 
     // TODO: field 'Design Specification ZoneHVAC Sizing' isn't implemented since the object isn't wrapped in SDK
+
+    // Design Specification Multispeed Object Name
+    if (boost::optional<UnitarySystemPerformanceMultispeed> designSpecificationMultispeedObject = modelObject.designSpecificationMultispeedObject()) {
+      boost::optional<IdfObject> _unitarySystemPerformance = translateAndMapModelObject(designSpecificationMultispeedObject.get());
+
+      if (_unitarySystemPerformance && _unitarySystemPerformance->name()) {
+        idfObject.setString(ZoneHVAC_WaterToAirHeatPumpFields::DesignSpecificationMultispeedObjectType,
+                                _unitarySystemPerformance->iddObject().name());
+        idfObject.setString(ZoneHVAC_WaterToAirHeatPumpFields::DesignSpecificationMultispeedObjectName, _unitarySystemPerformance->name().get());
+      }
+    }
 
     return idfObject;
   }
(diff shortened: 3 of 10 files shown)


resources/model/OpenStudio.idd
```diff
@@ -29063,13 +29063,17 @@ OS:ZoneHVAC:TerminalUnit:VariableRefrigerantFlow,
...
+  A15, \field Controlling Zone or Thermostat Location
        \type object-list
        \object-list ThermalZoneNames
        \note Used only for AirloopHVAC equipment on a main branch and defines zone name where thermostat is located.
        \note Not required for zone equipment. Leave blank if terminal unit is used in AirLoopHVAC:OutdoorAirSystem:EquipmentList.
        \note Required when terminal unit is used on main AirloopHVAC branch and coils are not set point controlled.
        \note When terminal unit is used in air loop and is load controlled, this zone's thermostat will control operation.
+  A16; \field Design Specification Multispeed Object Name
+       \type object-list
+       \object-list UnitarySystemPerformaceNames
```
jmarrec, line 29075: I know this is in E+ too, but 

```diff
-UnitarySystemPerformaceNames
+UnitarySystemPerformanceNames
```

Do a full file replace in there.

src/model/ZoneHVACWaterToAirHeatPump.cpp
```diff
@@ -77,6 +79,10 @@ namespace model {
 
       wahpClone.setSupplementalHeatingCoil(supplementalHeatingCoilClone);
 
+      if (auto designSpec = designSpecificationMultispeedObject()) {
+        wahpClone.setDesignSpecificationMultispeedObject(designSpec->clone(model).cast<UnitarySystemPerformanceMultispeed>());
+      }
```
jmarrec, line 84: Ok, AirLoopHVACUNitarySystem already clones / remove, so let's do the same

src/model/ZoneHVACWaterToAirHeatPump.cpp
```diff
@@ -1101,6 +1135,10 @@ namespace model {
     return getImpl<detail::ZoneHVACWaterToAirHeatPump_Impl>()->supplyAirFanOperatingModeSchedule();
   }
 
+  boost::optional<ZoneHVACWaterToAirHeatPump> AirLoopHVACUnitarySystem::designSpecificationMultispeedObject() const {
```
jmarrec, line 1138:
```suggestion
  boost::optional<UnitarySystemPerformanceMultispeed> ZoneHVACWaterToAirHeatPump::designSpecificationMultispeedObject() const {
```

src/energyplus/ForwardTranslator/ForwardTranslateZoneHVACTerminalUnitVariableRefrigerantFlow.cpp
```diff
@@ -432,6 +434,18 @@ namespace energyplus {
       }
     }
 
+    // Design Specification Multispeed Object Name
```
jmarrec, line 437: I don't see any FT testing for this one?

src/energyplus/Test/ZoneHVACWaterToAirHeatPump_GTest.cpp
```diff
@@ -269,6 +270,8 @@ TEST_F(EnergyPlusFixture, ForwardTranslator_ZoneHVACWaterToAirHeatPump) {
     EXPECT_TRUE(idf_hp.isEmpty(ZoneHVAC_WaterToAirHeatPumpFields::AvailabilityManagerListName));
     EXPECT_EQ("ConstantOnDemand", idf_hp.getString(ZoneHVAC_WaterToAirHeatPumpFields::HeatPumpCoilWaterFlowMode).get());
     EXPECT_TRUE(idf_hp.isEmpty(ZoneHVAC_WaterToAirHeatPumpFields::DesignSpecificationZoneHVACSizingObjectName));
+    EXPECT_TRUE(idf_hp.isEmpty(ZoneHVAC_WaterToAirHeatPumpFields::DesignSpecificationMultispeedObjectType));
+    EXPECT_TRUE(idf_hp.isEmpty(ZoneHVAC_WaterToAirHeatPumpFields::DesignSpecificationMultispeedObjectName));
```
jmarrec, line 274: Please actually assign one and check that it's all good.

jmarrec (changes requested):

src/energyplus/ForwardTranslator/ForwardTranslateZoneHVACTerminalUnitVariableRefrigerantFlow.cpp
joseph-robertson: @jmarrec I didn't get too far on this PR yet. I sort of got tripped up on FT. Appears that for AirLoopHVACUnitarySystem, it FTs the UnitarySystemPerformanceMultispeed object, but also _creates_ one if one doesn't exist. Under what conditions should we create one for ZoneHVACWaterToAirHeatPump and ZoneHVACTerminalUnitTVariableRefrigerantFlow?

jmarrec: AirLoopHVACUnitarySystem I believe throws if you use multispeed / variable speed coils without a UnitarySystemPerformanceMultispeed object, hence why the FT creates one if needed.

I'm not 100% sure whether the VRF TU and the ZoneHVAC:WaterToAirHeatPump have the same behavior now. If so, we should create one in FT. In that case, should move the default creation to a helper function so we can reuse in other modules.

jmarrec: oh, that's quite confusing, but it seems that this applies not to the coils but to the fans. The original NFP proposed a new object "FanPerformance:Multispeed", but apparently UnistarySystemPerformance:Multispeed was kept.

https://github.com/NREL/EnergyPlus/blob/develop/design/FY2023/NFP_MultispeedFans.md

My understanding is that this is purely optional, regardless of the coils you use for the ZoneHVAC:WAHP or ZoneHVAC:TU:VRF objects. We should probably make some dummy IDF tests to confirm. I'll ask the E+ dev team right now too.

joseph-robertson: @jmarrec I stubbed UnitarySystemPerformance FT to handle additional coil types `CoilXXXWaterToAirHeatPumpEquationFit`, `CoilXXXDXVariableRefrigerantFlow`, and `CoilXXXDXVariableRefrigerantFlowFluidTemperatureControl`. But I'm a little unclear on how to determine `NumberofSpeedsforXXX`. The speeds should not correspond to the coils, but to the supply fan? Are we going to need to create a new setter method for speeds on the UnitarySystemPerformance object?

jmarrec: My earlier post was misleading I think. The object varies the fan airflow in conjunction with the heating/cooling capacity.
So Number of Speeds for Heating should match the heating coil number of speeds, when the coil does have speeds.

I don't think we need to create a UnitarySystemPerformance object for VRF / ZoneHVACWaterToAirHeatPump if they don't throw when one isn't present. In that case, we're just left translating whatever the user **chose** to specify.


Also, take `VRFMultispeedFan.idf`for eg. You have one `ZoneHVAC:TerminalUnit:VariableRefrigerantFlow` (`TU1`) object with some `COIL:Cooling:DX:VariableRefrigerantFlow` and `COIL:Heating:DX:VariableRefrigerantFlow`. These coils do not have speed-related fields. Yet there is a UnitarySystemPerformance assigned to it.

joseph-robertson: I'm looking at `VRFMultispeedFan.idf`...
```
  UnitarySystemPerformance:Multispeed,
    Sys 1 WaterToAirHP System MultiSpeed Performance,  !- Name
    2,                       !- Number of Speeds for Heating
    2,                       !- Number of Speeds for Cooling
```
Where does the "2" come from? Should this value be hardcoded when coil object types are `COIL:*:DX:VariableRefrigerantFlow`?

jmarrec: That is arbitrary I believe.

src/energyplus/ForwardTranslator/ForwardTranslateUnitarySystemPerformanceMultispeed.cpp
```diff
@@ -34,44 +54,78 @@ namespace openstudio {
...
+      ZoneHVACTerminalUnitVariableRefrigerantFlow& tuVRF = tuVRFs[0];
+      heatingCoil = tuVRF.heatingCoil();
+      coolingCoil = tuVRF.coolingCoil();
+    }
+
+    if (hps.size() == 1) {
+      ZoneHVACWaterToAirHeatPump& hp = hps[0];
+      heatingCoil = hp.heatingCoil().optionalCast<HVACComponent>();
+      coolingCoil = hp.coolingCoil().optionalCast<HVACComponent>();
+    }
```
jmarrec, line 94:
```suggestion
    if (unitarySystems.size() == 1) {
      AirLoopHVACUnitarySystem& unitarySystem = unitarySystems.front();
      heatingCoil = unitarySystem.heatingCoil();
      coolingCoil = unitarySystem.coolingCoil();
    } else if (tuVRFs.size() == 1) {
      ZoneHVACTerminalUnitVariableRefrigerantFlow& tuVRF = tuVRFs.front();
      heatingCoil = tuVRF.heatingCoil();
      coolingCoil = tuVRF.coolingCoil();
    } else {  // if (hps.size() == 1)
      ZoneHVACWaterToAirHeatPump& hp = hps.front();
      heatingCoil = hp.heatingCoil().optionalCast<HVACComponent>();
      coolingCoil = hp.coolingCoil().optionalCast<HVACComponent>();
    }
```

src/energyplus/ForwardTranslator/ForwardTranslateUnitarySystemPerformanceMultispeed.cpp
```diff
@@ -34,44 +54,78 @@ namespace openstudio {
...
         auto heatingCoilGasMultiStage = heatingCoil->cast<CoilHeatingGasMultiStage>();
         sysPerf.setInt(UnitarySystemPerformance_MultispeedFields::NumberofSpeedsforHeating, heatingCoilGasMultiStage.stages().size());
-      } else {
+      } else if (heatingCoil->iddObjectType() == openstudio::IddObjectType::OS_Coil_Heating_DX_VariableSpeed) {
+        auto heatingCoilDXVarspeed = heatingCoil->cast<CoilHeatingDXVariableSpeed>();
+        sysPerf.setInt(UnitarySystemPerformance_MultispeedFields::NumberofSpeedsforHeating, heatingCoilDXVarspeed.speeds().size());
+      } else if (heatingCoil->iddObjectType() == openstudio::IddObjectType::OS_Coil_Heating_WaterToAirHeatPump_VariableSpeedEquationFit) {
+        auto heatingCoilWaterToAirHeatingVar = heatingCoil->cast<CoilHeatingWaterToAirHeatPumpVariableSpeedEquationFit>();
+        sysPerf.setInt(UnitarySystemPerformance_MultispeedFields::NumberofSpeedsforHeating, heatingCoilWaterToAirHeatingVar.speeds().size());
+      } else {  // e.g., CoilHeatingWaterToAirHeatPumpEquationFit, CoilHeatingDXVariableRefrigerantFlow, CoilHeatingDXVariableRefrigerantFlowFluidTemperatureControl
```
jmarrec, line 110: If you run an AirLoopHVACUnitarySystem (or a ZoneHVACWAHP) with these coils and no UnitarySystemPerformance object, does E+ throw?

Just curious if you tested it, and same question for the below cooling coils below

src/energyplus/ForwardTranslator/ForwardTranslateZoneHVACTerminalUnitVariableRefrigerantFlow.cpp
```diff
@@ -432,6 +434,18 @@ namespace energyplus {
...
+    if (boost::optional<UnitarySystemPerformanceMultispeed> designSpecificationMultispeedObject = modelObject.designSpecificationMultispeedObject()) {
+      boost::optional<IdfObject> _unitarySystemPerformance = translateAndMapModelObject(designSpecificationMultispeedObject.get());
+
+      if (_unitarySystemPerformance && _unitarySystemPerformance->name()) {
+        idfObject.setString(ZoneHVAC_TerminalUnit_VariableRefrigerantFlowFields::DesignSpecificationMultispeedObjectType,
+                            _unitarySystemPerformance->iddObject().name());
+        idfObject.setString(ZoneHVAC_TerminalUnit_VariableRefrigerantFlowFields::DesignSpecificationMultispeedObjectName,
+                            _unitarySystemPerformance->name().get());
+      }
+    }
```
jmarrec, line 447: This is pointless here for the VRF object, because you don't have a way to set the number of Speeds during Heating/Cooling, and the VRF coils do not bear the requested information to autofill them.

Given what we see in `VRFMultispeedFan.idf` that means:

* We actually need to expose the number of heating/cooling speeds
* We also need the extensible portion `<Heating|Cooling> Speed <N> Supply Air Flow Ratio`
* We probably need to throw or overwrite when writing for a coil that has an actual number of speeeds
* We probably need to reach out to the E+ devs responsible for this feature to understand it fully. I'm quite fuzzy on how this is used with coils that don't have an actual number of speeds

src/energyplus/ForwardTranslator/ForwardTranslateZoneHVACWaterToAirHeatPump.cpp
```diff
@@ -355,6 +357,17 @@ namespace energyplus {
...
+    // Design Specification Multispeed Object Name
+    if (boost::optional<UnitarySystemPerformanceMultispeed> designSpecificationMultispeedObject = modelObject.designSpecificationMultispeedObject()) {
+      boost::optional<IdfObject> _unitarySystemPerformance = translateAndMapModelObject(designSpecificationMultispeedObject.get());
+
+      if (_unitarySystemPerformance && _unitarySystemPerformance->name()) {
+        idfObject.setString(ZoneHVAC_WaterToAirHeatPumpFields::DesignSpecificationMultispeedObjectType,
+                            _unitarySystemPerformance->iddObject().name());
+        idfObject.setString(ZoneHVAC_WaterToAirHeatPumpFields::DesignSpecificationMultispeedObjectName, _unitarySystemPerformance->name().get());
+      }
+    }
```
jmarrec, line 369: Unlike VRF, there are a varierty of coils when we can autofill the `Number of Speeds for <Heating|Cooling>` fields nor the `<Heating|Cooling> Speed <N> Supply Air Flow Ratio`.
One, or maybe the only one, where we cannot is the CoilHeatingWaterToAirHeatPumpEquationFit, and perhaps that's fine. I'm not 100% sure

jmarrec (changes requested):
This is a tough one and I'm not sure I completely understand how this object is supposed to be used when not in a AirLoopHVAC:UnitarySystem.

This does not enable using the object with VRF objects at least (eg, `VRFMultispeedFan.idf` is not reproducible). Seems like it'd work for the ZoneHVACWaterToAirHeatPump in most cases though.

For VRF we'd need to expose the `Number of Speeds for <Heating|Cooling>` fields nor the `<Heating|Cooling> Speed <N> Supply Air Flow Ratio` fields. Which may be problematic for other objects because that means we need to handle the case where what the user manually entered is inconsistent with the VSD coil's number of Speeds.

src/energyplus/ForwardTranslator/ForwardTranslateZoneHVACTerminalUnitVariableRefrigerantFlow.cpp
joseph-robertson: Couple of things:
(1) I had noticed the "gap" on the heating/cooling speeds - that's why I had posed the question "Are we going to need to create a new setter method for speeds on the UnitarySystemPerformance object?" above.
(2) I thought you had said that the unitary system performance speeds in `VRFMultispeedFan.idf` were arbitrary?

But generally speaking, yes, I agree that we should probably punt on this new feature until later.

src/energyplus/ForwardTranslator/ForwardTranslateUnitarySystemPerformanceMultispeed.cpp
joseph-robertson: Because of the `if (((unitarySystems.size() + tuVRFs.size() + hps.size()) != 1)) {` line above, yes, I think we'd be able to collapse to your suggestion.

PR #4973, opened 2023-09-19
Improve OpenStudio's support for Python unitttest

### LINKED ISSUE #4906: Support Testing of Python Measures with standard OpenStudio install
Enhancement Request
--------------
example Python measures included with OpenStudio use `pytest` for the measure tests. `pytest` isn't included in the EnergyPlus Python that is used by the OpenStudio CLI when `openstudio labs my_test.py` is run.

## Detailed Description
While a clean OpenStudio installer can run OSW file with labs that runs fine, there isn't way currently to run measure tests. We want to make it easy for novice measure writes to write and use tests when developing or altering Python measures, as we have done for Ruby measures. The goal here is to provide update the example measure tests, and make any changes necessary so the example Python measure, and other Python measures that do not use external packages can be tested with out of the box OpenStudio installer.

## Possible Implementation
Preferred solution is to change to use `unittest` and fix issues that exist. When I try to `import unittest` on non-developer mac OpenStudio I get the following error. I have not tested on Windows.
```
  File "/Users/dgoldwas/Documents/github/nrel/openstudio-common-measures-gem/lib/measures/IncreaseRoofRValuePython/tests/test_measure.py", line 2, in <module>
    import unittest
  File "/Applications/OpenStudio-3.6.0/EnergyPlus/python_standard_lib/unittest/__init__.py", line 60, in <module>
    from .async_case import IsolatedAsyncioTestCase
  File "/Applications/OpenStudio-3.6.0/EnergyPlus/python_standard_lib/unittest/async_case.py", line 1, in <module>
    import asyncio
  File "/Applications/OpenStudio-3.6.0/EnergyPlus/python_standard_lib/asyncio/__init__.py", line 8, in <module>
    from .base_events import *
  File "/Applications/OpenStudio-3.6.0/EnergyPlus/python_standard_lib/asyncio/base_events.py", line 23, in <module>
    import socket
  File "/Applications/OpenStudio-3.6.0/EnergyPlus/python_standard_lib/socket.py", line 49, in <module>
    import _socket
ModuleNotFoundError: No module named '_socket'
```

If we can't get `unittest` to work, it would be fine as an alternative to add `pytest` to OpenStudio's Python, so that the existing example Python Measure tests work as written.

kbenne:
In the Python unittest library, unittest.main() was not working as expected, because tests defined in the __main__ scope where not being detected. This meant that `openstudio labs execute_python_script foo_measure_test.rb` did not work as expected.

The solution here is to register the __main__ module properly within Python.

* An example Measure test is located at https://github.com/NREL/openstudio-common-measures-gem/commit/0c73cdb0dcde1eac960628e3c4ee853f18ef0f3f
* A Measure test can be invoked with `openstudio labs --python_path=. execute_python_script tests/test_measure.py`

close #4906

Pull request overview
---------------------

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/cli/UpdateCommand.cpp
+++ b/src/cli/UpdateCommand.cpp
@@ -104,8 +104,10 @@ sys.argv.append("{}")
     }
     cmd += fmt::format(R"python(
 import importlib.util
-spec = importlib.util.spec_from_file_location('__main__', r'{}')
+module_name = '__main__'
+spec = importlib.util.spec_from_file_location(module_name, r'{}')
 module = importlib.util.module_from_spec(spec)
+sys.modules[module_name] = module
 spec.loader.exec_module(module)
 )python",
                        pythonScriptPath.generic_string());


jmarrec:
I don't want to sound like a broken record, but I would urge we standardize around `pytest` and not `unittest`. yes, I now this isn't part of the standard library, but I think we should just pip install it into the EnergyPlus/python_standard_lib directory.

src/cli/UpdateCommand.cpp
```diff
@@ -104,8 +104,10 @@ sys.argv.append("{}")
     }
     cmd += fmt::format(R"python(
 import importlib.util
-spec = importlib.util.spec_from_file_location('__main__', r'{}')
+module_name = '__main__'
+spec = importlib.util.spec_from_file_location(module_name, r'{}')
 module = importlib.util.module_from_spec(spec)
+sys.modules[module_name] = module
```
jmarrec, line 110: Do you have a link to somewhere where they do this `sys.modules["__main__"] = module` or something to convince me this is the right way to do it? Don't want to go chasing info if you already have it.

I have a _completely non-justified_ feeling this might cause trouble in some edge case we haven't foreseen, when this isn't a unittest file or something.

kbenne: You could be right. Here is my justification though.

https://docs.python.org/3/library/__main__.html#import-main

```
Python inserts an empty __main__ module in [sys.modules](https://docs.python.org/3/library/sys.html#sys.modules) at interpreter startup, and populates it by running top-level code....
```

jmarrec: https://stackoverflow.com/a/62492464/2543372

jmarrec: https://github.com/python/cpython/blob/754519a9f8c2bb06d85ff9b3e9fe6f967ac46d5c/Lib/unittest/loader.py#L336-L338

jmarrec (approved):
Assuming CI comes clean, this will merge

jmarrec:
Just the usual suspect OpenStudioCLI.Labs.test_logger_rb (Failed) on windows. Thanks @kbenne

Issue #5002, opened 2023-10-21
v370-rc1 : too stringent polyhedron checks?


brgix:
Context
--------------
[OSut](https://github.com/rd2/osut) holds a number of geometry methods, as well as supporting RSpecs like "[checks for space/surface convexity](https://github.com/rd2/osut/blob/63719558293794ee569f79d78e16bc1ad11817f1/spec/osut_tests_spec.rb#L2789)". The latter tests OpenStudio behaviour when adding _holes_ cut in (larger) surfaces. The intention here is to first test, prior to developing a general OSut method (accommodating either other smaller surfaces or shafts, e.g. skylight wells). Roughly half of the comments/illustrations of this merged [PR](https://github.com/rd2/osut/pull/13) describe encountered pitfalls/workarounds for a handful of scenarios. The (simple) solution rests on properly winding [extra](https://github.com/rd2/osut/pull/13#discussion_r1276124965) vertices (of the larger surface) around a new _hole_.

<img width="725" alt="osut" src="https://github.com/NREL/OpenStudio/assets/98104985/4c918476-da33-4f3c-b6d8-783d784361ca">

## Current Behavior
As reported [here](https://github.com/rd2/osut/pull/15), the RSpec fails with v3.7.0-rc1 calculated volume for irregular geometries (SmallOffice, Attic):

```
Failure/Error: expect(attic.volume).to be_within(TOL).of(720.19) unless v < 350
       expected 798.4139451428566 to be within 0.01 of 720.19
```

720.19 m3 is the correct volume. It is my understanding that v3.7.0-rc1 reverts to pre-v3.5.0 volume calculations due to (revised) polyhedron checks (i.e. more stringent than v3.6.1, than EnergyPlus v23.2.0 ?):

```
[utilities.Polyhedron] <0> Polyhedron is not enclosed in original testing. Trying to add missing colinear points.
[utilities.Polyhedron] <0> Polyhedron is not enclosed.
[openstudio.model.Space] <0> Object of type 'OS:Space' and named 'Core_ZN' is not enclosed, there are 2 edges that aren't used exactly twice. Falling back to ceilingHeight * floorArea. Volume calculation will be potentially inaccurate.
[utilities.Polyhedron] <0> Polyhedron is not enclosed in original testing. Trying to add missing colinear points.
[utilities.Polyhedron] <0> Polyhedron is not enclosed.
[...]

PR #5003, opened 2023-10-22
Allow edges to have same surface on both sides

macumber:
Pull request overview
---------------------

 - Fixes https://github.com/NREL/OpenStudio/issues/5002

Supports polygons with interior holes cut out with reverse winding, this fixes a regression from 3.6.0 that was discovered in 3.7.0-rc1.

### Pull Request Author

No API or IDD changes needed

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/utilities/geometry/Polyhedron.cpp
+++ b/src/utilities/geometry/Polyhedron.cpp
@@ -242,9 +242,6 @@ void Polyhedron::performEdgeMatching() {
 
   for (size_t i = 0; i < m_surfaces.size(); ++i) {
     for (size_t j = 0; j < m_surfaces.size(); ++j) {
-      if (i == j) {
-        continue;
-      }
       auto& surface1 = m_surfaces[i];
       auto& surface2 = m_surfaces[j];
       for (Surface3dEdge& edge1 : surface1.edges) {
--- a/src/utilities/geometry/Polyhedron.cpp
+++ b/src/utilities/geometry/Polyhedron.cpp
@@ -243,7 +243,7 @@ void Polyhedron::performEdgeMatching() {
   m_hasAnySurfaceWithIncorrectOrientation = false;
 
   // We use **Combinations** (rather than Permutations) to avoid traversing unnecessarily
-  for (size_t i = 0; i < m_surfaces.size() - 1; ++i) {
+  for (size_t i = 0; i < m_surfaces.size(); ++i) {
     for (size_t j = i + 1; j < m_surfaces.size(); ++j) {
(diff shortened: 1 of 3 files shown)


src/utilities/geometry/Polyhedron.cpp
```diff
@@ -268,6 +268,25 @@ void Polyhedron::performEdgeMatching() {
       }
     }
   }
+
+  // special case to find edges that are used to "cut" in to a surface to remove an interior hole
```
jmarrec, line 272:
```suggestion
  // #5002 - special case to find edges that are used to "cut" in to a surface to remove an interior hole
```

jmarrec (commented):
Just going to make a dumb change to retrigger a CI pass

jmarrec:
@macumber thanks for the fix. I wonder if we shouldn't just erase these edges in the polyhedron instead of retaining them and putting them being shared twice to the same surface. Have you tried it by any change?

src/utilities/geometry/Polyhedron.cpp
```diff
@@ -268,6 +268,25 @@ void Polyhedron::performEdgeMatching() {
       }
     }
   }
+
+  // #5002 - special case to find edges that are used to "cut" in to a surface to remove an interior hole
+  // we allow these edges to double count the first surface since they bound the same surface on two sides
+  for (auto& surface : m_surfaces) {
+    auto& edges = surface.edges;
+    for (size_t i = 0; i < edges.size(); ++i) {
+      for (size_t j = 0; j < edges.size(); ++j) {
```
jmarrec, line 277: I know I'm guilty here because I've done it a couple of lines above when looping on surfaces, but we can avoid traversing that much by just doing **combinations** (instead of **permutations**)

```suggestion
    for (size_t i = 0; i < edges.size() - 1; ++i) {
      for (size_t j = i + 1; j < edges.size(); ++j) {
```

jmarrec:
I guess I botched something in f2c25372e25533c98c8015cb874769600bf779e1 since tests are failing now, but I can't see what I did wrong right now...

jmarrec:
Ok I see. https://github.com/NREL/OpenStudio/blob/6fb3dedab53cc873f2b9a53417cbc361af335015/src/gbxml/Test/ForwardTranslator_GTest.cpp#L999-L1002

This has zero surfaces.

So the loop goes: `for (size_t i = 0; i < -1; ++i)`, so trying to compare a size_t to a negative thing... `static_cast<size_t>(-1) ==. 18446744073709551615`

PR #5011, opened 2023-10-31
Improve weather file handling in OS Workflow

### LINKED ISSUE #5009: The new (C++ based) OpenStudio workflow fails if weather_file is not defined in the osw
The Ruby based workflow-gem would fallback to looking for the weather file location using the OpenStudio Model's `Model::weatherFile`. The new C++ workflow does not appear to do this, so if `weather_file` is not defined in the osw the workflow will fail. 

See WFG logic here: https://github.com/NREL/OpenStudio-workflow-gem/blob/696306b5b24cd597edc6ab04faa1e618f99db190/lib/openstudio/workflow/jobs/run_initialization.rb#L147

kbenne:
Now we make sure to look in the Model for the final weather file defined after all Measures have run.

We also still iniitialize the weather file from the OSW if one is defined.

close #5009

Pull request overview
---------------------

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/workflow/RunTranslator.cpp
+++ b/src/workflow/RunTranslator.cpp
@@ -38,8 +38,6 @@ void OSWorkflow::runTranslator() {
   auto runDir = workflowJSON.absoluteRunDir();
   OS_ASSERT(openstudio::filesystem::is_directory(runDir));
 
-  updateLastWeatherFileFromModel();
-
   // Copy in the weather file
   if (!epwPath.empty()) {
     openstudio::filesystem::copy_file(epwPath, runDir / "in.epw", openstudio::filesystem::copy_options::overwrite_existing);
--- a/src/workflow/RunTranslator.cpp
+++ b/src/workflow/RunTranslator.cpp
@@ -38,6 +38,8 @@ void OSWorkflow::runTranslator() {
   auto runDir = workflowJSON.absoluteRunDir();
   OS_ASSERT(openstudio::filesystem::is_directory(runDir));
 
+  applyWeatherFileFromModel();
+
   // Copy in the weather file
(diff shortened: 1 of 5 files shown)


src/workflow/RunTranslator.cpp
```diff
@@ -38,6 +38,8 @@ void OSWorkflow::runTranslator() {
   auto runDir = workflowJSON.absoluteRunDir();
   OS_ASSERT(openstudio::filesystem::is_directory(runDir));
 
+  applyWeatherFileFromModel();
```
jmarrec, line 41: Workflow gem seem to be doing it in APplyMeasure when it's a ModelMeasure: https://github.com/NREL/OpenStudio-workflow-gem/blob/696306b5b24cd597edc6ab04faa1e618f99db190/lib/openstudio/workflow/util/measure.rb#L551-L571

jmarrec: I believe you actually need to complete this unfinished block instead: https://github.com/NREL/OpenStudio/blob/develop/src/workflow/ApplyMeasure.cpp#L215-L224

kbenne: Right. I forgot about Runner::lastEpwFile. But I think it still needs to happen before translation too, because don't we need to account for the fact that the last ModelMeasure could have changed the weather location?

jmarrec: Updating last runner epwfile is done prior to measure run currently. The unfinished code block I link to is supposed to be executed after the measure run.
I'm not against changing what workflow-gem does, but it's 1) easier to Match and 2) clearer and probably 3) less error prone

PR #5023, opened 2023-11-10
Fix #5018 - C++ CLI: Calling a non existing method in a measure ends up with a crash and stack trace

### LINKED ISSUE #5018: C++ CLI: Calling a non existing method in a measure ends up with a crash and stack trace
Issue overview
--------------

## Current Behavior

For example, if you put in a measure something like:

```
  def run(model, runner, user_arguments)
    super(model, runner, user_arguments)

    runner.nonExistingMethod()
```

You end up with a hard crash.

## Expected Behavior

It should gracefully fail and report the error.

## Steps to Reproduce

See Current behavior

## Possible Solution

The issue is that we do `ScriptObject->run(args)` whereas the ruby CLI was all under a `rb_protect` block.

We need to find a way to do things like RubyInterpreter::makeProtectedCall or RubyEval's `evalString` (the Ruby Engine has exec / eval that call evalString).

### Environment

 - Version of OpenStudio (if using an intermediate build, include SHA): 3.7.0-rc2

### Context

- #5016

jmarrec:
Pull request overview
---------------------

 - Fixes #5018 
 - @kbenne saved the day by figuring out we could use the SWIG `%feature("director:except")`
 - I just implemented the error messages

* [ ] **TODO:** Does this stuff need to be moved at a (much) higher level, so it applies to every `SwigDirector` we have, not just in `Measure.i`. Maybe in utilities/core?
    * I think it isn't needed, because the only ruby/python stuff we call directly from C++ is the measure methods, but I could be wrong.
    * @kbenne thoughts please 

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/python/engine/test/BadMeasure/measure.py
+++ b/python/engine/test/BadMeasure/measure.py
@@ -1,0 +1,33 @@
+import typing
+
+import openstudio
+
+
+class BadMeasure(openstudio.measure.ModelMeasure):
+    def name(self):
+        return "Bad Measure"
+
+    def modeler_description(self):
+        return "The arguments method calls another_method which does a raise ValueError"
+
+    def another_method(self):
+        raise ValueError("oops")
+
+    def arguments(self, model: typing.Optional[openstudio.model.Model] = None):
+        self.another_method()
+        args = openstudio.measure.OSArgumentVector()
+
+        return args
+
+    def run(
+        self,
+        model: openstudio.model.Model,
+        runner: openstudio.measure.OSRunner,
+        user_arguments: openstudio.measure.OSArgumentMap,
+    ):
+        """
+        define what happens when the measure is run
+        """
+        super().run(model, runner, user_arguments)  # Do **NOT** remove this line
+
+BadMeasure().registerWithApplication()
--- a/python/engine/test/PythonEngine_GTest.cpp
+++ b/python/engine/test/PythonEngine_GTest.cpp
@@ -51,15 +51,13 @@ TEST_F(PythonEngineFixture, BadMeasure) {
   ASSERT_EQ(measurePtr->name(), "Bad Measure");
 
   std::string expected_exception = fmt::format(R"(SWIG director method error. In method 'arguments': `ValueError('oops')`
-Python traceback follows:
-```
+
 Traceback (most recent call last):
   File "{0}", line 17, in arguments
     self.another_method()
   File "{0}", line 14, in another_method
     raise ValueError("oops")
-ValueError: oops
-```)",
+ValueError: oops)",
                                                scriptPath.generic_string());
 
   openstudio::model::Model model;
@@ -84,13 +82,47 @@ TEST_F(PythonEngineFixture, WrongMethodMeasure) {
 
   std::string expected_exception =
     fmt::format(R"(SWIG director method error. In method 'arguments': `AttributeError("'Model' object has no attribute 'nonExistingMethod'")`
-Python traceback follows:
-```
+
 Traceback (most recent call last):
   File "{}", line 14, in arguments
     model.nonExistingMethod()
-AttributeError: 'Model' object has no attribute 'nonExistingMethod'
-```)",
+AttributeError: 'Model' object has no attribute 'nonExistingMethod')",
+                scriptPath.generic_string());
+
+  openstudio::model::Model model;
+  try {
+    measurePtr->arguments(model);
+    ASSERT_FALSE(true) << "Expected measure arguments(model) to throw";
+  } catch (std::exception& e) {
+    std::string error = e.what();
+    EXPECT_EQ(expected_exception, error);
+  }
+}
+
+TEST_F(PythonEngineFixture, StackLevelTooDeepMeasure) {
+
+  const std::string classAndDirName = "StackLevelTooDeepMeasure";
+
+  const auto scriptPath = getScriptPath(classAndDirName);
+  auto measureScriptObject = (*thisEngine)->loadMeasure(scriptPath, classAndDirName);
+  auto* measurePtr = (*thisEngine)->getAs<openstudio::measure::ModelMeasure*>(measureScriptObject);
+
+  ASSERT_EQ(measurePtr->name(), "Stack Level Too Deep Measure");
+
+  std::string expected_exception =
+    fmt::format(R"(SWIG director method error. In method 'arguments': `RecursionError('maximum recursion depth exceeded')`
+
+Traceback (most recent call last):
+  File "{0}", line 16, in arguments
+    s(10)
+  File "{0}", line 6, in s
+    return s(x)
+  File "{0}", line 6, in s
+    return s(x)
+  File "{0}", line 6, in s
+    return s(x)
+  [Previous line repeated 996 more times]
+RecursionError: maximum recursion depth exceeded)",
                 scriptPath.generic_string());
 
   openstudio::model::Model model;
--- a/python/engine/test/PythonEngine_GTest.cpp
+++ b/python/engine/test/PythonEngine_GTest.cpp
@@ -1,0 +1,104 @@
+/***********************************************************************************************************************
+*  OpenStudio(R), Copyright (c) Alliance for Sustainable Energy, LLC.
+*  See also https://openstudio.net/license
+***********************************************************************************************************************/
+
+#include <gtest/gtest.h>
+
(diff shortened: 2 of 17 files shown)


src/measure/Measure.i
```diff
@@ -96,6 +96,55 @@
 };
 
 #if defined SWIGRUBY
+  %feature("director:except") {
+    // Mimic openstudio::evalString for a possible approach to reporting more info or a callstack.
```
jmarrec, line 100: For ruby, mimic the https://github.com/NREL/OpenStudio/blob/c0cbe73b51f34c459d67a17da36739795bbf25fc2/ruby/interpreter/RubyEval.hpp#L18

src/measure/Measure.i
```diff
@@ -96,6 +96,55 @@
...
+
+    // I just **cannot** figure out a way to get the error location in C, without calling $@.to_s. Seems nothing is available here
+    int locationError;
+    VALUE locval = rb_eval_string_protect("$@.to_s", &locationError);
+    std::string loc;
+    if (locationError == 0) {
+      loc = StringValuePtr(locval);
+    } else {
+      loc = "Failed to get VM location";
+    }
```
jmarrec, line 119: This bit seems redundant with the backtrace block below, you get the same info twice.

src/measure/Measure.i
```diff
@@ -107,4 +156,76 @@
...
+          }
+
+          // PyList_Size returns a borrowed reference, do not decrement
+          Py_DECREF(pyth_val); // PyObject_CallFunction returns new reference, decrement
+      }
+      Py_DECREF(pyth_func); // PyObject_GetAttrString returns a new reference, decrement it
+      throw Swig::DirectorMethodException(err_msg.c_str());
+    }
+  }
+#endif
```
jmarrec, line 229: For python, I repurposed what Edwin and I wrote for EnergyPlus's `PluginInstance::reportPythonError` (cf https://github.com/NREL/EnergyPlus/pull/8181 for where I implemented the traceback bit)

python/engine/test/WrongMethodMeasure/measure.py
```diff
@@ -0,0 +1,30 @@
...
+class WrongMethodMeasure(openstudio.measure.ModelMeasure):
+    def name(self):
+        return "Wrong Method Measure"
+
+    def modeler_description(self):
+        return "The arguments method calls a non existing method on the model passed as argument"
+
+    def arguments(self, model: typing.Optional[openstudio.model.Model] = None):
+        model.nonExistingMethod()
+        args = openstudio.measure.OSArgumentVector()
```
jmarrec, line 15: Added a test for both python and ruby that calls a wrong method on model passed as argument in arguments()

python/engine/test/BadMeasure/measure.py
```diff
@@ -0,0 +1,33 @@
...
+        return "Bad Measure"
+
+    def modeler_description(self):
+        return "The arguments method calls another_method which does a raise ValueError"
+
+    def another_method(self):
+        raise ValueError("oops")
+
+    def arguments(self, model: typing.Optional[openstudio.model.Model] = None):
+        self.another_method()
```
jmarrec, line 17: Both Python and Ruby have this failed measure, that calls another method that throws, so we can check if the traceback is good

ruby/engine/test/RubyEngine_GTest.cpp
```diff
@@ -0,0 +1,106 @@
...
+    (*thisEngine)->registerType<openstudio::measure::ModelMeasure*>("openstudio::measure::ModelMeasure *");
+    (*thisEngine)->exec("OpenStudio::init_rest_of_openstudio()");
+  }
+  // tear down after each test
+  virtual void TearDown() override {
+    // Want to ensure the engine is reset regardless of the test outcome, or it may throw a segfault
+    thisEngine->reset();
+  }
+
+  std::unique_ptr<openstudio::ScriptEngineInstance> thisEngine;
```
jmarrec, line 49: FYI @kbenne I slightly changed the test you wrote so that the fixture will always reset the engine, regardless of whether the test passed or failed.

ruby/engine/test/RubyEngine_GTest.cpp
```diff
@@ -0,0 +1,106 @@
...
+{0}:16:in `arguments')",
+                                               scriptPath.generic_string());
+
+  openstudio::model::Model model;
+  try {
+    measurePtr->arguments(model);
+    ASSERT_FALSE(true) << "Expected measure arguments(model) to throw";
+  } catch (std::exception& e) {
+    std::string error = e.what();
+    EXPECT_EQ(expected_exception, error);
```
jmarrec, line 76: Ruby traceback example
```
SWIG director method error. RuntimeError: oops
["/Users/julien/Software/Others/OpenStudio2/ruby/engine/test/BadMeasure/measure.rb:12:in `another_method'", "/Users/julien/Software/Others/OpenStudio2/ruby/engine/test/BadMeasure/measure.rb:16:in `arguments'"]

Traceback:
/Users/julien/Software/Others/OpenStudio2/ruby/engine/test/BadMeasure/measure.rb:12:in `another_method'
/Users/julien/Software/Others/OpenStudio2/ruby/engine/test/BadMeasure/measure.rb:16:in `arguments'
```

jmarrec: Hence why I just said the `"$@.to_s"` in EvalString (and now measure.i) is redundant if the traceback is available.

ruby/engine/test/RubyEngine_GTest.cpp
```diff
@@ -0,0 +1,106 @@
...
+{0}:12:in `arguments')",
+                scriptPath.generic_string());
+
+  openstudio::model::Model model;
+  try {
+    measurePtr->arguments(model);
+    ASSERT_FALSE(true) << "Expected measure arguments(model) to throw";
+  } catch (std::exception& e) {
+    std::string error = e.what();
+    EXPECT_EQ(expected_exception, stripAddressFromErrorMessage(error));
```
jmarrec, line 104: I gotta strip the "0x00283498743eff" address here because it naturally changes everry time

python/engine/test/PythonEngine_GTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+  std::string expected_exception = fmt::format(R"(SWIG director method error. In method 'arguments': `ValueError('oops')`
+Python traceback follows:
+```
+Traceback (most recent call last):
+  File "{0}", line 17, in arguments
+    self.another_method()
+  File "{0}", line 14, in another_method
+    raise ValueError("oops")
+ValueError: oops
+```)",
```
jmarrec, line 62: Python tracebacks!

```
RuntimeError: SWIG director method error. In method 'arguments': `ValueError('oops')`
Python traceback follows:
` ` `
Traceback (most recent call last):
  File "/Users/julien/Software/Others/OpenStudio2/python/engine/test/BadMeasure/./measure.py", line 14, in arguments
    self.another_method()
  File "/Users/julien/Software/Others/OpenStudio2/python/engine/test/BadMeasure/./measure.py", line 11, in another_method
    raise ValueError("oops")
ValueError: oops
` ` `
```

python/engine/test/PythonEngine_GTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+  std::string expected_exception =
+    fmt::format(R"(SWIG director method error. In method 'arguments': `AttributeError("'Model' object has no attribute 'nonExistingMethod'")`
+Python traceback follows:
+```
+Traceback (most recent call last):
+  File "{}", line 14, in arguments
+    model.nonExistingMethod()
+AttributeError: 'Model' object has no attribute 'nonExistingMethod'
+```)",
+                scriptPath.generic_string());
```
jmarrec, line 94: WrongMethodMeasure for python

src/measure/Measure.i
```diff
@@ -96,6 +96,71 @@
 };
 
 #if defined SWIGRUBY
+  %feature("director:except") {
```
kbenne, line 99: I also contemplated moving this up to a higher level. I would be fine with that, but I opted to keep it "local" to where the directors are defined. I thought there was a bit of logic in that. Either way is really ok.

kbenne (approved):
Thanks for putting this togethor @jmarrec !

PR #5035, opened 2023-11-14
Fix #5030 - C++ CLI: run.log is huge and missing the runner.registerWarning messages

### LINKED ISSUE #5030: C++ CLI: run.log is huge and missing the runner.registerWarning messages
Issue overview
--------------

## Current Behavior

- [x] The run.log is huge -- used to have three messages in it, now it has hundreds of messages.
- [x] If a reporting measure issues a warning (`runner.registerWarning("WARNING TEST")`), the warning does not end up in the run.log file.

## Expected Behavior

Should match the classic (Ruby) CLI.

## Possible Solution

I think the piggy backing of runner.registerWarning and similar is happening here: 

https://github.com/NREL/OpenStudio-workflow-gem/blob/696306b5b24cd597edc6ab04faa1e618f99db190/lib/openstudio/workflow_runner.rb#L189-L217

### Environment

 - Version of OpenStudio (if using an intermediate build, include SHA): 3.7.0-rc2

### Context

- #5016

jmarrec:
Pull request overview
---------------------

 - Fixes #5030
- Addresses a bunch more issues that @shorowit reported after testing the installer of this PR (thanks a lot for the testing!)

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


src/cli/test/test_with_analysis.py
```diff
@@ -80,5 +97,100 @@ def test_run_with_analysis(osclipath, is_labs: bool):
...
+    runDir = run_workflow(osclipath=osclipath, suffix=suffix, is_labs=is_labs, verbose=True, debug=False)
+    assert runDir.is_dir()
+    run_log_path = runDir / "run.log"
+    assert run_log_path.is_file()
+    run_log = run_log_path.read_text()
+    assert "No results for objective function IsNonExisting.NonExisting" in run_log
+    assert "runner.registerWarning called" in run_log
+    if not is_labs:
+        pytest.skip("classic (Ruby) CLI just ignored the top level --verbose flag for run.log")
+    assert 50 < len(run_log.splitlines()) < 100
```
jmarrec, line 189: If you did `openstudio --verbose run -w workflow.osw` , would you expect the `run.log` to contain **Debug** log messages?

As a reminder, we have ALL these ways of callling the CLI

```
openstudio --verbose ...
openstudio run --debug ...
openstudio run -w workflow.osw with workflow.osw having "run_options" > debug in there
# Labs only
openstudio --loglevel Debug ...
```


Follow up question: knowing that the Ruby CLI just didn’t set the run.log to Debug in that case (it only cares about run --debug ), would you:

* A) Use the historical behavior.
* B) Change it to respect the top level log level flag.
     * if `--debug` is passed, it will increase the log level to `Debug` if not already, but for `openstudio --loglevel Trace run --debug -w in.osw` , the Log Level for stdout and run.log would stay at `Trace`)

src/measure/OSRunner.hpp
```diff
@@ -274,6 +274,11 @@ namespace measure {
 
     void resetLanguagePreference();
 
+    /** Whether registerInfo, registerWarning, registerError, registerInitialCondition, registerFinalCondition and registerAsNotApplicable
+      * also LOG an actual message */
+    bool registerMsgAlsoLogs() const;
+    void setRegisterMsgAlsoLogs(bool registerMsgAlsoLogs);
```
jmarrec, line 280: Here

src/workflow/OSWorkflow.cpp
```diff
@@ -78,6 +80,8 @@ OSWorkflow::OSWorkflow(const WorkflowRunOptions& t_workflowRunOptions, ScriptEng
     m_add_timings(t_workflowRunOptions.add_timings),
     m_style_stdout(t_workflowRunOptions.style_stdout) {
 
+  runner.setRegisterMsgAlsoLogs(true);
```
jmarrec, line 83: To mimic https://github.com/NREL/OpenStudio-workflow-gem/blob/696306b5b24cd597edc6ab04faa1e618f99db190/lib/openstudio/workflow_runner.rb#L189-L217

src/cli/test/test_with_analysis.py
kbenne: If this is going to be merged into our imminent release, then I think we need to take option A and mirror the historical behavior (plus now the --loglevel option). 

However, I think post release (and early in the cycle) we should make a change to remove the logging option under the run subcommand.

@brianlball @shorowit @joseph-robertson

src/measure/OSRunner.hpp
kbenne: I feel like option like this is just punting on making a decision. I think I would prefer that we just make a call one way or the other and avoid giving an option like this.

jmarrec: Until we remove the Ruby CLI (classic), we need an option, otherwise if always enabled, it'll log TWICE int he ruby cli

kbenne: ok.

kbenne (approved):
It seems like this need to merge before release, but after release I think we should gather togethor and talk about the big picture of logging. Perhaps a wiki document or something to outline the design of logging in OpenStudio CLI and Workflow would be helpful. It is getting hard for me to track the options and I'm guessing I'm not the only one.

brianlball:
yeah, it would also be good to do a quick study on the performance impact of logging.  I've seen performance go to sh*t when there is too much logging going on, espically with runner.registerINFO etc

Issue #5063, opened 2023-12-19
Found issue in doxygen generating documentation


wenyikuang:
Righ now the 3.7.0 [documentation](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.7.0-doc/airflow/html/index.html) is generated without line: https://github.com/NREL/OpenStudio/blob/develop/CMakeLists.txt#L1142

And we could notice the search bar is not working properly. May be due to 
- wrong Doxygen
- Error in patch file

TODO:
- Short term fix: require the 1.8.7 version of the Doxygent to generate the documentation properly.
- Mid-term fix: update the corresponding Doxygen files (idf and CMakeList) and dependency to recent released version
- Long-term fix: conver the mechanism of doc generating make it tracable.

PR #5066, opened 2023-12-21
Specify the requirement of Doxygen.

wenyikuang:
short term solution of issue #5063 

https://github.com/NREL/OpenStudio/issues/5063

https://github.com/NREL/OpenStudio/wiki/Configuring-OpenStudio-SDK-Build-Environments-(3.X.X-or-greater)#windows-1

Since it required 1.8.7 or 1.8.8 I specify the requirement.
Since it only effect the documentation generation flag, not much test need to be done.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1089,8 +1089,16 @@ mark_as_advanced(MAXIMIZE_CPU_USAGE)
 # Doxygen
 if(BUILD_DOCUMENTATION)
   # need doxygen
-  find_package(Doxygen REQUIRED)
+  find_package(Doxygen 1.8.7 REQUIRED)
 
+  if(NOT Doxygen_FOUND)
+    find_package(Doxygen 1.8.8 REQUIRED)
+  endif()
+
+  if(NOT Doxygen_FOUND)
+    message(FATAL_ERROR "Doxygen version 1.8.7 or 1.8.8 not found")
+  endif()
+  
   # if have dot you have more options
   #if(DOXYGEN_DOT_FOUND)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1089,8 +1089,16 @@ mark_as_advanced(MAXIMIZE_CPU_USAGE)
...
+  find_package(Doxygen 1.8.7 REQUIRED)
 
+  if(NOT Doxygen_FOUND)
+    find_package(Doxygen 1.8.8 REQUIRED)
+  endif()
+
+  if(NOT Doxygen_FOUND)
+    message(FATAL_ERROR "Doxygen version 1.8.7 or 1.8.8 not found")
+  endif()
+
```
jmarrec, line 1101: If REQUIRED, it'll throw directly, so no need for NOT Doxygen_FOUND. Also, cmake find_package has a version range available.

```suggestion
  find_package(Doxygen 1.8.7...<1.8.9 REQUIRED)
```

jmarrec (changes requested):

CMakeLists.txt
jmarrec: cf https://cmake.org/cmake/help/latest/command/find_package.html

jmarrec: > By specifying <, the upper end point will be excluded. Version ranges are only supported with CMake 3.19 or later

We request 3.20 already, so it's ok.

wenyikuang:
Thanks @jmarrec ,that's some syntax I haven't known. Let me merge it.

PR #5078, opened 2024-01-15
Properly set up CCACHE + fix doxygen config for versions up to 1.10.0

### LINKED ISSUE #5063: Found issue in doxygen generating documentation

jmarrec:
Pull request overview
---------------------
 
* Fix #5063

This should work on windows. Haven't had time to confirm, but I did enable it successfully on OpenStudioApplication with the same change.

(One of) the old issue was that MSVC would fail to build RC files when ccache was picked up, and this won't.

`RULE_LAUNCH_COMPILE` is now flagged as an internal implementation detail, and it makes the compile of the RC file fail on windows.

For incremental rebuilds on the OSApp using github actions, for something that had very little change (or no change, only a workflow change for eg), I took the run time from 1h15 to about 15min (the build time itself went from 50 to 3min). So I'd say it's definitely worth trying to implement ccache on windows for Jenkins.

It's possible this won't work as is due to C# or something like this. **Regardless, this is the correct way of enabling ccache on Unix as well, so this should be merged anyways**

See the note on https://cmake.org/cmake/help/latest/prop_gbl/RULE_LAUNCH_COMPILE.html

> Note This property is intended for internal use by ctest(1). Projects and developers should use the <LANG>_COMPILER_LAUNCHER target properties or the associated CMAKE_<LANG>_COMPILER_LAUNCHER variables instead.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


jmarrec:
The windows incremental runner was broken to due having doxygen 1.8.15 and us strongly requiring >=1.8.8,<1.8.9, so just fix the config to make it work.

PR #5099, opened 2024-02-26
v24.1.0-IOFreeze: HeatExchangerAirToAirSensibleAndLatent

joseph-robertson:
Pull request overview
---------------------

Remove:
- Sensible Effectiveness at 75% Heating Air Flow
- Latent Effectiveness at 75% Heating Air Flow
- Sensible Effectiveness at 75% Cooling Air Flow
- LatentEffectiveness at 75% Cooling Air Flow

Add:
- Sensible Effectiveness of Heating Air Flow Curve Name
- Latent Effectiveness of Heating Air Flow Curve Name
- Sensible Effectiveness of Cooling Air Flow Curve Name
- Latent Effectiveness of Cooling Air Flow Curve Name

Questions:
- **Do we want to support deprecated 75 methods? If so, seems like it would get tricky. They would need to operate on TableLookup objects, checking for existence, creating if not in existence, etc. I think even the 100 methods would need to modify the TableLookup objects...**

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/model/HeatExchangerAirToAirSensibleAndLatent.cpp
+++ b/src/model/HeatExchangerAirToAirSensibleAndLatent.cpp
@@ -283,24 +283,28 @@ namespace model {
     }
 
     bool HeatExchangerAirToAirSensibleAndLatent_Impl::setSensibleEffectivenessat100HeatingAirFlow(double sensibleEffectivenessat100HeatingAirFlow) {
+      // FIXME: wouldn't this value need to get set on the tablelookup if it exists?
       bool result = setDouble(OS_HeatExchanger_AirToAir_SensibleAndLatentFields::SensibleEffectivenessat100_HeatingAirFlow,
                               sensibleEffectivenessat100HeatingAirFlow);
       return result;
     }
 
     bool HeatExchangerAirToAirSensibleAndLatent_Impl::setLatentEffectivenessat100HeatingAirFlow(double latentEffectivenessat100HeatingAirFlow) {
+      // FIXME: wouldn't this value need to get set on the tablelookup if it exists?
       bool result =
         setDouble(OS_HeatExchanger_AirToAir_SensibleAndLatentFields::LatentEffectivenessat100_HeatingAirFlow, latentEffectivenessat100HeatingAirFlow);
       return result;
     }
 
     bool HeatExchangerAirToAirSensibleAndLatent_Impl::setSensibleEffectivenessat100CoolingAirFlow(double sensibleEffectivenessat100CoolingAirFlow) {
+      // FIXME: wouldn't this value need to get set on the tablelookup if it exists?
       bool result = setDouble(OS_HeatExchanger_AirToAir_SensibleAndLatentFields::SensibleEffectivenessat100_CoolingAirFlow,
                               sensibleEffectivenessat100CoolingAirFlow);
       return result;
     }
 
     bool HeatExchangerAirToAirSensibleAndLatent_Impl::setLatentEffectivenessat100CoolingAirFlow(double latentEffectivenessat100CoolingAirFlow) {
+      // FIXME: wouldn't this value need to get set on the tablelookup if it exists?
       bool result =
         setDouble(OS_HeatExchanger_AirToAir_SensibleAndLatentFields::LatentEffectivenessat100_CoolingAirFlow, latentEffectivenessat100CoolingAirFlow);
       return result;
@@ -921,13 +925,13 @@ namespace model {
     if (boost::optional<TableLookup> sensibleEffectivenessofCoolingAirFlowCurve_ =
           sensibleEffectivenessofCoolingAirFlowCurve()->optionalCast<TableLookup>()) {
       double sensibleEffectivenessat100CoolingAirFlow =
-        getImpl<detail::HeatExchangerAirToAirSensibleAndLatent_Impl>()->latentEffectivenessat100HeatingAirFlow();
+        getImpl<detail::HeatExchangerAirToAirSensibleAndLatent_Impl>()->sensibleEffectivenessat100CoolingAirFlow();
       sensibleEffectivenessofCoolingAirFlowCurve_->setOutputValues(
         {sensibleEffectivenessat75CoolingAirFlow, sensibleEffectivenessat100CoolingAirFlow});
       return setSensibleEffectivenessofCoolingAirFlowCurve(sensibleEffectivenessofCoolingAirFlowCurve_.get());
     } else if (!sensibleEffectivenessofCoolingAirFlowCurve()) {
       double sensibleEffectivenessat100CoolingAirFlow =
-        getImpl<detail::HeatExchangerAirToAirSensibleAndLatent_Impl>()->latentEffectivenessat100HeatingAirFlow();
+        getImpl<detail::HeatExchangerAirToAirSensibleAndLatent_Impl>()->sensibleEffectivenessat100CoolingAirFlow();
 
       TableIndependentVariable var(model());
       var.setInterpolationMethod("Linear");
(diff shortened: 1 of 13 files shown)


src/model/HeatExchangerAirToAirSensibleAndLatent.cpp
```diff
@@ -534,6 +539,59 @@ namespace model {
...
+    TableLookup l75cooling(model);
+    l75cooling.setNormalizationMethod("DivisorOnly");
+    l75cooling.setNormalizationDivisor(0.68);
+    l75cooling.setMinimumOutput(0.0);
+    l75cooling.setMaximumOutput(10.0);
+    l75cooling.setOutputUnitType("Dimensionless");
+    l75cooling.addOutputValue(0.73);
+    l75cooling.addOutputValue(0.68);
+    l75cooling.addIndependentVariable(var);
+    setLatentEffectivenessofCoolingAirFlowCurve(l75cooling);
```
jmarrec, line 594: Ok, so I get why this is here: to maintain the same effectiveness at different airflows as we had before.

* We could also not do this, similar to what E+ would do if you create this object in the IDF Editor
    * Leave them empty so the effectiveness is linear.
    * You could extract that bit of code into a public method, something like "applyDefaultEffectivenessCurves" or something for convenience, so that the user has the ability to use the former defaults with a single method call.
    * VT (already done here) + modification of the OS resources ruby tests would ensure we don't show EUI deviations.
* If we want these curves to be set, then please name the Table / Vars

src/model/HeatExchangerAirToAirSensibleAndLatent.cpp
```diff
@@ -744,5 +804,196 @@ namespace model {
...
+  // DEPRECATED
+  double HeatExchangerAirToAirSensibleAndLatent::sensibleEffectivenessat75HeatingAirFlow() const {
+    DEPRECATED_AT_MSG(3, 8, 0, "As of EnergyPlus 24.1.0, this property ...");
+    if (boost::optional<TableLookup> sensibleEffectivenessofHeatingAirFlowCurve_ =
+          sensibleEffectivenessofHeatingAirFlowCurve()->optionalCast<TableLookup>()) {
+      return sensibleEffectivenessofHeatingAirFlowCurve_->outputValues()[0];
+    } else {
+      return getImpl<detail::HeatExchangerAirToAirSensibleAndLatent_Impl>()->sensibleEffectivenessat100HeatingAirFlow();
+    }
+  }
```
jmarrec, line 816: * You have no proof that the `outputValues()[0] `corresponds to X=0.75. If I created that TableLookup manually, this may very well **NOT** be the case.

* You MUST check if the `sensibleEffectivenessofHeatingAirFlowCurve()` is initialized before you deference it or it'll throw if not. Your testing didn't catch it because the curves are optional BUT your constructor is initializing them.

* If I have set something other than a TableLookup (meaning a regular Curve:XXX object), you're also returning the 100 airflow without warning.

src/model/HeatExchangerAirToAirSensibleAndLatent.cpp
```diff
@@ -744,5 +804,196 @@ namespace model {
     return getImpl<detail::HeatExchangerAirToAirSensibleAndLatent_Impl>()->autosizedNominalSupplyAirFlowRate();
   }
 
+  // DEPRECATED
+  double HeatExchangerAirToAirSensibleAndLatent::sensibleEffectivenessat75HeatingAirFlow() const {
+    DEPRECATED_AT_MSG(3, 8, 0, "As of EnergyPlus 24.1.0, this property ...");
```
jmarrec, line 809: TODO: update the message.

```
DEPRECATED_AT_MSG(3, 8, 0, "Use sensibleEffectivenessofHeatingAirFlowCurve instead.");
```

src/model/HeatExchangerAirToAirSensibleAndLatent.cpp
jmarrec: All the getters would work fine if we had `TableLookup_Impl::evaluate`, but we don't.

src/model/HeatExchangerAirToAirSensibleAndLatent.cpp
```diff
@@ -744,5 +804,196 @@ namespace model {
...
+      tableLookup.setNormalizationMethod("DivisorOnly");
+      tableLookup.setNormalizationDivisor(sensibleEffectivenessat100HeatingAirFlow);
+      tableLookup.setMinimumOutput(0.0);
+      tableLookup.setMaximumOutput(10.0);
+      tableLookup.setOutputUnitType("Dimensionless");
+      tableLookup.addOutputValue(sensibleEffectivenessat75HeatingAirFlow);
+      tableLookup.addOutputValue(sensibleEffectivenessat100HeatingAirFlow);
+      tableLookup.addIndependentVariable(var);
+
+      return setSensibleEffectivenessofHeatingAirFlowCurve(tableLookup);
```
jmarrec, line 880: No ambiguity in this case: if the curve isnt there, we **can** make a table lookup indeed.

src/model/HeatExchangerAirToAirSensibleAndLatent.cpp
```diff
@@ -744,5 +804,196 @@ namespace model {
...
+
+  bool HeatExchangerAirToAirSensibleAndLatent::setSensibleEffectivenessat75HeatingAirFlow(double sensibleEffectivenessat75HeatingAirFlow) {
+    DEPRECATED_AT_MSG(3, 8, 0, "As of EnergyPlus 24.1.0, this property ...");
+    if (boost::optional<TableLookup> sensibleEffectivenessofHeatingAirFlowCurve_ =
+          sensibleEffectivenessofHeatingAirFlowCurve()->optionalCast<TableLookup>()) {
+      double sensibleEffectivenessat100HeatingAirFlow =
+        getImpl<detail::HeatExchangerAirToAirSensibleAndLatent_Impl>()->sensibleEffectivenessat100HeatingAirFlow();
+      sensibleEffectivenessofHeatingAirFlowCurve_->setOutputValues(
+        {sensibleEffectivenessat75HeatingAirFlow, sensibleEffectivenessat100HeatingAirFlow});
+      return setSensibleEffectivenessofHeatingAirFlowCurve(sensibleEffectivenessofHeatingAirFlowCurve_.get());
```
jmarrec, line 856: same problems: can't dereference the curve object if not initialized, and the setting of the output values isn't good since you're not also modifying the independent variable.

jmarrec: Side note, but related to a potential final conclusion on whether we should try to maintain backward compatibilty: The `TableLookup` interface is crap from a user experience perspective. The former `TableMultiVariableLookup` was better in that sense since you could do `addPoint(x, y)` and it'd detect if the x value was already there, and you could call `boost::optional<double> yValue(const std::vector<double>& xValues)`

I think we should re-add that capability...

src/sdd/MapHVAC.cpp
```diff
@@ -2719,7 +2719,12 @@ namespace sdd {
     auto htgSensEff75Element = element.child("HtgSensEff75");
     boost::optional<double> _htgSensEff75 = lexicalCastToDouble(htgSensEff75Element);
     if (_htgSensEff75) {
-      hx.setSensibleEffectivenessat75HeatingAirFlow(_htgSensEff75.get());
+      if (boost::optional<model::Curve> curve_ = hx.sensibleEffectivenessofHeatingAirFlowCurve()) {
+        if (boost::optional<model::TableLookup> sensibleEffectivenessofHeatingAirFlowCurve_ = curve_->optionalCast<TableLookup>()) {
+          double sensibleEffectivenessat100HeatingAirFlow = hx.sensibleEffectivenessat100HeatingAirFlow();
+          sensibleEffectivenessofHeatingAirFlowCurve_->setOutputValues({_htgSensEff75.get(), sensibleEffectivenessat100HeatingAirFlow});
+        }
+      }
```
jmarrec, line 2727: Ok, so it seems we do want to support **some** amount of backward compatibility in setting the 75% effectivness due to this MapHVAC stuff.

I don't think the current implementation here is good though. You're creating the object, so you're in control. You should create the curve instead of checking whether it exists.

jmarrec: I think the opposite way (OSM > SDD) is missing too.

jmarrec (changes requested):
I was tempted to say we shouldn't even try to support backward compat here (just log a message saying the method does nothing now and you should call xxxCurve() method), but it seems some amount of backward compat is needed due to the SDD stuff.

I think I'm going to sleep on it, I'll try to think overnight about the amount of work that'd be necessary to properly support it.

jmarrec:
Options:

* Remove backward compatibility altogether. This would require the SDD stuff to be updated... Is that an option @kbenne ?
* Implement backward compatibility manually. This requires a bit of code to do it right, I'm guessing I can do it in about 4 hours.
* Implement TableLookup_Impl::evaluate: this would ideally require using https://github.com/bigladder/btwxt like E+ does, but the packaging is problematic. It has submodules dependencies (google test, fmt and bigladder's courrier) and therefore creating a conan recipe is quite problematic. Using CMAke's ExternalPackageAdd isn't really helping either since some of these dependencies we also do have and we don't want to mix up deps versions.

jmarrec:
@joseph-robertson I made some changes, this is good enough to go into the main IO freeze branch

mdahlhausen:
@joseph-robertson @jmarrec we may need to modify the work done in this PR. Some models made in OpenStudio are hitting a failure in EnergyPlus with the default curve set: https://github.com/NREL/EnergyPlus/issues/10464

PR #5111, opened 2024-03-13
ScheduleDay: new getValues method

### LINKED ISSUE #5001: ScheduleDay::getValue applies interpolation differently than EnergyPlus
Issue overview
--------------
For a ScheduleDay with 'Interpolate to Timestep' = 'Yes', calling [`getValue`](https://github.com/NREL/OpenStudio/blob/7a987d3fa223bd8319aac2e09db9ff7ce09a9258/src/model/ScheduleDay.cpp#L174) 
 with a Time value that does not equal one of the scheduled Hour+Minute values will return an interpolated value, which is inconsistent with how EnergyPlus interprets that input.

## Current Behavior
EnergyPlus interprets '[Interpolate to Timestep](https://bigladdersoftware.com/epx/docs/23-2/input-output-reference/group-schedules.html#field-interpolate-to-timestep)' = 'Average' to mean "when the simulation timestep doesn't coincide with the scheduled interval, the value at the intermediate timestep is interpolated".  I.e., for this schedule: 
```
Schedule:Day:Interval,
    Day Schedule,            !- Name
    Fractional,              !- Schedule Type Limits Name
    Average,                 !- Interpolate to Timestep
    8:15,            	     !- Time 1 {hh:mm}
    0,                       !- Value Until Time 1
    21:45,                   !- Time 2 {hh:mm}
    1,                       !- Value Until Time 2
    24:00,                   !- Time 3 {hh:mm}
    0;                       !- Value Until Time 3
```
If the simulation timesteps per hour are 4, will evaluate like: 
image

With 'Average' interpolation, for 10 timesteps per hour, it would evaluate like: 
image

However `ScheduleDay::getValue` in OpenStudio **interpolates between schedule times** (i.e., 8:15 to 21:45), so calling `getValue` with intermediate times results in a plot like this: 
image

## Expected Behavior
I expect `getValue` to return the same value that EnergyPlus would for a schedule at a given time.

## Steps to Reproduce
Create a ScheduleDay with 'Interpolate To Timestep' = 'Yes', and make successive calls to `getValue` like: 
```
def print_vals(sch)
  (0..24).each do |hr|
    ["00","15","30","45"].each do |min|
      hr_pad = "%02d" % hr
      time = OpenStudio::Time.new("#{hr_pad}:#{min}:00")
      val = day_sch.getValue(time)
      puts "#{time}, #{val}"
    end
  end
end

sch = OpenStudio::Model::ScheduleDay.new(model)
sch.setInterpolatetoTimestep(true)
sch.addValue(OpenStudio::Time.new("08:15:00"),0)
sch.addValue(OpenStudio::Time.new("21:45:00"),1)
print_vals(sch)

```
-> ...
08:00:00, 0.0
08:15:00, 0.0
08:30:00, 0.01851851851851845
08:45:00, 0.03703703703703701
09:00:00, 0.05555555555555555
...
21:00:00, 0.9444444444444444
21:15:00, 0.9629629629629629
21:30:00, 0.9814814814814814
21:45:00, 1.0
22:00:00, 0.8888888888888893
22:15:00, 0.7777777777777786

 - Version of OpenStudio (if using an intermediate build, include SHA): 3.6.1 but this goes back a long ways

### Context
This impacts multiple uses of `getValue`, e.g. for determining HVAC occupancy schedules in [openstudio-standards](https://github.com/NREL/openstudio-standards/blob/682d3b2187d541fdeb91949e5bacbc6bc83952ff/lib/openstudio-standards/standards/Standards.ThermalZone.rb#L345), schedule reporting in [OpenStudio Results](https://github.com/NREL/openstudio-common-measures-gem/blob/0e3558cf96b446e8a6f5041ae21cc1fd93bc4661/lib/measures/openstudio_results/resources/os_lib_reporting.rb#L4657), etc.

joseph-robertson:
Pull request overview
---------------------

 - Fixes #5001.

A new `getValues` method for returning a vector of schedule values given a `Timestep`.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/model/ScheduleDay.cpp
+++ b/src/model/ScheduleDay.cpp
@@ -16,6 +16,8 @@
 #include "ScheduleRuleset_Impl.hpp"
 #include "ScheduleRule.hpp"
 #include "ScheduleRule_Impl.hpp"
+#include "Timestep.hpp"
+#include "Timestep_Impl.hpp"
 
 #include "../utilities/idf/IdfExtensibleGroup.hpp"
 #include <utilities/idd/OS_Schedule_Day_FieldEnums.hxx>
@@ -209,6 +211,26 @@ namespace model {
 
       double result = interp(x, y, time.totalDays(), interpMethod, NoneExtrap);
 
+      return result;
+    }
+
+    std::vector<double> getValues(const openstudio::model::Timestep& timestep) const {
+      std::vector<double> values = this->values();          // these are already sorted
+      std::vector<openstudio::Time> times = this->times();  // these are already sorted
+      std::vector<double> result;
+
+      int timesteps = 24.0 * timestep.numberOfTimestepsPerHour();
+      for (size_t timestep = 0; timestep < timesteps; ++timestep) {
+        openstudio::Time t(timestep / timesteps);
+        size_t i = 0;
+        for (const openstudio::Time& time : times) {
+          if (t <= time) {
+            result.push_back(values[i]);
+            break;
+          }
+          ++i;
+        }          
+      }
       return result;
     }
 
--- a/src/model/ScheduleDay.cpp
+++ b/src/model/ScheduleDay.cpp
@@ -229,7 +229,7 @@ namespace model {
             break;
           }
           ++i;
-        }          
+        }
       }
       return result;
     }
--- a/src/model/ScheduleDay.hpp
+++ b/src/model/ScheduleDay.hpp
@@ -69,6 +69,9 @@ namespace model {
     /// Returns the value in effect at the given time.  If time is less than 0 days or greater than 1 day, 0 is returned.
     double getValue(const openstudio::Time& time) const;
 
+    /// Returns the values for a given timestep.
+    std::vector<double> getValues(const openstudio::model::Timestep& timestep) const;
+
     //@}
     /** @name Setters */
     //@{
--- a/src/model/ScheduleDay_Impl.hpp
+++ b/src/model/ScheduleDay_Impl.hpp
@@ -72,6 +72,9 @@ namespace model {
       /// Returns the value in effect at the given time.  If time is less than 0 days or greater than 1 day, 0 is returned.
       double getValue(const openstudio::Time& time) const;
 
+      /// Returns the values for a given timestep.
+      std::vector<double> getValues(const openstudio::model::Timestep& timestep) const;
+
       //@}
       /** @name Setters */
       //@{
--- a/src/model/test/ScheduleDay_GTest.cpp
+++ b/src/model/test/ScheduleDay_GTest.cpp
@@ -10,6 +10,8 @@
 #include "../ScheduleDay_Impl.hpp"
 #include "../ScheduleTypeLimits.hpp"
 #include "../ScheduleTypeLimits_Impl.hpp"
+#include "../Timestep.hpp"
+#include "../Timestep_Impl.hpp"
 
 #include "../../utilities/time/Date.hpp"
 #include "../../utilities/time/Time.hpp"
@@ -319,4 +321,32 @@ TEST_F(ModelFixture, Schedule_Day_addValue_NaN_Infinity) {
   EXPECT_FALSE(sch_day.addValue(t, std::numeric_limits<double>::quiet_NaN()));
   EXPECT_FALSE(sch_day.addValue(t, std::numeric_limits<double>::infinity()));
   EXPECT_FALSE(sch_day.addValue(t, -std::numeric_limits<double>::infinity()));
+}
+
+TEST_F(ModelFixture, Schedule_Day_getValues) {
+  Model model;
+
+  ScheduleDay sch_day(model);
+
+  Time t1("08:15:00");
+  sch_day.addValue(t1, 0);
+  Time t2("21:45:00");
+  sch_day.addValue(t2, 1);
+
+  auto timestep = model.getUniqueModelObject<Timestep>():
+
+  timestep.setNumberOfTimestepsPerHour(1);
+  std::vector<double> values24 = sch_day.getValues(timestep);
+  EXPECT_EQ(24 * 1, values24.size());
+
+  timestep.setNumberOfTimestepsPerHour(4);
+  std::vector<double> values15 = sch_day.getValues(timestep);
+  EXPECT_EQ(24 * 4, values15.size());
+
+  timestep.setNumberOfTimestepsPerHour(6);
+  std::vector<double> values10 = sch_day.getValues(timestep);
+  EXPECT_EQ(24 * 6, values10.size());
+
+  // TODO: check for 0s and 1s
+  // TODO: check interpolatetoTimestep doesn't matter
 }
--- a/src/model/test/ScheduleDay_GTest.cpp
+++ b/src/model/test/ScheduleDay_GTest.cpp
@@ -333,9 +333,11 @@ TEST_F(ModelFixture, Schedule_Day_getValues) {
   Time t2("21:45:00");
   sch_day.addValue(t2, 1);
 
-  auto timestep = model.getUniqueModelObject<Timestep>():
+  auto timestep = model
+                    .getUniqueModelObject<Timestep>()
+    :
 
-  timestep.setNumberOfTimestepsPerHour(1);
+      timestep.setNumberOfTimestepsPerHour(1);
   std::vector<double> values24 = sch_day.getValues(timestep);
   EXPECT_EQ(24 * 1, values24.size());
 


src/model/ScheduleDay.hpp
```diff
@@ -66,8 +66,19 @@ namespace model {
     /// Returns a vector of values in the same order and with the same number of elements as times.
     std::vector<double> values() const;
 
-    /// Returns the value in effect at the given time.  If time is less than 0 days or greater than 1 day, 0 is returned.
-    double getValue(const openstudio::Time& time) const;
+    /// Returns the value in effect at the given time.
+    /// If time is less than 0 days or greater than 1 day, 0 is returned.
+    double getValue(const openstudio::Time& time, int numberOfTimestepsPerHour = 0) const;
```
jmarrec, line 71: I find the numberOfTimestepPerHour argument weird.

The ScheduleDay is part of the model, so we **know** the Number of Timestep per Hour (`model.getTimestep.numberOfTimestepsPerHour`)

src/model/ScheduleDay.cpp
```diff
@@ -176,18 +181,64 @@ namespace model {
...
+          double value = values[i + 1];
+        } else if ((t0 < time) && (time < t1)) {
+          double value = values[i + 1];
+        }
+      }
+
+      return value;
+    }
+
+    openstudio::TimeSeries ScheduleDay_Impl::timeSeries() const {
```
joseph-robertson, line 208: This returns timeseries with timesteps that align with the `Timestep` object.

src/model/ScheduleDay.cpp
```diff
@@ -200,14 +251,48 @@ namespace model {
...
+                double int1 = (times[i] - t1).totalDays() / interval;  // fraction that is the value before the schedule change
+                double int2 = (t0 - times[i]).totalDays() / interval;  // fraction that is the value after the schedule change
+                tsValues[j] = (values[i] * int1) + (values[i + 1] * int2);
+                break;
+              } else if (t0 <= times[i]) {
+                tsValues[j] = values[i];
+                break;
+              }
+            }
+          }
```
joseph-robertson, line 288: I suppose this could live in `Vector.cpp`, although you'd need to pass in the timestep?

src/model/ScheduleDay.cpp
```diff
@@ -176,18 +181,64 @@ namespace model {
...
+          t1 = openstudio::Time(0, 24, 0);
+        }
+
+        if (t0 == time) {
+          double value = values[i];
+        } else if (t1 == time) {
+          double value = values[i + 1];
+        } else if ((t0 < time) && (time < t1)) {
+          double value = values[i + 1];
+        }
```
joseph-robertson, line 202: Should we still linearly interpolate for requested times _between_ timesteps? Seems like E+ doesn't see/use them, so we return the value corresponding to the following timestep?

src/model/ScheduleDay.cpp
joseph-robertson: Moved to `Vector.cpp`.

src/model/ScheduleDay.cpp
joseph-robertson: Could we just use `HoldNextInterp` here?

src/model/ScheduleDay.cpp
```diff
@@ -369,6 +447,7 @@ namespace model {
     void ScheduleDay_Impl::clearCachedVariables() {
       m_cachedTimes.reset();
       m_cachedValues.reset();
+      m_cachedTimeSeries.reset();
```
joseph-robertson, line 450: @jmarrec We need to somehow call this reset whenever the `numberOfTimestepsPerHour` changes on the `Timestep` object.

eringold:
Although this PR contains an API breaking change, I believe the negative impacts should be fairly minimal. The primary API changes are: 
- what `ScheduleDay::getValue` returns when `ScheduleDay::setInterpolateToTimestep` is `true` (which acted the same as 'Linear' in EnergyPlus, although FT would set the option to 'Average')
- change `ScheduleDay::setInterpolateToTimestep` to take the appropriate options to instead of a boolean

In openstudio-standards, ComStock and a few measure repos I've looked at, there are many uses of `ScheduleDay::getValue`, but the only uses of `ScheduleDay::setInterpolateToTimestep` are to set as false (with one exception, which is the use in openstudio-standards from which I identified this issue, but which I [removed](https://github.com/NREL/openstudio-standards/pull/1632) because it was not needed). Since the default ScheduleDay interpolation method is 'none', all the existing uses of `getValue` would be unaffected by these changes. 

Adding the appropriate options to `setInterpolateToTimestep` and the Average interpolation calculation is an improvement in that it aligns the OpenStudio user experience with how EnergyPlus interprets that input, and the addition of the `ScheduleDay::timeSeries` method adds further utility to users in processing of schedules, such as in equivalent full-load hour calculations and parametric schedule modifications. 

In short, these are good and necessary changes. Thanks @joseph-robertson!

src/model/ScheduleDay.cpp
jmarrec: Easiest is probably just to do it inside Timestep::setNumberOfTimestepsPerHour and resetNumberOfTimestepsPerHour.

Make ScheduleDay_Impl friend the Timestep_Impl.

```
    bool Timestep_Impl::setNumberOfTimestepsPerHour(int numberOfTimestepsPerHour) {
       if oldValue != numberOfTimestepsPerHour:
           for (auto& sch_day: model.getConcreteModelObjects<ScheduleDay>()) { clearCachedVariables }
      setInt as usual
```

jmarrec: MMMH, you may need to make ScheduleDay (public) have a protected clearCachedVariables that forward to _Impl.

But getImpl is public I **think**


           for (auto& sch_day: model.getConcreteModelObjects<ScheduleDay>()) { sch_day.getImpl<ScheduleDay_Impl>()->clearCachedVariables() }

jmarrec: Yep https://github.com/NREL/OpenStudio/blob/ab2d186f5afe7e95da38a26f87b8cf6f495ff64a/src/utilities/idf/IdfObject.hpp#L361-L370

resources/model/OpenStudio.idd
```diff
@@ -4594,12 +4594,13 @@ OS:Schedule:Day,
        \type object-list
        \object-list ScheduleTypeLimitsNames
   A4, \field Interpolate to Timestep
-       \note when the interval does not match the user specified timestep a Yes choice will average between the intervals request (to
-       \note timestep resolution.  a No choice will use the interval value at the simulation timestep without regard to if it matches
-       \note the boundary or not.
+       \note when the interval does not match the user specified timestep a Average choice will average between the intervals request (to
+       \note timestep resolution. A No choice will use the interval value at the simulation timestep without regard to if it matches
+       \note the boundary or not. A Linear choice will interpolate linearly between successive values.
```
jmarrec, line 4599: Add VersionTranslator + test for it

src/model/Timestep.cpp
```diff
@@ -66,11 +68,20 @@ namespace model {
     }
 
     bool Timestep_Impl::setNumberOfTimestepsPerHour(int numberOfTimestepsPerHour) {
+      int currentNumberOfTimestepsPerHour = numberOfTimestepsPerHour();
+      if (currentNumberOfTimestepsPerHour != numberOfTimestepsPerHour) {
+        for (auto& sch_day : model().getConcreteModelObjects<ScheduleDay>()) {
+          sch_day.getImpl<ScheduleDay_Impl>()->clearCachedVariables();
+        }
```
jmarrec, line 75: I don't think you need to clear the cached times and values. Just the timeseries

src/model/ScheduleDay.cpp
jmarrec: you need to make      `void clearCachedVariables();`protected instead of private, and friend Timeseries_Impl

src/model/ScheduleDay.hpp
```diff
@@ -66,14 +66,18 @@ namespace model {
...
 
+    /// Returns the timeseries corresponding to simulation timestep and chosen interpolation method.
+    openstudio::TimeSeries timeSeries() const;
+
     //@}
     /** @name Setters */
     //@{
 
-    bool setInterpolatetoTimestep(bool interpolatetoTimestep);
+    bool setInterpolatetoTimestep(const std::string& interpolatetoTimestep);
```
jmarrec, line 80: You can avoid this API break by keeping the bool overload. Not sure whether we want that or not. The getter is an API break for sure though.

joseph-robertson: Decided not to go for this.

src/model/ScheduleDay.hpp
```diff
@@ -54,7 +54,7 @@ namespace model {
    *  inherits one from a Schedule. */
     bool isScheduleTypeLimitsDefaulted() const;
 
-    bool interpolatetoTimestep() const;
+    std::string interpolatetoTimestep() const;
```
jmarrec, line 57: Document the API break in the Release Notes markdown file please

src/model/Timestep.cpp
```diff
@@ -66,11 +68,19 @@ namespace model {
     }
 
     bool Timestep_Impl::setNumberOfTimestepsPerHour(int numberOfTimestepsPerHour) {
+      if (numberOfTimestepsPerHour != this->numberOfTimestepsPerHour()) {
+        for (auto& sch_day : model().getConcreteModelObjects<ScheduleDay>()) {
+          sch_day.getImpl<ScheduleDay_Impl>()->clearCachedTimeSeries();
+        }
+      }
```
jmarrec, line 75: Reset the cachedTImeSeries when we change the number of timesteps per hour in both Timestep::setNumberOfTimesPerHour and the resetNum... ok!

src/osversion/VersionTranslator.cpp
```diff
@@ -9160,6 +9160,32 @@ namespace osversion {
         ss << newObject;
         m_refactored.emplace_back(std::move(object), std::move(newObject));
 
+      } else if (iddname == "OS:Schedule:Day") {
+
+        // 1 Field has been added from 3.7.0 to 3.8.0:
+        // ----------------------------------------------
+        // * Interpolate to Timestep * 3
```
jmarrec, line 9167: pedantic man to the rescue!

```suggestion
        // 1 Field has been modified from 3.7.0 to 3.8.0:
        // ----------------------------------------------
        // * Interpolate to Timestep * 3 - Changed from bool to string choice
```

jmarrec: Anyways, VT + VT test, :+1:

src/model/ScheduleDay.cpp
```diff
@@ -176,42 +181,123 @@ namespace model {
         return 0.0;
       }
 
-      std::vector<double> values = this->values();          // these are already sorted
-      std::vector<openstudio::Time> times = this->times();  // these are already sorted
+      TimeSeries ts = this->timeSeries();
```
jmarrec, line 184: So, I'm probably overthinking/bikeshedding but conceptually it pains me that we calculate everything for the entire day when we want one value.

I'm going to write the rationale for why I think that's still a good thing (and probably make it a comment for future ref)

* we're talking about max 60 timesteps * 24 hours = 1440 points, but realistically more often than not you'll have a timestep of 6 (our default) or 4 so 96 to 144 points. So not a lot of points
* We cache the timeSeries
* More often than not, the use case will be to get the values for an entire day, such as what openstudio-standards does with it (eg https://github.com/NREL/openstudio-standards/blob/682d3b2187d541fdeb91949e5bacbc6bc83952ff/lib/openstudio-standards/standards/Standards.ThermalZone.rb#L345)

```suggestion
      // We'll calculate the entire day when we request a single value but that's on purpose:
      // * We're talking about max 60 timesteps * 24 hours = 1440 points,
      //   but realistically more often than not you'll have a timestep of 6 (our default) or 4 so 96 to 144 points. So not a lot of points
      // * We cache the timeSeries, and,
      // * More often than not, the use case is to do it for the entire day anyways (eg: openstudio-standards to determine occupancy schedules)
      TimeSeries ts = this->timeSeries();
```

src/model/ScheduleDay.cpp
```diff
@@ -176,42 +181,123 @@ namespace model {
...
 
-      unsigned N = times.size();
+      DateTimeVector dateTimes = ts.dateTimes();
+      Vector values = ts.values();
+
+      unsigned N = dateTimes.size();
```
jmarrec, line 189: Mmh, starting there you lost me. I hadn't seen this bit when I wrote the comment above, and I hadn't though about a time being requested in between actual timesteps, but this looks wrong to me.

Let me see if I follow:

* You get timeSeries, which is already something interpolated, but aligns on the Timeseries::timestep
* Then you scan the **entire** day for that, and you call another interpolation call

I will go look into Vector/Matrix next, but this doesn't sit right with me.

jmarrec, line 189: Also, const

src/model/ScheduleDay.cpp
```diff
@@ -176,42 +181,123 @@ namespace model {
...
 
-      InterpMethod interpMethod;
-      if (this->interpolatetoTimestep()) {
-        interpMethod = LinearInterp;
-      } else {
-        interpMethod = HoldNextInterp;
-      }
-
-      double result = interp(x, y, time.totalDays(), interpMethod, NoneExtrap);
+      double result = interp(x, y, time.totalDays(), HoldNextInterp, NoneExtrap);
```
jmarrec, line 222: I really feel like you ever need to pass two values to that call...

As I wrote on https://github.com/NREL/OpenStudio/issues/5001#issuecomment-1998619932

> doesn't the getValues routine look inefficient? If I'm understanding correctly, you only need to find the time before and the time after and the corresponding, you can break early. I know, this is a scheduleday, so we're talking about 60 (timestep max in E+ If I recall correctly) * 24 hours datapoints max, so it's not noticeable, but still


And you should probably just two that on values & times, not on timeSeries (which is interpolated already, and contains the entire day)

src/model/ScheduleDay.cpp
```diff
@@ -176,42 +181,123 @@ namespace model {
...
+              openstudio::Time t(0, hour, minute);
+              tsDateTimes.push_back(startDateTime + t);
+            }
+          }
+        }
+
+        std::vector<double> values = this->values();          // these are already sorted
+        std::vector<openstudio::Time> times = this->times();  // these are already sorted
+
+        unsigned N = times.size();
```
jmarrec, line 253: const

src/utilities/data/Matrix.hpp
```diff
@@ -33,22 +33,22 @@ UTILITIES_API bool operator!=(const Matrix& lhs, const Matrix& rhs);
 /// linear interpolation of the function v = f(x, y) at point xi, yi
 /// assumes that x and y are strictly increasing
 UTILITIES_API double interp(const Vector& x, const Vector& y, const Matrix& v, double xi, double yi, InterpMethod interpMethod = LinearInterp,
-                            ExtrapMethod extrapMethod = NoneExtrap);
+                            ExtrapMethod extrapMethod = NoneExtrap, double ti = 0.0);
```
jmarrec, line 36: Please adjust the docstring of all of these to explain what the ti is, and **when** you should pass a non-zero value.

src/utilities/data/Vector.cpp
```diff
@@ -92,16 +96,28 @@ InterpInfo interpInfo(const Vector& x, double xi) {
...
+    }
   }
 
   return result;
 }
 
 /// linear interpolation of the function y = f(x) at point xi
 /// assumes that x is strictly increasing
-double interp(const Vector& x, const Vector& y, double xi, InterpMethod interpMethod, ExtrapMethod extrapMethod) {
+double interp(const Vector& x, const Vector& y, double xi, InterpMethod interpMethod, ExtrapMethod extrapMethod, double ti) {
```
jmarrec, line 120: I'd like to see an assertion that `ti` must be valid when interpMethod is AverageInterp (I think that's the only one where it's needed)

src/utilities/data/Vector.cpp
```diff
@@ -130,6 +146,10 @@ double interp(const Vector& x, const Vector& y, double xi, InterpMethod interpMe
         // linear interpolation
         result = info.wa * y(info.ia) + info.wb * y(info.ib);
         break;
+      case AverageInterp:
+        // average interpolation
+        result = (info.wa * y(info.ia) + info.wb * y(info.ib)) / info.ti;
+        break;
```
jmarrec, line 152: Reading this makes me thing that the default ti = 0.0 is weird, shouldn't it be 1.0?


**Edit**: given that do you equality check == 0.0 in interpInfo I don't think it should be either

src/utilities/data/Vector.cpp
```diff
@@ -92,16 +96,28 @@ InterpInfo interpInfo(const Vector& x, double xi) {
 
     result.ia = (unsigned)(it - begin - 1);
     result.ib = (unsigned)(it - begin);
-    result.wa = (x(result.ib) - xi) / (x(result.ib) - x(result.ia));
-    result.wb = (xi - x(result.ia)) / (x(result.ib) - x(result.ia));
+    if (ti == 0.0) {
```
jmarrec, line 99: Floating point equality... That triggers a lot of red flags for me

src/utilities/data/Vector.hpp
```diff
@@ -75,21 +76,22 @@ struct UTILITIES_API InterpInfo
...
 /** Linear interpolation of the function y = f(x) at point xi. Assumes that x is strictly
  *  increasing. */
-UTILITIES_API InterpInfo interpInfo(const Vector& x, double xi);
+UTILITIES_API InterpInfo interpInfo(const Vector& x, double xi, double ti);
 
 /** Linear interpolation of the function y = f(x) at point xi. Assumes that x is strictly
  *  increasing */
 UTILITIES_API double interp(const Vector& x, const Vector& y, double xi, InterpMethod interpMethod = LinearInterp,
-                            ExtrapMethod extrapMethod = NoneExtrap);
+                            ExtrapMethod extrapMethod = NoneExtrap, double ti = 0.0);
```
jmarrec, line 89: After scanning the code, I think this is really trying to replace a `boost::optional<double>` only needed for InterpMethod == AverageInterp

Save for actually using a boost::optional<double>, I think you should make the default a negative value, like `-9999.0` (abritrary E+ convention)

src/model/ScheduleDay.cpp
joseph-robertson: The second one (i.e., getValue) is not _really_ an interpolation call. It is a call with `HoldNextInterp`.

src/model/ScheduleDay.cpp
```diff
@@ -176,42 +181,123 @@ namespace model {
...
+      double result = interp(x, y, time.totalDays(), HoldNextInterp, NoneExtrap);
 
       return result;
     }
 
+    openstudio::TimeSeries ScheduleDay_Impl::timeSeries() const {
+      if (!m_cachedTimeSeries) {
+
+        auto timestep = this->model().getUniqueModelObject<Timestep>();
+        int numberOfTimestepsPerHour = timestep.numberOfTimestepsPerHour();
```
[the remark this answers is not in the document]
jmarrec: Yeah.

src/model/ScheduleDay.cpp
jmarrec: So, I'm really still not a big fan of this. But I don't think it warrants blocking the PR... I expect most people will just use timeSeries anyways.

jmarrec:
@joseph-robertson Thanks for addressing some of the review comments.

I'm still not happy about getValue, and maybe we should open a ticket to revisit the getValue and avoid recomputing the entire day, but maybe not.

Anyways, if CI is happy with this, we can merge.

jmarrec:
/srv/jenkins/openstudio/docker-volumes/ubuntu-2004/Openstudio/src/utilities/data/Vector.cpp:103:7: error: ‘OS_ASSERT’ was not declared in this scope; did you mean ‘BOOST_ASSERT’?

src/utilities/data/Vector.cpp
```diff
@@ -96,10 +96,11 @@ InterpInfo interpInfo(const Vector& x, double xi, double ti) {
 
     result.ia = (unsigned)(it - begin - 1);
     result.ib = (unsigned)(it - begin);
-    if (ti == 0.0) {
+    if (ti < 0.0) {
       result.wa = (x(result.ib) - xi) / (x(result.ib) - x(result.ia));
       result.wb = (xi - x(result.ia)) / (x(result.ib) - x(result.ia));
     } else {
+      OS_ASSERT(ti > 0.0);
```
jmarrec, line 105: I think the OS_ASSERT here is pointless. You want it in interp

```
if (interpMethod == AverageInterp && ti <= 0.0) {
  LOG_FREE_AND_THROW(...)
```

jmarrec (changes requested):

jmarrec (approved):

PR #5118, opened 2024-03-22
Fix #1472 - Respect user's SimulationControl choices and move smart defaults to the modelObject

### LINKED ISSUE #1472: SimulationControl choices not translating to E+
Looks like the choices from SimulationControl aren't being translated to E+ correctly (ref: [UnmetHours](https://unmethours.com/question/3054/disable-zoneplantsystem-sizing/)).

openstudio_issue1472-1b
openstudio_issue1472-2b

openstudio_issue1p
openstudio_issue2p

jmarrec:
Pull request overview
---------------------

 - Fixes #1472

- We always respect a value that the user has picked
    - I don't care if this means that the E+ simulation will crash. It's going to be clear that's the problem
    - I do issue a Warning if the user has deliberately picked "No" but it looks like it should be enabled
- When defaulted, the smart default is in the model object itself

@kbenne does that make sense to you?

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/model/SimulationControl.cpp
+++ b/src/model/SimulationControl.cpp
@@ -7,22 +7,30 @@
 #include "SimulationControl_Impl.hpp"
 
 #include "Model.hpp"
+
+#include "AirLoopHVAC.hpp"
+#include "AirLoopHVAC_Impl.hpp"
 #include "ConvergenceLimits.hpp"
 #include "ConvergenceLimits_Impl.hpp"
 #include "HeatBalanceAlgorithm.hpp"
 #include "HeatBalanceAlgorithm_Impl.hpp"
 #include "InsideSurfaceConvectionAlgorithm.hpp"
 #include "InsideSurfaceConvectionAlgorithm_Impl.hpp"
+#include "PlantLoop.hpp"
+#include "PlantLoop_Impl.hpp"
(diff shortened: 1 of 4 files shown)


src/model/SimulationControl.cpp
```diff
@@ -169,29 +179,94 @@ namespace model {
...
+        }
+        return result;
+      }
+
+      const auto result = shouldItBeOn();
+      if (result) {
+        LOG(Debug, "You have zonal equipment and design days, and SimulationControl::DoZoneSizingCalculation is defaulted: turning on.");
+      }
+
+      return result;
```
jmarrec, line 211: Example with doZoneSizing: the model object has smart logic.

* If not defaulted, but we think it should be on: warn
* If defaulted, we check whether we think it should be on. If so, since this is the opposite of the IDD default, we issue a Debug message explaining why we picked it

kbenne:
@jmarrec I like the principles that you laid out and I expect that most other people will too. I think this is a good pattern to use as much as we can.

* We always respect a value that the user has picked, even if the E+ simulation will crash. 
* Issue a Warning if the user has deliberately picked a value that will result in EnergyPlus crashing.
* When defaulted, the smart default is in the model object itself

kbenne (approved):

PR #5119, opened 2024-03-22
4847 logger singleton 3.1

jmarrec:
Pull request overview
---------------------

I just want CI to run on it to I'm targeting develop for now, but really this is just a modification of #5110

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


src/utilities/core/Logger.hpp
```diff
@@ -8,18 +8,14 @@
 
 #include "../UtilitiesAPI.hpp"
 
-#include "Singleton.hpp"
-#include "Exception.hpp"
-#include "Compare.hpp"
+#include "Compare.hpp"  // NOTE that this this is not needed for this header but so many compilation errors if I omit it
+                        // since other files transitively use Compare
```
jmarrec, line 12: An unfortunate discovery. Cleaning up ours headers would be good, but that's not what I'm trying to achieve here.

src/utilities/core/Logger.hpp
```diff
@@ -94,40 +93,10 @@ class UTILITIES_API LoggerImpl
...
-  static LoggerImpl& instance() {
-    if (!obj) {
-      obj = std::shared_ptr<LoggerImpl>(new LoggerImpl());
-    }
-    return *obj;
-  }
-
- private:
-  static std::shared_ptr<LoggerImpl> obj;
+  std::shared_ptr<detail::Logger_Impl> m_impl;
```
jmarrec, line 99: Given that the Logger is a Meyers singleton, I'm not sure whether the Logger_Impl class is really needed at all. But it feels familiar at least, and can allow us to shorten the public includes

src/utilities/core/Logger.cpp
```diff
@@ -4,134 +4,175 @@
...
+    // }
   }
+
+}  // namespace detail
+
+Logger::Logger() : m_impl(std::shared_ptr<detail::Logger_Impl>(new detail::Logger_Impl())) {}
+
+Logger& Logger::instance() {
+  static Logger instance;
+  return instance;
```
jmarrec, line 140: Private constructor, and public ::instance()

src/utilities/core/Logger.hpp
kbenne: I agree, I don't think we really need an Impl at all.

Issue #5121, opened 2024-03-25
Strange crash when removing Unitheater from thermalzone equipment list


jmarrec:
Issue overview
--------------

I am extremely puzzled about what's happening but I can reproduce a crash originally reported at https://github.com/openstudiocoalition/OpenStudioApplication/issues/686 on windows (it works on Ubuntu).

## Current Behavior

It seems that when the extensible group corresponding to the UH is deleted, something very strange happens.

In the following group, which points correctly to a ZoneVentilationDesignFlowRate object to begin with, the handle is changed to be the handle of that object's Schedule Name (?!)

## Steps to Reproduce

I'll provide a shortened file asap: edit: Here is a mcve.osm:

[mcve.osm.txt](https://github.com/NREL/OpenStudio/files/14742397/mcve.osm.txt)

```shell
C:\openstudio-3.7.0\bin\openstudio -e "m = OpenStudio::Model::Model::load('mcve.osm').get; uh = m.getZoneHVACUnitHeaters.first; uh.remove; puts 'ok'"
[BOOST_ASSERT] <2> Assertion eg failed on line 346 of bool __cdecl openstudio::model::detail::ZoneHVACEquipmentList_Impl::removeEquipment(const class openstudio::model::ModelObject &) in file D:\OSN\src\model\ZoneHVACEquipmentList.cpp.
[BOOST_ASSERT] <2> Assertion eg failed on line 346 of bool __cdecl openstudio::model::detail::ZoneHVACEquipmentList_Impl::removeEquipment(const class openstudio::model::ModelObject &) in file D:\OSN\src\model\ZoneHVACEquipmentList.cpp.
[BOOST_ASSERT] <2> Assertion eg failed on line 356 of bool __cdecl openstudio::model::detail::ZoneHVACEquipmentList_Impl::removeEquipment(const class openstudio::model::ModelObject &) in file D:\OSN\src\model\ZoneHVACEquipmentList.cpp.
[BOOST_ASSERT] <2> Assertion eg failed on line 356 of bool __cdecl openstudio::model::detail::ZoneHVACEquipmentList_Impl::removeEquipment(const class openstudio::model::ModelObject &) in file D:\OSN\src\model\ZoneHVACEquipmentList.cpp.
ok
```

```ruby
m = OpenStudio::Model::Model::load('mcve.osm')
uh.remove
```

test2:

```ruby
require 'openstudio'

include OpenStudio::Model

def getGroupForModelObject(eqlist, modelObject)
[...]

PR #5122, opened 2024-03-25
#5121 - Extensible Groups problems in ModelObject/WorkspaceObject

jmarrec:
Pull request overview
---------------------

 - Fix for #5121 
 
Seems like the problem is that ModelObject, via WorkspaceObject resolves any handle field to the name the target, and during eraseExtensibleGroup, that goes nuclear, and if the Schedule is named like the object, and the schedule is found first in the workspace, then you end up with the Schedule being on the ZoneHVACEquipmentList...

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/utilities/idf/IdfExtensibleGroup.cpp
+++ b/src/utilities/idf/IdfExtensibleGroup.cpp
@@ -22,6 +22,17 @@ StringVector IdfExtensibleGroup::fields(bool returnDefault) const {
   return result;
 }
 
+StringVector IdfExtensibleGroup::fieldsWithHandles(bool returnDefault) const {
+  StringVector result;
+  const unsigned n = numFields();
+  result.reserve(n);
+  for (unsigned i = 0; i < n; ++i) {
+    OptionalString str = getField(i, returnDefault);
+    result.push_back(str.get_value_or(""));
+  }
+  return result;
+}
+
 StringVector IdfExtensibleGroup::fieldComments(bool returnDefault) const {
   StringVector result;
   for (unsigned i = 0, n = numFields(); i < n; ++i) {
@@ -55,6 +66,13 @@ OptionalString IdfExtensibleGroup::getString(unsigned fieldIndex, bool returnDef
     return boost::none;
   }
   return m_impl->getString(mf_toIndex(fieldIndex), returnDefault);
+}
+
+boost::optional<std::string> IdfExtensibleGroup::getField(unsigned index, bool returnDefault) const {
+  if (!isValid(index)) {
+    return boost::none;
+  }
+  return m_impl->getField(mf_toIndex(index), returnDefault);
 }
 
 OptionalDouble IdfExtensibleGroup::getDouble(unsigned fieldIndex, bool returnDefault) const {
@@ -169,7 +187,7 @@ IdfExtensibleGroup IdfExtensibleGroup::pushClone() const {
     return {p, 0};
   }
 
-  StringVector values = fields();
+  StringVector values = fieldsWithHandles();
   OS_ASSERT(values.size() == numFields());
   return m_impl->pushExtensibleGroup(values);
 }
@@ -180,7 +198,7 @@ IdfExtensibleGroup IdfExtensibleGroup::insertClone(unsigned groupIndex) const {
     return {p, 0};
   }
 
-  StringVector values = fields();
+  StringVector values = fieldsWithHandles();
   OS_ASSERT(values.size() == numFields());
   return m_impl->insertExtensibleGroup(groupIndex, values);
 }
--- a/src/utilities/idf/IdfExtensibleGroup.hpp
+++ b/src/utilities/idf/IdfExtensibleGroup.hpp
@@ -40,9 +40,12 @@ class UTILITIES_API IdfExtensibleGroup
   /** @name Getters */
   //@{
 
-  /** Returns this extensible group's fields. Return value will be empty() if this group is
-   *  empty(). */
+  /** Returns this extensible group's fields. Return value will be empty() if this group is empty(). */
   std::vector<std::string> fields(bool returnDefault = false) const;
+
+  /** Returns this extensible group's fields. Return value will be empty() if this group is empty().
+    * Unlike fields(), in the case it's a WorkspaceObject_Impl, it uses handles as string and not the name of the target object */
+  std::vector<std::string> fieldsWithHandles(bool returnDefault = false) const;
 
   /** Returns the comments associated with this extensible group's fields. */
   std::vector<std::string> fieldComments(bool returnDefault = false) const;
@@ -53,6 +56,10 @@ class UTILITIES_API IdfExtensibleGroup
   /** Returns the string value of this extensible group's field fieldIndex, if that field
    *  exists (isValid(fieldIndex)). */
   boost::optional<std::string> getString(unsigned fieldIndex, bool returnDefault = false) const;
+
+  /** Like getString except for reference fields getString will return the name of the referenced object.
+   * This method, getField, will always return the string value  of the field.   */
+  boost::optional<std::string> getField(unsigned index, bool returnDefault = false) const;
 
   /** Returns true if the field is empty. */
   bool isEmpty(unsigned fieldIndex) const;
--- a/src/utilities/idf/IdfObject.cpp
+++ b/src/utilities/idf/IdfObject.cpp
@@ -209,6 +209,10 @@ namespace detail {
     return boost::none;
   }
 
+  boost::optional<std::string> IdfObject_Impl::getField(unsigned index, bool returnDefault) const {
+    return IdfObject_Impl::getString(index, returnDefault, false);
+  }
+
   boost::optional<double> IdfObject_Impl::getDouble(unsigned index, bool returnDefault) const {
     OptionalDouble result;
     OptionalString value = getString(index, returnDefault, false);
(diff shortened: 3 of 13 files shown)


src/utilities/idf/IdfObject.cpp
```diff
@@ -825,14 +829,14 @@ namespace detail {
       StringVector temp = result;
       IdfExtensibleGroup eg = getExtensibleGroup(i);
       OS_ASSERT(!eg.empty());
-      result = eg.fields();
+      result = eg.fieldsWithHandles();
```
jmarrec, line 832: Ok, so here is develop version of eraseExtensibleGroup: https://github.com/NREL/OpenStudio/blob/f916fc13ef0f90bf42b100e2050455f75c404667/src/utilities/idf/IdfObject.cpp#L806-L852


Basically it goes:

* popExtensibleGroup (remove last extensible group). This will actually resolve to WorkspaceObject_Impl::popExtensibleGroup which nullify the pointers too
* Then loop from last to groupIndex and copy groups / rollup until overwrite groupIndex
    * That rollup for each does basically this 
```
for i in [numExtensibleGroups() -1, groupIndex):
   prev_fields = extensibleGroups[i+1].fields() 
   extensibleGroups[i].setFields(prev_fields)
```
   
* The issue is that IdfExtensibleGroup::fields()  calls m_impl->getString which as reminder for a WorkspaceObject, actually **resolves** any objectList field to the Target workspaceobject's name, rather than return the handle.
 
* setFields in turns calls setString, which for a workspace object, if passed an object name for an objectList field, will convert to a target handle via `WorkspaceObject_Impl::convertToTargetHandle` which calls `WorkspaceObject_Impl::getObjectsByName`. If you have more than one candidate, it looks at reference lists too (but `OS:ZoneHVAC:EquipmentList` 's Equipment Name has `\reference AllObjects` (maybe that's a mistake too, but outside the point here), but bottom line is that if there are **more than one candidate, it returns the first**.

This means that in the case of a ZoneHVACEquipmentList that points to an Equipment named `XXXX` and you do have another modelObject named `XXXX` and it is found first, the call sequence goes:

```
prev_fields = ["XXXX", "", "", ""]
eg.setFields(prev_fields)
   eg.setString(0, "XXXX")
     handleOfFirstModelObjectNamedXXXX = "the handle of the wrong modelobject"
      eg.setString(0, handleOfFirstModelObjectNamedXXXX) => succeed because AllObjects allows it
```

src/utilities/idf/IdfExtensibleGroup.hpp
```diff
@@ -40,10 +40,13 @@ class UTILITIES_API IdfExtensibleGroup
...
   //@{
 
-  /** Returns this extensible group's fields. Return value will be empty() if this group is
-   *  empty(). */
+  /** Returns this extensible group's fields. Return value will be empty() if this group is empty(). */
   std::vector<std::string> fields(bool returnDefault = false) const;
 
+  /** Returns this extensible group's fields. Return value will be empty() if this group is empty().
+    * Unlike fields(), in the case it's a WorkspaceObject_Impl, it uses handles as string and not the name of the target object */
+  std::vector<std::string> fieldsWithHandles(bool returnDefault = false) const;
```
jmarrec, line 48: New method to return always the handles as strings

src/utilities/idf/IdfExtensibleGroup.cpp
```diff
@@ -22,6 +22,17 @@ StringVector IdfExtensibleGroup::fields(bool returnDefault) const {
...
+StringVector IdfExtensibleGroup::fieldsWithHandles(bool returnDefault) const {
+  StringVector result;
+  const unsigned n = numFields();
+  result.reserve(n);
+  for (unsigned i = 0; i < n; ++i) {
+    OptionalString str = getField(i, returnDefault);
+    result.push_back(str.get_value_or(""));
+  }
+  return result;
+}
```
jmarrec, line 34: It relies on getField

src/utilities/idf/IdfObject_Impl.hpp
```diff
@@ -101,6 +101,11 @@ namespace detail {
      */
     virtual boost::optional<std::string> getString(unsigned index, bool returnDefault = false, bool returnUninitializedEmpty = false) const;
 
+    /** Like getString except for reference fields getString will return the name of the referenced object.
+     * This method, getField, will always return the string value of the field.
+     * For IdfObject, this is the same as getString */
+    virtual boost::optional<std::string> getField(unsigned index, bool returnDefault = false) const;
```
jmarrec, line 107: `getField` was only a WorkspaceObject method before. Now it's on IdfObject for the public side, IdfObject_Impl on the impl side and overriden in WorkspaceObject_Impl.

This allows IdfExtensibleGroup, since it uses a std::shared_ptr<IdfObject_Impl>, to use polymorphism to "do the right thing(TM)"

src/utilities/idf/IdfObject.cpp
```diff
@@ -209,6 +209,10 @@ namespace detail {
     return boost::none;
   }
 
+  boost::optional<std::string> IdfObject_Impl::getField(unsigned index, bool returnDefault) const {
+    return IdfObject_Impl::getString(index, returnDefault, false);
+  }
```
jmarrec, line 214: For IdfObject, getField does getString.

src/utilities/idf/WorkspaceObject.cpp
```diff
@@ -188,7 +188,7 @@ namespace detail {
     return IdfObject_Impl::getString(index, returnDefault, returnUninitializedEmpty);
   }
 
-  boost::optional<std::string> WorkspaceObject_Impl::getField(unsigned index) const {
+  boost::optional<std::string> WorkspaceObject_Impl::getField(unsigned index, bool returnDefault) const {
```
jmarrec, line 191: for WorkspaceObject, it does **IdfObject_Impl::getString** for any non-source field, and grabs the handle otherwise

src/utilities/idf/WorkspaceObject.hpp
```diff
@@ -66,12 +66,6 @@ class UTILITIES_API WorkspaceObject : public IdfObject
   /** Returns all objects of type that point to this object, filtering for duplicate objects. */
   std::vector<WorkspaceObject> getSources(IddObjectType type) const;
 
-  /** Like getString except for reference fields getString will return the
-   *  name of the referenced object. This method, getField, will always return the string value
-   *  of the field.
-   */
-  boost::optional<std::string> getField(unsigned index) const;
```
jmarrec, line 73: The public forward to impl is done on IdfObject.

src/utilities/idf/WorkspaceObject_Impl.hpp
```diff
@@ -129,7 +129,7 @@ namespace detail {
      *  (non-extensible) fields and fields with empty data, if a default exists. */
     virtual boost::optional<std::string> getString(unsigned index, bool returnDefault = false, bool returnUninitializedEmpty = false) const override;
 
-    boost::optional<std::string> getField(unsigned index) const;
+    virtual boost::optional<std::string> getField(unsigned index, bool returnDefault = false) const override;
```
jmarrec, line 132: I added a returnDefault param to match the others, with the original default of false

src/utilities/idf/Test/WorkspaceObject_GTest.cpp
```diff
@@ -585,6 +591,59 @@ TEST_F(IdfFixture, WorkspaceObject_setName_allObjects) {
   }
 }
 
+TEST_F(IdfFixture, WorkspaceObject_eraseExtensibleGroups) {
```
jmarrec, line 594: new test

src/utilities/idf/Test/ExtensibleGroup_GTest.cpp
```diff
@@ -488,17 +491,19 @@ TEST_F(IdfFixture, ExtensibleGroup_Erase) {
   ASSERT_FALSE(eg.empty());  // Brick, Insulation, Insulation, Gypsum
 
   // erase and check pointers
-  EXPECT_EQ(static_cast<unsigned>(2), temp[2].numSources());
+  EXPECT_EQ(static_cast<unsigned>(2), wsInsulation.numSources());
   values = wsConstruction.eraseExtensibleGroup(1);
   ASSERT_EQ(static_cast<unsigned>(1), values.size());
-  EXPECT_EQ("Insulation", values[0]);  // Brick, Insulation, Gypsum
-  EXPECT_EQ(static_cast<unsigned>(1), temp[2].numSources());
+  EXPECT_EQ(openstudio::toString(wsInsulation.handle()), values[0]);  // Brick, Insulation, Gypsum
```
jmarrec, line 497: Adjust existing test since return of eraseExtensibleGroup(index) for **WorkspaceObject** changes from object name to handle string for source fields. I don't think we care about this change.

jmarrec (commented):
This was a doozy to track down, but I think the fix is sound. I added a thorough review to explain what's going on and why I did these changes, since I can't expect anyone to waste the required amount of time needed to figure it out themselves.

src/utilities/idf/IdfObject.cpp
jmarrec: My fix is to always grab source fields as handles.

```
prev_fields = ["{handle of equipment}", "", "", ""]
eg.setFields(prev_fields)
      eg.setPointer(0, "{handle of equipment}")
```

kbenne: Is this also a peformance improvement? Surely yes. 

Thanks for the narrative Julien.

PR #5134, opened 2024-04-01
Handle incomplete EPW design conditions header

### LINKED ISSUE #5132: EPWDesignCondition getters should return boost::optional
Issue overview
--------------
Missing design condition (double?) fields in the header of the EPW file return a zero, and they shouldn't.

See https://github.com/NREL/OpenStudio/pull/3086 for the original implementation.

## Steps to Reproduce

1.
2.
3.
4.

### Environment

joseph-robertson:
Pull request overview
---------------------

 - Fixes #5132.
 - `EpwDesignCondition` has many **API-breaking changes** related to its getters. The previous behavior was to misleadingly return a value of 0 for any empty design condition header field. The types for the getters are now either `boost::optional<double>` or `boost::optional<int>`.

Notes:
- Does `stringToDouble` return 0.0 if an empty string is passed in? Yes.
- Should the following changes be made to EpwFile.hpp / EpwFile.cpp:
  - check for empty strings in setters; e.g., check `heatingDryBulb99` is empty string in `setHeatingDryBulb99`? Yes.
  - change all getters to return `boost::optional<double>`; e..g, `double EpwDesignCondition::heatingDryBulb99() const {` to `boost::optional<double> EpwDesignCondition::heatingDryBulb99() const {`? Yes.
  - change, e.g., `double m_heatingDryBulb99` to `boost::optional<double> m_heatingDryBulb99` in EpwDesignCondition class? Yes.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


jmarrec:
Rebased onto develop and resolved conflicts.

src/utilities/filetypes/EpwFile.cpp
```diff
@@ -2941,7 +2956,11 @@ void EpwDesignCondition::setHeatingDryBulb99pt6(double heatingDryBulb99pt6) {
 bool EpwDesignCondition::setHeatingDryBulb99(const std::string& heatingDryBulb99) {
   bool ok;
   double value = stringToDouble(heatingDryBulb99, &ok);
-  setHeatingDryBulb99(value);
+  if (!ok) {
+    m_heatingDryBulb99 = boost::none;
+  } else {
+    setHeatingDryBulb99(value);
+  }
   return ok;
```
jmarrec, line 2964: Not in the scope of this PR, but I dislike this. `stringToDouble` should be `boost::optional<double> stringToDouble(const std::string& str)`.

Also a lot of fields in `EpwDataPoint` internally store a `std::string` representation of a number, use magic string numbers for unknown values (constants would be better, whether that's a string or an actual number), which I also don't like.

https://github.com/NREL/OpenStudio/blob/cae94352f7d5ce2c44602dbda113fcb7bf085ab8/src/utilities/filetypes/EpwFile.cpp#L1718-L1744

But this isn't something we need to address now.

jmarrec (approved):
Made a couple of minor changes

jmarrec:
I didn't find occurrences of this being used anywhere in the common openstudio-gems I have on my machine right now.

The only one I can find is ResStock's weather.rb, but I'm sure @joseph-robertson is well aware he'll need to change that so I'm fine with the API Break. @kbenne, grand protector of the API, could you chime in for confirmation?

https://github.com/NREL/resstock/blob/280a58c2686f30a1b09f10ba3640e49c265d06f3/resources/hpxml-measures/HPXMLtoOpenStudio/resources/weather.rb#L261-L276)

joseph-robertson:
> I didn't find occurrences of this being used anywhere in the common openstudio-gems I have on my machine right now.
> 
> The only one I can find is ResStock's weather.rb, but I'm sure @joseph-robertson is well aware he'll need to change that so I'm fine with the API Break. @kbenne, grand protector of the API, could you chime in for confirmation?
> 
> https://github.com/NREL/resstock/blob/280a58c2686f30a1b09f10ba3640e49c265d06f3/resources/hpxml-measures/HPXMLtoOpenStudio/resources/weather.rb#L261-L276)

@jmarrec I think OS-HPXML's weather.rb was originally what motivated this PR; there's no way to know if it's actually zero vs. missing.

However, did you see my "-9999" question here? https://github.com/NREL/OpenStudio/pull/5318#issuecomment-2551894811

jmarrec:
It's debatable whether -9999 or boost::optional is better than the other, and I'd be fine with both probably. As long as we convey that it's missing, that's probably fine.

The -9999 option does allow not breaking API, which is a plus. And even for dry bulb temperature, you could always check if < -100C and you'd be guaranted to know it's missing. 
The downside is that it's not as clear that it's missing (dereferencing an empty optional will produce a crash, while you could technically just that -9999 value for calculations and just get bogus results without noticing). But as a mitigation factor: 1) the EPW format itself includes some -99 or -9999 values to indicate missing values (and so does EnergyPlus internally for AutoSized fields for eg), and 2) I don't think this is an area of the SDK that is largely used.

joseph-robertson:
> It's debatable whether -9999 or boost::optional is better than the other, and I'd be fine with both probably. As long as we convey that it's missing, that's probably fine.
> 
> The -9999 option does allow not breaking API, which is a plus. And even for dry bulb temperature, you could always check if < -100C and you'd be guaranted to know it's missing. The downside is that it's not as clear that it's missing (dereferencing an empty optional will produce a crash, while you could technically just that -9999 value for calculations and just get bogus results without noticing). But as a mitigation factor: 1) the EPW format itself includes some -99 or -9999 values to indicate missing values (and so does EnergyPlus internally for AutoSized fields for eg), and 2) I don't think this is an area of the SDK that is largely used.

@jmarrec Hm. Well what's your vote here?

Should we at least be consistent with the new EPW ground temperatures implementation? I.e., change all ground temps getters to  boost::optionals if we decide to go with this PR as-is?

Issue #4993, opened 2023-10-04
Add native support for measure argument methods in OsLib_HelperMethods 


mdahlhausen:
Enhancement Request
--------------
There are several methods in `OsLib_HelperMethods` in the extension gem that should be native to the `Measure` or `Measure::OSArgument` objects in OpenStudio.

## Detailed Description
These are the specific methods that **_could_** be implemented in OpenStudio:
```
- check_upstream_measure_for_arg(runner, arg_name)
- checkChoiceArgFromModelObjects(object, variableName, to_ObjectType, runner, user_arguments)
- checkDoubleAndIntegerArguments(runner, user_arguments, arg_check_hash)
- checkOptionalChoiceArgFromModelObjects(object, variableName, to_ObjectType, runner, user_arguments)
- createRunVariables(runner, model, user_arguments, arguments)
- populateChoiceArgFromModelObjects(model, modelObject_args_hash, includeBuilding)
```

## Possible Implementation
To be clear, I DO NOT think these methods should be adopted as is. Some are already partially or mostly addressed (checkDoubleAndIntegerArguments with #4551). @DavidGoldwasser - can you analyze which methods should live in OpenStudio, and which are best left to code in the measure, then comment below? @eringold may have thoughts as well.

joseph-robertson:
Another method:
```
- getAbsoluteAzimuthForSurface(surface, model)
```
(/openstudio-extension-gem/lib/openstudio/extension/core/os_lib_geometry.rb)

joseph-robertson:
Perhaps we could create an overloaded `runner.validateUserArguments(script_arguments, user_arguments, model)` method (note the `model` is new) that checks, for any user argument that has a handle (i.e., a UUID), whether that handle exists in `model`.

Would this cover some of the functionality of the `OsLib_HelperMethods.checkXXX` methods above?

jmarrec:
`populateChoiceArgFromModelObjects`. How about we do it this way:

static OSArgument makeChoiceArgument(const std::string& name, const std::map<std::string, std::string>& choices_to_display_values_map, bool required = true,  bool modelDependent = false);

That way you could do something like

```ruby
choices_to_display_values_map = (
  model.getInteriorPartitionSurfaces
       .reject{|s| s.construction.empty?}
       .map(&:construction).map(&:get)
       .map{|c| [c.handle.to_s, c.nameString]}.to_h
)
[...]

PR #5135, opened 2024-04-02
Add native support for measure argument methods in OsLib_HelperMethods, try 2

joseph-robertson:
Pull request overview
---------------------

 * Addresses #4993.

Add methods that return `Json::Value`:
* `runner.getArgumentValues(script_arguments, user_arguments)`
* `runner.getPastStepValuesForMeasure(measure_name)`
* `runner.getPastStepValuesForName(step_name)`

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


src/measure/OSRunner.cpp
```diff
@@ -784,6 +786,68 @@ namespace measure {
...
+        Json::Value root = it->second.toJSON();
+        if (auto value = root["value"]) {
+          argument_values[name] = value;
+        }
+      }
+    }
+    return argument_values;
+  }
+
+  Json::Value OSRunner::getPastStepValuesForMeasure(const std::string& measure_name) {
```
jmarrec, line 805: * I'm still confused about the difference between the getPastStepValuesForMeasure and getPastStepValuesForName, but I'll figure that out.
* The above is mostyl due to how unreadable both methods are. The flow logic places an unnecessarily high cognitive burden on the reader. I'll revamp   
* Both of these should be marked const.
    * Generally speaking, the entire existing OSRunner class is completely messed up as far as const-correctness and staticness of methods that can be (like getOptionalXXXArgument), but I'm wary of changing this since it's an ABI break

src/measure/OSRunner.cpp
```diff
@@ -784,6 +786,68 @@ namespace measure {
     return result;
   }
 
+  Json::Value OSRunner::getArgumentValues(std::vector<OSArgument>& script_arguments, const std::map<std::string, OSArgument>& user_arguments) {
```
jmarrec, line 789: This should be static actually.

Issue #5153, opened 2024-04-18
Interior windows missing required Outside Boundary Condition Object when ModelMerger


macumber:
Issue overview
--------------
Originally reported as an issue with FloorspaceJS translator at https://github.com/NREL/OpenStudio/issues/4670

@manuvarkey found the issue actually occurs in ModelMerger

## Current Behavior
Adjacent Surfaces are matched after cloning in  ModelMerger but adjacent SubSurfaces are not

## Expected Behavior
Adjacent surfaces and SubSurfaces are matched after cloning in ModelMerger 

## Steps to Reproduce

```
import openstudio as os

# Open floorplan and get translated OSM model
with open('floorplan.json') as fp:
    json_str = fp.read()
fp_json = os.FloorplanJS(json_str)
ts_scene = fp_json.toThreeScene(True)
rt = os.model.ThreeJSReverseTranslator()
os_model_tr = rt.modelFromThreeJS(ts_scene).get()
model_handle_mapping = rt.handleMapping()

# Merge translated model with blank model and save
model = os.model.Model()
mm = os.model.ModelMerger()
mm.mergeModels(model, os_model_tr, model_handle_mapping)
model.save('out2.osm')
```

## Possible Solution
On going through the code for OpenStudio/src/model/ModelMerger.cpp, the following code segment was found dealing with setting adjacentSurface for Surface elements.

https://github.com/NREL/OpenStudio/blob/d2f0bdd21a70e6a9711f6da8f5aabecadc4a1a1c/src/model/ModelMerger.cpp#L349C1-L359C6

```
      // DLM: this should probably be moved to a mergeSurface method
      auto clone = newSurface.clone(m_currentModel).cast<Surface>();
      clone.setSpace(currentSpace);

      m_newMergedHandles.insert(newSurface.handle());
      m_currentToNewHandleMapping[clone.handle()] = newSurface.handle();
      m_newToCurrentHandleMapping[newSurface.handle()] = clone.handle();

      boost::optional<Surface> newAdjacentSurface = newSurface.adjacentSurface();
      if (newAdjacentSurface) {
        boost::optional<UUID> currentAdjacentSurfaceHandle = getCurrentModelHandle(newAdjacentSurface->handle());
        if (currentAdjacentSurfaceHandle) {
          boost::optional<Surface> currentAdjacentSurface = m_currentModel.getModelObject<Surface>(*currentAdjacentSurfaceHandle);
          if (currentAdjacentSurface) {
[...]

PR #5154, opened 2024-04-18
Preserve SubSurface Outside Boundary Condition Object in ModelMerger

macumber:
Pull request overview
---------------------

Fixes https://github.com/NREL/OpenStudio/issues/5153

Preserve SubSurface Outside Boundary Condition Object in ModelMerger

### Pull Request Author

 - [X] All new and existing tests passes

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/model/ModelMerger.cpp
+++ b/src/model/ModelMerger.cpp
@@ -357,16 +357,16 @@ namespace model {
         }
       }
 
-      // match new to cloned subsurfaces
+      // setAdjacentSurface resets the AdjacentSubSurface on all child subsurfaces
       for (const auto& newSubSurface : newSurface.subSurfaces()) {
-        m_newMergedHandles.insert(newSubSurface.handle());
-        for (auto& cloneSubSurface : clone.subSurfaces()) {
-          if (circularEqual(newSubSurface.vertices(), cloneSubSurface.vertices(), 0.01)) {
-            m_currentToNewHandleMapping[cloneSubSurface.handle()] = newSubSurface.handle();
-            m_newToCurrentHandleMapping[newSubSurface.handle()] = cloneSubSurface.handle();
-
-            boost::optional<SubSurface> newAdjacentSubSurface = newSubSurface.adjacentSubSurface();
-            if (newAdjacentSubSurface) {
+        // for performance reasons, only find matching subsurfaces if there is an AdjacentSubSurface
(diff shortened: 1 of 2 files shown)


src/model/ModelMerger.cpp
```diff
@@ -356,6 +356,29 @@ namespace model {
...
+                boost::optional<SubSurface> currentAdjacentSubSurface = m_currentModel.getModelObject<SubSurface>(*currentAdjacentSubSurfaceHandle);
+                if (currentAdjacentSubSurface) {
+                  cloneSubSurface.setAdjacentSubSurface(*currentAdjacentSubSurface);
+                }
+              }
+            }
+            break;
+          }
+        }
+      }
```
jmarrec, line 381: Ok, so:

1. We always emplace subSurfaces in m_newMergedHandles
2. we do resolve m_currentToNewHandleMapping and m_newToCurrentHandleMapping
3. Only if newAdjacentSubSurface then we try to setAdjacentSubSurface

Do we _need_ 1 & 2 if `newAdjacentSubSurface` is empty?

I can see it follows the intent, **but** matching via `circularEqual` is expensive and it could be a big slow down on large models with lots of surfaces & more importantly lots of subsurfaces on each surface.

macumber: The issue was that `clone` clones all the children without any support for the handle mapping so I tried to find a way to find the same sub surfaces in both models after the clone.  I thought about matching sub surfaces based on names but there are some degenerate cases that would fail on (e.g. merge model {Space1 with SubSurface1, Space2 with SubSurface2} with model {Space1 with SubSurface2, Space2 with SubSurface1} might change object names based on the order spaces are processed).

My thought was to add all sub surfaces to the handle mappings for completeness, but we don't 100% need it.  I would be ok to only do the checks if `adjacentSubSurface` is not empty.  I'll update this tonight.

Changes to clone to support the handle mapping would be possible, and might be generally useful, just a bigger change.

macumber: I made the change and tested it locally.  I updated my branch and pushed the change however I can no longer build after the branch was updated to latest develop, I don't have my environment set up for 3.8.0 yet.

jmarrec: https://github.com/NREL/OpenStudio/blob/develop/BUILDING.md

PR #5165, opened 2024-04-23
IRB Debug Init + byebug

jmarrec:
Pull request overview
---------------------

* Companion PR: https://github.com/NREL/openstudio-gems/pull/84

- Opening as Draft because I'm not sure this is going to be wanted

This PR does two things:
- When loglevel is Trace, it preloads the IRB code so you can put `binding.irb` anywhere and it'll work. That includes inside our gem initialization code, but will work in any external script too. This allows you to inspect variables, but you don't have frame execution, because "debug" gem is built shared actually... see https://github.com/jmarrec/conan-recipes/issues/6
- After initialization is done, meaning in an external script for eg, you can `require 'byebug'` and put `byebug` and this **will allow you to have frame execution**, see this example

byebug

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


kbenne:
@jmarrec. This would be very useful for the kind of issues we just worked through with bundler. I'm sure that was part of the inspiration. What would be the downside of merging this? It seems like a win/win.

jmarrec:
Downsides include:
* people who do ask for trace loglevel get code they don't need (they will not notice and noone asks for trace level output unlesd they're in trouble so big they wouldn't care), so noop here
* byebug is a new native ext gem

CMakeLists.txt
```diff
@@ -835,7 +835,7 @@ if(BUILD_CLI)
   set(OPENSTUDIO_GEMS_BASEURL "http://openstudio-resources.s3.amazonaws.com/dependencies")
 
   # TODO: temp
-  set(OPENSTUDIO_GEMS_BASEURL "https://github.com/NREL/openstudio-gems/releases/download/v3.8.0-RC1-5")
+  set(OPENSTUDIO_GEMS_BASEURL "https://github.com/NREL/openstudio-gems/releases/download/v3.8.0-RC1-6")
```
jmarrec, line 838: @kbenne done

PR #5166, opened 2024-04-23
Fix #5162 - Capture `exit` from Ruby

### LINKED ISSUE #5162: "exit 1" in ruby code is silently dismissed
Issue overview
--------------

Due to this change https://github.com/NREL/OpenStudio/commit/a52f8e00bf8cdfacd316d2c6e249b94d81fa7f9b

in **test.rb**, just put `exit 1`

`openstudio test.rb`

In 3.7.0, it was capturing the error, now it thinks it's the help command...

## Steps to Reproduce

1.
2.
3.
4.

### Environment

jmarrec:
Pull request overview
---------------------

 - Fixes #5162

**Note: this _does_ fix the Minitest return value, even with the `minitest/autorun` using at_exit**

https://github.com/minitest/minitest/blob/ea9caafc0754b1d6236a490d59e624b53209734a/lib/minitest.rb#L66-L89

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/ruby/engine/RubyEngine.cpp
+++ b/ruby/engine/RubyEngine.cpp
@@ -75,7 +75,13 @@ void RubyEngine::setupEmbeddedGems(const std::vector<openstudio::path>& includeD
 }
 
 RubyEngine::~RubyEngine() {
-  ruby_finalize();
+  // ruby_cleanup calls ruby_finalize
+  int ex = ruby_cleanup(0);
+  if (ex != 0) {
+    fmt::print("RubyEngine return code was {}\n", ex);
+    exit(ex);
+  }
+  //ruby_finalize();
 }
 
 ScriptObject RubyEngine::eval(std::string_view sv) {
--- a/ruby/engine/RubyEngine.cpp
+++ b/ruby/engine/RubyEngine.cpp
@@ -95,26 +95,18 @@ void* RubyEngine::getAs_impl(ScriptObject& obj, const std::type_info& ti) {
 
   void* return_value = nullptr;
 
-  // TODO: this sucks, and probably needs memory management
-  if (ti == typeid(std::string*)) {
-    char* cstr = StringValuePtr(val);
-    size_t size = RSTRING_LEN(val);  // + 1;  From trial and eror, if I had + 1 I get a string with one two many size
-    return_value = new std::string(cstr, size);
+  const auto& type_name = getRegisteredTypeName(ti);
 
-    // std::string s = rb_string_value_cstr(&val);
-  } else {
-    const auto& type_name = getRegisteredTypeName(ti);
-    auto* type = SWIG_TypeQuery(type_name.c_str());
+  auto* type = SWIG_TypeQuery(type_name.c_str());
 
-    if (!type) {
-      throw std::runtime_error("Unable to find type in SWIG");
-    }
+  if (!type) {
+    throw std::runtime_error("Unable to find type in SWIG");
+  }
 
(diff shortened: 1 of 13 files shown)


kbenne:
I think we might get some new failures from ctest as a result of this. (which is a feature not a bug of course)

jmarrec:
Hummmm I no longer have any minitest output.

wenyikuang:
Eventually I need build the binary from scratch so its not a big difference.

jmarrec:
I'd say wait, I have some issues with it. Maybe it'll go in RC2.

ruby/engine/RubyEngine.cpp
```diff
@@ -75,7 +75,13 @@ void RubyEngine::setupEmbeddedGems(const std::vector<openstudio::path>& includeD
...
 
 RubyEngine::~RubyEngine() {
-  ruby_finalize();
+  // ruby_cleanup calls ruby_finalize
+  int ex = ruby_cleanup(0);
+  if (ex != 0) {
+    fmt::print("RubyEngine return code was {}\n", ex);
+    exit(ex);
+  }
+  //ruby_finalize();
```
jmarrec, line 84: @kbenne here is the most important change.

Originally I had fixed it by using an at_exit block in executeRubyScriptCommand. THe Kernel.exit! was sufficient to propagate it to C++. But the issue was that this would be returning zero : `openstudio -e "at_exit { exit 12 }"`. So I thought "Wouldn't it be more logical to install that at_exit in the RubyEngine itself?" and that's when I realized ruby has a ruby_cleanup which returns the final exit code, so we should use that

```ruby
at_exit {{
  exit_code = ($! and $!.kind_of? SystemExit) ? $!.status : 0
  STDOUT.flush
  STDERR.flush
  if exit_code != 0
    Kernel.exit!(exit_code)
  end
}}
```


https://github.com/ruby/ruby/blob/af471c0e0127eea0cafa6f308c0425bbfab0acf5/include/ruby/internal/interpreter.h#L160-L181

src/cli/CMakeLists.txt
```diff
@@ -476,6 +476,17 @@ if(BUILD_TESTING)
...
+  if (Python_EXECUTABLE)
+    add_test(NAME OpenStudioCLI.minitest_fail
+      COMMAND ${Python_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/test/run_minitest_fail.py" $<TARGET_FILE:openstudio>
+    )
+    if (SYSTEM_RUBY_EXECUTABLE)
+      add_test(NAME RubyTest.minitest_fail
+        COMMAND ${Python_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/test/run_minitest_fail.py" "${SYSTEM_RUBY_EXECUTABLE}"
+      )
+    endif()
+  endif()
```
jmarrec, line 488: new tests

ruby/engine/RubyEngine.cpp
```diff
@@ -75,7 +75,13 @@ void RubyEngine::setupEmbeddedGems(const std::vector<openstudio::path>& includeD
 }
 
 RubyEngine::~RubyEngine() {
-  ruby_finalize();
+  // ruby_cleanup calls ruby_finalize
+  int ex = ruby_cleanup(0);
+  if (ex != 0) {
+    fmt::print("RubyEngine return code was {}\n", ex);
```
shorowit, line 81: @jmarrec Did you mean to leave this in or was it for debugging? Seems weird that this is now printed.

PR #5200, opened 2024-05-15
Comment out debug print statements

shorowit:
Pull request overview
---------------------

Comment out debug print statements


src/cli/RunCommand.cpp
```diff
@@ -34,7 +34,7 @@ namespace cli {
     app->add_flag(
       "-m,--measures_only",
       [opt](std::int64_t val) {
-        fmt::print("val={}\n", val);
+        // fmt::print("val={}\n", val);
```
jmarrec, line 37: Totally ok here

ruby/engine/RubyEngine.cpp
```diff
@@ -78,7 +78,7 @@ RubyEngine::~RubyEngine() {
   // ruby_cleanup calls ruby_finalize
   int ex = ruby_cleanup(0);
   if (ex != 0) {
-    fmt::print("RubyEngine return code was {}\n", ex);
+    // fmt::print("RubyEngine return code was {}\n", ex);
```
jmarrec, line 81: I'm not sure I want this one gone...

shorowit: It doesn't seem like standard practice to print the exit code? After all, anyone can easily obtain the exit code themself.

jmarrec: The message was to explain why it returned a non zero exit code.

When running something like https://github.com/NREL/OpenStudio/blob/20d91f56c68b1ef37b1c4ee9b2d59cf6899503d2/src/cli/test/at_exit.rb#L1-L3

But sure, I suppose that's fine removing it

jmarrec (approved):

PR #5236, opened 2024-08-16
Add Utility for Generating Alfalfa Metadata to OpenStudio

kbenne:
Alfalfa is a platform which allows users to interact with OpenStudio/EnergyPlus models in real time. In order to accomplish this users must specify what components of the model to connect with Alfalfa as input and output points. This PR moves the mechanisms for creating "Alfalfa Points" into the OpenStudio runner so that they may be accessed during the measure workflow.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok


TShapinsky:
@kbenne what do you think about the language of `add` vs something like `export`.

Right now all the `AlfalfaJSON` methods are phrased as `add` like `addGlobalVariable`, `addMeter`, etc. I wonder if it might make more sense to replace `add` with `export` since what you're doing is exporting a `GlobalVariable`, `Meter`, `OutputVariable`, etc. to Alfalfa so it may be consumed there.

src/utilities/alfalfa/AlfalfaComponent.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace openstudio {
+namespace alfalfa {
+  enum ComponentCapabilities
+  {
+    Input = 1,
+    Output = 2
+  };
+
+  class AlfalfaComponent
```
TShapinsky, line 15: This component may be better as just a struct?

TShapinsky:
@kbenne this is pretty much there.
My TODO:
- [ ] Finish json serialization saving and connection with existing OS workflow.
- [ ] Audit if there are additional Model Objects of IDF objects that we should support automatic point creation for
- [ ] Add methods for exposing lists of variables from one object, as seen in `EnergyManagementSystem:GlobalVariable`

kbenne:
@TShapinsky as we discussed, it would be a good idea to test a workflow that excerses this new feature. You can see an example of workflow tests [here](https://github.com/NREL/OpenStudio/blob/ba1b94b6542c1f1d4cdcdb775886852b9a82bc4f/src/cli/CMakeLists.txt#L212). Note they are not Google Tests, but instead these are implemented purely in the CTest wrapper.

kbenne:
I created a new [issue](https://github.com/NREL/OpenStudio/issues/5250) to create documentation for this new API.

TShapinsky:
@kbenne I'm happy where everything is, with the exception of the behaviour for dealing with duplicate ids. One possible approach is to modify duplicate ids on export so they become unique. This would work, but it could be confusing and it is also non-deterministic. If someone was expecting a certain set of IDs from a measure they might not get them if a previous measure adds points of the same id. 

Another approach is to throw an error on export which could be annoying. It also runs the possibility of weird behaviour where based on naming of model objects certain models would error while others didn't. Are names unique across an IDF? or are they just unique to an idd type?

A possible third approach would be to refactor the id out of the point and have it be set in the AlfalfaJSON. This is probably the most correct as it is a key used to refer to the point and thus not something that the point needs to know about. This would allow for runtime checking of point Id against the existing points.

I think the best option may be a combination of the first and third. Refactoring the point id to be stored in the AlfalfaJSON object and dynamically renaming point ids to prevent overlap. A warning could also be printed to alert of this happening. But that would only be shown locally, so it wouldn't include weird behaviour that happens when the measures interact with the baked in alfalfa measures.

Maybe after this we need to reevaluate if we want to be automatically baking points into alfalfa or if we should just provide the measures we create for generating generic points as a resource for people putting together workflows.

kbenne:
@wenyikuang can you investigate the CI failures?

wenyikuang:
Looks the windows-incremental is lost connection from the jenkins, let me look

TShapinsky:
If i remember correctly, the osx build fails were from tests failing for BCL measures.

python/engine/test/AlfalfaMeasure/measure.py
```diff
@@ -0,0 +1,39 @@
...
+    ):
+        """
+        define what happens when the measure is run
+        """
+
+        alfalfa = runner.alfalfa()
+        alfalfa.exposeConstant(10, "safe value")
+        alfalfa.exposeMeter("Facility:Electricity", "Facility Electricity")
+        alfalfa.exposeActuator("somehting", "another thing", "key", "example actuator")
+        alfalfa.exposeOutputVariable("Whole Building", "Facility Total Purchased Electricity Energy", "output variable")
```
kbenne, line 33: Is this a change? I seem to recall that these originally took ModelObject instances. I think when we @TShapinsky last spoke you were contemplating a change.

kbenne: A risk I can think of is that the final names might not be known at ModelMeasure time.

resources/Examples/compact_osw/measures/AlfalfaEPlusPython/measure.py
```diff
@@ -0,0 +1,97 @@
...
+        if not (runner.validateUserArguments(self.arguments(workspace), user_arguments)):
+            return False
+
+        alfalfa = runner.alfalfa()
+
+        # Test Meters
+        alfalfa.exposeMeter("Electricity:Facility", "Electricity Meter String:EPlus:Python")
+
+        meter_object = openstudio.IdfObject.load("Output:Meter, Electricity:Facility;").get()
+        alfalfa.exposeFromObject(meter_object, "Electricity Meter IDF:Eplus:Python")
```
kbenne, line 70: Does this method work with a ModelObject ?

resources/Examples/compact_osw/measures/AlfalfaModelPython/measure.py
```diff
@@ -0,0 +1,104 @@
...
+
+        alfalfa = runner.alfalfa()
+
+        # Test Meters
+        alfalfa.exposeMeter("Electricity:Facility", "Electricity Meter String:Model:Python")
+
+        meter_object = openstudio.model.OutputMeter(model)
+        meter_object.setFuelType(openstudio.FuelType("Electricity"))
+        meter_object.setInstallLocationType(openstudio.InstallLocationType("Facility"))
+        alfalfa.exposeFromObject(meter_object, "Electricity Meter OSM:Model:Python")
```
kbenne, line 74: I see. Questions answered.

src/utilities/CMakeLists.txt
```diff
@@ -125,6 +128,24 @@ set(filetypes_src
   filetypes/WorkflowStepResult.cpp
 )
 
+set(alfalfa_src
```
kbenne, line 131: Perhaps src/utilities/alfalfa should come up a level as utilities is taking on a lot. Perhaps src/alfalfa?

wenyikuang:
Resolve the issue of connection between jenkins and windows-incre-on-aws

https://ci.openstudio.net/blue/organizations/jenkins/openstudio-incremental-windows/detail/PR-5255/7/pipeline

(From another PR but using the same node)

Or are you mentioning these one?


```
The following tests FAILED:

	665 - BCLFixture.RemoteBCL_BCLSearchResult (Failed)

	1877 - ModelFixture.PythonPluginInstance_NotPYFile (Failed)

	2304 - ModelFixture.ScheduleFile (Failed)

	3771 - OpenStudioCLI.Check_Alfalfa (Failed)
```

src/utilities/CMakeLists.txt
TShapinsky: That's what i originally had, but then i got humble. I'm happy to move it back up to that level again. It would also make some of the ruby bindings a bit easier.

src/utilities/alfalfa/AlfalfaComponent.hpp
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  inline Capability operator&(const Capability& a, const Capability& b) {
+    return static_cast<Capability>(static_cast<int>(a) & static_cast<int>(b));
+  }
+
+  class UTILITIES_API AlfalfaComponent
+  {
+   public:
+    AlfalfaComponent(const std::string& type, const Capability capabilities);
```
jmarrec, line 27: Should this be public? I'm getting hints of "this is just a base class and you're supposed to only instantiate the derived ones"

src/utilities/alfalfa/AlfalfaComponent.hpp
```diff
@@ -0,0 +1,59 @@
...
+     */
+    bool canOutput() const;
+
+    bool operator==(const AlfalfaComponent& rhs) const;
+
+    /**
+     * A string which reflects the type of the component
+     */
+    std::string type;
+
```
jmarrec, line 45: Enum class instead of string? Seems like this is a limited number of valid things.

src/utilities/alfalfa/AlfalfaGlobalVariable.cpp
```diff
@@ -0,0 +1,42 @@
...
+  AlfalfaGlobalVariable::AlfalfaGlobalVariable(const IdfObject& global_variable)
+    : AlfalfaComponent("GlobalVariable", Capability::Output | Capability::Input) {
+    IddObjectType idd_type = global_variable.iddObject().type();
+    boost::optional<std::string> variable_name;
+
+    if (idd_type == IddObjectType::OS_EnergyManagementSystem_GlobalVariable) {
+      variable_name = global_variable.getString(OS_EnergyManagementSystem_GlobalVariableFields::Name);
+    } else if (idd_type == IddObjectType::EnergyManagementSystem_GlobalVariable) {
+      variable_name = global_variable.getString(EnergyManagementSystem_GlobalVariableExtensibleFields::ErlVariableName);
+    } else {
```
jmarrec, line 31: K so second time I see this pattern so it's a thing. We have one Ctor that takes IdfObject and tests if this is a ModelObject type or an actual IdfObject/Workspace object.

This is a couple of types, so it's probably fine... and since you're living inside utilities/ you don't have access to Model namespace really.

If you move up one folder as @kbenne asked about, then I'd probably specialize the ctor for `modelObject` and have it call the getters instead of getString

src/utilities/alfalfa/AlfalfaPoint.hpp
```diff
@@ -0,0 +1,111 @@
...
+namespace alfalfa {
+  namespace detail {
+    class AlfalfaPoint_Impl;
+  }
+  class UTILITIES_API AlfalfaPoint
+  {
+   public:
+    AlfalfaPoint();
+
+    AlfalfaPoint(const std::string& display_name);
```
jmarrec, line 28: I'm noticing this is the only class that has is under the pImpl idiom in that folder. I'm not sure what an Alfalfa point is and why a pImpl makes sense yet.

src/utilities/alfalfa/AlfalfaPoint_Impl.hpp
```diff
@@ -0,0 +1,30 @@
...
+  class AlfalfaPoint;
+  namespace detail {
+    class UTILITIES_API AlfalfaPoint_Impl
+    {
+     public:
+      AlfalfaPoint_Impl();
+
+     protected:
+      friend class openstudio::alfalfa::AlfalfaPoint;
+      boost::optional<AlfalfaComponent> m_input, m_output;
```
jmarrec, line 21: https://clang.llvm.org/extra/clang-tidy/checks/readability/isolate-declaration.html

src/utilities/alfalfa/AlfalfaPoint_Impl.hpp
```diff
@@ -0,0 +1,30 @@
...
+      std::string m_display_name;
+      boost::optional<std::string> m_units, m_id;
```
jmarrec, line 23: same, break it into two declarations.

src/utilities/alfalfa/AlfalfaPoint.cpp
```diff
@@ -0,0 +1,137 @@
...
+    return boost::regex_match(id, boost::regex("^[A-Za-z0-9_\\-\\[\\]:()]*$")) && id.size() > 0;
+  }
+
+  std::string AlfalfaPoint::toIdString(const std::string& str) const {
+    return boost::regex_replace(str, boost::regex(" "), "_");
+  }
+
+  bool AlfalfaPoint::operator==(const AlfalfaPoint& rhs) const {
+    return toJSON() == rhs.toJSON();
+  }
```
jmarrec, line 135: I'm a bit puzzled by this. Is the operator== needed? for what?

And should it be based on m_impl instead of being by json value?

src/utilities/alfalfa/AlfalfaPoint.cpp
```diff
@@ -0,0 +1,137 @@
...
+      point["units"] = units().get();
+    }
+
+    point["optional"] = optional();
+
+    return point;
+  }
+
+  bool AlfalfaPoint::isValidId(const std::string& id) const {
+    return boost::regex_match(id, boost::regex("^[A-Za-z0-9_\\-\\[\\]:()]*$")) && id.size() > 0;
```
jmarrec, line 126: Just an FYI but using a raw string litteral you can avoid the double slashing which improves readability a lot IMHO.

```suggestion
    return boost::regex_match(id, boost::regex(R"(^[A-Za-z0-9_\-\[\]:()]*$)")) && id.size() > 0;

```

jmarrec: Also worth noting that boost::regex is truly awful in terms of performance (like std::regex), so if you expect this to be used a lot, make it a static regex.

src/utilities/alfalfa/AlfalfaPoint.cpp
```diff
@@ -0,0 +1,137 @@
+#include "AlfalfaPoint.hpp"
+#include "AlfalfaPoint_Impl.hpp"
+
+#include <boost/regex.hpp>
+#include <fmt/format.h>
+
+namespace openstudio {
+namespace alfalfa {
+  namespace detail {
+    AlfalfaPoint_Impl::AlfalfaPoint_Impl() {}
```
jmarrec, line 10: `= default`

https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-equals-default.html

src/utilities/alfalfa/AlfalfaPoint.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+namespace openstudio {
+namespace alfalfa {
+  namespace detail {
+    class AlfalfaPoint_Impl;
+  }
+  class UTILITIES_API AlfalfaPoint
+  {
+   public:
+    AlfalfaPoint();
```
jmarrec, line 26: why is this one here? For convenience in vector or a swig thing?
or is displayName meant to be optional?

src/utilities/alfalfa/AlfalfaPoint.hpp
jmarrec, line 28: should this ctor be explicit?

src/utilities/alfalfa/AlfalfaPoint.cpp
```diff
@@ -0,0 +1,137 @@
...
+    point["optional"] = optional();
+
+    return point;
+  }
+
+  bool AlfalfaPoint::isValidId(const std::string& id) const {
+    return boost::regex_match(id, boost::regex("^[A-Za-z0-9_\\-\\[\\]:()]*$")) && id.size() > 0;
+  }
+
+  std::string AlfalfaPoint::toIdString(const std::string& str) const {
```
jmarrec, line 129: This method and `isValidId` should either take no param, or they should be static methods...

As is, they don't use the state

src/utilities/alfalfa/AlfalfaPoint.hpp
```diff
@@ -0,0 +1,111 @@
...
+    bool operator==(const AlfalfaPoint& rhs) const;
+
+   private:
+    std::shared_ptr<detail::AlfalfaPoint_Impl> m_impl;
+
+    std::string toIdString(const std::string& str) const;
+
+    bool isValidId(const std::string& id) const;
+
+    inline static const std::string ID_VALID_CHARS_MSG = "IDs can only contain letters, numbers, and the following special characters _-[]():";
```
jmarrec, line 103: how bout static constexpr std::string_view here? Can be defined directly in the cpp file.

src/utilities/alfalfa/AlfalfaPoint.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  void AlfalfaPoint::setOptional(bool optional) {
+    m_impl->m_optional = optional;
+  }
+
+  Json::Value AlfalfaPoint::toJSON() const {
+    Json::Value point;
+
+    if (id().is_initialized()) {
+      point["id"] = id().get();
```
jmarrec, line 101: Two calls to id() here. Same comment applies to all other occurences of this pattern below.

```suggestion
    if (auto id_ = id().is_initialized()) {
      point["id"] = *id_;
```

notes:
* you can also `std::move(*id_)` if relevant
* .get() will internally check if is_initialized to handle failure, which you're already doing. Dereferencing the optional does not so it's faster.

src/utilities/alfalfa/AlfalfaPoint.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * Set display name of point.
+     */
+    void setDisplayName(const std::string& display_name);
+
+    /**
+     * Get whether a point is marked as optional or not.
+     */
+    bool optional() const;
```
jmarrec, line 85: I find the naming confusing, but maybe that's just me. I'd rather have `isOptional`

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+#include <iostream>
+namespace openstudio {
+namespace alfalfa {
+  namespace detail {
+    AlfalfaJSON_Impl::AlfalfaJSON_Impl() {}
+
+    AlfalfaJSON_Impl::AlfalfaJSON_Impl(const std::string& s) {}
+
+    AlfalfaJSON_Impl::AlfalfaJSON_Impl(const openstudio::path& p) {}
```
jmarrec, line 32: that **cannot** be right. You're not using the string/path.

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -0,0 +1,223 @@
...
+      return m_points;
+    }
+
+    bool AlfalfaJSON_Impl::saveAs(const openstudio::path& path) {
+      m_JSONPath = path;
+      return save();
+    }
+
+    bool AlfalfaJSON_Impl::save() const {
+      if (makeParentFolder(m_JSONPath)) {
```
jmarrec, line 48: m_JSONPath could be an empty path here.

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -0,0 +1,223 @@
...
+      }
+      LOG(Error, "Unable to write file to path '" << toString(m_JSONPath) << "', because parent directory "
+                                                  << "could not be created.");
+
+      return false;
+    }
+
+    Json::Value AlfalfaJSON_Impl::toJSON() const {
+      Json::Value root;
+      for (AlfalfaPoint point : m_points) {
```
jmarrec, line 71: Making a copy of each object, probably not on purpose. Yes, this is a pImpl class, but still, there's no point not taking a `const AlfalfaPoint& point` here

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+  boost::optional<AlfalfaPoint> AlfalfaJSON::exposeConstant(float value, const std::string& display_name) {
+    AlfalfaConstant component(value);
+    return exposeFromComponent(component, display_name);
+  }
+
+  boost::optional<AlfalfaPoint> AlfalfaJSON::exposeMeter(const std::string& meter_name, const std::string& display_name) {
+    AlfalfaMeter component(meter_name);
+    return exposeFromComponent(component, display_name);
+  }
```
jmarrec, line 93: Ok, so if I follow correctly, really the only public class is this AlfalfaJSON class, which exposes a bunch of methods that will:
* Instantiate the derived class of AlfalfaComponent
* Call its own exposeFromComponent with that derived comp..

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -0,0 +1,223 @@
...
+    try {
+      switch (idd_type.value()) {
+        case IddObjectType::OS_Output_Meter:
+        case IddObjectType::Output_Meter:
+          component = new AlfalfaMeter(idf_object);
+          break;
+        case IddObjectType::OS_EnergyManagementSystem_Actuator:
+        case IddObjectType::EnergyManagementSystem_Actuator:
+          component = new AlfalfaActuator(idf_object);
+          break;
```
jmarrec, line 137: I'm noting that you're comparing the iddObjectType once here, then a second time in the Derived AlfalfaComponent's ctor.

While I'm sure this can work, I'm just going to say I dislike it because:
* Doing work twice, but that is probably plently fine, as the slowdown isn't likely to be noticeable
* Mostly because that duplicates the code. Say AlfalfaMeter tomorrow needs to support Output:Meter:CustomDecrement, we have to go modify two files.

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -0,0 +1,223 @@
...
+      } else if (component.type == "GlobalVariable") {
+        _display_name = "Global Variable for " + component.parameters["variable_name"].asString();
+      } else if (component.type == "OutputVariable") {
+        _display_name =
+          "Output Variable for " + component.parameters["variable_key"].asString() + ":" + component.parameters["variable_name"].asString();
+      } else {
+        LOG(Error, "Invalid component type");
+        return boost::none;
+      }
+    }
```
jmarrec, line 179: Kinda similar. I guess I'm questionning whether this logic should instead be inside the Derived class of AlfalfaComponent.

AlfalfaComponent would have a `virtual Json::Value exposeToAlfalfa() const = 0` method or whatever the name (maybe just toJSON? or alfalfaParameters?), and each derived class would implement the business logic where it's most logical to place it? Making it a pure virtual method would make AlfalfaComponent an abstract class, which I guess is what it should be.

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -0,0 +1,223 @@
...
+    AlfalfaPoint point(_display_name);
+    if (component.canInput()) {
+      point.setInput(component);
+    }
+    if (component.canOutput()) {
+      point.setOutput(component);
+    }
+    exposePoint(point);
+    boost::optional<AlfalfaPoint> result = point;
+    return result;
```
jmarrec, line 190: just return point, or std::move it to the optional.

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -0,0 +1,223 @@
...
+  boost::optional<AlfalfaPoint> AlfalfaJSON::exposeGlobalVariable(const std::string& variable_name, const std::string& display_name) {
+    AlfalfaGlobalVariable component(variable_name);
+    return exposeFromComponent(component, display_name);
+  }
+
+  void AlfalfaJSON::exposePoint(const AlfalfaPoint& point) {
+    m_impl->exposePoint(point);
+  }
+
+  boost::optional<AlfalfaPoint> AlfalfaJSON::exposeFromObject(const openstudio::IdfObject& idf_object, const std::string& display_name) {
```
jmarrec, line 116: THe vast majority of the code I've read was "throwy", but here you go with LOG and return boost::none. That's fine, just noting.

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+  boost::optional<AlfalfaPoint> AlfalfaJSON::exposeFromComponent(const AlfalfaComponent& component, const std::string& display_name) {
+    std::string _display_name = display_name;
+    if (display_name.size() == 0) {
+      if (component.type == "Actuator") {
+        _display_name = "Actuator for " + component.parameters["component_name"].asString() + ":" + component.parameters["component_type"].asString()
+                        + ":" + component.parameters["control_type"].asString();
+      } else if (component.type == "Constant") {
+        LOG(Error, "Constant points must be provided with a display name");
+        return boost::none;
```
jmarrec, line 167: There's something bothering me with error handling here, or maybe public-ness of the exposeFromComponent method.

Why ever allow a component.type Constant to NOT have a displayName to begin with?!

src/utilities/filetypes/AlfalfaJSON.hpp
```diff
@@ -0,0 +1,114 @@
...
+     * Expose a point in the alfalfa API from an AlfalfaComponent.
+     * display_name is optional and by default will be infered from the contents of the component.
+     * Input and Ouput of the point will reflect the capabilities of the component.
+     */
+    boost::optional<AlfalfaPoint> exposeFromComponent(const AlfalfaComponent& component, const std::string& display_name = std::string());
+
+    /**
+     * Get a vector of all points currently exported to the Alfalfa API
+     */
+    std::vector<AlfalfaPoint> getPoints();
```
jmarrec, line 100: I think `points()` better matches our general convention.

src/workflow/RunPostProcess.cpp
```diff
@@ -18,6 +18,13 @@ void OSWorkflow::runPostProcess() {
     openstudio::workflow::util::gatherReports(workflowJSON.absoluteRunDir(), workflowJSON.absoluteRootDir());
     LOG(Info, "Finished gathering reports");
   }
+  // If no points have been exported, skip file creation
+  if (runner.alfalfa().getPoints().size() > 0) {
```
jmarrec, line 22: std::vector::empty()

jmarrec (changes requested):
This is good work. I'm just generally picky.

There are a couple of constructs I'd change, I'd also suggest taking your files for a spin with clang-tidy to modernize the code a bit and follow best practices.

My main point is more around the articulation of the classes between themselves, and what public API should be like (probably use actual type values to store stuff and not JSON, provide getters, some methods should probably be private, and error handling should be reworked a bit).

python/engine/test/PythonEngine_GTest.cpp
```diff
@@ -134,3 +135,38 @@ RecursionError: maximum recursion depth exceeded)",
...
+    \"steps\": [\
+      {\
+        \"arguments\": {},\
+        \"description\": \"The method attempts to build an alfalfa json in the measure\",\
+        \"measure_dir_name\": \"AlfalfaMeasure\",\
+        \"modeler_description\": \"The method attempts to build an alfalfa json in the measure\",\
+        \"name\": \"AlfalfaMeasure\"\
+      }\
+    ]\
+  }";
```
jmarrec, line 161: Use a Raw string litteral here. I'm also seeing extra empty lines, so you should use clang-format. (**Why did our CI not catch this?** the python/ folder is outside... geez)


```suggestion
  std::string workflow_json = R"json(
  {
    "seed": "../seed.osm",
    "weather_file": "../weather.epw",
    "steps": [
      {
        "arguments": {},
        "description": "The method attempts to build an alfalfa json in the measure",
        "measure_dir_name": "AlfalfaMeasure",
        "modeler_description": "The method attempts to build an alfalfa json in the measure",
        "name": "AlfalfaMeasure"
      }
    ]
  })json";
```

python/engine/test/PythonEngine_GTest.cpp
```diff
@@ -134,3 +135,38 @@ RecursionError: maximum recursion depth exceeded)",
...
+
+
+
+  openstudio::model::Model model;
+  openstudio::WorkflowJSON workflow = *openstudio::WorkflowJSON::load(workflow_json);
+  openstudio::measure::OSRunner runner(workflow);
+
+  openstudio::measure::OSArgumentMap arguments;
+  measurePtr->run(model, runner, arguments);
```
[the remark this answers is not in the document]
jmarrec: Check the JSON is written at least.

jmarrec:
@wenyikuang cppcheck is broken, you should fix that: https://github.com/NREL/OpenStudio/actions/runs/10853524579/job/30432985755#step:3:16

python/engine/test/PythonEngine_GTest.cpp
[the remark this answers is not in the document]
jmarrec: Check the runner's alfalfa has points at least.

src/utilities/filetypes/AlfalfaJSON.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include <vector>
+#include <json/json.h>
+
+#include "../UtilitiesAPI.hpp"
+#include "../alfalfa/AlfalfaPoint.hpp"
+#include "../core/Path.hpp"
+#include "../core/Logger.hpp"
+#include "../core/Optional.hpp"
+
+#include "model/ModelObject.hpp"
```
jmarrec, line 18: ModelObject unused

jmarrec: but IdfObject is

src/utilities/alfalfa/AlfalfaComponent.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include "../UtilitiesAPI.hpp"
+
+namespace openstudio {
+namespace alfalfa {
+  enum Capability
+  {
+    Input = 1,
+    Output = 2
+  };
```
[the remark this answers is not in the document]
jmarrec: Interesting, I have a warning saying

```
  bool AlfalfaComponent::canInput() const {
    return capabilities & Capability::Input;
  }
```

> Conversion of 'Capability' into 'bool' will always return 'true', enum does not have a zero-value enumerator


I guess *I am* one of the devs that aren't too familiar with the bitwise operations on enums...

I had to create a compiler-explorer testbed to ensure this is working as intended... https://godbolt.org/z/YMMasa1Pv

jmarrec:
* Made a few cosmetic changes in #5259

src/utilities/alfalfa/AlfalfaComponent.hpp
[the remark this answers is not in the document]
TShapinsky: Weird that it's giving that warning. In the base alfalfa source I just have 3 enums: `INPUT`, `OUTPUT`, and `BIDIRECTIONAL`. I think I just enjoy bitwise operations a little too much due to embedded C. I'll switch to the more natural enum expression.

src/utilities/alfalfa/AlfalfaComponent.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * The raw input/output capabilities of the component
+     */
+    Capability capabilities;
+
+    /**
+     * JSON encoded parameters of the component
+     */
+    Json::Value parameters;
```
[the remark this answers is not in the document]
TShapinsky: yeah, this is the part I'm least comfortable with. I want the fields of the of each of the independent children of `AlfalfaComponent` to be accessible from an instance of the superclass as there are cases where a method may just return an `AlfalfaComponent` type when it can be any of the children. It seems like the pattern used in OS is the pImpl, but I though that was a bit complex for something so simple.

src/utilities/alfalfa/AlfalfaComponent.hpp
TShapinsky: That is a good question, it could just as easily be protected without requiring any change, right?

src/utilities/alfalfa/AlfalfaComponent.hpp
TShapinsky: Yeah, I also didn't like this. Usually it would all be type inference so you wouldn't have to maintain in multiple places. Depending on where we decide to go with the component it could go one of a couple different ways.

src/utilities/alfalfa/AlfalfaGlobalVariable.cpp
TShapinsky: I'm happy to move it up a namespace, I just didn't want to be too entitled. I actually started with it on the top level and then moved it here later.

src/utilities/alfalfa/AlfalfaPoint.cpp
TShapinsky: I guess my thought was that the only thing that matters at the end of the day is how it is serialized, so comparing the serialization is enough to get if they are functionally the same, and it makes it so someone can't add an extra field to the class for serialization but forget to include it in the equality operator.

src/utilities/alfalfa/AlfalfaPoint.hpp
TShapinsky: Sounds good, I prefer that pattern, I was just trying to fit within the current paradigms of setters and getters in OS.

src/utilities/filetypes/AlfalfaJSON.cpp
TShapinsky: Yeah, this is a current debate I'm having on whether or not you should be able to create this object from JSON. It is not required by OpenStudio, but it might be nice to have for use of the file outside the workflow.

Either way I wanted to get the rest of the API figured out before writing a load of parsers.

src/utilities/filetypes/AlfalfaJSON.cpp
TShapinsky: That is the basic pattern, users will mostly be interacting with these helper methods, however, I want to keep the rest of the API available for advanced users to do stuff.

src/utilities/filetypes/AlfalfaJSON.cpp
TShapinsky: Yeah, this is the tension I always run up against. I considered making different constructors for the different model object types, but you still need to do idd_type checking to prevent bad arguments since idfObjects don't have separate c++ types. You could completely remove this function, but given that this method will probably only run 100ish times per workflow, I doubt the performance impact will be noticeable at all.

src/utilities/filetypes/AlfalfaJSON.cpp
TShapinsky: If you look through the git commit history I originally had everything going off a base class with fully virtual functions. The problem was I couldn't access those in functions that just took a reference to the base class so any child type could be passed in.

src/utilities/filetypes/AlfalfaJSON.cpp
TShapinsky: The general pattern was I log and return `boost::none` everywhere possible, but in ctors you can't do that (to my knowledge) to I throw a `std::runtime_error`.

src/utilities/filetypes/AlfalfaJSON.cpp
TShapinsky: It was mostly in the interest of making the base programming interface as simple as possible for users to get up and running with. Maybe I went too far.

resources/Examples/compact_osw/measures/AlfalfaEPlusPython/measure.py
TShapinsky: yes

src/utilities/CMakeLists.txt
TShapinsky: done

python/engine/test/AlfalfaMeasure/measure.py
```diff
@@ -0,0 +1,39 @@
...
+        alfalfa.exposeGlobalVariable("global_1", "global variable")
+        super().run(model, runner, user_arguments)  # Do **NOT** remove this line
```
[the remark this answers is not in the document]
TShapinsky: done

src/utilities/filetypes/AlfalfaJSON.cpp
TShapinsky: this has now been improved

src/utilities/filetypes/AlfalfaJSON.hpp
TShapinsky: done

src/utilities/alfalfa/AlfalfaComponent.hpp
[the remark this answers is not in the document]
TShapinsky: this has now been migrated to an `OPENSTUDIO_ENUM`

TShapinsky:
@jmarrec I'm starting to get into the cleaning phase. I think it would be useful if you could go over the Component code (ComponentBase, AlfalfaComponent, AflalfaActuator, AlfalfaConstant, AlfalfaMeter, AlfalfaOutputVariable, AlfalfaGlobalVariable) and let me know what you think. I'm decently happy with the new architecture, but it isn't exactly something that is elsewhere in the code base so I'd like your opinion.

python/module/openstudio.py
```diff
@@ -28,6 +28,7 @@
     from . import openstudioosversion as osversion
     from . import openstudioradiance as radiance
     from . import openstudiosdd as sdd
+    from . import openstudioualfalfa as alfalfa
```
jmarrec, line 31: what's that extra "u" in there? I think you meant `openstudioalfalfa` @TShapinsky

src/alfalfa/AlfalfaComponent.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include "AlfalfaAPI.hpp"
+#include "ComponentBase.hpp"
+
+namespace openstudio {
+namespace alfalfa {
+  class ALFALFA_API AlfalfaComponent
+  {
+   public:
+    template <typename T, std::enable_if_t<std::is_base_of<ComponentBase, T>::value, bool> = true>
+    AlfalfaComponent(T component) : m_component(std::make_unique<T>(std::move(component))) {}
```
jmarrec, line 16: seeing this while looking diagonally at this PR changes since I was out of office. Makes me wonder if I should investigate what's up or not. @kbenne I'm assuming you reviewed the changes already while I wasn't around and I don't need to?

jmarrec: (I'm failing to see directly why this AlfalfaComponent class is needed. Instead of having ComponentBase (renamed to AlfalfaComponent probably, but basically an abstract class with the pure virtual methods) and just be done with it)

src/alfalfa/AlfalfaJSON_Impl.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+      bool saveAs(const openstudio::path& p);
+
+      void setJSONPath(const openstudio::path& p);
+
+      Json::Value toJSON() const;
+
+      void exposePoint(const AlfalfaPoint& point);
+
+      std::vector<AlfalfaPoint> points();
```
jmarrec, line 33: should be const. cppcheck catches this (now that I fixed the workflow on a separate PR...)

src/alfalfa/AlfalfaComponent.hpp
TShapinsky: Ahh, the biggest issue appears to be that you can't `boost::optional` an abstract base class. I can obviously use a `unique_ptr` but I don't think that translates to swig very well...

kbenne: @jmarrec It wouldn't hurt to have your touch on this for one more round. If you and @TShapinsky agree to a simplification then I'm all for that.

PR #5259, opened 2024-09-20
[Developer] Adjust alfalfa_utility

jmarrec:
Pull request overview
---------------------

- Adjustments for #5236

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/python/engine/test/PythonEngine_GTest.cpp
+++ b/python/engine/test/PythonEngine_GTest.cpp
@@ -145,28 +145,28 @@ TEST_F(PythonEngineFixture, AlfalfaMeasure) {
 
   ASSERT_EQ(measurePtr->name(), "Alfalfa Measure");
 
-  std::string workflow_json = "\
-  {\
-    \"seed\": \"../seed.osm\",\
-    \"weather_file\": \"../weather.epw\",\
-    \"steps\": [\
-      {\
-        \"arguments\": {},\
-        \"description\": \"The method attempts to build an alfalfa json in the measure\",\
-        \"measure_dir_name\": \"AlfalfaMeasure\",\
-        \"modeler_description\": \"The method attempts to build an alfalfa json in the measure\",\
-        \"name\": \"AlfalfaMeasure\"\
-      }\
-    ]\
-  }";
-
-
+  std::string workflow_json = R"json(
+{
+  "seed": "../seed.osm",
+  "weather_file": "../weather.epw",
+  "steps": [
+    {
+      "arguments": {},
+      "description": "The method attempts to build an alfalfa json in the measure",
+      "measure_dir_name": "AlfalfaMeasure",
+      "modeler_description": "The method attempts to build an alfalfa json in the measure",
+      "name": "AlfalfaMeasure"
+    }
+  ]
+}
+  )json";
 
   openstudio::model::Model model;
   openstudio::WorkflowJSON workflow = *openstudio::WorkflowJSON::load(workflow_json);
   openstudio::measure::OSRunner runner(workflow);
+  EXPECT_TRUE(runner.alfalfa().getPoints().empty());
 
   openstudio::measure::OSArgumentMap arguments;
   measurePtr->run(model, runner, arguments);
-
+  EXPECT_EQ(5, runner.alfalfa().getPoints().size());
 }
--- a/src/utilities/alfalfa/AlfalfaActuator.cpp
+++ b/src/utilities/alfalfa/AlfalfaActuator.cpp
@@ -1,6 +1,6 @@
 #include "AlfalfaActuator.hpp"
-#include "utilities/idd/IddObject.hpp"
-#include "utilities/idd/IddEnums.hpp"
+#include "../idd/IddObject.hpp"
+#include "../idd/IddEnums.hpp"
 
 #include <utilities/idd/OS_EnergyManagementSystem_Actuator_FieldEnums.hxx>
 #include <utilities/idd/EnergyManagementSystem_Actuator_FieldEnums.hxx>
@@ -12,11 +12,11 @@ namespace openstudio {
 namespace alfalfa {
   AlfalfaActuator::AlfalfaActuator(const std::string& component_name, const std::string& component_type, const std::string& control_type)
     : AlfalfaComponent("Actuator", Capability::Input | Capability::Output) {
-    if (component_name.size() == 0) {
+    if (component_name.empty()) {
       throw std::runtime_error("Error creating AlfalfaActuator: component_name must be non-empty");
-    } else if (component_type.size() == 0) {
+    } else if (component_type.empty()) {
       throw std::runtime_error("Error creating AlfalfaActuator: component_type must be non-empty");
-    } else if (control_type.size() == 0) {
+    } else if (control_type.empty()) {
       throw std::runtime_error("Error creating AlfalfaActuator: control_type must be non-empty");
     }
     parameters["component_name"] = component_name;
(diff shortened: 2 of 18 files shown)


src/utilities/alfalfa/AlfalfaActuator.hpp
```diff
@@ -2,8 +2,7 @@
 #define ALFALFA_COMPONENT_ACTUATOR_HPP
 
 #include "AlfalfaComponent.hpp"
-#include "utilities/idf/IdfObject.hpp"
-#include "utilities/core/Logger.hpp"
```
jmarrec, line 6: Logger was included everywhere but not used. Because you throw std::runtime_error.

Maybe you should use a proper Logger and use LOG_AND_THROW, one of the two.

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -21,15 +23,20 @@
...
+
 namespace openstudio {
 namespace alfalfa {
   namespace detail {
-    AlfalfaJSON_Impl::AlfalfaJSON_Impl() {}
+    AlfalfaJSON_Impl::AlfalfaJSON_Impl() = default;
 
-    AlfalfaJSON_Impl::AlfalfaJSON_Impl(const std::string& s) {}
+    AlfalfaJSON_Impl::AlfalfaJSON_Impl(const std::string& s) {
+      // TODO: use s
```
jmarrec, line 34: TODO

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -21,15 +23,20 @@
...
+    }
 
-    AlfalfaJSON_Impl::AlfalfaJSON_Impl(const openstudio::path& p) {}
+    AlfalfaJSON_Impl::AlfalfaJSON_Impl(const openstudio::path& p) {
+      // TODO: use p
```
jmarrec, line 38: TODO

src/utilities/alfalfa/AlfalfaPoint_Impl.hpp
```diff
@@ -17,10 +17,13 @@ namespace alfalfa {
       AlfalfaPoint_Impl();
 
      protected:
+      // TODO: is this pImpl is needed? If so, add getters and setters and make all members private.
```
jmarrec, line 20: TODO

src/utilities/filetypes/AlfalfaJSON.hpp
```diff
@@ -100,7 +99,7 @@ namespace alfalfa {
     std::vector<AlfalfaPoint> getPoints();
 
    private:
-    boost::optional<std::string> getName(const openstudio::IdfObject& idf_object);
+    static boost::optional<std::string> getName(const openstudio::IdfObject& idf_object);
```
jmarrec, line 102: static a few methods

src/utilities/filetypes/AlfalfaJSON.cpp
```diff
@@ -68,45 +75,40 @@ namespace alfalfa {
...
-  AlfalfaJSON::AlfalfaJSON(const std::string& s) : m_impl(std::shared_ptr<detail::AlfalfaJSON_Impl>(new detail::AlfalfaJSON_Impl(s))) {}
+  AlfalfaJSON::AlfalfaJSON(const std::string& s) : m_impl(std::make_shared<detail::AlfalfaJSON_Impl>(s)) {}
 
-  AlfalfaJSON::AlfalfaJSON(const openstudio::path& p) : m_impl(std::shared_ptr<detail::AlfalfaJSON_Impl>(new detail::AlfalfaJSON_Impl(p))) {}
+  AlfalfaJSON::AlfalfaJSON(const openstudio::path& p) : m_impl(std::make_shared<detail::AlfalfaJSON_Impl>(p)) {}
 
   boost::optional<AlfalfaPoint> AlfalfaJSON::exposeConstant(float value, const std::string& display_name) {
-    AlfalfaConstant component(value);
-    return exposeFromComponent(component, display_name);
+    return exposeFromComponent(AlfalfaConstant{value}, display_name);
```
TShapinsky, line 93: I'm not familiar with this method of constructing, this is just a way to construct the class inline in the stack?

TShapinsky (approved):
Thanks for this @jmarrec just a couple questions if you have time.

python/engine/test/PythonEngine_GTest.cpp
```diff
@@ -145,28 +145,28 @@ TEST_F(PythonEngineFixture, AlfalfaMeasure) {
...
+    {
+      "arguments": {},
+      "description": "The method attempts to build an alfalfa json in the measure",
+      "measure_dir_name": "AlfalfaMeasure",
+      "modeler_description": "The method attempts to build an alfalfa json in the measure",
+      "name": "AlfalfaMeasure"
+    }
+  ]
+}
+  )json";
```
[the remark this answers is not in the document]
jmarrec: Nothing, it's an optional arbitrary string, at most 16 characters long. I like to use it, when this is a json thing I put json, when it's a sql query I put sql... etc.

(technically speaking you could configure your editor to pick that up and apply a different syntax highlighting based on it)

https://en.cppreference.com/w/cpp/language/string_literal

```
 R"d-char-seq ﻿(optional)(r-char-seq ﻿(optional))d-char-seq ﻿(optional)"
```

> d-char-seq 	- 	A sequence of one or more d-char ﻿s, at most 16 characters long
> d-char 	- 	A character from the [basic character set](https://en.cppreference.com/w/cpp/language/charset#Basic_character_set), except parentheses, backslash and [spaces](https://en.cppreference.com/w/cpp/string/byte/isspace)

src/utilities/alfalfa/AlfalfaActuator.cpp
```diff
@@ -1,6 +1,6 @@
 #include "AlfalfaActuator.hpp"
-#include "utilities/idd/IddObject.hpp"
-#include "utilities/idd/IddEnums.hpp"
+#include "../idd/IddObject.hpp"
+#include "../idd/IddEnums.hpp"
```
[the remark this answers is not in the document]
jmarrec: That's the way all other files are as far as I can tell. It's debatable whether it's a good idea, but I'd prefer consistency.

(Bit of context: We don't move things often (if at all), and I am not afraid of spending 15 minutes to write a python script with a regex or two to fixup the paths if we need to move things)

src/utilities/alfalfa/AlfalfaComponent.cpp
```diff
@@ -1,10 +1,9 @@
 #include "AlfalfaComponent.hpp"
-#include <iostream>
 
 namespace openstudio {
 namespace alfalfa {
 
-  AlfalfaComponent::AlfalfaComponent(const std::string& type, const Capability capabilities) : type(type), capabilities(capabilities) {}
+  AlfalfaComponent::AlfalfaComponent(const std::string& type, Capability capabilities) : type(type), capabilities(capabilities) {}
```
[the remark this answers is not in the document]
jmarrec: yes, it's a POD type [1] that'll fit in a register. I tried adding a clang-tidy check for that in LLVM (and someone else did independently) but it never got in.

[1] This may be an unsigned int, and it is on GCC for eg (cf https://gcc.godbolt.org/z/47z5z3bP8), but it's actually implementation defined per the cpp standards.

src/utilities/alfalfa/AlfalfaPoint.cpp
```diff
@@ -97,36 +103,36 @@ namespace alfalfa {
   Json::Value AlfalfaPoint::toJSON() const {
     Json::Value point;
 
-    if (id().is_initialized()) {
-      point["id"] = id().get();
+    if (auto id_ = id()) {
```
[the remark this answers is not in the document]
jmarrec: boost::optional has an `operator bool()`, which is going to be used here.

FYI, C++17 added init-statement to `if`. So you could do something like

`if (auto x = func(); x.condition()) {`

https://en.cppreference.com/w/cpp/language/if

src/utilities/alfalfa/AlfalfaPoint.cpp
```diff
@@ -4,15 +4,21 @@
 #include <boost/regex.hpp>
 #include <fmt/format.h>
 
+#include <memory>
+#include <string_view>
+
 namespace openstudio {
 namespace alfalfa {
+
+  static constexpr std::string_view ID_VALID_CHARS_MSG = "IDs can only contain letters, numbers, and the following special characters _-[]():";
```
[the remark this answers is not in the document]
jmarrec: this is a global static constexpr variable. I could omit the `static`, and it'd still be `static` because it's declared at global scope and it's constexpr, but this is more explicit.

And so it initializes it in static storage (the `.rodata` section of the binary). The string_view is materialized in the binary.

see this short example: https://gcc.godbolt.org/z/o7veTMETP

and if you want to learn more about constexpr, I suggest C++ weekly, the start of this episode maybe: https://www.youtube.com/watch?v=IDQ0ng8RIqs

src/utilities/filetypes/AlfalfaJSON.cpp
jmarrec: This is by definition a temporary (because it does not have a name), so it's an R-value, and I know that there will be no copies.

In this specific case, this is unnecessary, because the `exposeFromComponent` actually takes a const-ref, so no copy or move will happen, but that may change... **It didn't hinder readability much here**, so I just went with it.

This is heavily related to the pass-by-value and move idiom.

Here is something to play for you, switch the defines on line 41 and 42: https://gcc.godbolt.org/z/911xK1bT6

You'll see that if exposeFromComponent takes by-value:
* If you NAME your variable, the Ctor is called, then the Copy Ctor
* If you use a temporary (like I did here), the variable is actually constructed directly in the exposeFromComponent scope and you save a copy.

src/utilities/alfalfa/AlfalfaActuator.cpp
```diff
@@ -27,35 +27,35 @@ namespace alfalfa {
...
-    boost::optional<std::string> control_type;
+    std::string component_name;
+    std::string component_type;
+    std::string control_type;
 
     if (idd_type == IddObjectType::OS_EnergyManagementSystem_Actuator) {
-      component_name = actuator.getString(OS_EnergyManagementSystem_ActuatorFields::ActuatedComponentName);
-      component_type = actuator.getString(OS_EnergyManagementSystem_ActuatorFields::ActuatedComponentType);
-      control_type = actuator.getString(OS_EnergyManagementSystem_ActuatorFields::ActuatedComponentControlType);
+      component_name = actuator.getString(OS_EnergyManagementSystem_ActuatorFields::ActuatedComponentName).value_or("");
```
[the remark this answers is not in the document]
jmarrec: https://live.boost.org/doc/libs/1_79_0/libs/optional/doc/html/boost_optional/reference/header__boost_optional_optional_hpp_/detailed_semantics___optional_values.html

image

src/utilities/alfalfa/AlfalfaPoint_Impl.hpp
jmarrec: It's annoying to write for sure. For ModelObjects we actually have ruby scripts that will do 90% of the work for us and generate the pimpl class: https://github.com/NREL/OpenStudio/blob/develop/developer/ruby/GenerateClass.rb and https://github.com/NREL/OpenStudio/blob/develop/developer/doc/Adding_Model_Object_to_OpenStudio.docx

I had written this when I first touched the OpenStudio codebase: https://github.com/jmarrec/OpenStudio-Tutorial-for-Cpp-Noobs?tab=readme-ov-file#pimpl

a few years later, I think the pImpl pattern in the case of model objects at least is useful for us because:

* The Impl methods are not exposed to SWIG (ruby, python, C# bindings)
* Technically you can change the Impl header without changing the public header, which is good because:
    * If a class never included the Impl header (and we have to do that too often for casting especially), then there's a net compilation time benefit
    * It forces us to treat the public header as a sacred API, and non-backward compatible changes must be really  motivated. We often remove methods from the Impl when they become moot, and we'll put a `[[deprecated]]` method in the public side for 2 or 3 releases with a clear message that it'll be removed.

jmarrec:
@TShapinsky I took some time to give (probably too) detailed explanations in the hope that it's useful context and references.

Anyways, you're responsible for the PR / branch I'm pointing this to, so it's up to you to merge this one or not (I don't know if you have merge rights though, but I can click the button if you ask me to)

src/utilities/alfalfa/AlfalfaPoint_Impl.hpp
TShapinsky: That makes sense. The reason I used the pImpl pattern here was to allow the point object a user is given after creation and the point object in the points vector within `AlflalfaJSON` to all point to the same underlying data. This seemed like the best option outside of dealing with just raw or lightly wrapped pointers. If there is a better paradigm you have in mind for this I would be happy to implement it.

src/utilities/alfalfa/AlfalfaActuator.cpp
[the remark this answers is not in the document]
TShapinsky: that sounds good, currently there a question on the other PR from kyle as to whether `alfalfa` should be moved out of `utilities` to be on the same level. Ideally I think I'd wait for that before switching everything back to relative, but as you noted it's not that big of a lift to fix.

PR #5266, opened 2024-10-03
Make pip install run only once instead of everytime CMake reconfigures

jmarrec:
Pull request overview
---------------------

Make pip install run only once. This takes 15s to a minute (depending on pip cache and internet connection) and would run everytime cmake reconfigures, which annoyed the crap out of me.

@kbenne please review.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -689,7 +689,7 @@ if(UNIX)
     endif()
 
     execute_process(COMMAND ${CMAKE_COMMAND} -E remove_directory "${PROJECT_BINARY_DIR}/${ENERGYPLUS_PATH}")
-
+    file(REMOVE "${PROJECT_BINARY_DIR}/python/engine/pip_install_done.stamp")
   endif()
 
 elseif(WIN32)
@@ -715,7 +715,7 @@ elseif(WIN32)
          EXPECTED_MD5 ${ENERGYPLUS_EXPECTED_HASH})
 
     execute_process(COMMAND ${CMAKE_COMMAND} -E remove_directory "${PROJECT_BINARY_DIR}/${ENERGYPLUS_PATH}")
-
+    file(REMOVE "${PROJECT_BINARY_DIR}/python/engine/pip_install_done.stamp")
   endif()
 
 endif()
--- a/python/engine/CMakeLists.txt
+++ b/python/engine/CMakeLists.txt
@@ -45,6 +45,27 @@ target_compile_options(pythonengine PRIVATE
 
 target_compile_definitions(pythonengine PRIVATE openstudio_scriptengine_EXPORTS SHARED_OS_LIBS)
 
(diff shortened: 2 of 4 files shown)


python/engine/CMakeLists.txt
```diff
@@ -45,6 +45,27 @@ target_compile_options(pythonengine PRIVATE
...
+# And a target that calls the above command. It is NOT called by default
+add_custom_target(pip_install
+  DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/pip_install_done.stamp"
+)
+
+if(BUILD_CLI)
+  # When you build CLI, we want the pip_install to run, so add it as a dependency
+  # which will ensure it's run (beforehand, at **build** not configure time)
+  add_dependencies(pythonengine pip_install)
+endif()
```
jmarrec, line 67: This is the gist, and it's commented enough that I don't need to say anything else.

I have tested that this works, and that changing requirements.txt triggers another pip install correctly.

CMakeLists.txt
```diff
@@ -689,7 +689,7 @@ if(UNIX)
     endif()
 
     execute_process(COMMAND ${CMAKE_COMMAND} -E remove_directory "${PROJECT_BINARY_DIR}/${ENERGYPLUS_PATH}")
-
+    file(REMOVE "${PROJECT_BINARY_DIR}/python/engine/pip_install_done.stamp")
```
jmarrec, line 692: When we change E+ directory (update), remove the .stamp file to trigger another install

python/engine/CMakeLists.txt
jmarrec: It's pretty much the same as https://github.com/NREL/OpenStudio/blob/ba1b94b6542c1f1d4cdcdb775886852b9a82bc4f/ProjectMacros.cmake#L153-L180 excepts I use `DEPENDS` instead of `SOURCES` in [add_custom_target](https://cmake.org/cmake/help/latest/command/add_custom_target.html), because I think that's the right one.

> **DEPENDS**
>    Reference files and outputs of custom commands created with [add_custom_command()](https://cmake.org/cmake/help/latest/command/add_custom_command.html#command:add_custom_command) command calls in the same directory (CMakeLists.txt file). They will be brought up to date when the target is built.


> **SOURCES**
>  Specify additional source files to be included in the custom target. Specified source files will be added to IDE project files for convenience in editing even if they have no build rules.

kbenne: I think I've setup similar build rules with a custom target combined with a custom command. Each time I think I go back to the documentation looking for a way to do it with a single custom target, as I did again today, but in the end I think what you've done is the correct way. The add_custom_target command is always out of date, so we have to do this in two steps.

PR #5296, opened 2024-11-07
Relax Alfalfa Point ID Requirements

TShapinsky:
Pull request overview
---------------------
After testing the new interface in the real world I found the requirements on id characters to be too onerous. This PR removes all requirements outside of non emptiness.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/alfalfa/AlfalfaPoint.cpp
+++ b/src/alfalfa/AlfalfaPoint.cpp
@@ -92,9 +92,6 @@ namespace alfalfa {
         throw std::runtime_error("Display name must have non-zero length");
       }
       m_display_name = display_name;
-      if (m_id.empty()) {
-        setId(toIdString(display_name));
-      }
     }
 
     std::string AlfalfaPoint_Impl::displayName() const {
--- a/src/alfalfa/AlfalfaPoint.cpp
+++ b/src/alfalfa/AlfalfaPoint.cpp
@@ -11,8 +11,6 @@
 namespace openstudio {
 namespace alfalfa {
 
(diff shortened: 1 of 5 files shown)


src/alfalfa/AlfalfaPoint.cpp
```diff
@@ -125,17 +106,11 @@ namespace alfalfa {
       return m_optional;
     }
 
-    bool AlfalfaPoint_Impl::isValidId(const std::string& id) {
-      return !id.empty() && boost::regex_match(id, boost::regex(R"(^[A-Za-z0-9_\-\[\]:()]*$)"));
-    }
```
jmarrec, line 130: @TShapinsky can you clarify what you meant by "onerous"?

Was it just annoying / too restrictive?
Or was is slow?

Boost (and std) regexes are awful in terms of performance, but using a static regex makes it better (we use Meyers singletons in IddRegex.hpp for eg)

TShapinsky: It was mostly annoying / too restrictive. For some background, the point ids become path parameters when interacting with the Alfalfa/BOPTest REST API. Originally I thought it would be neat to restrict the character set to avoid needing a lot of percent encoded characters. However, it's pretty common for names in open studio to include various different characters and since the requirement isn't hard I decided to get rid of it in the interest of making the alfalfa interface more user friendly.

kbenne (approved):

TShapinsky:
@kbenne can you merge it? I don't have permission to do so.

PR #5367, opened 2025-03-14
#4830 - Add a new ReportingMeasure::modelOutputRequests(model, runner, argument_map) that runs before E+ FT

### LINKED ISSUE #4830: Allow model instead of workspace in `energyPlusOutputRequests`
Enhancement Request
--------------
When writing a reporting measure and requesting outputs (using output variables like `Output:Variable` or other output configuration objects like `OutputControl:Timestamps`), you only have access to `workspace` and have to insert E+ objects. The initial reason for this may have been because the objects were not wrapped in the SDK, but almost all output-related objects are wrapped in OpenStudio these days.

So can we allow `model` instead of `workspace` for output requests? If it needs to be non-breaking, it could be a new method such that the existing method is preserved.

Pros:
1. It's cleaner, better, and more maintainable for a writer of a reporting measure to interact with OpenStudio objects (vs writing E+ objects as strings).
2. It allows the writer of a reporting measure to set object properties _individually_. With the E+ objects, you have to insert an entire valid object, and it's [not obvious what behavior should be expected for unique objects that have already been assigned some values upstream](https://github.com/NREL/OpenStudio-workflow-gem/pull/147#discussion_r1140146099).
3. The OS development team [won't have to maintain a list of valid E+ objects that can be used (and write special code for some of them)](https://github.com/NREL/OpenStudio/blob/64d5ff2e1d5250dd9563d669de71fe1aa4609c84/src/workflow/Util.cpp#L162-L219).

Cons:
1. A writer of a reporting measure might interact with the model in ways that are not desired/expected (and which have nothing to do with output).

There may be other pros/cons, but that's what immediately came to mind.

jmarrec:
Pull request overview
---------------------

 - Fixes #4830

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/python/engine/PythonEngine.cpp
+++ b/python/engine/PythonEngine.cpp
@@ -413,18 +413,33 @@ int PythonEngine::numberOfArguments(ScriptObject& methodObject, std::string_view
   return numberOfArguments;
 }
 
-bool PythonEngine::hasMethod(ScriptObject& methodObject, std::string_view methodName) {
+bool PythonEngine::hasMethod(ScriptObject& methodObject, std::string_view methodName, bool overriden_only) {
   auto val = std::any_cast<PythonObject>(methodObject.object);
   if (PyObject_HasAttrString(val.obj_, methodName.data()) == 0) {
     return false;
   }
   PyObject* method = PyObject_GetAttrString(val.obj_, methodName.data());  // New reference
-  bool result = false;
-  if (PyMethod_Check(method)) {
-    result = true;
+  if (PyMethod_Check(method) == 0) {
+    // Should never happen with modelOutputRequests since the Base class (C++) has it
+    return false;
+  }
+  Py_DECREF(method);
+  if (!overriden_only) {
+    return true;
   }
 
-  Py_DECREF(method);
+  // equivalent to getattr(instance_obj.__class__, method_name) == getattr(instance_obj.__class__.__bases__[0], method_name)
+  PyTypeObject* class_type = Py_TYPE(val.obj_);  // PyObject_Type returns a strong (New) reference, not needed for us
+  PyObject* class_method = PyObject_GetAttrString((PyObject*)class_type, methodName.data());  // New reference
+
+  assert(class_type->tp_base != nullptr);
+  auto* base = (PyTypeObject*)class_type->tp_base;
+  PyObject* base_method = PyObject_GetAttrString((PyObject*)base, methodName.data());  // New reference
+
+  bool result = class_method != base_method;
+  Py_DECREF(class_method);
+  Py_DECREF(base_method);
+
   return result;
 }
 
--- a/python/engine/PythonEngine.cpp
+++ b/python/engine/PythonEngine.cpp
@@ -413,6 +413,21 @@ int PythonEngine::numberOfArguments(ScriptObject& methodObject, std::string_view
   return numberOfArguments;
 }
 
+bool PythonEngine::hasMethod(ScriptObject& methodObject, std::string_view methodName) {
+  auto val = std::any_cast<PythonObject>(methodObject.object);
+  if (PyObject_HasAttrString(val.obj_, methodName.data()) == 0) {
+    return false;
+  }
+  PyObject* method = PyObject_GetAttrString(val.obj_, methodName.data());  // New reference
+  bool result = false;
+  if (PyMethod_Check(method)) {
+    result = true;
+  }
+
+  Py_DECREF(method);
+  return result;
+}
+
 }  // namespace openstudio
 
 extern "C"
--- a/ruby/engine/RubyEngine.cpp
+++ b/ruby/engine/RubyEngine.cpp
@@ -237,6 +237,12 @@ int RubyEngine::numberOfArguments(ScriptObject& methodObject, std::string_view m
   return rb_obj_method_arity(val, method_id);
 }
 
+bool RubyEngine::hasMethod(ScriptObject& methodObject, std::string_view methodName) {
+  auto val = std::any_cast<VALUE>(methodObject.object);
+  ID method_id = rb_intern(methodName.data());
+  return rb_respond_to(val, method_id) == 1;
+}
+
 }  // namespace openstudio
 
 extern "C"
--- a/ruby/engine/RubyEngine.cpp
+++ b/ruby/engine/RubyEngine.cpp
@@ -237,10 +237,23 @@ int RubyEngine::numberOfArguments(ScriptObject& methodObject, std::string_view m
   return rb_obj_method_arity(val, method_id);
 }
 
-bool RubyEngine::hasMethod(ScriptObject& methodObject, std::string_view methodName) {
+bool RubyEngine::hasMethod(ScriptObject& methodObject, std::string_view methodName, bool overriden_only) {
   auto val = std::any_cast<VALUE>(methodObject.object);
   ID method_id = rb_intern(methodName.data());
-  return rb_respond_to(val, method_id) == 1;
+  if (rb_respond_to(val, method_id) == 0) {
+    return false;
+  }
+  if (!overriden_only) {
+    return true;
+  }
+
+  // I'd have prefered to do the equivalent of `instance_obj.method(:methodName).owner == instance_obj.class` but that is borderline impossible
+  // Instead, this is equivalent to `instance_obj.class.instance_methods(false).include?(:methodName)`
+  VALUE klass = rb_obj_class(val);
+  // include_methods_from_ancestors: false;
+  VALUE argv[1] = {Qfalse};
+  VALUE methods_without_ancestors = rb_class_instance_methods(1, argv, klass);
+  return rb_ary_includes(methods_without_ancestors, ID2SYM(method_id)) == Qtrue;
 }
 
 }  // namespace openstudio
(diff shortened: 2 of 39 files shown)


src/measure/ReportingMeasure.hpp
```diff
@@ -55,6 +55,15 @@ namespace measure {
...
 
+    /** This method is called on all reporting measures immediately before the translation to E+ IDF.
+     *  There is an implicit contract that this method should NOT be modifying your model in a way that would produce
+     *  different results, meaning it should only add or modify reporting-related elements (eg: OutputTableSummaryReports, OutputControlFiles, etc)
+     *  but that is not going to be enforced.
+     *  If you mean to modify the model in a significant way, use a ModelMeasure.
+     */
+    virtual bool modelOutputRequests(openstudio::model::Model& model, OSRunner& runner,
+                                     const std::map<std::string, OSArgument>& user_arguments) const;
+
```
jmarrec, line 66: New

src/scriptengine/ScriptEngine.hpp
```diff
@@ -77,8 +77,12 @@ class ScriptEngine
   // issue for the underlying ScriptObject (and VALUE or PyObject), so just return the ScriptObject
   virtual ScriptObject loadMeasure(const openstudio::path& measureScriptPath, std::string_view className) = 0;
 
+  // Returns number of arguments for methodName of the object methodObject
   virtual int numberOfArguments(ScriptObject& methodObject, std::string_view methodName) = 0;
 
+  // Check if methodObject has a method called methodName. If overriden_only, will only look into the current subclass, not the parent ones
+  virtual bool hasMethod(ScriptObject& methodObject, std::string_view methodName, bool overriden_only = true) = 0;
```
jmarrec, line 84: I really thought this would be needed to run older measures. And it was by far the hardest thing to achieve.
Only to realize that the base class having it makes it safe to call it even if not implemented/overidden in the actual measure...

src/workflow/ApplyMeasure.cpp
```diff
@@ -276,7 +276,16 @@ end
...
+        if (apply_measure_type == ApplyMeasureType::ModelOutputRequests) {
+          if ((*thisEngine)->hasMethod(measureScriptObject, "modelOutputRequests")) {
+            auto n = model.numObjects();
+            LOG(Debug, "Calling measure.modelOutputRequests for '" << measureDirName << "'");
+            static_cast<openstudio::measure::ReportingMeasure*>(measurePtr)->modelOutputRequests(model, runner, argmap);
+            LOG(Debug, "Finished measure.modelOutputRequests for '" << measureDirName << "', " << (model.numObjects() - n) << " objects added");
+          } else {
+            LOG(Debug, "Reporting Measure '" << measureDirName << "' does not have a modelOutputRequests method");
+          }
+        } else if (apply_measure_type == ApplyMeasureType::EnergyPlusOutputRequest) {
```
jmarrec, line 288: Since I did the work, I'm keeping it to output a debug message

src/workflow/OSWorkflow.hpp
```diff
@@ -124,7 +124,15 @@ class OSWorkflow
...
-  void applyMeasures(MeasureType measureType, bool energyplus_output_requests = false);
+
+  enum class ApplyMeasureType
+  {
+    Regular = 0,
+    ModelOutputRequests,
+    EnergyPlusOutputRequest
+  };
+
+  void applyMeasures(MeasureType measureType, ApplyMeasureType = ApplyMeasureType::Regular);
```
jmarrec, line 135: Func def tweak

src/workflow/RunOpenStudioMeasures.cpp
```diff
@@ -19,9 +19,13 @@ void OSWorkflow::runOpenStudioMeasures() {
...
 
   LOG(Info, "Beginning to execute OpenStudio Measures");
-  applyMeasures(MeasureType::ModelMeasure, false);
+  applyMeasures(MeasureType::ModelMeasure, ApplyMeasureType::Regular);
   LOG(Info, "Finished applying OpenStudio Measures.");
 
+  LOG(Info, "Beginning to execute Reporting Measures's Model Output Requests");
+  applyMeasures(MeasureType::ReportingMeasure, ApplyMeasureType::ModelOutputRequests);
+  LOG(Info, "Finished applying Reporting Measures's Model Output Requests.")
+
```
jmarrec, line 28: I made it run right after openstudio modelmeasures. I should probably make a separate job for it for Timings purposes.

python/engine/PythonEngine.cpp
```diff
@@ -413,6 +413,39 @@ int PythonEngine::numberOfArguments(ScriptObject& methodObject, std::string_view
...
+  auto* base = (PyTypeObject*)class_type->tp_base;
+  // cppcheck-suppress cstyleCast
+  PyObject* base_method = PyObject_GetAttrString((PyObject*)base, methodName.data());  // New reference
+
+  bool result = class_method != base_method;
+  Py_DECREF(class_method);
+  Py_DECREF(base_method);
+
+  return result;
+}
```
jmarrec, line 447: Here the python hasMethod version

ruby/engine/RubyEngine.cpp
```diff
@@ -237,6 +237,25 @@ int RubyEngine::numberOfArguments(ScriptObject& methodObject, std::string_view m
...
+  }
+
+  // I'd have prefered to do the equivalent of `instance_obj.method(:methodName).owner == instance_obj.class` but that is borderline impossible
+  // Instead, this is equivalent to `instance_obj.class.instance_methods(false).include?(:methodName)`
+  VALUE klass = rb_obj_class(val);
+  // include_methods_from_ancestors: false;
+  VALUE args[1] = {Qfalse};
+  VALUE methods_without_ancestors = rb_class_instance_methods(1, args, klass);
+  return rb_ary_includes(methods_without_ancestors, ID2SYM(method_id)) == Qtrue;
+}
```
jmarrec, line 257: The worst was this. Ruby's C API is what it is... i found lots of methods that were better suited but none public, and it took a long time to come up with this

shorowit:
Thank you, @jmarrec!!

(The title should say `user_arguments` instead of `argument_map`, right?)

I went to try this on a [reasonably complex reporting measure](https://github.com/NREL/OpenStudio-HPXML/blob/38bcc7678ed7ea3ec57b984a8242aa0354bf0bad/ReportSimulationOutput/measure.rb#L342), only to get 95% of the way through it before discovering that we use an E+ object (`Output:Table:Monthly`) that isn't wrapped by OpenStudio. Doh!

jmarrec:
That one could still be added as an energyPlusOutputRequests.

File an issue to wrap it if there's none and you want it, should be simple enough (2-4 hours of work top I think)

shorowit:
Yeah, I thought about that. Could do if need be.

Added an issue: https://github.com/NREL/OpenStudio/issues/5368

shorowit (approved):
Tested it w/ two OS-HPXML reporting measures and it is working as expected.

shorowit:
@jmarrec Not a big deal, but I was a little surprised to see that a new reporting measure created from the CLI (`openstudio measure new -t ReportingMeasure ...`) has both the `modelOutputRequests` and `energyPlusOutputRequests` methods. I see there are some comments trying to help the user choose the former over the latter, but why not exclude `energyPlusOutputRequests` altogether? Seems like we should discourage its use going forward.

jmarrec:
That's a good point that I didn't really consider to be honest.

Thinking about it I guess we don't wrap all object types in the OS SDK, so I think it makes sense to keep energyPlusOutputRequests in the template.

Another thing is that we still offer the `classic` (Ruby) CLI which will not run it. I actually think we should warn on the workflow-gem when we detect a modelOutputRequests and **not** run it (instead of silently not run it), I had a plan to make a PR there.

src/measure/ReportingMeasure.hpp
kbenne: Thanks for the code comment @jmarrec. You're speaking to my comments over in the original GitHub [issue](https://github.com/NREL/OpenStudio/issues/4830#issuecomment-1557628854).

src/scriptengine/ScriptEngine.hpp
kbenne: But the check remains? Seems like a switch statement on has "modelOutputRequests" is unneessary. Not a big deal in any case.

src/workflow/ApplyMeasure.cpp
kbenne: oh i see.

kbenne:
@DavidGoldwasser can we remove classic CLI now?

jmarrec:
* https://github.com/NREL/OpenStudio-PAT/pull/279

DavidGoldwasser:
> @DavidGoldwasser can we remove classic CLI now?

@brianlball I know we can't get rid of Workflow gem but can we get rid of Classic CLI? I don't believe you are using that in OSAF or PAT, but wasn't sure. If we need to we can push this back to remove in 3.11

jmarrec:
PAT is definitely using classic, cf the linked PR above: 

* https://github.com/NREL/OpenStudio-PAT/pull/279

PR #5394, opened 2025-04-07
Fix #5388 - --bundle options not working for CLI in docker-openstudio, possibly OS itself + Fix #5190 - Don't pick up system gem in CLI

### LINKED ISSUE #5190: CLI includes system paths in the gem search path
```
➜  OpenStudio git:(develop) ✗ ./build/Products/openstudio -e "puts Gem.paths.path"
/home/kbenne/.gem/ruby/3.2.0
//lib/ruby/gems/3.2.0
```

In some cases bundler installs to these paths and it pollutes the CLI. There have been various reports from @joseph-robertson and @asparke2. I believe this is the root cause.

The tell-tale sign is when 
```
openstudio list_gems
```

shows gems that are not in the embedded filesystem.

### LINKED ISSUE #5388: --bundle options not working for CLI in docker-openstudio, possibly OS itself
Issue overview
--------------
since Ruby 3.2.2, the `--bundle`, `--bundle_path` and `--bundle_without` args dont seem to be working right.
This impacts the ability to use a custom `Gemfile` in OSAF.

## Current Behavior
a min rep example is to 
`docker pull nrel/openstudio:3.9.0`
`docker run -it nrel/openstudio:3.9.0`

Then use the  `--bundle`, `--bundle_path` and `--bundle_without` args in the CLI with command `gem_list` with the already installed `/var/oscli/Gemfile` and `/var/oscli/gems`

`/usr/local/bin/openstudio --bundle /var/oscli/Gemfile --bundle_path /var/oscli/gems --bundle_without native_ext gem_list`
output:
`>Bundler::GemNotFound: Could not find gem 'oga (= 3.2)' in cached gems or installed locally.`

if you remove the `--bundle_without` arg you get

`/usr/local/bin/openstudio --bundle /var/oscli/Gemfile --bundle_path /var/oscli/gems gem_list`
output:
`Warn: Bundle activated but ENV['BUNDLE_WITHOUT'] is not set`
`Info: Setting BUNDLE_WITHOUT to 'test'`
`Bundler::GemNotFound: Could not find sqlite3-1.7.2-x86_64-linux in locally installed gems`

If you just do it without the `--bundle` args it works and finds the gems:
`/usr/local/bin/openstudio gem_list`
output:
`...`
`oga (3.2) ':/ruby/3.2.0/gems/oga-3.2'`
`sqlite3 (1.7.2) ':/ruby/3.2.0/gems/sqlite3-1.7.2-x86_64-linux'`
`...`

## Expected Behavior
The `--bundle`, `--bundle_path` and `--bundle_without` args to the CLI should work and use the passed in Gemfile and gem bundle

## Steps to Reproduce
see above

### Environment
OpenStudio 3.8+
I'm seeing it in nrel/openstudio:3.9.0 from https://github.com/NREL/docker-openstudio
but I'm not sure its limited to just these docker images.

 - Ubuntu 22.04
 - Version of OpenStudio (if using an intermediate build, include SHA): 3.9.0

### Context
using custom versions of OpenStudio-Standards in OSAF is broken right now since users cannot easily override the gems that come with OpenStudio.

jmarrec:
Pull request overview
---------------------

 - Fixes #5388
 - Fixes #5190 

I am extremely unhappy with what I came up with, but that's the result of a day of hitting the same wall.

I'm not even 100% sure this fixes all corner cases yet.

We ought to remove the embedded filesystem.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/ruby/bindings/InitRubyBindings.cpp
+++ b/ruby/bindings/InitRubyBindings.cpp
@@ -851,6 +851,8 @@ ENV.reject! { |k, _| ['GEM', 'BUNDLE'].any? { |x| k.start_with?(x) } }
 
     if (use_bundler) {
       initCmd += R"ruby(
+  $logger.trace "embedded_gems_to_activate=#{embedded_gems_to_activate.map{ |s| s.name} }"
+
   # Load the bundle before activating any embedded gems
   embedded_gems_to_activate.each do |spec|
     if spec.name == "bundler"
@@ -884,6 +886,7 @@ ENV.reject! { |k, _| ['GEM', 'BUNDLE'].any? { |x| k.start_with?(x) } }
     RbConfig::CONFIG['arch'] = 'x64-mingw32'
   end
 
+  $logger.trace "Requiring bundler"
   # require bundler
   # have to do some forward declaration and pre-require to get around autoload cycles
   require 'bundler/errors'
@@ -962,9 +965,16 @@ ENV.reject! { |k, _| ['GEM', 'BUNDLE'].any? { |x| k.start_with?(x) } }
     locked_specs = Bundler.definition.instance_variable_get(:@locked_specs)
 
     embedded_gems_to_activate.each do |spec|
-      next if spec.extensions.empty?
+      $logger.trace "embedded_gems_to_activate.each: spec: #{spec.name}"
+      if spec.extensions.empty?
+        $logger.trace "Spec #{spec.name} has no extensions"
+        next
+      end
 
-      next unless locked_specs.any? {|locked_spec| locked_spec.name == spec.name}
+      unless locked_specs.any? {|locked_spec| locked_spec.name == spec.name}
+        $logger.trace "#{spec.name} not in locked_specs"
+        next
+      end
 
       dep_to_requirements = {}
       locked_specs.each do |locked_spec|
--- a/ruby/bindings/InitRubyBindings.cpp
+++ b/ruby/bindings/InitRubyBindings.cpp
@@ -662,6 +662,10 @@ end)ruby";
     // in hash.c there's lot of env-related functions, but all private, aside from rb_env_clear();
     const std::string clearEnvs = R"ruby(
 ENV.reject! { |k, _| ['GEM', 'BUNDLE'].any? { |x| k.start_with?(x) } }
+# No env variable isn't the same as empty. Gem::PathSupport does stuff like @home = env["GEM_HOME"] || Gem.default_dir
+ENV['GEM_HOME'] = ''
+ENV['GEM_PATH'] = ''
+ENV['GEM_SPEC_CACHE'] = ''
   )ruby";
 
     openstudio::evalString(clearEnvs);
@@ -749,6 +753,21 @@ ENV.reject! { |k, _| ['GEM', 'BUNDLE'].any? { |x| k.start_with?(x) } }
     end
   end
 
+#  module Gem
+#    def self.default_dir
+#      @default_dir = ""
+#    end
+#
+#    def self.find_home
+#      ""
+#    end
+#
+#    def self.default_specifications_dir
+#      @default_specifications_dir = ""
+#    end
+#  end
+#
+  Gem.clear_paths
   Gem::Deprecate.skip = true
 
   # find all the embedded gems
--- a/ruby/bindings/InitRubyBindings.cpp
+++ b/ruby/bindings/InitRubyBindings.cpp
@@ -630,7 +630,7 @@ module IRB # :nodoc:
 
     @CONF[:IRB_RC].call(irb.context) if @CONF[:IRB_RC]
     # puts @CONF[:PROMPT_MODE]
-    os_version = "3.8.0" # OpenStudio::openStudioVersion
+    os_version = "3.10.0" # OpenStudio::openStudioVersion
     @CONF[:PROMPT][:OPENSTUDIO] = {
       :PROMPT_I=>"(os #{os_version}) :%03n > ",
       :PROMPT_S=>"(os #{os_version}) :%03n%l> ",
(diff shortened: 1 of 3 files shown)


ruby/bindings/InitRubyBindings.cpp
```diff
@@ -964,6 +964,8 @@ ENV.reject! { |k, _| ['GEM', 'BUNDLE'].any? { |x| k.start_with?(x) } }
     # Filter out dependencies with native extensions that we already have embedded in the CLI and that satisfy the requirements
     locked_specs = Bundler.definition.instance_variable_get(:@locked_specs)
 
+    extra_load_paths = []
```
jmarrec, line 986: Store

ruby/bindings/InitRubyBindings.cpp
```diff
@@ -986,27 +988,67 @@ ENV.reject! { |k, _| ['GEM', 'BUNDLE'].any? { |x| k.start_with?(x) } }
...
+        Gem::Specification.add_spec(spec)
         raise "Failed to activate #{spec.name}" unless spec.activate
+        Bundler.rubygems.mark_loaded(spec)
+        stub = Gem::Specification.stubs_for(spec.name).first
+        bundler_stub = Bundler::StubSpecification.from_stub(stub)
+        raise "That shouldn't happen" if locked_specs[spec.name].nil?
+        # locked_specs.delete_by_name(spec.name)
+        # locked_specs[spec.name] = bundler_stub
+
+        extra_load_paths += spec.full_require_paths
```
jmarrec, line 1051: Pfff. Seriously. That needs cleaning up  I guess, I'm not really using the stubs: I thought I could swap the Bundler::LazySpecification with the Bundler::StubSpecification instead, but that doesn't work.

Anyways, the idea is to try to activate the dependencies of the spec before the spec is activated...

And I'm storing the load paths that are added

ruby/bindings/InitRubyBindings.cpp
```diff
@@ -986,27 +988,67 @@ ENV.reject! { |k, _| ['GEM', 'BUNDLE'].any? { |x| k.start_with?(x) } }
...
     remaining_specs = []
     Bundler.definition.specs_for(keep_groups).each {|s| remaining_specs << s.name}
 
     $logger.info "Specs to be included [#{remaining_specs.join(',')}]"
 
-
+    # This is going to remove some stuff from the LOAD_PATH that we already loaded, when we really don't want to, so we readd them...
     Bundler.setup(*keep_groups)
+
+    Bundler.rubygems.add_to_load_path(extra_load_paths - $LOAD_PATH)
```
jmarrec, line 1069: This is bad, but I guess you get the idea.

jmarrec (commented):
Review for #5388

ruby/bindings/InitRubyBindings.cpp
```diff
@@ -662,6 +662,10 @@ end)ruby";
     // in hash.c there's lot of env-related functions, but all private, aside from rb_env_clear();
     const std::string clearEnvs = R"ruby(
 ENV.reject! { |k, _| ['GEM', 'BUNDLE'].any? { |x| k.start_with?(x) } }
+# No env variable isn't the same as empty. Gem::PathSupport does stuff like @home = env["GEM_HOME"] || Gem.default_dir
+ENV['GEM_HOME'] = ''
+ENV['GEM_PATH'] = ''
+ENV['GEM_SPEC_CACHE'] = ''
```
jmarrec, line 668: #5190 Do not leave GEM_HOME etc to be undefined, it's not the same as having a blank. + Call Gem::clear_paths

https://github.com/ruby/ruby/tree/v3_2_2lib/rubygems/path_support.rb#L8

ruby/engine/embedded_help.rb
```diff
@@ -94,6 +94,7 @@ module Kernel
   # puts $LOAD_PATH
 
   # TODO: double check and update platform-specific includes
+  $LOAD_PATH.clear
```
jmarrec, line 97: Clear the LOAD_PATH early on, so we don't pick up filesystem stuff.

ruby/engine/embedded_help.rb
```diff
@@ -700,6 +706,54 @@ def self.chdir(dir)
...
+# if RbConfig::CONFIG['prefix'] == '/'
+#   # Normally CONFIG["libdir"] = "$(prefix)/lib" but we want just $(prefix)
+#   puts "Fixing RbConfig prefix from #{RbConfig::CONFIG['prefix']} to ':'"
+#   libdir = RbConfig::CONFIG['libdir']
+#   RbConfig::CONFIG.transform_values!{ |val| val.gsub(libdir, ':').gsub('//', ':/') }
+#   RbConfig::CONFIG['bindir'] = File.dirname(EmbeddedScripting::applicationFilePath)
+# end
+raise "rubylibprefix isn't correct, it's '#{RbConfig::CONFIG["rubylibprefix"]}' but should be ':/ruby' "unless RbConfig::CONFIG["rubylibprefix"] == ':/ruby'
+
+# NOTE: fileutils requires rbconfig, so we have to do our RbConfig shenanigans beforehand
```
jmarrec, line 756: Pfff.

I really only needed to fix rubylibprefix 

https://github.com/ruby/ruby/blob/e51014f9c05aa65cbf203442d37fef7c12390015/lib/rubygems/defaults.rb#L36-L38
```

https://github.com/ruby/ruby/blob/v3_2_2/lib/rubygems/defaults.rb#L36-L38

or to path Gem.default_dir. But having the wrong prefix through RbConfig::CONFIG is recipe for disaster. So I chose to patch it fully.

```
class Gem
  def self.default_dir
    @default_dir ||= File.join(RbConfig::CONFIG["rubylibprefix"], "gems", RbConfig::CONFIG["ruby_version"])
  end
end
```

RbCONFIG is weird as hell but it'll expand some `$(var_name)` when it's first loaded, so I'm going to be using that to minimize the chances of screwing up. cf commit message on 03009e09524944ca88e9821f4bb5f1a5e1e61fc9 to see the rbconfig.rb that I have via conan ruby.

File: ruby/bindings/InitRubyBindings.cpp [hunk 7]
```diff
@@ -749,6 +753,21 @@ ENV.reject! { |k, _| ['GEM', 'BUNDLE'].any? { |x| k.start_with?(x) } }
...
+#
+#    def self.find_home
+#      ""
+#    end
+#
+#    def self.default_specifications_dir
+#      @default_specifications_dir = ""
+#    end
+#  end
+#
```
jmarrec, line 769: I didn't need this after fixing the RbConfig stuff, but we might...

ruby/bindings/InitRubyBindings.cpp
```diff
@@ -749,6 +753,21 @@ ENV.reject! { |k, _| ['GEM', 'BUNDLE'].any? { |x| k.start_with?(x) } }
...
+  Gem.clear_paths
```
jmarrec, line 770: This is going to force recomputing Gem @paths

Could have done `Gem.paths = {'GEM_HOME' => xxx, 'GEM_PATH' => xxx, 'GEM_SPEC_CACHE' => xx}`, but this is much better.

RubyEngine::setEmbeddedGems does:

* setupEmbeddedGemsClearEnvVars: set all three vars in ENV to ''
* initEmbeddedGems: will load rubygems
* setGemPathDir, setRubyEnvVarPath(various including GEM_HOME): all of these tweak ENV only if the path ins't empty, so it'll stay a '' blank string if not supplied
* locateEmbeddedGems calls Gem.clear_paths here. So paths is recomputed using the ENV variables via path_support, and will pick up the variables that were added (or not) to do so
*

jmarrec: https://github.com/ruby/ruby/blob/e51014f9c05aa65cbf203442d37fef7c12390015/lib/rubygems.rb#L313-L323

```ruby
  ##
  # Reset the +dir+ and +path+ values.  The next time +dir+ or +path+
  # is requested, the values will be calculated from scratch.  This is
  # mainly used by the unit tests to provide test isolation.


  def self.clear_paths
    @paths         = nil
    @user_home     = nil
    Gem::Specification.reset
    Gem::Security.reset if defined?(Gem::Security)
  end
```

https://github.com/ruby/ruby/blob/e51014f9c05aa65cbf203442d37fef7c12390015/lib/rubygems.rb#L358-L363

```ruby
  # Retrieve the PathSupport object that RubyGems uses to
  # lookup files.


  def self.paths
    @paths ||= Gem::PathSupport.new(ENV)
  end
```

jmarrec (commented):
Review for #5190

jmarrec:
@anchapin please fixup the ubuntu CI, they seem offline

shorowit:
I confirmed that the system gems no longer show up under `gem_list`.

kbenne:
This whole thing makes me shudder and I'm sure it doesn't please you either @jmarrec. I don't really know what to say, except time will tell if there are any unwanted side effects.

PR #5466, opened 2025-07-30
Fix floorArea method for Space returning unsorted surfaces

### LINKED ISSUE #5465: The floorArea method for Space can return a non deterministic result based on order of Surfaces
Issue overview
--------------
If OS:Space's "Floor Area" is not set in the OSM file, the `floorArea` method uses `this->surfaces()` to get the sum of surface gross areas. This can lead to a non deterministic result.

## Current Behavior
It's possible for `space.floorArea` to return slightly different values.

## Expected Behavior
The value should be the same every time the method is called.

## Steps to Reproduce

1.
2.
3.
4.

## Possible Solution
Is there a way we can step through surfaces in a deterministic order?

### Environment

joseph-robertson:
Pull request overview
---------------------

 - Fixes #5465

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/model/Space.cpp
+++ b/src/model/Space.cpp
@@ -698,7 +698,11 @@ namespace model {
     }
 
     SurfaceVector Space_Impl::surfaces() const {
-      return getObject<ModelObject>().getModelObjectSources<Surface>(Surface::iddObjectType());
+      // get all surfaces, sort so results are repeatable
+      std::vector<Surface> surfaces = getObject<ModelObject>().getModelObjectSources<Surface>(Surface::iddObjectType());
+      std::sort(surfaces.begin(), surfaces.end(), IdfObjectNameLess());
+
+      return surfaces;
     }
 
     std::vector<InternalMass> Space_Impl::internalMass() const {
@@ -2827,14 +2831,10 @@ namespace model {
     std::vector<Point3d> Space_Impl::floorPrint() const {
       double tol = 0.01;  // 1 cm tolerance
 
-      // get all surfaces, sort so results are repeatable
-      std::vector<Surface> surfaces = this->surfaces();
-      std::sort(surfaces.begin(), surfaces.end(), IdfObjectNameLess());
-
       // find all floors
       boost::optional<double> z;
       std::vector<Surface> floors;
-      for (const Surface& surface : surfaces) {
+      for (const Surface& surface : this->surfaces()) {
         if (surface.vertices().size() < 3) {
           LOG(Warn, "Skipping floor with fewer than 3 vertices");
           continue;


src/model/Space.cpp
```diff
@@ -698,7 +698,11 @@ namespace model {
     }
 
     SurfaceVector Space_Impl::surfaces() const {
-      return getObject<ModelObject>().getModelObjectSources<Surface>(Surface::iddObjectType());
+      // get all surfaces, sort so results are repeatable
+      std::vector<Surface> surfaces = getObject<ModelObject>().getModelObjectSources<Surface>(Surface::iddObjectType());
+      std::sort(surfaces.begin(), surfaces.end(), IdfObjectNameLess());
```
joseph-robertson: Is there a significant performance hit by putting the sort here? If so, we can just sort from within the `floorPrint` and `floorArea` methods.

shorowit: There could be a lot of surfaces in a model, but hard to imagine that there would that many for a single space.

src/model/Space.cpp
```diff
@@ -2827,14 +2831,10 @@ namespace model {
...
 
-      // get all surfaces, sort so results are repeatable
-      std::vector<Surface> surfaces = this->surfaces();
-      std::sort(surfaces.begin(), surfaces.end(), IdfObjectNameLess());
-
       // find all floors
       boost::optional<double> z;
       std::vector<Surface> floors;
-      for (const Surface& surface : surfaces) {
+      for (const Surface& surface : this->surfaces()) {
```
jmarrec: I can see how floorPrint *could* be affected by order of surfaces, since it's doing geometry things with it, but the original issue says it's floorArea that was affected, and as I said at https://github.com/NREL/OpenStudio/issues/5465#issuecomment-3179744342 I don't see how that's possible.

src/model/Space.cpp
joseph-robertson: https://github.com/NREL/OpenStudio/pull/5466/commits/f3b89705e5fec9f537720acb1633d2ed8814ca28

Like `floorPrint` which already sorts surfaces, update `floorArea` to also sort surfaces. There should be less of a performance hit risk having the sort downstream of the parent `surfaces` method.

src/model/Space.cpp
```diff
@@ -1071,8 +1071,12 @@ namespace model {
         return value.get();
       }
 
+      // get all surfaces, sort so results are repeatable
+      std::vector<Surface> surfaces = this->surfaces();
+      std::sort(surfaces.begin(), surfaces.end(), IdfObjectNameLess());
+
       double result = 0;
-      for (const Surface& surface : this->surfaces()) {
+      for (const Surface& surface : surfaces) {
```
jmarrec, line 1079: Ok, so now this is the only change, and it's only done in Space_Impl::floorArea

I was asked

> [@jmarrec](https://github.com/jmarrec) Are you ok with the approach described in [#5466 (comment)](https://github.com/NREL/OpenStudio/pull/5466#discussion_r2417478624)?

My _personnal_ opinion is that this is still unecessary, and this should be handled by the person trying to compare things. 

At least the scope / impact is smaller here.

@joseph-robertson and @shorowit seem to be pretty set on doing it it seems, so I'll leave it up to you to hit the merge button if that's the case.

jmarrec:
I really need to clean up the number of open PRs, I'm going with the assumption that this is really wanted. Merging.

Issue #2601, opened 2017-05-08
Ability to skip NA simulations


shorowit:
For ResStock, it would be extremely helpful to be able to skip simulations. The context is that we have measures (e.g., apply wall insulation) that only apply to a subset of the building stock (e.g., buildings with uninsulated walls). If the simulation continues without applying this measure, it means we are running a huge number of additional simulations that we don't need to be.

Skipping simulations should work by using:
```
runner.registerAsNotApplicable
return false
```
Using `registerAsNotApplicable; return true` should work as it does now (measure is not applied but simulation continues).

Note that one can of course use `registerError; return false` as a way to not run simulations, but then it becomes really difficult to determine what's a "true" error (e.g., measure syntax error) and what is simply us trying to skip a simulation.

@macumber @rHorsey @asparke2 @brianlball

macumber:
https://github.com/NREL/OpenStudio/commit/97a52031d59c1e0e70b6db20ad424d6be5ac5377

macumber:
https://github.com/NREL/OpenStudio-workflow-gem/commit/77e5a95b3c631278ec1c47c5fce557997b710f2d

rHorsey:
This should be done. Testing it well is difficult until we have 2.1.2 builds, but that should be doable shortly.

shorowit:
Worked well in my tests!

PR #5539, opened 2025-12-02
For halted workflow, return before FT and OSM save

joseph-robertson:
Pull request overview
---------------------

File run_translation.rb in the OpenStudio-workflow-gem [has a return before FT](https://github.com/NREL/OpenStudio-workflow-gem/blob/develop/lib/openstudio/workflow/jobs/run_translation.rb#L24), but OSWorkflow.cpp in OpenStudio's src/workflow folder does not. We'll now expect no in.idf or pre-preprocess.idf.

Furthermore, the OpenStudio-workflow-gem saves the model to in.osm even if the workflow was halted. This PR maintains that behavior.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/.github/workflows/incremental-build.yml
+++ b/.github/workflows/incremental-build.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
   push:
     branches:
-      - main
+      - develop
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
@@ -50,6 +50,15 @@ jobs:
           cd ${{ env.OPENSTUDIO_DOCKER_VOLUME }}/${{ env.OPENSTUDIO_SOURCE_NAME }}
           # make safe repository
           git config --global --add safe.directory "${{ env.OPENSTUDIO_DOCKER_VOLUME }}/${{ env.OPENSTUDIO_SOURCE_NAME }}"
+
+      - name: Remove old artifacts
+        run: |
+          BUILD_DIR="${{ env.OPENSTUDIO_DOCKER_VOLUME }}/${{ env.OPENSTUDIO_SOURCE_NAME }}/${{ env.OPENSTUDIO_BUILD_NAME }}"
+          if [ -d "$BUILD_DIR" ]; then
+            echo "Cleaning up old artifacts in $BUILD_DIR"
+            find "$BUILD_DIR" -maxdepth 1 -name "*.deb" -delete
+            rm -rf "$BUILD_DIR/_CPack_Packages"
+          fi
 
       - name: Install ccache
         run: |
--- a/ruby/engine/embedded_help.rb
+++ b/ruby/engine/embedded_help.rb
@@ -667,6 +667,8 @@ class Dir
     override_args_extglob = true
 
     flags = flags | File::FNM_EXTGLOB if override_args_extglob
+    # Without this 'path/**/*.rb' will match 'path/sub/file.rb' but not 'path/file.rb', cf #5546
+    flags = flags | File::FNM_PATHNAME
     result = []
     pattern_array.each do |pattern|
 
--- a/src/cli/test/test_embedded_ruby.rb
+++ b/src/cli/test/test_embedded_ruby.rb
@@ -242,4 +242,30 @@ class EmbeddedRuby_Test < Minitest::Test
     assert(!schemer.valid?({ 'abc' => 10 }))
   end
 
+  def test_json_schema
+    require 'json-schema'
+
+    schema = {
+      "type" => "object",
+      "required" => ["a"],
+      "properties" => {
+        "a" => {
+          "type" => "integer"
+        }
+      }
+    }
+
+    data_valid = { "a" => 5 }
+    JSON::Validator.validate(schema, data_valid)
+
+    data_invalid = { "a" => "taco" }
+
+    error = assert_raises(JSON::Schema::ValidationError) do
+      # "The property '#/a' of type String did not match the following type: integer"
+      JSON::Validator.validate!(schema, data_invalid)
+    end
+    assert_match(/The property '#\/a' of type string did not match the following type: integer/, error.message)
+
+  end
+
 end


src/workflow/RunTranslator.cpp
```diff
@@ -33,6 +33,7 @@ void OSWorkflow::runTranslator() {
     LOG(Info, "Workflow halted, skipping OSM to IDF translation");
     // This allows model arguments to still be calculated
     workspace_ = Workspace{};
+    return;
```
jmarrec, line 36: that's ok.

src/workflow/OSWorkflow.cpp
```diff
@@ -194,6 +194,12 @@ void OSWorkflow::saveOSMToRootDirIfDebug() {
     return;
   }
 
+  // Skip saving the OSM if halted
+  if (runner.halted()) {
+    LOG(Info, "Workflow halted, skipping saving the OSM");
+    return;
+  }
```
jmarrec: Not sure whether we actually want this or not.

This is saving to root dir only if debug.

And the in.osm is saved in the run/ folder anyways.


Also, It's say the source of truth should be the classic CLI. Run this example https://github.com/NREL/OpenStudio-workflow-gem/blob/4a799dbf4c5477610c9e9665dfa7729fe74fc52e/spec/files/halt_workflow_osw/halt_workflow.osw

```
cd OpenStudio-workflow-gem/spec/files/halt_workflow_osw
openstudio classic --verbose run --show-stdout --debug -w halt_workflow.osw
```

I do end up with the "in.osm" (and in.idf, empty) in the root dir.

jmarrec:
```
git clean -fxd
openstudio classic --verbose run --show-stdout -w halt_workflow.osw
git ls-files . --ignored --exclude-standard --others
```

```
out.osw
run/data_point.zip
run/data_point_out.json
run/finished.job
run/in.osm
run/measure_attributes.json
run/objectives.json
run/pre-preprocess.idf
run/results.json
run/run.log
run/started.job
```

jmarrec: Before fix

```
git clean -fxd
openstudio run --show-stdout -w halt_workflow.osw
git ls-files . --ignored --exclude-standard --others
```

```diff
out.osw
run/data_point.zip
run/data_point_out.json
run/finished.job
+ run/in.idf
run/in.osm
run/measure_attributes.json
run/pre-preprocess.idf  # And this is full of stuff
run/results.json
run/run.log
run/started.job
```

jmarrec: With this PR:

```
git clean -fxd
$os_build/Products/openstudio run --show-stdout -w halt_workflow.osw
git ls-files . --ignored --exclude-standard --others
```

I still have run/in.idf and run/pre-processed.idf but at least they contain only the "Version" object.

```
(py312)(3.2.2)julien@halt_workflow_osw (master *=)$ cat run/in.idf 

Version,
  25.2.0;                                 !- Version Identifier

(py312)(3.2.2)julien@halt_workflow_osw (master *=)$ cat run/pre-preprocess.idf 

Version,
  25.2.0;                                 !- Version Identifier
```

jmarrec:
@joseph-robertson I'm tempted to say that the RunPreprocess check for halted is misplaced in the original workflow gem: https://github.com/NREL/OpenStudio-workflow-gem/blob/4a799dbf4c5477610c9e9665dfa7729fe74fc52e/lib/openstudio/workflow/jobs/run_preprocess.rb#L46-L49

the C++ implementation does it exactly the same.

https://github.com/NREL/OpenStudio/blob/887ed96f7135b1fbc44744806b942fbbcd809db4/src/workflow/RunPreProcess.cpp#L50-L54

I don't see why you would create pre-process.idf and apply energyplus output requests if you halted, and didn't create an IDF to begin with.

(But then again, I completely fail to see what haltWorkflow is all about since I never had a use case for it)

shorowit:
> (But then again, I completely fail to see what haltWorkflow is all about since I never had a use case for it)

It’s possible that ResStock (and I assume ComStock) are the only users of it. It’s used to skip simulations when an upgrade doesn’t apply to a given building sample. See https://github.com/NREL/OpenStudio/issues/2601

src/workflow/OSWorkflow.cpp
joseph-robertson: So with halt workflow (debug or not), we'll now expect no in.idf or pre-preprocess.idf, but we'll still expect an (almost empty) in.osm?

jmarrec: Yes, in.osm would not be necessarily empty
Seed + whatever measure ran until halted.

joseph-robertson: And the reason for preserving in.osm is mainly to stay consistent with the classic workflow? You don't think it makes more sense to not produce the in.osm?

jmarrec: Consistent with workflow-gem.

And I could see myself wanting to look at in.osm (I could halt workflow before FT for that reason: don't fail, but produce the model with measures applied). Anyways, if I even think there's a valid application out there and I don't see a strong drawback, I'd want to keep the existing behavior in check.

And I do not see a strong drawback here (only one I can think of is disk space, but that's not a very pressing concern).

jmarrec:
@anchapin the windows node is offline and can't be launched from jenkins admin panel

https://ci.openstudio.net/computer/openstudio%2Dwin%2Dserver%2Dvs%2D2019%2Dssh%2Dincr%2Daws/log

```
SSHLauncher{...}
[12/09/25 08:07:16] [SSH] Opening SSH connection to xxx.xxx.xxx.xxx

No route to host (Host unreachable)
SSH Connection failed with IOException: "No route to host (Host unreachable)", retrying in 30 seconds. There are 3 more retries left.
```

src/workflow/OSWorkflow.cpp
shorowit: Note that you can specify the "completed status" when you halt the workflow, and [it can be "Success", "Fail", "Invalid", and "Cancel"](https://github.com/NREL/OpenStudio/commit/97a52031d59c1e0e70b6db20ad424d6be5ac5377). We are using "Invalid" in ResStock, which arguably could mean that you shouldn't get an in.osm. But it's not the end of the world. The most important thing is to not get an in.idf, considering that you won't be running the simulation.

anchapin:
> @anchapin the windows node is offline and can't be launched from jenkins admin panel
> 
> https://ci.openstudio.net/computer/openstudio%2Dwin%2Dserver%2Dvs%2D2019%2Dssh%2Dincr%2Daws/log
> 
> ```
> SSHLauncher{...}
> [12/09/25 08:07:16] [SSH] Opening SSH connection to xxx.xxx.xxx.xxx
> 
> No route to host (Host unreachable)
> SSH Connection failed with IOException: "No route to host (Host unreachable)", retrying in 30 seconds. There are 3 more retries left.
> ```

I'm working on making a new windows node on AWS since our old one had to be shut down. I'm hoping it will be ready soon but I'm dealing with 'packer' difficulties.

joseph-robertson:
Run folder:

**Before**
image

**After**
image

PR #5561, opened 2025-12-18
Fix segfault in ruby bindings

kbenne:
```
/**
 * Register a Ruby end_proc hook that tears down ANTLR singletons safely when the interpreter exits.
 *
 * Why does this exist? Linking the Modelica parser into openstudio.so drags in ANTLR's global
 * singletons (notably CommonTokenFactory::DEFAULT). Ruby unloads extensions with dlclose() and runs
 * them through its GC at exit, so ANTLR's destructors fire while the interpreter is already winding
 * down, which consistently segfaults. We cannot delete the singleton from a global static
 * destructor either because every executable that links
 * openstudio_modelica (gtests, CLI, etc.) would explode during __cxa_finalize() before Ruby was
 * ever involved, and we cannot register Ruby callbacks from such a static because rb_set_end_proc
 * must run after the VM boots.
 *
 * Instead, we register a Ruby end_proc callback once per interpreter that releases the singleton
 * *only* when the Ruby VM shuts down. That keeps the workaround scoped to Ruby, preserves the
 * default teardown order for other binaries, and mirrors Ruby's own extension lifecycle.
 */
```

### DIFF
--- a/ruby/bindings/ModelicaAntlrCleanup.cpp
+++ b/ruby/bindings/ModelicaAntlrCleanup.cpp
@@ -1,0 +1,42 @@
+#include "ModelicaAntlrCleanup.hpp"
+
+#include <antlr4-runtime.h>
+#include <ruby.h>
+
+#include <memory>
+#include <mutex>
+
+namespace {
+
+using FactoryPtr = std::unique_ptr<antlr4::TokenFactory<antlr4::CommonToken>>;
+
+extern "C" void modelica_antlr_cleanup(VALUE) {
+  // This runs as part of Ruby's end_proc chain before the VM dlcloses openstudio.so. We explicitly
+  // release the singleton so ANTLR's own static destructor becomes a no-op; otherwise Ruby would
+  // run the destructor after parts of the interpreter are gone and we'd segfault while unloading.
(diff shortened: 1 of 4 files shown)


ruby/bindings/ModelicaAntlrCleanup.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+void registerModelicaAntlrCleanup() {
+  // Ruby's Init_* entry points are called once per interpreter, so guard the end_proc registration
+  // with std::call_once to avoid stacking duplicate callbacks when Init_openstudiomodelica is
+  // invoked multiple times (eg. when tests reload the bindings). This static also satisfies the
+  // "just use a registrar" idea without running Ruby APIs during global constructors; rb_set_end_proc
+  // must execute while the VM is alive, so the call_once guard lives inside the Init path instead of
+  // at the translation-unit scope.
+  static std::once_flag registeredFlag;
+  std::call_once(registeredFlag, []() { rb_set_end_proc(modelica_antlr_cleanup, Qnil); });
```
jmarrec, line 39: Yay! more ruby shenanigans.

Anyways, this looks legit, is properly documented, and the code quality is nice.

I'm sure it was such a blast to track down the cause and fix it... good job @kbenne!

PR #5565, opened 2025-12-23
Vcpkg Option

anchapin:
Create a new github action workflow that utilizes vcpkg instead of Conan.

I'm hoping that this will speed up the Windows build.

### DIFF
--- a/.github/workflows/full-build-vcpkg.yml
+++ b/.github/workflows/full-build-vcpkg.yml
@@ -1,0 +1,1200 @@
+# on:
+#   push:
+#     branches:
+#       - develop
+#     tags:
+#       - "v*"
+#   schedule:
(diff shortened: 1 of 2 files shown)


jmarrec:
@anchapin I don't get the point of this PR.

Conan has been used so we have a cross platform way of pinning dependencies. If this is too slow it probably just means we're missing binaries on our artifactory (https://conan.openstudio.net)

Deciding we want to change from conan to vcpkg is really not a light decision to make at all, and it should be weighted and discussed (I don't see this PR having a review).

(This is also going to affect downstream projects such as a the OpenStudio Application)

jmarrec:
Ok I see that #5566 reverted this

anchapin:
I was just trying to test vcpkg in parallel hoping that it would speed up the builds on the free github runners but I ended up giving up trying to get it to work for the time being.

Issue #5574, opened 2026-01-01
Missing `@loader_path/../lib` in openstudio.bundle


macumber:
### Description

openstudio.bundle is installed in ./Ruby and libopenstudiolib.dylib is installed in ./lib.  However, there is no `@loader_path/../lib` in openstudio.bundle which means that trying to load openstudio in system ruby fails.

### Current Behavior

`ruby -I /path/to/os/Ruby/ -e "require 'openstudio'` will fail due to libopenstudiolib.dylib not found

### Expected Behavior

`@loader_path/../lib` in openstudio.bundle will allow libopenstudiolib.dylib to be found

### Steps to reproduce

1. `ruby -I /path/to/os/Ruby/ -e "require 'openstudio'` 

### Possible Solution

Could possibly set this similar to where it is set for Python bindings: https://github.com/NREL/OpenStudio/blob/develop/ProjectMacros.cmake#L392

### Operating System affected

macOS arm64, macOS x86_64

### Environment

N/A

### Version of OpenStudio

3.11.0-rc2

jmarrec:
@macumber do you know if this is a regression or was it the same in earlier OS SDK versions?

brianlball:
I think its new and related to the issues we are seeing in OSAF tests now
https://github.com/NREL/OpenStudio-server/actions/runs/20727664604/job/59507203493

jmarrec:
@kbenne I think that's your modelica work.

https://github.com/NREL/OpenStudio/blob/7d1f8bdb36e78b39b8d0ac511db510b75db89d85/ruby/module/CMakeLists.txt#L71

The issue is likely that this explicitly links to openstudiolib

https://github.com/NREL/OpenStudio/blob/7d1f8bdb36e78b39b8d0ac511db510b75db89d85/src/modelica/CMakeLists.txt#L47-L57


And since it doesn't specify PUBLIC/PRIVATE/INTERACE, with this signature Library dependencies are transitive by default

https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-both-a-target-and-its-dependents

PR #5575, opened 2026-01-05
Add macOS build improvements

anchapin:
Enhance the CMake build configuration for macOS by adding the macOS version specification and updating the installation RPATH to include the library path. This ensures better compatibility and functionality for macOS bundles.

### DIFF
--- a/.github/workflows/incremental-build.yml
+++ b/.github/workflows/incremental-build.yml
@@ -123,11 +123,13 @@ jobs:
             mkdir -p $HOME/bin
             ln -sf $SWIG_BIN $HOME/bin/swig
             export PATH=$HOME/bin:$PATH
+            SWIG_EXEC_ARG="-DSWIG_EXECUTABLE=$HOME/bin/swig"
           fi
           
           cmake -G Ninja \
                 -DCMAKE_TOOLCHAIN_FILE=conan_toolchain.cmake \
                 -DCMAKE_BUILD_TYPE:STRING=Release \
+                $SWIG_EXEC_ARG \
                 -DBUILD_TESTING:BOOL=ON \
                 -DCPACK_BINARY_DEB:BOOL=ON \
                 -DCPACK_BINARY_TGZ:BOOL=ON \


.github/workflows/full-build.yml
```diff
@@ -824,7 +825,7 @@ jobs:
           -DCPACK_BINARY_IFW:BOOL=ON \
           -DBUILD_PYTHON_BINDINGS:BOOL=ON \
           -DBUILD_PYTHON_PIP_PACKAGE:BOOL=OFF \
-          -DCMAKE_INSTALL_RPATH="@loader_path;@executable_path" \
+          -DCMAKE_INSTALL_RPATH="@loader_path;@executable_path;@loader_path/../lib" \
```
jmarrec, line 828: Is this for #5574 ?

This is the wrong solution IMHO. This should be done in cmakelists, not via a command line parameter (which anyone building is likely to forget to add).

jmarrec: It could be that #5574 exists is because you are passing this variable CMAKE_INSTALL_RPATH to begin with (or something similar). In any case I don't think we want to do that.

We never explicitly added libopenstudiolib.dylib into the openstudio.bundle and it has always worked fine, until the full-build workflow was added.

```
$ otool -L /Applications/OpenStudio-3.10.0/Ruby/openstudio.bundle
/Applications/OpenStudio-3.10.0/Ruby/openstudio.bundle:
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 61439.101.1)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 3423.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1900.178.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
```


```
$ otool -L /Applications/OpenStudio-3.11.0-rc2/Ruby/openstudio.bundle
/Applications/OpenStudio-3.11.0-rc2/Ruby/openstudio.bundle:
	@rpath/libopenstudiolib.dylib (compatibility version 0.0.0, current version 0.0.0)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 61439.120.27)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 3502.1.255)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1900.180.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
```

Should track where this started popping.

jmarrec: I don't get why passing CMAKE_INSTALL_RPATH, seems to be stemming from https://github.com/NREL/OpenStudio/commit/40d062508380918cf945ca44bb68a4a69c149522 and the commit message doesn't explain anything

anchapin: I see that you made a PR #5583 to fix this so that CMAKE_INSTALL_RPATH is no longer provided.

PR #5615, opened 2026-04-28
Os 4.0 dev epmodel swig

kbenne:
# Add EPModel SWIG bindings for Python and Ruby

* Implement EPModel SWIG binding generation using CMake's native SWIG support. Each EPModel subdirectory now owns a local EPModel*.i file, and the root EPModel module aggregates submodules for Python import exposure.

* Add EPModel binding support for both Python and Ruby, including Ruby initialization wiring and a dedicated EPModel Ruby init source. The Python bindings now expose EPModel submodules through openstudio.epmodel and maintain explicit dependency ordering between parent/base modules to avoid SWIG inheritance race conditions.

* Add shared EPModel SWIG macros for forward declarations, vector/optional templates, and wrapped object entrypoints. Root-owned types such as ModelObject, ResourceObject, and Schedule emit their container wrappers from EPModel.i so APIs returning std::vector<ModelObject> produce usable Python sequence values instead of opaque SwigPyObject pointers.

* Add EPModelObjectIncludes.hpp as the common concrete/base object include include file used by generated wrappers, and document the binding entrypoints in the EPModel README.

* Add a targeted EPModel to_Node helper for Python ModelObject casts. This avoids colliding with canonical model's shared IdfObject.to_Node monkey-patch while allowing EPModel ModelObject values that are actually Nodes to be converted to EPModel Node wrappers before passing to APIs requiring Node&.

* The generated bindings have been smoke-tested from the build tree, but install and packaging behavior has not yet been vetted.


src/epmodel/Loop/EPModelLoop.i
```diff
@@ -0,0 +1,16 @@
...
+
+%include <epmodel/EPModel_Common_Include.i>
+%import <epmodel/EPModel.i>
+%import <epmodel/ParentObject/EPModelParentObject.i>
+
+EPMODELOBJECT_WRAP(Loop, <epmodel/Loop/Loop.hpp>)
+EPMODELOBJECT_WRAP(AirLoopHVAC, <epmodel/Loop/AirLoopHVAC.hpp>)
+EPMODELOBJECT_WRAP(PlantLoop, <epmodel/Loop/PlantLoop.hpp>)
+
+#endif
```
jmarrec, line 16: Here is an example.

I notice the same macro is called for an abstract class (Loop) and for actual Concrete classes. fishy?

src/epmodel/ModelObject/EPModelModelObject.i
```diff
@@ -0,0 +1,176 @@
...
+#ifdef SWIGPYTHON
+  %module openstudioepmodelmodelobject
+#endif
+
+%include <epmodel/EPModel_Common_Include.i>
+%import <epmodel/EPModel.i>
+
+%include <epmodel/ModelObject/ModelObject.hpp>
+
+EPMODELOBJECT_WRAP(Timestep, <epmodel/ModelObject/Timestep.hpp>)
```
jmarrec: The same macro is called for a Unique ModelObject

src/epmodel/CMakeLists.txt
jmarrec, line 1: I guess this only got EPModelObjectIncludes.hpp added, the rest is just reordering alphabetically, with the line break between subfolders?

src/epmodel/EPModel_Common_Include.i
```diff
@@ -0,0 +1,602 @@
...
+
+%include <utilities/core/CommonInclude.i>
+%import <utilities/core/CommonImport.i>
+%import <utilities/Utilities.i>
+
+%ignore openstudio::epmodel::detail;
+
+%{
+  #include <epmodel/Model.hpp>
+  #include <epmodel/EPModelObjectIncludes.hpp>
```
jmarrec, line 15: If I'm folllowing correctly, we include EVERY hpp file everywhere now?

jmarrec: I guess Model_Common_Include was kinda doing it via

https://github.com/NatLabRockies/OpenStudio/blob/8008ef767fdc0f9d3dd3fabd383da15d009aef76/src/model/Model_Common_Include.i#L30-L31

src/epmodel/EPModel_Common_Include.i
```diff
@@ -0,0 +1,602 @@
...
+%define EPMODELOBJECT_FORWARD_DECLARE(_name)
+namespace openstudio {
+namespace epmodel {
+class _name;
+%feature("valuewrapper") _name;
+%nodefaultctor _name;
+%ignore _name::_name();
+}
+}
+%enddef
```
jmarrec, line 323: Not props for the intendation/readability here :)

Ok so... if I follow.


```suggestion
%define EPMODELOBJECT_FORWARD_DECLARE(_name)
namespace openstudio {
  namespace epmodel {
    class _name;
    %feature("valuewrapper") _name; // Allow by-value returns without needing a default ctor
    %nodefaultctor _name; // Do not generate **implicit** default constructors
     %ignore _name::_name(); // Ignore **explicit** default constructors too
  }
}
%enddef
```

Our classes shouldn't have an **explicit** default ctor (they all take at least model), so I wonder if the `%ignore _name::_name();` line is 1) correct in intent to begin with, and 2) just useless

src/epmodel/EPModel_Common_Include.i
```diff
@@ -0,0 +1,602 @@
...
+
+%define EPMODELOBJECT_WRAP(_name, _header)
+  %include _header
+  EPMODELOBJECT_TEMPLATES(_name)
+%enddef
```
jmarrec: Lot's missing here.

jmarrec: 1. EPMODELOBJECT_TO_HELPER is not called whatsoever (only Node calls it).
2. Unique and ModelObjects need to be handled differently in TEMPLATES (no need for vectors!)
3. ModelObject extensions also needed: In the end, we want a to_XXX, but also epmodel.Model getters
     * model.getXXX(handle)
     * model.getXXXs()
     * model.getXXXByName(str)
     * model.getXXXByNames(str, exactMatch)
     *  For Unique: model.getXXX() and model.getOptionalXXX

src/epmodel/EPModel_Common_Include.i
```diff
@@ -0,0 +1,602 @@
...
+#if defined SWIGPYTHON
+  %pythoncode %{
+def _to_##_name(self):
+    """Try to cast the EPModel object to _name."""
+    return to##_name(self)
+# Patch EPModel's ModelObject instead of the shared IdfObject base. Canonical
+# model bindings also install to_<Type> helpers on IdfObject, and using that
+# shared surface here would make same-named helpers such as to_Node collide
+# across openstudio.model and openstudio.epmodel.
+openstudioepmodel.ModelObject.to_##_name = _to_##_name
```
jmarrec:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-1-098070d69a53> in ?()
----> 1 load_openstudio_interactive()

~/.ipython/profile_default/startup/openstudio_helpers.py in ?()
    108     else:
    109         p = "openstudio"
    110         is_package = True
    111     global openstudio
--> 112     import openstudio
    113 
    114     print(f"Loading '{p}': {openstudio.openStudioLongVersion()}")
    115 

~/Software/Others/OS-build-release/Products/python/openstudio.py in ?()
     56             os.add_dll_directory(bin_dir)
     57 
     58     import openstudioairflow as airflow
     59     import openstudioenergyplus as energyplus
---> 60     import openstudioepmodel as epmodel
     61     import openstudioepjson as epjson
     62     import openstudiogbxml as gbxml
     63     import openstudiogltf as gltf

~/Software/Others/OS-build-release/Products/python/openstudioepmodel.py in ?()
    716 from openstudioepmodelcurve import *
    717 from openstudioepmodelhvaccomponent import *
    718 from openstudioepmodelscheduleinterval import *
    719 from openstudioepmodelloop import *
--> 720 from openstudioepmodelstraightcomponent import *
    721 from openstudioepmodelwatertoaircomponent import *
    722 from openstudioepmodelzonehvaccomponent import *
    723 from openstudioepmodelwatertowatercomponent import *

~/Software/Others/OS-build-release/Products/python/openstudioepmodelstraightcomponent.py in ?()
  27463 #Patch EPModel's ModelObject instead of the shared IdfObject base. Canonical
  27464 # model bindings also install to_<Type> helpers on IdfObject, and using that
  27465 # shared surface here would make same-named helpers such as to_Node collide
  27466 # across openstudio.model and openstudio.epmodel.
> 27467 openstudioepmodel.ModelObject.to_##Node = _to_##Node
  27468 
  27469 

AttributeError: type object 'ModelObject' has no attribute 'to_'
```

src/epmodel/EPModel_Common_Include.i
kbenne: yep. I questioned this too, but I think we always have been. It would be nice if we didn't though.

PR #5619, opened 2026-05-05
Fix workflow: add workflow_dispatch support and use /mnt directly instead of symlinks

anchapin:
This PR makes two important improvements to the full-build workflow:

1. **Adds workflow_dispatch support**: Allows the build jobs (linux-build, macos-build, windows-build) to run when the workflow is triggered manually via the workflow_dispatch interface, in addition to the existing push, schedule, and pull_request events.

2. **Replaces symlink approach with direct /mnt usage**: 
   - Eliminates the complex Prepare workspace step that was creating symlinks to /mnt
   - Changes all build-related paths to use /mnt directly:
     - OPENSTUDIO_SOURCE: /mnt/OpenStudio
     - OPENSTUDIO_BUILD: /mnt/build
     - CCACHE_BASEDIR: /mnt/OpenStudio
     - Cache directories: /mnt/.ccache, /mnt/.conan2, /mnt/.sccache
   - This resolves Boost filesystem::canonical() path resolution issues that were causing test failures
   - Simplifies the workflow while maintaining the benefit of using /mnt for additional build space

The changes ensure that when manually triggering the workflow for branches like 'os-4.0-dev-epmodel-workflow', all build jobs will execute properly without being skipped, and without the path resolution issues that were affecting tests.

### DIFF
--- a/.github/workflows/full-build.yml
+++ b/.github/workflows/full-build.yml
@@ -56,24 +56,24 @@ permissions:
   id-token: write
 
 env:
-  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  BUILD_TYPE: Release
-  OPENSTUDIO_BUILD: build
-  OPENSTUDIO_SOURCE: OpenStudio
-  PYTHON_REQUIRED_VERSION: "3.12.2"
-  SDKROOT: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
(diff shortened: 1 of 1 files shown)


jmarrec:
I would like to understand why you want to use /mnt for the build/source, and where that /mnt folder is documented?

Is this the undocummented /mnt from Azure? If so, I don't think we really want to use it for build/source. We can put the extra swap and ccache/conan cache on it if you want, but I don't think it has more space that the regular workspace.

Also, be mindful of what impact that has on caching, as I think actions/cache deals only with workspace-internal paths

PR #5629, opened 2026-07-11
Fix #5614: Handle UTF-8 BOM in EPW files

wnykuang:
## Summary

- Fix #5614
- Strip a leading UTF-8 byte order mark before parsing an EPW `LOCATION` header.
- Add a regression test proving BOM-prefixed EPW content remains loadable.

## Root cause

`EpwFile::parse` passed the first line through unchanged, while `parseLocation` requires the first comma-delimited field to equal `LOCATION`. For UTF-8-with-BOM files, that field begins with the bytes `EF BB BF`, so a valid header was reported as missing its `LOCATION` specifier.

The fix only removes a UTF-8 BOM at the beginning of the first line. Existing validation remains unchanged, and this does not attempt to decode UTF-16 or UTF-32 files.

## User impact

EPW files saved as UTF-8 with BOM, including the file attached to #5614, can now be processed normally.

## Validation

- Added `Filetypes.EpwFile_UTF8BOM` and observed it fail before the implementation change and pass afterward.
- Ran all 25 `Filetypes.EpwFile*` tests successfully.
- Re-ran the original 2.5 MB EPW attachment from #5614 and confirmed the parsing failure no longer occurs.
- Ran `git diff --check` successfully.

Fixes #5614.

PR #5633, opened 2026-07-16
Hotfix #5629 - read_as_string should open as binary

jmarrec:
Pull request overview
---------------------

- Hotfix #5629

Root cause: read_as_string(const openstudio::path& t_path) in src/utilities/core/FilesystemHelpers.cpp opened the file with openstudio::filesystem::ifstream f(t_path); — no std::ios_base::binary flag. On Windows this defaults to text mode, which translates \r\n → \n while reading.

The underlying read() helper computes file_len from tellg()/seekg() (the raw on-disk byte count) and then does a single t_file.read(&buf.front(), file_len). In text mode, the CRLF translation means fewer characters actually get read than the buffer was sized for, so the string gets truncated/corrupted with trailing garbage — silently, no error. This only manifests on Windows since POSIX has no text/binary distinction, which matches the "Windows and only Windows" symptom, and explains why parsing barreled all the way to line 8769 before choking on garbage.

The sibling function read(const path&) already correctly used std::ios_base::binary — the string-returning overload just missed it.

Fix: src/utilities/core/FilesystemHelpers.cpp:38 — added std::ios_base::binary to match.

### Pull Request Author

**Labels:**

### Review Checklist

 - [ ] Labeling is ok

### DIFF
--- a/src/utilities/core/FilesystemHelpers.cpp
+++ b/src/utilities/core/FilesystemHelpers.cpp
@@ -35,7 +35,7 @@ namespace filesystem {
   }
 
   std::string read_as_string(const openstudio::path& t_path) {
-    openstudio::filesystem::ifstream f(t_path);
+    openstudio::filesystem::ifstream f(t_path, std::ios_base::binary);
     return read_as_string(f);
   }
 


jmarrec:
CI "Build and Test" reports `EPModelFixture` failures, when this is on `develop` and epmodel doesn't exist at all. @anchapin FYI. I think this is a cleaning step that is missing, it's picking up failed tests from a previous run...

